PR #1, opened 2020-12-08
Fix the range of for loop in memory_delete()

sammer1107:
This pull request fix the range of `for` loop in memory_delete() to avoid memory leak.


jserv:
Thank @sammer1107 for contributing!

PR #2, opened 2020-12-29
Use computed-goto to lower instruction dispatch overhead

sammer1107:
+ This pull request applies the computed goto technique to the
main loop. This modification is tested to have 15 to 23
percent improvement on performance.
+ The old implementation of rv_step can be enabled
by setting the environment variable ENABLE_COMPUTED_GOTO to 0.
+ Only clang and gcc is supported by computed goto feature,
ENABLE_COMPUTED_GOTO will be ignored if other compiler is used.
+ An op_unimp handler is also added in order to acheive
this. This also allows better handling of unimplemented
opcodes instead of jumping into NULL.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,8 @@
-CFLAGS = -std=gnu99 -O2 -Wall -Wextra
+ifeq ("$(origin CC)", "default")
+	CC = gcc
+endif
+
+CFLAGS = -std=gnu99 -Wall -Wextra
 CFLAGS += -include common.h
 
 # Base configurations for RISC-V extensions
@@ -12,6 +16,17 @@ CFLAGS += -D DEFAULT_STACK_ADDR=0xFFFFF000
 CFLAGS += -D ENABLE_SDL
 CFLAGS += `sdl2-config --cflags`
 LDFLAGS += `sdl2-config --libs`
+
+# Whether to enable computed goto in riscv.c
+ENABLE_COMPUTED_GOTO ?= 1
+ifeq ("$(ENABLE_COMPUTED_GOTO)", "1")
+ifneq ($(filter $(CC), gcc clang),)
+riscv.o: CFLAGS += -D ENABLE_COMPUTED_GOTO
+	ifeq ("$(CC)", "gcc")
+riscv.o: CFLAGS += -fno-gcse -fno-crossjumping 
+	endif
+endif
+endif
 
 # Control the build verbosity
 ifeq ("$(VERBOSE)","1")
--- a/riscv.c
+++ b/riscv.c
@@ -809,7 +809,7 @@ void rv_step(struct riscv_t *rv, int32_t cycles)
 #define DISPATCH()                                      \
     {                                                   \
         if (rv->csr_cycle >= cycles_target || rv->halt) \
-            goto exit;                                  \
+            goto quit;                                  \
         /* fetch the next instruction */                \
         inst = rv->io.mem_ifetch(rv, rv->PC);           \
         /* standard uncompressed instruction */         \
@@ -826,7 +826,7 @@ void rv_step(struct riscv_t *rv, int32_t cycles)
     {                                 \
         /* dispatch this opcode */    \
         if (!op_##instr(rv, inst))    \
-            goto exit;                \
+            goto quit;                \
         /* increment the cycles csr*/ \
         rv->csr_cycle++;              \
     }
@@ -856,7 +856,7 @@ void rv_step(struct riscv_t *rv, int32_t cycles)
 #endif
     TARGET(unimp)
 
-exit:
+quit:
     return;
 
 #undef DISPATCH
--- a/riscv.c
+++ b/riscv.c
@@ -785,16 +785,17 @@ void rv_step(struct riscv_t *rv, int32_t cycles)
     assert(rv);
     const uint64_t cycles_target = rv->csr_cycle + cycles;
     uint32_t inst, index;
-// clang-format off
+
 #define OP_UNIMP op_unimp
 #ifdef ENABLE_COMPUTED_GOTO
-    #define OP(instr) &&op_##instr
-    #define TABLE_TYPE const void *
-#else
-    #define OP(instr) op_##instr
-    #define TABLE_TYPE const opcode_t
+#define OP(instr) &&op_##instr
+#define TABLE_TYPE const void *
+#else  // ENABLE_COMPUTED_GOTO = false
+#define OP(instr) op_##instr
+#define TABLE_TYPE const opcode_t
 #endif
 
+    // clang-format off
     TABLE_TYPE jump_table[] = {
     //  000         001           010        011           100         101        110   111
         OP(load),   OP(load_fp),  OP(unimp), OP(misc_mem), OP(op_imm), OP(auipc), OP(unimp), OP(unimp), // 00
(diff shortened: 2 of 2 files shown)


jserv:
Change the subject (of pull request and git commit message) to "Use computed-goto to lower instruction dispatch overhead"

Makefile
```diff
@@ -13,6 +13,11 @@ CFLAGS += -D ENABLE_SDL
 CFLAGS += `sdl2-config --cflags`
 LDFLAGS += `sdl2-config --libs`
 
+# Whether to enable computed goto in riscv.c
+ifeq ("$(NO_COMPUTED_GOTO)", "")
+riscv.o: CFLAGS += -fno-gcse -fno-crossjumping -D ENABLE_COMPUTED_GOTO
```
jserv: The option `-fno-crossjumping` is gcc specific. And, clang would reject it.
See https://github.com/peberlein/interpreter

riscv.c
```diff
@@ -758,47 +758,124 @@ static bool op_amo(struct riscv_t *rv, uint32_t inst)
...
+        inst = rv->io.mem_ifetch(rv, rv->PC);
 
         // standard uncompressed instruction
         if ((inst & 3) == 3) {
-            const uint32_t index = (inst & INST_6_2) >> 2;
+            index = (inst & INST_6_2) >> 2;
 
-	    // dispatch this opcode
-            const opcode_t op = opcodes[index];
+            // dispatch this opcode
```
jserv, line 874: Use `clang-format -i` to ensure the consistent style.

Makefile
```diff
@@ -13,6 +13,11 @@ CFLAGS += -D ENABLE_SDL
 CFLAGS += `sdl2-config --cflags`
 LDFLAGS += `sdl2-config --libs`
 
+# Whether to enable computed goto in riscv.c
+ifeq ("$(NO_COMPUTED_GOTO)", "")
```
jserv: Use the consistent name scheme, `ENABLE_COMPUTED_GOTO`.

riscv.c
```diff
@@ -758,47 +758,124 @@ static bool op_amo(struct riscv_t *rv, uint32_t inst)
...
+        OP(branch), OP(jalr),     OP(unimp), OP(jal),      OP(system), OP(unimp), OP(unimp), OP(unimp), // 11
+    };
+    // clang-format on
+
+#ifdef ENABLE_COMPUTED_GOTO
+#define DISPATCH()                                      \
+    {                                                   \
+        if (rv->csr_cycle >= cycles_target || rv->halt) \
+            goto exit;                                  \
+        else {                                          \
```
jserv: The `else {` is redundant.

riscv.c
sammer1107: + I used clang-format but it doesn't complain about this line. 
+ It did complain about nested #if at [line 790](https://github.com/sysprog21/rv32emu-next/pull/2/files/7087c7f3ecdd630ef92198109b655dab816d50e6#diff-76a818eab913b98b885f954273a467dea476f7ddbfd5886f97b1e56cb0e2b58dR790-R796). However, I think it looks better with indentation.
+ It also tried to make [the label](https://github.com/sysprog21/rv32emu-next/pull/2/files/7087c7f3ecdd630ef92198109b655dab816d50e6#diff-76a818eab913b98b885f954273a467dea476f7ddbfd5886f97b1e56cb0e2b58dR837) and `EXEC()` into single line but it looks better separated for me.

Should I just use clang-format or leave it as is ?

jserv: Use `clang-format off` and `clang-format on` to switch manual indentation.

Makefile
jserv: You should detect the compiler. If it is gcc, apply the options. Otherwise, omit.

Makefile
jserv: The computed-goto should be supported only to clang and gcc. You have to detect the compilers.

jserv (changes requested):
Change the subject (of pull request and git commit message) to "Use computed-goto to lower instruction dispatch overhead"

Makefile
jserv: It is not necessary to check `NO_COMPUTED_GOTO`. Instead, the user can manually set `ENABLE_COMPUTED_GOTO`. You only have to validate its availability.

riscv.c
jserv: Remove the `else {` statement here, which is redundant.

jserv:
The commit message should be amended, mentioning the benefit of the proposed computed-goto change.

riscv.c
```diff
@@ -758,47 +758,122 @@ static bool op_amo(struct riscv_t *rv, uint32_t inst)
...
+
+#define EXEC(instr)                   \
+    {                                 \
+        /* dispatch this opcode */    \
+        if (!op_##instr(rv, inst))    \
+            goto exit;                \
+        /* increment the cycles csr*/ \
+        rv->csr_cycle++;              \
+    }
+// clang-format off
```
jserv: Is `clang-format off` necessary?

jserv (changes requested):
The commit message should provide the measurement (numbers) for speedup.

riscv.c
sammer1107: Because I think it looks better that line 834 and 835 are separated at different line, so I added `clang-format off` here.

riscv.c
```diff
@@ -758,47 +758,122 @@ static bool op_amo(struct riscv_t *rv, uint32_t inst)
...
-    op_branch, op_jalr,     NULL,     op_jal,      op_system, NULL,     NULL, NULL, // 11
-};
-// clang-format on
-
 void rv_step(struct riscv_t *rv, int32_t cycles)
 {
     assert(rv);
     const uint64_t cycles_target = rv->csr_cycle + cycles;
+    uint32_t inst, index;
+// clang-format off
```
jserv: Let's remove the `clang-format` mark. Stick on the formatting.
```cpp

#define OP_UNIMP op_unimp                                                                                                                                     
#ifdef ENABLE_COMPUTED_GOTO
#define OP(instr) &&op_##instr
#define TABLE_TYPE const void *
#else
#define OP(instr) op_##instr
#define TABLE_TYPE const opcode_t
#endif
```

The above is acceptable and common.

riscv.c
jserv: Remove `clang-format` mark. You should add one blank line before the macro.

riscv.c
jserv: Alternatively, you can add some visual annotation as following:
```c
#ifdef ENABLE_COMPUTED_GOTO
#define OP(instr) &&op_##instr
#define TABLE_TYPE const void *
#else /* without computed goto */
#define OP(instr) op_##instr
#define TABLE_TYPE const opcode_t
#endif
```

riscv.c
```diff
@@ -758,47 +758,121 @@ static bool op_amo(struct riscv_t *rv, uint32_t inst)
...
+            /* TODO: compressed instruction*/           \
+            assert(!"Unreachable");                     \
+        }                                               \
+    }
+
+#define EXEC(instr)                   \
+    {                                 \
+        /* dispatch this opcode */    \
+        if (!op_##instr(rv, inst))    \
+            goto exit;                \
```
jserv: Rename `exit` to `quit` or `bail` because `exit` was meant to be a system call in general.

jserv:
@sammer1107, would you continue this work recently?

sammer1107:
> @sammer1107, would you continue this work recently?

Yes, I would like to finish the work here. I fixed the naming problem of 'exit'.

jserv:
Thank @sammer1107 for contributing!

sammer1107:
@jserv Should I also send the pull request for the compliance test ?

jserv:
> @jserv Should I also send the pull request for the compliance test ?

Please go ahead.

PR #4, opened 2021-01-18
Implement RV32C instruction decompressor

kaeteyaruyo:
I implemented the RV32C extension on this emulator. Excepts `C.EBREAK`, all the other instructions are passed the compliance test. The detail of implementation are recorded in the commit message.

~~I understand that the way I implement the decompressor is very different from the instruction decoder in this project. I will try to reorganize the code and make it more modularized.~~
Edit at 1/20: modularization has been done.

There are somethings that I am not sure whether I implement correctly:
* The error handling on the invalid 16-bits instruction (`0x0000`). For now, if the emulator encounters an invalid 16-bits instruction, it will halt immediately.
* The address alignment in the `memory_read_ifetch()`. I wonder if the function returns correct data at the boundary of a chunk.
* When should I increase the `rv->csr_cycle`?


jserv:
I would like to ask ccs100203 to review.

kaeteyaruyo:
> I would like to ask ccs100203 to review.

Sorry for late reply, but should I contact ccs100203 myself to ask for a review?

riscv.c
```diff
@@ -6,6 +6,10 @@
 #include "riscv.h"
 #include "riscv_private.h"
 
+#ifdef ENABLE_RV32C
+#include "rvc.h"
```
jserv: Change `rvc.h` to `compressed.h` for readability.

jserv:
> I implemented the RV32C extension on this emulator. Excepts `C.EBREAK`, all the other instructions are passed the compliance test. The detail of implementation are recorded in the commit message.

Add corresponding FIXME/TODO in the comment of source files.

> There are somethings that I am not sure whether I implement correctly:
> * The error handling on the invalid 16-bits instruction (`0x0000`). For now, if the emulator encounters an invalid 16-bits instruction, it will halt immediately.
> * The address alignment in the `memory_read_ifetch()`. I wonder if the function returns correct data at the boundary of a chunk.
> * When should I increase the `rv->csr_cycle`?

Instead of halting, can you raise verbose error messages when executing invalid 16-bits instruction?

kaeteyaruyo:
> > I implemented the RV32C extension on this emulator. Excepts `C.EBREAK`, all the other instructions are passed the compliance test. The detail of implementation are recorded in the commit message.
> 
> Add corresponding FIXME/TODO in the comment of source files.
> 
I think I implement the decoder for `C.EBREAK` correctly. I tried to print some message to see if the emulator executes `EBREAK` when it encounters `C.EBREAK`, and the result shows that it actually does. But in compliance test, it checks the correctness of implementation by verifying the memory dump, and it shows that the case `C.EBREAK` doesn't manipulate memory correctly (there should be something in the memory, but the range it tests are all filled with `deadbeef`). I am not sure whether the problem comes from emulator or the device configuration in compliance test. Should I still add the TODO comment?
> > There are somethings that I am not sure whether I implement correctly:
> > 
> > * The error handling on the invalid 16-bits instruction (`0x0000`). For now, if the emulator encounters an invalid 16-bits instruction, it will halt immediately.
> > * The address alignment in the `memory_read_ifetch()`. I wonder if the function returns correct data at the boundary of a chunk.
> > * When should I increase the `rv->csr_cycle`?
> 
> Instead of halting, can you raise verbose error messages when executing invalid 16-bits instruction?
>
I raise `rv_except_illegal_inst` error when encountering that.

kaeteyaruyo:
I noticed that there is a `ENABLE_COMPUTED_GOTO` feature in current implementation, but the `DISPATCH()` part do not support compressed instruction yet. Should I also add the decoder part into `DISPATCH()`?

jserv:
Drop in favor of #11

PR #5, opened 2021-03-15
Integrate riscv-compliance

sammer1107:
+ This pull request integrate riscv-compliance (riscv-arch-test) 2.0 into rv32emu-next
+ The Makefile target support running all test cases or a specific test case. The command line options are explained in README.md
+ Some tests does not pass because it still doesn't have a correct value defined yet.


Makefile
```diff
@@ -60,6 +60,19 @@ check: $(BIN)
 	(cd $(OUT); ../$(BIN) hello.elf)
 	(cd $(OUT); ../$(BIN) puzzle.elf)
 
+# variables for compliance
+COMPLIANCE_DIR ?= ./riscv-compliance
+export RISCV_PREFIX ?= riscv32-unknown-elf-
+export RISCV_TARGET = test-rv32emu
```
jserv: Change the target to `compliance-tests` or something alike.

README.md
```diff
@@ -30,6 +30,17 @@ make demo
 The build script will then download data file for Doom automatically. SDL2 based window
 should appear when Doom is loaded and executed.
 
+### Run riscv-compliance test
+Once the submodule `riscv-compliance` is pulled, run all the available compliance test via command:
```
jserv, line 34: In `Makefile`, you can add submodule specific operations to pull automatically. See https://github.com/sysprog21/jitboy/blob/master/Makefile

test-rv32emu/device/rv32i_m/M
```diff
@@ -0,0 +1 @@
+I
```
jserv: Why?

test-rv32emu/device/rv32i_m/I/Makefile.include
```diff
@@ -0,0 +1,23 @@
+RUN_TARGET= $(TARGETDIR)/build/rv32emu $(<) \
+    $(RISCV_TARGET_FLAGS) \
+    --compliance $(*).signature.output \
+    1>$(@) 2>&1
+
+RISCV_PREFIX   ?= riscv32-unknown-elf-
```
jserv: Why did you set `RISCV_PREFIX` here?

test-rv32emu/device/rv32i_m/M
sammer1107: Because the file is the same as the file rv32i_m/I, so i used symbolic link.

test-rv32emu/device/rv32i_m/I/Makefile.include
sammer1107: It's a mistake. I will fix that.

jserv:
Can you avoid using symbolic links in source tree? In directory `build`, we can create symbolic links. However, if symbol links exists outside `build` directory, it is a bit difficult and confusing to remove the linked entries.

Makefile
```diff
@@ -60,6 +60,22 @@ check: $(BIN)
...
 	(cd $(OUT); ../$(BIN) puzzle.elf)
 
+# variables for compliance
+COMPLIANCE_DIR ?= ./riscv-compliance
+export RISCV_PREFIX ?= riscv32-unknown-elf-
+export RISCV_TARGET = compliance-target
+export TARGETDIR = $(shell pwd)
+export XLEN = 32
+export JOBS ?= -j
+export WORK = $(TARGETDIR)/build/compliance
```
jserv, line 70: I am not sure why `pwd` is necessary. Can you explain?

main.c
```diff
@@ -101,7 +105,8 @@ static void print_usage(const char *filename)
             "RV32I[MA] Emulator which loads an ELF file to execute.\n"
             "Usage: %s [options] [filename]\n"
             "Options:\n"
-            "  --trace : print executable trace\n",
+            "  --trace : print executable trace\n"
+            "  --compliance [signature filename] : dump signature to the given file for compliance test\n",
```
jserv: Simplified as 
```
"  --compliance [filename] : dump signature to the given file, required bycompliance test\n",

```

Makefile
sammer1107: Because the build process of compliance involves multiple makefiles in different places, I found out that the makefile from the compliance will fail to find our target file if I just set `TARGETDIR` to `.`. Therefore, I set it to the absolute path of our project root.

jserv:
Close in favor of #12

PR #9, opened 2022-01-06
Fix csrrwi command referring risc-v spec

feathertw:


jserv:
`CSRRWI` is an instruction rather than a command.

feathertw:
I will fix it.

PR #24, opened 2022-07-31
Detect NaN in FADD and FSUB

2011eric:
I used the pattern specified in the RISC-V ISA manual: 0x7fc00000.


jserv (changes requested):
List the difference of the results generated by `make arch-test RISCV_DEVICE=F` between the original implementation and the proposed one.

emulate.c
```diff
@@ -5,6 +5,7 @@
 
 #ifdef ENABLE_RV32F
 #include <math.h>
+#define RISCV_NAN 0x7fc00000
```
jserv: Move the definition `RISCV_NAN` to header `riscv_private.h`, so that the incoming JIT compiler can make use of it as well.

2011eric:
Moved the definition to `riscv_private.h`

jserv:
> Moved the definition to `riscv_private.h`

Check [CI regressions](https://github.com/sysprog21/rv32emu/actions/runs/2774871094) and attempt to fix.

jserv (changes requested):
Use "git rebase -i" to squash/rework the commits.

jserv:
> List the difference of the results generated by `make arch-test RISCV_DEVICE=F` between the original implementation and the proposed one.

@2011eric, you should mention the results within git commit messages.
See https://cbea.ms/git-commit/

emulate.c
```diff
@@ -134,6 +134,14 @@ static void rv_except_illegal_insn(struct riscv_t *rv, uint32_t insn)
     rv->csr_mcause = code;
 }
 
+/* no trap supports on floating point exceptions */
+/* TODO: implement fp_exception functions */
+static void rv_except_fp_invalid_operation(struct riscv_t *rv);
```
jserv: Don't put incomplete functions here. If you want to declare something, ensure the availability.

jserv: Also, make git commit messages informative!

riscv_private.h
```diff
@@ -4,6 +4,7 @@
 #include "riscv.h"
 
 #define RV_NUM_REGS 32
+#define RV_NAN 0x7fc00000
```
jserv: `RV_NAN` should not appear in the region. Move it to somewhere after `FMASK_FRAC`.

2011eric: Should I put it here?
```
#ifdef ENABLE_RV32F
enum {
    //             ....xxxx....xxxx....xxxx....xxxx
    FMASK_SIGN = 0b10000000000000000000000000000000,
    FMASK_EXPN = 0b01111111100000000000000000000000,
    FMASK_FRAC = 0b00000000011111111111111111111111,
    //             ....xxxx....xxxx....xxxx....xxxx
    RV_NAN     = 0b01111111110000000000000000000000
};
#endif
```

jserv: It looks fine.

2011eric:
Passed 5 more tests in RV32F, listed in the commit message.

jserv:
> https://cbea.ms/git-commit/

Read [How to Write a Git Commit Message](https://cbea.ms/git-commit/) carefully and enforce the rules:
* Do not end the subject line with a period

2011eric:
I noticed that in the expected output_signature of test `fadd_b1-01`, one of the results is ff800000, which should be -INF.
However,  the overflow flag (OF) and inexact flag (NX) are not set.

jserv:
> I noticed that in the expected output_signature of test `fadd_b1-01`, one of the results is ff800000, which should be -INF. However, the overflow flag (OF) and inexact flag (NX) are not set.

You can submit issue(s) about floating-point compliance tests if you can recognize.

2011eric:
> > I noticed that in the expected output_signature of test `fadd_b1-01`, one of the results is ff800000, which should be -INF. However, the overflow flag (OF) and inexact flag (NX) are not set.
> 
> You can submit issue(s) about floating-point compliance tests if you can recognize.

I'm not sure if I can state the problem.
In some cases, the result of the operation differs in LSB. And I don't know if it is related to rounding.
In other cases, our simulator reports zero while the golden data is 'subnormal 0' .
And in some cases, the result of FADD is correct, but the flag is set incorectly.

emulate.c
```diff
@@ -651,7 +653,7 @@ static uint32_t csr_csrrs(struct riscv_t *rv, uint32_t csr, uint32_t val)
     if (!c)
         return 0;
 
-    const uint32_t out = *c;
+    const uint32_t out = (csr == CSR_FFLAGS) ? *c & FFLAG_MASK : *c;
```
jserv: Rewrite as following:
```c
uint32_t out = *c;
if (csr == CSR_FFLAGS)
    out &= FFLAG_MASK;
```

2011eric: I rewrote the function and also add the same code in `csr_csrrw` and `csr_csrrc`.

Since FCSR is defined only if ENABLE_RV32F is defined:
```
static uint32_t *csr_get_ptr(struct riscv_t *rv, uint32_t csr)
{
    switch (csr) {
    case CSR_CYCLE:
        return (uint32_t *) (&rv->csr_cycle) + 0;
    case CSR_CYCLEH:
        return (uint32_t *) (&rv->csr_cycle) + 1;
    case CSR_MSTATUS:
        return (uint32_t *) (&rv->csr_mstatus);
    case CSR_MTVEC:
        return (uint32_t *) (&rv->csr_mtvec);
    case CSR_MISA:
        return (uint32_t *) (&rv->csr_misa);
    case CSR_MSCRATCH:
        return (uint32_t *) (&rv->csr_mscratch);
    case CSR_MEPC:
        return (uint32_t *) (&rv->csr_mepc);
    case CSR_MCAUSE:
        return (uint32_t *) (&rv->csr_mcause);
    case CSR_MTVAL:
        return (uint32_t *) (&rv->csr_mtval);
    case CSR_MIP:
        return (uint32_t *) (&rv->csr_mip);
#ifdef ENABLE_RV32F
    case CSR_FFLAGS:
        return (uint32_t *) (&rv->csr_fcsr);
    case CSR_FCSR:
        return (uint32_t *) (&rv->csr_fcsr);
#endif
    default:
        return NULL;
    }
}
```

Should I also use macro inside `csr_csrrs` ?

jserv: > Should I also use macro inside `csr_csrrs` ?

You can do it in another pull request.

jserv (changes requested):
If `ENABLE_RV32F` is turned off, build system would complain:
```
emulate.c: In function ‘csr_csrrw’:
emulate.c:644:16: error: ‘FFLAG_MASK’ undeclared (first use in this function)
  644 |         out &= FFLAG_MASK;
      |                ^~~~~~~~~~
emulate.c:644:16: note: each undeclared identifier is reported only once for each function it appears in
emulate.c: In function ‘csr_csrrs’:
emulate.c:661:16: error: ‘FFLAG_MASK’ undeclared (first use in this function)
  661 |         out &= FFLAG_MASK;
      |                ^~~~~~~~~~
emulate.c: In function ‘csr_csrrc’:
emulate.c:678:16: error: ‘FFLAG_MASK’ undeclared (first use in this function)
  678 |         out &= FFLAG_MASK;
      |                ^~~~~~~~~~
```

Resolve the above compilation errors.

jserv:
gcc-9 raises the following warnings:
```
emulate.c:918:15: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
  918 |             *((uint32_t *) &rv->F[rd]) = RV_NAN;
      |              ~^~~~~~~~~~~~~~~~~~~~~~~~
emulate.c:930:15: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
  930 |             *((uint32_t *) &rv->F[rd]) = RV_NAN;
      |              ~^~~~~~~~~~~~~~~~~~~~~~~~
```

jserv:
Also, rebase the latest `master` branch.

PR #25, opened 2022-08-01
Fix uninitialized memory causing invalid access

RinHizakura:
It looks like some uninitialized memory problem leading me to fail on `make check` command(segmentation fault).

It may just work without the fix sometime. In fact, I couldn't reproduce the error after I reclone the project and run it again, but I believe these are the errors that should be fixed.

### DIFF
--- a/jit.c
+++ b/jit.c
@@ -895,8 +895,7 @@ struct block_map_t *block_map_alloc(uint32_t bits)
 {
     struct block_map_t *map = malloc(sizeof(struct block_map_t));
     const uint32_t capacity = (1 << bits);
-    void *ptr = malloc(capacity * sizeof(struct block_t *));
-    memset(ptr, 0, capacity);
+    void *ptr = calloc(capacity, sizeof(struct block_t *));
     map->bits = bits;
     map->map = (struct block_t **) ptr;
     map->capacity = capacity;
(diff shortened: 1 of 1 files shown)


jit.c
```diff
@@ -896,7 +896,7 @@ struct block_map_t *block_map_alloc(uint32_t bits)
     struct block_map_t *map = malloc(sizeof(struct block_map_t));
     const uint32_t capacity = (1 << bits);
     void *ptr = malloc(capacity * sizeof(struct block_t *));
-    memset(ptr, 0, capacity);
+    memset(ptr, 0, capacity * sizeof(struct block_t *));
```
jserv: Always use `calloc` instead of `malloc` + `memset`.

In most cases, they will be the same. In some cases, calloc() will do less work because it can skip memset() entirely. In other cases, calloc() can even cheat and not allocate any memory on modern operating systems.  However, the use of malloc() + memset() will always do the full amount of work.

RinHizakura:
There's the fix.

By the way, I am not sure if this is just an in-developing branch for repo owners to test some features only. It's fine to tell me if this is not the branch to contribute to.

jserv:
> There's the fix.
> 
> By the way, I am not sure if this is just an in-developing branch for repo owners to test some features only. It's fine to tell me if this is not the branch to contribute to.

Branch `wip/jit` is targeted in JIT compilation prototyping. And, welcome!

PR #26, opened 2022-08-02
Fix rv_step to correct access jit code block

RinHizakura:
There's some strange flow on the use of code block that I think they have to be fixed:

1. I think we should do the instruction count check before running the block's related function.
2.   `block_find_or_translate` will be called outside the while loop first, and called again when entering the loop. This will suppose to:
    * Having a duplicate call of `block_find_or_translate`  which will return the same block since the rv->PC won't change before executing any block
    * The block related to `rv->PC` is not run, instead, we will run the `block->predict`. I think this might cause some error although I don't have test pattern now to produce the corner case (the `if (block->predict && block->predict->pc_start == pc)` case)

### DIFF
--- a/emulate.c
+++ b/emulate.c
@@ -1593,8 +1593,7 @@ rv_func import_funcs[] = {
 void rv_step(struct riscv_t *rv, int32_t cycles)
 {
     /* find or translate a block for our starting PC */
-    struct block_t *block = block_find_or_translate(rv, NULL);
-    assert(block);
+    struct block_t *prev = NULL;
 
     const uint64_t cycles_start = rv->csr_cycle;
     const uint64_t cycles_target = rv->csr_cycle + cycles;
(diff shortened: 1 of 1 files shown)


LambertWSJ:
I have run your code with my toy profiler. both produce same blocks and execute same times
https://github.com/LambertWSJ/rv32emu/tree/profiler
if you want run both version rv_step please remove `qsort` in `profile_log` 
```shell
make ENABLE_PROFILE=1
./build/rv32emu ./build/hello.elf
```

>Having a duplicate call of block_find_or_translate which will return the same block since the rv->PC won't change before executing any block

you are right. it should write like your rv_step, avoid duplicate find same block at first time.

>The block related to rv->PC is not run, instead, we will run the block->predict. I think this might cause some error although I don't have test pattern now to produce the corner case (the if (block->predict && block->predict->pc_start == pc) case)

start_pc of predict is come from previous executed block, the end of insn in block is jalr or branch, so when execute block(whatever block or predict), pc will update to next block. it can reduce call `block_find_or_translate`

RinHizakura:
> start_pc of predict is come from previous executed block, the end of insn in block is jalr or branch, so when execute block(whatever block or predict), pc will update to next block. it can reduce call block_find_or_translate

Sure, I know what the `predict` is used for, but the problem I am wondering about before my fix is that the code may run like this:
1. We first do  `block_find_or_translate(rv, NULL)` and get a block out of the loop
2. Going into the loop for the first iteration, and let's say `if (prev && prev->predict && prev->predict->pc_start == pc)` is true. In the end, we will then execute the `block->predict` directly, which means the `block` we get before the loop is never executed

Although I don't have any test pattern to make the prediction become true in the first iteration currently, I think the logic doesn't make sense, so my PR actually fix this. Am I thinking this wrong?

PR #37, opened 2022-08-10
Fix infinite loop issue of block_find

LambertWSJ:
Since origin block map capacity is 512(2^9), but execute coremark need about 1200 buckets.
the buckets is cleared in block map when size is 80% of capacity. it cause block map cannot
find specific block.

repair some codegen format
    * jal/jalr: remove redundant rd
    * CSRRW: repair rd to format

### DIFF
--- a/jit.c
+++ b/jit.c
@@ -929,22 +929,25 @@ static void rv_jit_clear(struct riscv_t *rv)
     }
 }
 
+/* free block map but not clear all entries */
 static void block_map_free(struct block_map_t *map)
 {
     assert(map->map);
-    for (uint32_t i = 0; i < map->size; i++)
-        free(map->map[i]);
+    free(map->map);
+    map->map = NULL;
 }
 
-/* clear all entries in the block map
- * NOTE: will not clear the code buffer
- */
+/* clear all entries in the block map */
 static void block_map_clear(struct block_map_t *map)
 {
     assert(map->map);
-    for (uint32_t i = 0; i < map->size; i++) {
-        free(map->map[i]);
-        map->map[i] = NULL;
+    for (uint32_t i = 0; i < map->capacity; i++) {
+        struct block_t *block = map->map[i];
+        if (block) {
+            free(block->code);
+            free(block);
+            map->map[i] = NULL;
+        }
     }
     map->size = 0;
 }
@@ -982,12 +985,20 @@ static struct block_t *block_find(struct block_map_t *map, uint32_t addr)
     return NULL;
 }
 
+/* expand block_map capacity 2 times and reinsert entries */
 static void block_map_enlarge(struct riscv_jit_t *jit)
 {
-    struct block_map_t *map = jit->block_map;
-    map->bits++;
-    map->capacity = (1 << map->bits);
-    map = realloc(map, sizeof(struct block_t *) * map->capacity);
+    struct block_map_t *old_map = jit->block_map;
+    struct block_map_t *new_map = block_map_alloc(old_map->bits + 1);
+
+    for (size_t i = 0; i < old_map->capacity; i++) {
+        struct block_t *block = old_map->map[i];
+        if (block)
+            block_map_insert(new_map, block);
+    }
+
+    block_map_free(old_map);
+    jit->block_map = new_map;
 }
 
 static void rv_translate_block(struct riscv_t *rv, struct block_t *block)
@@ -1231,8 +1242,8 @@ struct riscv_jit_t *rv_jit_init(uint32_t bits)
 void rv_jit_free(struct riscv_jit_t *jit)
 {
     if (jit->block_map) {
+        block_map_clear(jit->block_map);
         block_map_free(jit->block_map);
-        jit->block_map->map = NULL;
     }
 
     MIR_context_t ctx = jit->ctx;


jit.c
```diff
@@ -932,8 +932,8 @@ static void rv_jit_clear(struct riscv_t *rv)
 static void block_map_free(struct block_map_t *map)
 {
     assert(map->map);
-    for (uint32_t i = 0; i < map->size; i++)
-        free(map->map[i]);
+    free(map->map);
```
jserv, line 936: Why did you make this change?

jit.c
```diff
@@ -729,7 +729,7 @@ static void emit_op_system(rv_buffer *buff, uint32_t insn, struct riscv_t *rv)
     case 1: { /* CSRRW    (Atomic Read/Write CSR) */
         COMMENT("CSRRW");
         CODE("tmp_u32 = csr_csrrw(rv, %d, rv->X[%u]);\n", csr, rs1);
-        CODE("rv->X[%u] = rd ? tmp_u32 : rv->X[%u];\n", rd, rd);
+        CODE("rv->X[%u] = %u ? tmp_u32 : rv->X[%u];\n", rd, rd, rd);
```
jserv: If rd equals to `rv_reg_zero` (i.e., `x0`), it is NOP. You should exclude such case. Send pull request(s) for RV32C and zero register handling later.

jserv (changes requested):
Split this pull request into two:
1. Fix unmatched parameter list in code generation (my bad. too many compilation warnings!)
2. Block specific fixes

You can send a pull request which fixes block_find with minimal changes first.

jit.c
LambertWSJ: Since block will be used in latter, so I just free map instead of block in map.
May be I should write another function for different case.

jserv: If you change the semantics of a certain function, you need explicit motivation and explanations.

LambertWSJ: `block_map_free` should free block map, if I want free blocks that I should call `block_map_clear`.

For this case `block_map_enlarge`, the blocks will be used in latter, so I translate blocks and free block map to avoid segement falut.

If I want free both safely,  `rv_jit_free` should write like follow.
```
void rv_jit_free(struct riscv_jit_t *jit)
{
    if (jit->block_map) {
        block_map_clear(jit->block_map);
        block_map_free(jit->block_map);
        jit->block_map->map = NULL;
    }
...
}
```

jserv: Write down the corresponding comments and check the usage to adopt.

jit.c
```diff
@@ -1231,6 +1239,7 @@ struct riscv_jit_t *rv_jit_init(uint32_t bits)
 void rv_jit_free(struct riscv_jit_t *jit)
 {
     if (jit->block_map) {
+        block_map_clear(jit->block_map);
         block_map_free(jit->block_map);
         jit->block_map->map = NULL;
```
RinHizakura: It looks like we had set  `jit->block_map->map` to `NULL` already in `block_map_free`, so the statement can be removed.

LambertWSJ: Okay

RinHizakura:
I am not sure about the sentence "The buckets is cleared in block map when size is 80% of capacity. It cause block map cannot
find specific block."

I assume that the block map is used for caching some translated code. If we can't find the corresponding block on the address, we just do the codegen again for it. Does this patch seem to aim to reserve all generated blocks?

LambertWSJ:
Because `block_map_clear` not clear all blocks. it may be miss some blocks which index is greater than size.

```c
static void block_map_clear(struct block_map_t *map)
{
    assert(map->map);
    for (uint32_t i = 0; i < map->size; i++) {
        free(map->map[i]);
        map->map[i] = NULL;
    }
    map->size = 0;
}
```

```
[****         **********     *****  *****  *******  ****|*******]
```
after `block_map_clear`, blocks map remain some blocks. it may clear some blocks which used in latter
```
[                                                       |*******]
```
Since not clear all block, it may cause to follow block map, that why `block_find` occur infinite loop
```
[**********************************************]
```
So, if don't want to save all blocks, `block_map_clear` should write like follow.

```c
static void block_map_clear(struct block_map_t *map)
{
    assert(map->map);
    for (uint32_t i = 0; i < map->capacity; i++) {
        struct block_t *block = map->map[i];
        if (block) {
            free(block->code);
            free(block);
            map->map[i] = NULL;
        }
    }
    map->size = 0;
}
```

Thank you, You ask a right question, I should think this point first.

RinHizakura:
Yes, I think `for (uint32_t i = 0; i < map->capacity; i++)` is the actually right way to clear all entities. That's the true reason for the bug you reported. By the way, we can maintain a bitmap for the used block, but it's fine to just simply iterate the whole map and skip the block with `NULL` now.

I don't think that's a good idea to enlarge without any limitations, although we may also need it under some scenarios. We can leave the function and use it for some strategy/heuristic to maintain the block map in the future.

RinHizakura (approved):
The remaining looks good to me. Thanks for the nice explanation of my question :)

jserv (changes requested):
Improve the writing of git commit messages. You can make use of [QuillBot](https://quillbot.com/) to fix grammar mistakes.

PR #43, opened 2022-08-21
Add system call set_relative_mode

alanjian85:
Add a new system call to allow users to control the mouse relative mode of the SDL window and to synchronize the cursor with the game demo menu's current state. This commit also updates the ELF executable files of Quake and Doom demo.

P.S. I also opened change requests against [doom_riscv](https://github.com/sysprog21/doom_riscv/pull/2) and [quake-embedded](https://github.com/sysprog21/quake-embedded/pull/4)


jserv:
Can we stick to the principles of minimalist regarding the design of system? That is, instead of providing new system call, how about extending the semantics of `poll_event` by checking a certain bit which indicates the switch of the mouse relative mode in SDL.

jserv (changes requested):
Don't put the ELF files in this pull request since we should concentrate on the design of system calls. Also, keep the file permissions as it ought to be.

PR #46, opened 2022-09-02
Manage breakpoints with red-black tree

RinHizakura:
This patch enables `rv32emu` to set more than one breakpoint (should be infinite if we assume our memory is always enough to
create a new tree node).

Note that it's still bad to query the PC address in every instruction step. This patch only aims to fix the limitation of the breakpoint number in a short time. We may make use of the emulator's trap mechanism to avoid frequently querying the RB tree
1. Replace the instruction of the breakpoint's address with a breakpoint instruction
2. Query the RB tree when the emulator hit the trap when executing such instruction

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ GDBSTUB_LIB := $(GDBSTUB_OUT)/libgdbstub.a
 $(GDBSTUB_LIB): mini-gdbstub/Makefile
 	$(MAKE) -C $(dir $<) O=$(dir $@)
 $(OUT)/emulate.o: $(GDBSTUB_LIB)
-OBJS_EXT += gdbstub.o
+OBJS_EXT += gdbstub.o breakpoint.o
 CFLAGS += -D ENABLE_GDBSTUB -D'GDBSTUB_COMM="$(GDBSTUB_COMM)"'
 LDFLAGS += $(GDBSTUB_LIB)
 gdbstub-test: $(BIN)
--- a/breakpoint.c
+++ b/breakpoint.c
@@ -1,0 +1,60 @@
+#include "breakpoint.h"
+#include <assert.h>
+
+static inline int cmp(const void *arg0, const void *arg1)
+{
+    riscv_word_t *a = (riscv_word_t *) arg0, *b = (riscv_word_t *) arg1;
+    return (*a < *b) ? _CMP_LESS : (*a > *b) ? _CMP_GREATER : _CMP_EQUAL;
+}
+
+breakpoint_map_t breakpoint_map_new()
+{
+    return map_init(riscv_word_t, breakpoint_t, cmp);
+}
+
+bool breakpoint_map_insert(breakpoint_map_t map, riscv_word_t addr)
+{
+    breakpoint_t bp = (breakpoint_t){.addr = addr, .orig_insn = 0};
+    map_iter_t it;
+    map_find(map, &it, &addr);
+    /* We don't expect to set breakpoint at duplicate address */
+    if (!map_at_end(map, &it))
+        return false;
+    return map_insert(map, &addr, &bp);
+}
+
+static bool breakpoint_map_find_it(breakpoint_map_t map,
+                                   riscv_word_t addr,
+                                   map_iter_t *it)
+{
+    map_find(map, it, &addr);
+    if (map_at_end(map, it)) {
+        return false;
+    }
+
+    return true;
+}
+
+breakpoint_t *breakpoint_map_find(breakpoint_map_t map, riscv_word_t addr)
+{
+    map_iter_t it;
+    if (!breakpoint_map_find_it(map, addr, &it))
+        return NULL;
+
+    return (breakpoint_t *) map_iter_value(&it, breakpoint_t *);
+}
+
+bool breakpoint_map_del(breakpoint_map_t map, riscv_word_t addr)
+{
+    map_iter_t it;
+    if (!breakpoint_map_find_it(map, addr, &it))
+        return false;
+
+    map_erase(map, &it);
+    return true;
+}
+
+void breakpoint_map_destroy(breakpoint_map_t map)
+{
+    map_delete(map);
+}
--- a/breakpoint.h
+++ b/breakpoint.h
@@ -1,0 +1,17 @@
+#pragma once
+
+#include "map.h"
+#include "riscv.h"
+
+typedef struct {
+    riscv_word_t addr;
+    uint32_t orig_insn;
+} breakpoint_t;
+
+typedef map_t breakpoint_map_t;
+
+breakpoint_map_t breakpoint_map_new();
+bool breakpoint_map_insert(breakpoint_map_t map, riscv_word_t addr);
+breakpoint_t *breakpoint_map_find(breakpoint_map_t map, riscv_word_t addr);
+bool breakpoint_map_del(breakpoint_map_t map, riscv_word_t addr);
+void breakpoint_map_destroy(breakpoint_map_t map);
--- a/tests/gdbstub.sh
+++ b/tests/gdbstub.sh
@@ -7,24 +7,36 @@ PID=$!
 if ps -p $PID > /dev/null
 then
     tmpfile=/tmp/rv32emu-gdbstub.$PID
-    riscv32-unknown-elf-gdb --batch \
-        -ex "file build/puzzle.elf" \
-        -ex "target remote :1234" \
-        -ex "break *0x10700" \
-        -ex "continue" \
-        -ex "print \$pc" \
-        -ex "del 1" \
-        -ex "stepi" \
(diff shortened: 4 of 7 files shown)


Makefile
```diff
@@ -66,6 +66,7 @@ $(GDBSTUB_LIB): mini-gdbstub/Makefile
 	$(MAKE) -C $(dir $<) O=$(dir $@)
 $(OUT)/emulate.o: $(GDBSTUB_LIB)
 OBJS_EXT += gdbstub.o
+OBJS_EXT += brkp.o
```
jserv: Simply name it `breakpoint.o` instead of the abbreviated form.

brkp.c
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    return true;
+}
+
+brkp_map_t *brkp_map_find(brkp_map_t *brkp_map, riscv_word_t addr)
+{
+    assert(brkp_map);
+    map_iter_t it;
+    if(!brkp_map_find_it(brkp_map, addr, &it))
```
jserv: Indent.

brkp.h
```diff
@@ -0,0 +1,19 @@
...
+#include "riscv.h"
+
+typedef struct {
+    riscv_word_t addr;
+    uint32_t orig_instr;
+} brkp_t;
+
+typedef struct {
+    map_t map;
+} brkp_map_t;
```
jserv: Why not declaring `typedef map_t brkp_map_t;` straightforwardly?

brkp.c
```diff
@@ -0,0 +1,68 @@
+#include "brkp.h"
```
jserv: Let's use `breakpoint` instead. Since `brk` is used as a system call, someone might confuse between `brk` and `brkp`. The later was intentionally meant to breakpoint though.

brkp.c
```diff
@@ -0,0 +1,68 @@
...
+#include <assert.h>
+
+static inline int cmp(const void *arg0, const void *arg1)
+{
+    riscv_word_t *a = (riscv_word_t *) arg0, *b = (riscv_word_t *) arg1;
+    return (*a < *b) ? _CMP_LESS : (*a > *b) ? _CMP_GREATER : _CMP_EQUAL;
+}
+
+void brkp_map_init(brkp_map_t *brkp_map)
```
jserv: Ditto. I expect below:
```c
void breakpoint_map_init(breakpoint_map_t *map)
```

brkp.h
RinHizakura: I am not sure if we will extend this structure. I may want to record something like the reference to the previous hit breakpoint in the future.

breakpoint.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+#include "map.h"
+#include "riscv.h"
+
+typedef struct {
+    riscv_word_t addr;
+    uint32_t orig_instr;
```
jserv: Replace "orig_instr" with "orig_insn" for consistency.

breakpoint.c
```diff
@@ -0,0 +1,68 @@
...
+
+static inline int cmp(const void *arg0, const void *arg1)
+{
+    riscv_word_t *a = (riscv_word_t *) arg0, *b = (riscv_word_t *) arg1;
+    return (*a < *b) ? _CMP_LESS : (*a > *b) ? _CMP_GREATER : _CMP_EQUAL;
+}
+
+void breakpoint_map_init(breakpoint_map_t *breakpoint_map)
+{
+    assert(breakpoint_map);
```
jserv: Specify the name of parameter as "map" for short.

breakpoint.c
```diff
@@ -0,0 +1,68 @@
...
+
+    breakpoint_map->map = map_init(riscv_word_t, breakpoint_t, cmp);
```
jserv: If you simply wrap with `typedef`, you can avoid the indirection (`->map`).

breakpoint.c
```diff
@@ -0,0 +1,68 @@
...
+}
+
+bool breakpoint_map_insert(breakpoint_map_t *breakpoint_map, riscv_word_t addr)
+{
+    assert(breakpoint_map);
+
+    breakpoint_t breakpoint = (breakpoint_t) {.addr = addr, .orig_instr = 0};
```
jserv: Simply name the variable in shorter fashion. i.e., something like `bp` or `bkpt`.

tests/gdbstub.sh
```diff
@@ -11,20 +11,31 @@ then
...
     # check if we stop at the breakpoint
-    expected=$(grep -rw "Breakpoint 1 at" ${tmpfile} | awk {'print $4'})
-    ans=$(grep -r "$1 =" ${tmpfile} | awk {'print $5'})
-    if [ "$expected" != "$ans" ]; then
-        # Fail
-        exit 1
-    fi
+    for i in {1..3}
+    do
+        expected=$(grep -rw "Breakpoint ${i} at" ${tmpfile} | awk {'print $4'})
```
jserv, line 33: Bash array is powerful. Check it out: https://opensource.com/article/18/5/you-dont-know-bash-intro-bash-arrays
You can utilize this feature to shorten the above.

Makefile
```diff
@@ -66,6 +66,7 @@ $(GDBSTUB_LIB): mini-gdbstub/Makefile
 	$(MAKE) -C $(dir $<) O=$(dir $@)
 $(OUT)/emulate.o: $(GDBSTUB_LIB)
 OBJS_EXT += gdbstub.o
+OBJS_EXT += breakpoint.o
```
jserv: Merge into a single line. i.e., `OBJS_EXT += gdbstub.o breakpoint.o`

breakpoint.h
```diff
@@ -0,0 +1,20 @@
...
+} breakpoint_t;
+
+typedef struct {
+    map_t map;
+} breakpoint_map_t;
+
+void breakpoint_map_init(breakpoint_map_t *breakpoint_map);
```
jserv: How about refining as `breakpoint_map_new`? It would return the allocated object for breakpoint map rather than explicit pre-allocated one.

tests/gdbstub.sh
```diff
@@ -11,20 +11,31 @@ then
         -ex "file build/puzzle.elf" \
         -ex "target remote :1234" \
         -ex "break *0x10700" \
```
jserv: Once you utilize bash array, the break commands can be generated accordingly.

RinHizakura: But the repeat pattern only exists for a part of command arguments. I am not sure how could I use the bash array on this.

tests/gdbstub.sh
RinHizakura: Do you mean using bash array with element `(1 2 3)` here? Otherwise, I don't find any better place to make use of it.

tests/gdbstub.sh
jserv: Check this: https://gist.github.com/mateuszalmannai/8bc8cff1e84318d82c11004f86057966
You can iterative each element in a bash array (index -> breakpoint address) and generate the GDB commands associated with breakpoint addresses.

breakpoint.c
```diff
@@ -0,0 +1,60 @@
+#include "breakpoint.h"
+#include <assert.h>
+
+static inline int cmp(const void *arg0, const void *arg1)
+{
+    riscv_word_t *a = (riscv_word_t *) arg0, *b = (riscv_word_t *) arg1;
+    return (*a < *b) ? _CMP_LESS : (*a > *b) ? _CMP_GREATER : _CMP_EQUAL;
+}
+
+breakpoint_map_t breakpoint_map_init()
```
jserv: Here, you shall name it `breakpoint_map_new` instead of `breakpoint_map_init`. Quote from [Constructors in Python (`__init` vs `__new__`)](https://dev.to/pila/constructors-in-python-init-vs-new-2f9j)
> `__new__` is the first step of instance creation. It's called first and is responsible for returning a new instance of your class.
> In contrast, `__init__` doesn't return anything; it's only responsible for initializing the instance after it's been created. 

The above was dedicated to Python, but name scheme is still making sense to C APIs.

tests/gdbstub.sh
RinHizakura: I am still confused about this. Since it looks like we should run the  `riscv32-unknown-elf-gdb` command once for the test, and the part we may want to loop on is only a few of the `-ex` arguments. However we can't do something like this:

```
riscv32-unknown-elf-gdb --batch \
               for t in ${arr[@]}; do
                   -ex "break $t"  \
               done
               ...
```

jserv: Alternatively, you can simply operate as following
```shell
riscv32-unknown-elf-gdb --batch $(GDB_COMMANDS)
```

Here, `$(GDB_COMMANDS)` consists of generated sequences.

PR #51, opened 2022-09-08
Improve fmin/fmax

2011eric:
Passed 2 more tests in riscv-arch-test.
Support of fmin and fmax is complete.

### DIFF
--- a/src/riscv_private.h
+++ b/src/riscv_private.h
@@ -302,7 +302,8 @@ static inline uint32_t sign_extend_b(const uint32_t x)
 
 #ifdef ENABLE_RV32F
 /* compute the fclass result */
-static inline uint32_t calc_fclass(uint32_t f) {
+static inline uint32_t calc_fclass(uint32_t f) 
+{
   const uint32_t sign = f & FMASK_SIGN;
   const uint32_t expn = f & FMASK_EXPN;
   const uint32_t frac = f & FMASK_FRAC;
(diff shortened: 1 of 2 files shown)


riscv_private.h
```diff
@@ -333,6 +333,17 @@ static inline uint32_t calc_fclass(uint32_t f) {
 
   return out;
 }
+static inline bool is_nan(uint32_t f){
+  const uint32_t expn = f & FMASK_EXPN;
+  const uint32_t frac = f & FMASK_FRAC;
+  return (expn == FMASK_EXPN && frac);
+}
+
+static inline bool is_signan(uint32_t f){
```
jserv: In [Floating-point exceptions](https://www.ibm.com/docs/en/aix/7.3?topic=concepts-floating-point-exceptions) of IBM AIX Manual, signaling NaN is abbreviated as `snan`, which might represent better than "signan." In Linux, the prefix "sig" is used for UNIX signal specific identifiers. However, signaling NaN is quite different from them. Let's name it `is_snan` here.

riscv_private.h
```diff
@@ -333,6 +333,17 @@ static inline uint32_t calc_fclass(uint32_t f) {
 
   return out;
 }
+static inline bool is_nan(uint32_t f){
```
jserv: Run `clang-format-12 -i *.[ch]` to indent.

jserv (changes requested):
Rebase the latest "master" branch which checks C/C++ source files for consistent style.

jserv (changes requested):
You should explain the IEEE754-201x semantics of fmin/fmax in git commit messages.

2011eric:
I think I can pack the code into a function, since the code of fmin and the one of fmax is similar.

jserv:
> I think I can pack the code into a function, since the code of fmin and the one of fmax is similar.

You can introduce some macros as what [SoftFP](https://bellard.org/softfp/) does.

jserv:
Fix the conflicts caused by recent directory structure changes.

src/riscv_private.h
```diff
@@ -333,6 +333,17 @@ static inline uint32_t calc_fclass(uint32_t f) {
 
   return out;
 }
+static inline bool is_nan(uint32_t f){
```
jserv: Indent this file as well.

PR #56, opened 2022-09-17
gdbstub: Implement memory write and register write

RinHizakura:
The `mini-gdbstub` library supports the functionality to write register and write memory now. After this PR, `rv32emu` is enabled to do those commands that will try to write register and memory, e.g. the `load` command.

### DIFF
--- a/src/gdbstub.c
+++ b/src/gdbstub.c
@@ -18,12 +18,32 @@ static size_t rv_read_reg(void *args, int regno)
     return -1;
 }
 
+static void rv_write_reg(void *args, int regno, size_t data)
+{
+    if (unlikely(regno > 32))
+        return;
+
+    struct riscv_t *rv = (struct riscv_t *) args;
+    if (regno == 32)
+        rv_set_pc(rv, data);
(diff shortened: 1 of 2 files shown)


jserv (changes requested):
Update `README.md` accordingly.

src/gdbstub.c
```diff
@@ -18,6 +18,16 @@ static size_t rv_read_reg(void *args, int regno)
     return -1;
 }
 
+static void rv_write_reg(void *args, int regno, size_t data)
+{
+    struct riscv_t *rv = (struct riscv_t *) args;
```
jserv, line 26: Rewrite as following:
```c
if (unlikely(regno > 32))
    return;

struct riscv_t *rv = (struct riscv_t *) args;
if (regno == 32)
    rv_set_pc(rv, data);
else
    rv_set_reg(rv, regno, data);
```

RinHizakura:
> Update `README.md` accordingly.

The part of limitation is removed now since all of them are resolved.

PR #58, opened 2022-09-26
Lower branches in red-black tree

steven1lung:
The changes in this commit are aiming to remove if statements to achieve better performance by producing less branches.

Indirect pointers are used to store the pointer pointing to the left or right child node. So we would not need to check whether it is a right or left node.

### DIFF
--- a/src/map.c
+++ b/src/map.c
@@ -81,9 +81,7 @@ static inline void rb_set_parent_color(map_node_t *node,
  */
 static inline bool rb_is_red(map_node_t *node)
 {
-    if (node && (rb_color(node) == RB_RED))
-        return true;
-    return false;
+    return node && (rb_color(node) == RB_RED);
 }
 
 /* Create a node to be attached in the map internal tree structure */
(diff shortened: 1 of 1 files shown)


src/map.c
```diff
@@ -524,34 +522,22 @@ bool map_insert(map_t obj, void *key, void *value)
     }
 
     /* Traverse the tree until we hit the end or find a side that is NULL */
-    map_node_t *cur = obj->head;
+    map_node_t **indirect = &obj->head;
 
-    while (1) {
-        int res = obj->comparator(new_node->key, cur->key);
+    while (*indirect) {
```
jserv, line 527: Rewrite as following:
```c
    while (*indirect) {
        int res = obj->comparator(new_node->key, (*indirect)->key);
        if (res == 0) { /* If the key matches something, don't insert */
            map_delete_node(obj, new_node);
            return false;
        }
        indirect = res < 0 ? &(*indirect)->left : &(*indirect)->right;
    }
```

steven1lung:
Code updated, please check.

jserv:
Thank @steven1lung for contributing!

PR #59, opened 2022-09-27
Lower branches in red-black tree again

steven1lung:
This commit is similar with the one applied in insert_node, an indirect pointer can be used instead of checking left or right child nodes.

### DIFF
--- a/src/map.c
+++ b/src/map.c
@@ -580,38 +580,26 @@ void map_find(map_t obj, map_iter_t *it, void *key)
     }
 
     /* Basically a repeat of insert */
-    map_node_t *cur = obj->head;
+    map_node_t **indirect = &obj->head;
 
     /* binary search */
-    while (1) {
-        int res = obj->comparator(key, cur->key);
-        if (res == 0) /* If the key matches, we hit the target */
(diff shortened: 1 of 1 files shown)


src/map.c
```diff
@@ -580,38 +580,26 @@ void map_find(map_t obj, map_iter_t *it, void *key)
...
-                break;
-            }
-            cur = cur->right;
-        }
+        indirect = res < 0 ? &(*indirect)->left : &(*indirect)->right;
     }
 
-    if (cur) {
-        it->node = cur;
+    if (*indirect) {
```
jserv: Rewrite as following:
```c
    if (!*indirect) {
        it->node = NULL;
        return;
    }
    it->node = *indirect;

    /* Generate a "prev" as well */
    map_iter_t tmp = *it;
    map_prev(obj, &tmp);
    it->prev = tmp.node;
```

It results in slightly shorter code and less indentions.

jserv: Check https://github.com/sysprog21/rv32emu/blob/master/CONTRIBUTING.md (Avoid redundant else)

PR #62, opened 2022-09-29
Apply bidirectional queue design to the input system

alanjian85:
This commit splits the input system into two parts, event and submission, and introduces two system calls, setup_queue and submit, the former is used to assist in the creation of queues, which is part of the communication bridge between the user application code and the emulator, and the latter is used to notify the emulator that a submission (or a command) should be processed. The design of these system calls is heavily based on Linux io_uring interface.

### DIFF
--- a/docs/syscall.md
+++ b/docs/syscall.md
@@ -27,15 +27,32 @@ These system calls are solely for the convenience of accessing the [SDL library]
 
 If a window does not already exist, one will be created with the specified `width` and `height`. The `screen` buffer will replace the content of the framebuffer, passing a different `width` or `height` compared to the size of the window is undefined behavior. This system call additionally polls events from the SDL library, and, if necessary, update the internal input specific event queue.
 
-### `poll_event(event)` - Poll a input specific event from SDL
+### `setup_queue` - Setup input system's dedicated event and submission queue
 
 **system call number**: `0xC0DE`
 
-**synopsis**: `int poll_event(void *event)`
+**synopsis**: `void *setup_queue(int capacity, unsigned int* event_count)`
 
-`event` will be filled with the polled event data, it should have a 32-bit `type` field, and associated with an appropriately sized value buffer. The internal event queue will be updated everytime `draw_frame` is called.
+Allocate a continuous memory chunk that has the requested capacity and two closely packed queues, the event queue and the submission queue. The base address of the event queue is located in the returned address, and the submission queue is immediately after the event queue's last element, which is the event queue's base address plus the size of each event element times the given capacity. The capacity must be a power of 2; if it is not, it will be rounded up to the next highest power of 2. It is crucial to initialize the event counter variable before supplying its address to this system call because it serves as an interaction to the user that an event has been added to the event queue.
 
-Currently accepted event types include the following:
-* `KEY_EVENT` - `0x0`: Triggered when the status of keyboard changes, either when a key is pressed or released, and it returns a 32-bit keycode and a 8-bit key state, the values of the hexadecimal keycodes are listed in [SDL Keycode Lookup Table](https://wiki.libsdl.org/SDLKeycodeLookup)
-* `MOUSE_MOTION_EVENT` - `0x1`: A mouse move event, with relative position information, two 32-bit signed integers that correspond to the x and y delta value, the mouse is continually wrapped in the window border by default.
-* `MOUSE_BUTTON_EVENT` - `0x2`: the user code receives this event whenever the state of a mouse button changes, whether a button is pressed or released, and it returns a 8-bit value that indicates which button is updated(1 is left, 2 is middle, 3 is right and so on), as well as a 8-bit state value that indicates whether the button is pressed.
+#### Events
+
+An event entry is made up of a 32-bit value representing the event's type and a `union` buffer containing t1he event's parameters.
+
+* `KEY_EVENT`: Either a key is pressed or released. Its value buffer is made up of a 32-bit universal key code and an 8-bit state flag; if the corresponding character of the pressed key is not printable, the bit right after the most significant bit is set; for example, the "a" key's correspoding character is printable, so its keycode is the ASCII code of the "a" character, which is `0x61`. However, because the left shift key doesn't have a corresponding printable key, its hexadecimal value is `0x400000E1`, with the 31 bit is set.
+* `MOUSE_MOTION_EVENT`: The cursor is moved during the current frame. This event contains two signed integer value, which is the delta of x position and y position respectively. If the relative mouse mode is enabled, the mouse movement will never be 0 because the cursor is wrapped within the canvas and is repeated whenever the cursor reaches the border. 
+* `MOUSE_BUTTON_EVENT`: The state of a mouse button has been changed. Its value buffer contains a 8-bit button value(1 is left, 2 is middle, 3 is right and so on) and an 8-bit boolean flag that indicates whether the mouse button is pressed.
+
+### `submit_queue` - Notify the emulator a submission has been pushed into the submission queue
+
+**system call number**: `0xFEED`
+
+**synopsis**: `void submit_queue(int count)`
+
+To inform the emulator that a batch of submissions should be processed, the application code should push several submissions into the queue first, and then pass the size of the submissions batch to this system call; the submissions will be processed and executed sequentially and immediately.
+
+#### Submissions
+
+The submission entry is structured similarly to an event entry, with a 32-bit type field and an associated dynamic-sized value buffer whose width depends on the type of submission.
+
+* `RELATIVE_MODE_SUBMISSION`: Enable or disable the mouse relative mode. If the mouse relative mode is enabled, the mouse cursor is wrapped within the window border, it's associated with an 8-bit wide boolean value that indicates whether the relative mouse mode should be enbled.
--- a/src/syscall_sdl.c
+++ b/src/syscall_sdl.c
@@ -9,12 +9,6 @@
 
 #include "state.h"
 
-/* For optimization, the capcity of event queues must be the power of two to
- * avoid the expensive modulo operation, the details are explained here:
- * https://stackoverflow.com/questions/10527581/why-must-a-ring-buffer-size-be-a-power-of-2
- */
-#define EVENT_QUEUE_CAPACITY 128
-
 enum {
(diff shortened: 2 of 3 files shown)


src/syscall_sdl.c
```diff
@@ -193,34 +214,39 @@ void syscall_draw_frame(struct riscv_t *rv)
...
-    case MOUSE_MOTION_EVENT:
-        memory_write(s->mem, base + 4, (const uint8_t *) &event.mouse.motion,
-                     sizeof(mouse_motion_t));
-        break;
-    case MOUSE_BUTTON_EVENT:
-        memory_write(s->mem, base + 4, (const uint8_t *) &event.mouse.button,
-                     sizeof(mouse_button_t));
-        break;
-    }
+    void *base = malloc(sizeof(event_t) * queues_capacity +
```
jserv, line 244: Use `calloc` instead. It would result in shorter and slightly understandable code.

jserv: By the way, is there any chance to release the allocated memory `base`?

jserv (changes requested):
Check memory allocation.

docs/syscall.md
```diff
@@ -27,15 +27,32 @@ These system calls are solely for the convenience of accessing the [SDL library]
...
-### `poll_event(event)` - Poll a input specific event from SDL
+### `setup_queue` - Setup input system's dedicated event and submission queue
 
 **system call number**: `0xC0DE`
 
-**synopsis**: `int poll_event(void *event)`
+**synopsis**: `void *setup_queue(int capacity, unsigned int* event_count)`
 
-`event` will be filled with the polled event data, it should have a 32-bit `type` field, and associated with an appropriately sized value buffer. The internal event queue will be updated everytime `draw_frame` is called.
+Allocate a continuous memory chunk that contains two tightly-packed queues, event queue and submission queue, with the specified capacity, the base address of the former queue is reside in the returned address, and the latter is followed right after the event queue's last element, which is the event queue's base address plus the size of each event element times the given capacity. The provided event counter variable is used to inform the user that an event is pushed to the event queue, its value will be increased whenever there is a new event, so initializing it before passing its address to this system call is critical.
```
jserv: Rephrase as following:
> Allocate a continuous memory chunk that has the requested capacity and two closely packed queues, the event queue and the submission queue. The base address of the event queue is located in the returned address, and the submission queue is immediately after the event queue's last element.

src/syscall_sdl.c
```diff
@@ -47,38 +41,65 @@ typedef struct {
...
+static submission_t submission_pop(void)
 {
-    if (event_queue.start == event_queue.end)
-        return false;
-    *event = event_queue.events[event_queue.start++];
-    event_queue.start &= EVENT_QUEUE_CAPACITY - 1;
-    event_queue.full = false;
-    return true;
+    submission_t submission = submission_queue.base[submission_queue.start++];
+    if (queues_capacity & 1)
```
jserv: Can you simply guarantee that queue capacity is always power of 2?

jserv: Therefore, we can write down efficient and short code.

src/syscall_sdl.c
alanjian85: How can I make sure there won't be a space between the addresses allocated by two calloc operations, or how can I allocate both queues in a single calloc call without specifying the size of char to the size argument?

jserv: > How can I make sure there won't be a space between the addresses allocated by two calloc operations, or how can I allocate both queues in a single calloc call without specifying the size of char to the size argument?

You can utilize [atexit](https://man7.org/linux/man-pages/man3/atexit.3.html) to finalize the memory allocation.

src/syscall_sdl.c
alanjian85: Should we simply require the user to select a size that is a power of 2 and notify them when they don't; or should we just align the size up to a power of 2 while allocating and let the user push and pop with different sizes?

src/syscall_sdl.c
jserv: Memory de-allocation should take place when `SDL_Quit` is about to be invoked.

docs/syscall.md
```diff
@@ -27,15 +27,32 @@ These system calls are solely for the convenience of accessing the [SDL library]
...
-### `poll_event(event)` - Poll a input specific event from SDL
+### `setup_queue` - Setup input system's dedicated event and submission queue
 
 **system call number**: `0xC0DE`
 
-**synopsis**: `int poll_event(void *event)`
+**synopsis**: `void *setup_queue(int capacity, unsigned int* event_count)`
 
-`event` will be filled with the polled event data, it should have a 32-bit `type` field, and associated with an appropriately sized value buffer. The internal event queue will be updated everytime `draw_frame` is called.
+Allocate a continuous memory chunk that has the requested capacity and two closely packed queues, the event queue and the submission queue. The base address of the event queue is located in the returned address, and the submission queue is immediately after the event queue's last element, which is the event queue's base address plus the size of each event element times the given capacity. The provided event counter variable is used to inform the user that an event is pushed to the event queue, its value will be increased whenever there is a new event, so initializing it before passing its address to this system call is critical.
```
jserv: Rephrase as following:
> It is crucial to initialize the event counter variable before supplying its address to this system call because it serves as an interaction to the user that an event has been added to the event queue.

src/syscall_sdl.c
jserv: I prefer the way to align.

src/syscall_sdl.c
alanjian85: > By the way, is there any chance to release the allocated memory `base`?

The implementation is not completed yet, This main memory allocation will be replaced with an allocation in the emulator's address space, making it easy for users to release it using `free()`.

alanjian85: > By the way, is there any chance to release the allocated memory `base`?

The implementation is not yet complete. To make the returned address be accessible by the user, this main memory allocation will be replaced with an allocation in the emulator's address space.

jserv: Add appropriate `FIXME` or `TODO` comments.

src/syscall_sdl.c
```diff
@@ -193,34 +208,48 @@ void syscall_draw_frame(struct riscv_t *rv)
...
-                     sizeof(mouse_motion_t));
-        break;
-    case MOUSE_BUTTON_EVENT:
-        memory_write(s->mem, base + 4, (const uint8_t *) &event.mouse.button,
-                     sizeof(mouse_button_t));
-        break;
-    }
-
-    rv_set_reg(rv, rv_reg_a0, 1);
+    /* Bit Twiddling Hacks: Round up to the next highest power of 2 */
```
jserv: Move the below implementation to a dedicated helper function. In addition, you can use `__builtin_clz` for the fastpath (clang/gcc) and bitwise hacks as fallback. See https://jameshfisher.com/2018/03/30/round-up-power-2/

src/syscall_sdl.c
```diff
@@ -193,34 +208,48 @@ void syscall_draw_frame(struct riscv_t *rv)
...
+
+    rv_set_reg(
+        rv, rv_reg_a0,
+        (uint32_t) (uintptr_t) base); /* eliminate the "cast from pointer to
+                                         integer of different size" warning*/
 }
+
+void syscall_submit_queue(struct riscv_t *rv)
+{
+    /* submit(count) */
```
jserv: Should be `submit_queue`.

src/syscall_sdl.c
```diff
@@ -47,38 +41,75 @@ typedef struct {
...
+    state_t *s = rv_userdata(rv);
+    uint32_t count;
+    memory_read(s->mem, (void *) &count, event_count, sizeof(uint32_t));
+    count += 1;
+    memory_write(s->mem, event_count, (void *) &count, sizeof(uint32_t));
+}
+
+static uint32_t round_pow2(uint32_t arg)
+{
+#ifdef __GNUC__ || __clang__
```
jserv: Rewrite as
```c
```

Otherwise, it will not behave consistently.

src/syscall_sdl.c
```diff
@@ -47,38 +41,75 @@ typedef struct {
...
+    event_queue.end &= queues_capacity - 1;
+
+    state_t *s = rv_userdata(rv);
+    uint32_t count;
+    memory_read(s->mem, (void *) &count, event_count, sizeof(uint32_t));
+    count += 1;
+    memory_write(s->mem, event_count, (void *) &count, sizeof(uint32_t));
+}
+
+static uint32_t round_pow2(uint32_t arg)
```
jserv: Simply use `x` or something similar in parameter list.

src/syscall_sdl.c
```diff
@@ -47,38 +41,75 @@ typedef struct {
...
+{
+#ifdef __GNUC__ || __clang__
+    arg = 1 << (32 - __builtin_clz(arg - 1));
+#elif
+    arg--;
```
jserv: Mention [Bit Twiddling Hacks](https://graphics.stanford.edu/~seander/bithacks.html) for generic implementation.

src/syscall_sdl.c
```diff
@@ -47,38 +41,75 @@ typedef struct {
...
+    memory_read(s->mem, (void *) &count, event_count, sizeof(uint32_t));
+    count += 1;
+    memory_write(s->mem, event_count, (void *) &count, sizeof(uint32_t));
+}
+
+static uint32_t round_pow2(uint32_t arg)
+{
+#if defined __GNUC__ || defined __clang__
+    arg = 1 << (32 - __builtin_clz(arg - 1));
+#elif
```
jserv: Should be `#else` here.

src/syscall_sdl.c
```diff
@@ -47,38 +41,75 @@ typedef struct {
...
+
+typedef struct {
+    submission_t *base;
+    size_t start;
+} submission_queue_t;
+
 static SDL_Window *window = NULL;
 static SDL_Renderer *renderer;
 static SDL_Texture *texture;
+static uint32_t queues_capacity = 0;
```
jserv: Add a blank line to separate SDL related variables and event queue specific ones. In addition, comments should appear as well.

jserv:
Thank @alanjian85 for contributing!

PR #63, opened 2022-10-02
Avoid duplications in RISC-V exception handlers

### LINKED ISSUE #61: Avoid duplications in RISC-V exception handlers
Code duplication in the function body of the RISC-V exception handlers might be a maintenance headache. That is, we shall refine these functions in `src/emulate.c`:
* `rv_except_insn_misaligned`
* `rv_except_load_misaligned`
* `rv_except_store_misaligned`
* `rv_except_illegal_insn`
* `rv_except_breakpoint` proposed in #60

Code generation using preprocessor macros may be an approach to avoid such duplications.

Risheng1128:
Use preprocessor to avoid duplications in exception handlers, 
and create a list to define the RISC-V exception code.

Resolve #61

### DIFF
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -24,160 +24,56 @@ extern struct target_ops gdbstub_ops;
 #include "riscv.h"
 #include "riscv_private.h"
 
-static void rv_except_insn_misaligned(struct riscv_t *rv, uint32_t old_pc)
-{
-    /* mtvec (Machine Trap-Vector Base Address Register)
-     * mtvec[MXLEN-1:2]: vector base address
-     * mtvec[1:0] : vector mode
-     */
-    const uint32_t base = rv->csr_mtvec & ~0x3;
-    const uint32_t mode = rv->csr_mtvec & 0x3;
+/* RISC-V exception code list */
+#define GET_EXCEPTION_CODE(type) rv_exception_code_##type
+#define RV_EXCEPTION_LIST                                    \
+    _(insn_misaligned)  /* Instruction address misaligned */ \
+    _(insn_fault)       /* Instruction access fault */       \
+    _(illegal_insn)     /* Illegal instruction */            \
+    _(breakpoint)       /* Breakpoint */                     \
+    _(load_misaligned)  /* Load address misaligned */        \
+    _(load_fault)       /* Load access fault */              \
+    _(store_misaligned) /* Store/AMO address misaligned */
 
-    /* Exception Code: Instruction Address Misaligned */
-    const uint32_t code = 0;
+enum {
+#define _(type) GET_EXCEPTION_CODE(type),
+    RV_EXCEPTION_LIST
+#undef _
+};
 
-    /* mepc  (Machine Exception Program Counter)
-     * mtval (Machine Trap Value Register) : Misaligned Instruction
-     */
-    rv->csr_mepc = old_pc;
-    rv->csr_mtval = rv->PC;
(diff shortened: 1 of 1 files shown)


src/riscv.h
```diff
@@ -42,6 +42,17 @@ enum {
     rv_reg_t6,
 };
 
+/* RISC-V exception code list */
+enum {
```
jserv: Move the enumeration to C source.

jserv: With the [X macro](https://en.wikipedia.org/wiki/X_Macro) trick, we can write down shorter code.
Check https://github.com/sysprog21/rv32emu/blob/master/src/syscall.c (SUPPORTED_SYSCALLS)

Risheng1128:
Move the enumeration into source file.
Use [X macro](https://en.wikipedia.org/wiki/X_Macro) trick to construct the list.

src/emulate.c
```diff
@@ -24,160 +24,57 @@ extern struct target_ops gdbstub_ops;
...
-    switch (mode) {
-    case 0: /* DIRECT: All exceptions set PC to base */
-        rv->PC = base;
-        break;
-    case 1: /* VECTORED: Asynchronous interrupts set PC to base + 4 * code */
-        rv->PC = base + 4 * code;
-        break;
+/* RISC-V exception code list */
+#define GET_EXCEPTION_CODE(type) rv_exception_code_##type
+#define __EXCEPTION_LIST                                        \
```
jserv: Rename to `RV_EXCEPTION_LIST`.

src/emulate.c
```diff
@@ -24,160 +24,57 @@ extern struct target_ops gdbstub_ops;
...
+    _(insn_misaligned, 0)  /* Instruction address misaligned */ \
+    _(insn_fault, 1)       /* Instruction access fault */       \
+    _(illegal_insn, 2)     /* Illegal instruction */            \
+    _(breakpoint, 3)       /* Breakpoint */                     \
+    _(load_misaligned, 4)  /* Load address misaligned */        \
+    _(load_fault, 5)       /* Load access fault */              \
+    _(store_misaligned, 6) /* Store/AMO address misaligned */
+
+enum {
+#define _(type, code) GET_EXCEPTION_CODE(type) = code,
```
jserv: Using `enum`, it is not necessary to specify the assignment.

src/emulate.c
```diff
@@ -24,160 +24,57 @@ extern struct target_ops gdbstub_ops;
...
-        break;
-    case 1: /* VECTORED: Asynchronous interrupts set PC to base + 4 * code */
-        rv->PC = base + 4 * code;
-        break;
-    }
-
-    /* mcause (Machine Cause Register): store exception code */
-    rv->csr_mcause = code;
-}
+EXCEPTION_HANDLER_IMPL(insn_misaligned)
```
jserv: Rewrite as following:
```c
/* RISC-V exception handlers */
#define _(type) EXCEPTION_HANDLER_IMPL(type)
RV_EXCEPTION_LIST
#undef _
```

jserv (changes requested):
Append `Close #61` at the end of git commit message.

Risheng1128:
Becaue there are some exception handlers unused like `insn_fault` and 
`load_fault`, I add the function attributes `unused` to prevent compiler
warning.

jserv:
Thank @Risheng1128 for contributing!

PR #65, opened 2022-10-09
Pass several privilege tests and fix RV32I comment

Risheng1128:
This commit does the following things:
- Pass serveral privilege instruction and exceptions tests like ebreak, 
  misaligned load word and misaligned store word.
- Fix the comment of different type of RV32I instructions  format like 
  I, R and J-type and so on.

Because the compiler option `rv32gc` will make `ebreak` be compiled
as `c.ebreak` format, change it to `rv32g`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,19 +86,8 @@ Current progress of this emulator in riscv-arch-test(RV32):
     - `Zifencei`: Instruction-Fetch Fence
 * Failed Tests
     - `privilege`: RISCV Privileged Specification
-        + 2 system calls
-            * `ebreak`
+        + 1 system calls
             * `ecall`
-        + 5 misaligned CB-type instructions
-            * `misalign-beq`
-            * `misalign-bge`
-            * `misalign-bgeu`
-            * `misalign-blt`
-            * `misalign-bltu`
-        + 1 misaligned CL-type instructions
-            * `misalign-lw`
-        + 1 misaligned CS-type instructions
-            * `misalign-sw`
 * Unsupported tests (runnable but incomplete)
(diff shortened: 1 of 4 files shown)


jserv (changes requested):
Update README.md in top-level directory.

jserv:
I would anticipate the commit being divided into two:
1. Refine comments for RV32 instruction descriptions.
2. Improve compliance for privileged instructions.

Risheng1128:
The current result of privilege tests:
```shell
Check ebreak                    ... OK 
Check ecall                     ... FAIL 
Check misalign1-jalr-01         ... OK 
Check misalign2-jalr-01         ... OK 
Check misalign-beq-01           ... OK 
Check misalign-bge-01           ... OK 
Check misalign-bgeu-01          ... OK 
Check misalign-blt-01           ... OK 
Check misalign-bltu-01          ... OK 
Check misalign-bne-01           ... OK 
Check misalign-jal-01           ... OK 
Check misalign-lh-01            ... OK 
Check misalign-lhu-01           ... OK 
Check misalign-lw-01            ... OK 
Check misalign-sh-01            ... OK 
Check misalign-sw-01            ... OK 
```

jserv (changes requested):
Analyze the reason why ecall fails and describe in the git commit message.

Risheng1128:
The privilege instruction `ecall` doesn't pass the test because it only
dispatches the syscall number to corresponding handler in current
rv32emu.
Therefore, it needs to implement the exception handler with control
and status registers (CSRs).

jserv:
Thank @Risheng1128 for contributing!

PR #66, opened 2022-10-11
Implement environment call properly

Risheng1128:
Implement ecall instruction with control and status registers (CSRs) 
to pass the riscv-arch-test with all privilege tests.

riscv-arch-test result:
```shell
Check ebreak                    ... OK 
Check ecall                     ... OK 
Check misalign1-jalr-01         ... OK 
Check misalign2-jalr-01         ... OK 
Check misalign-beq-01           ... OK 
Check misalign-bge-01           ... OK 
Check misalign-bgeu-01          ... OK 
Check misalign-blt-01           ... OK 
Check misalign-bltu-01          ... OK 
Check misalign-bne-01           ... OK 
Check misalign-jal-01           ... OK 
Check misalign-lh-01            ... OK 
Check misalign-lhu-01           ... OK 
Check misalign-lw-01            ... OK 
Check misalign-sh-01            ... OK 
Check misalign-sw-01            ... OK
```

### DIFF
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -32,53 +32,59 @@ extern struct target_ops gdbstub_ops;
 #include "riscv_private.h"
 
 /* RISC-V exception code list */
-#define GET_EXCEPTION_CODE(type) rv_exception_code_##type
-#define RV_EXCEPTION_LIST                                    \
-    _(insn_misaligned)  /* Instruction address misaligned */ \
-    _(insn_fault)       /* Instruction access fault */       \
-    _(illegal_insn)     /* Illegal instruction */            \
-    _(breakpoint)       /* Breakpoint */                     \
-    _(load_misaligned)  /* Load address misaligned */        \
-    _(load_fault)       /* Load access fault */              \
-    _(store_misaligned) /* Store/AMO address misaligned */
+#define RV_EXCEPTION_LIST                                       \
+    _(insn_misaligned, 0)  /* Instruction address misaligned */ \
+    _(illegal_insn, 2)     /* Illegal instruction */            \
+    _(breakpoint, 3)       /* Breakpoint */                     \
+    _(load_misaligned, 4)  /* Load address misaligned */        \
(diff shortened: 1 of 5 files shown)


jserv (changes requested):
I failed to run `make check` because of no progress.

src/emulate.c
```diff
@@ -61,9 +72,15 @@ enum {
         const uint32_t code = GET_EXCEPTION_CODE(type);                     \
         /* mepc  (Machine Exception Program Counter)                        \
          * mtval (Machine Trap Value Register)                              \
+         * mcause (Machine Cause Register): store exception code            \
          */                                                                 \
         rv->csr_mepc = rv->PC;                                              \
         rv->csr_mtval = mtval;                                              \
+        rv->csr_mcause = code;                                              \
+        if (!rv->csr_mtvec) { /* elf doesn't implement csr instruction */   \
```
jserv: The comment was misleading. You can say,
> in case CSR is not configured.

src/emulate.c
```diff
@@ -40,14 +40,25 @@ extern struct target_ops gdbstub_ops;
     _(breakpoint)       /* Breakpoint */                     \
     _(load_misaligned)  /* Load address misaligned */        \
     _(load_fault)       /* Load access fault */              \
-    _(store_misaligned) /* Store/AMO address misaligned */
+    _(store_misaligned) /* Store/AMO address misaligned */   \
+    _(store_fault)      /* Store/AMO access fault */         \
+    _(ecall_U)          /* Environment call from U-mode */   \
+    _(ecall_S)          /* Environment call from S-mode */   \
+    _(reserved)         /* Reserved */                       \
```
jserv: Shall we care about "Reserved" item?

Risheng1128: I think the "reserved" item should be removed. Maybe I need to
modify the exception list with assigning exception code and remove
unuseful items.

jserv: In addition, check ecall_{U,S} as well since only `ecall_M` makes sense now.

PR #67, opened 2022-10-13
Refactor rbtree previous iterator

steven1lung:
The complicated if-else statements can be recuded using the early return mechanism. This commit would apply the mechanism to decrease code size and increase code readability.

### DIFF
--- a/src/map.c
+++ b/src/map.c
@@ -547,33 +547,27 @@ bool map_insert(map_t obj, void *key, void *value)
 
 static void map_prev(map_t obj, map_iter_t *it)
 {
-    if (!it->node) {
-        it->prev = NULL;
+    /* We have hit the end or null node. */
+    if (!it->node || it->node == obj->it_least.node) {
+        it->prev = it->node = NULL;
         return;
     }
 
-    if (it->node == obj->it_least.node) { /* We have hit the end. */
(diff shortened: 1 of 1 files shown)


src/map.c
```diff
@@ -547,33 +547,31 @@ bool map_insert(map_t obj, void *key, void *value)
...
         return;
     }
 
-    if (it->node == obj->it_least.node) { /* We have hit the end. */
-        it->prev = it->node = NULL;
-    } else {
-        if (it->node->left) { /* To the left, as far right as possible */
-            it->node = it->node->left;
+    if (it->node->left) { /* To the left, as far right as possible */
+        it->node = it->node->left;
```
jserv: Consider the following:
```c
    for (it->node = it->node->left; it->node->right; it->node = it->node->right)
        it->prev = it->node;
    return;
```

Later, to further simplify, we may create a `foreach` macro.

steven1lung: Shall the foreach macro be included in this pull request?

jserv: > Shall the foreach macro be included in this pull request?

Up to you. Since this pull request is intended to refactor the existing rbtree, we could think of feasible enhancements for both performance and simplicity.

PR #68, opened 2022-10-14
Eliminate memory allocations in the system call `setup_queue`

alanjian85:
Because the emulator lacks an address translation mechanism and is unable to allocate variables within the process's memory space, This commit grants the user the ability to create queues by passing an additional parameter to the system call that specifies the base address of queues.

### DIFF
--- a/docs/syscall.md
+++ b/docs/syscall.md
@@ -31,9 +31,9 @@ If a window does not already exist, one will be created with the specified `widt
 
 **system call number**: `0xC0DE`
 
-**synopsis**: `void *setup_queue(int capacity, unsigned int* event_count)`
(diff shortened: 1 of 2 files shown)


jserv (changes requested):
Always execute `clang-format -i` before submitting patches.

docs/syscall.md
```diff
@@ -31,9 +31,9 @@ If a window does not already exist, one will be created with the specified `widt
 
 **system call number**: `0xC0DE`
 
-**synopsis**: `void *setup_queue(int capacity, unsigned int* event_count)`
+**synopsis**: `void *setup_queue(void* base, int capacity, unsigned int* event_count)`
```
jserv: Check file `CONTRIBUTING.md` for consistent style.

PR #72, opened 2022-10-19
Fix red-black tree's implementation

alanjian85:
Even though the commit ff46dad7571d3d62852ca4b8722804c8324835bb has shortened the map code, I found that it may be incorrect because the second parameter passed to the function `rb_set_parent` on line 537 is `*indirect`, which was just assigned to the value of new node in the previous statement. I added a variable `parent` to store the previous value of `indirect` before any changes were made and set it as the parent of `new_node` as the old code intended.


steven1lung:
Looks good to me.

PR #74, opened 2022-10-19
Update demo `quake` and `doom`

alanjian85:
Replace the emulator's demo files with the most recent [quake-embedded](https://github.com/sysprog21/quake-embedded) and [doom_riscv](https://github.com/sysprog21/doom_riscv) executables.


jserv:
Duplicated.

PR #76, opened 2022-10-20
Refine relative mode submission mechanisms

alanjian85:
Remove the keyboard's control over mouse capturing and give the application the ability to set mouse relative mode. Also add a counter of deferred events to seamlessly delay execution of submissions delivered before the window was created.


jserv (changes requested):
Always run `clang-format -i` before submitting pull requests.

PR #77, opened 2022-10-22
Detect toolchain automatically

Risheng1128:
Detect toolchain automatically by creating toolchain list,
and check the items in list is installed or not.

If there are multiple toolchain installed, get the first item 
in valid toolchain list.

### DIFF
--- a/mk/toolchain.mk
+++ b/mk/toolchain.mk
@@ -9,24 +9,20 @@ else
 endif
 
 # Validate GNU Toolchain for RISC-V
-CROSS_COMPILE ?= riscv32-unknown-elf-
-RV32_CC = $(CROSS_COMPILE)gcc
-RV32_CC := $(shell which $(RV32_CC))
-ifndef RV32_CC
-    # xPack GNU RISC-V Embedded GCC
-    CROSS_COMPILE = riscv-none-elf-
-    RV32_CC = $(CROSS_COMPILE)gcc
-    RV32_CC := $(shell which $(RV32_CC))
-    ifndef RV32_CC
-        # DEPRECATED: Replaced by xpack-dev-tools/riscv-none-elf-gcc-xpack
-        CROSS_COMPILE = riscv-none-embed-
-        RV32_CC = $(CROSS_COMPILE)gcc
-        RV32_CC := $(shell which $(RV32_CC))
-        ifndef RV32_CC
-        $(warning No GNU Toolchain for RISC-V found.)
-        CROSS_COMPILE :=
-        endif
-    endif
+TOOLCHAIN_LIST := riscv-none-elf-      \
+		  riscv32-unknown-elf- \
+		  riscv64-unknown-elf- \
+		  riscv-none-embed-
+
+# TODO: add support to clang/llvm based cross compilers
+VALID_TOOLCHAIN := $(foreach toolchain,$(TOOLCHAIN_LIST),     \
+		   $(shell which $(toolchain)gcc > /dev/null) \
+		   $(if $(filter 0,$(.SHELLSTATUS)),$(toolchain)))
+
+# Get the first element in valid toolchain list
+CROSS_COMPILE ?= $(word 1,$(VALID_TOOLCHAIN))
+ifeq ($(CROSS_COMPILE),)
+$(warning GNU Toolchain for RISC-V is required. Please check package installation)
 endif
 
 export CROSS_COMPILE


mk/toolchain.mk
```diff
@@ -9,24 +9,14 @@ else
...
-        RV32_CC := $(shell which $(RV32_CC))
-        ifndef RV32_CC
-        $(warning No GNU Toolchain for RISC-V found.)
-        CROSS_COMPILE :=
-        endif
-    endif
-endif
+TOOLCHAIN_LIST := riscv-none-elf-      \
+		  riscv32-unknown-elf- \
+		  riscv64-unknown-elf-
```
jserv: Add `riscv-none-embed-` at the end of `TOOLCHAIN_LIST`.

mk/toolchain.mk
```diff
@@ -9,24 +9,14 @@ else
...
+
+VALID_TOOLCHAIN := $(foreach toolchain,$(TOOLCHAIN_LIST),     \
+		   $(shell which $(toolchain)gcc > /dev/null) \
+		   $(if $(filter 0,$(.SHELLSTATUS)),$(toolchain)))
 
+# Get the first element in valid toolchain list
+CROSS_COMPILE := $(word 1,$(VALID_TOOLCHAIN))
```
jserv: You should not use `:=` since someone might override `CROSS_COMPILE` by setting this environment variable. That is, the logic should honor the value specified by `make  CROSS_COMPILE=riscv-none-elf-`.

mk/toolchain.mk
```diff
@@ -9,24 +9,15 @@ else
...
+		  riscv32-unknown-elf- \
+		  riscv64-unknown-elf- \
+		  riscv-none-embed-
+
+VALID_TOOLCHAIN := $(foreach toolchain,$(TOOLCHAIN_LIST),     \
+		   $(shell which $(toolchain)gcc > /dev/null) \
+		   $(if $(filter 0,$(.SHELLSTATUS)),$(toolchain)))
 
+# Get the first element in valid toolchain list
+CROSS_COMPILE = $(word 1,$(VALID_TOOLCHAIN))
```
jserv: Use `?=` instead of `=`.
See https://www.gnu.org/software/make/manual/make.html (Variable Assignment)

mk/toolchain.mk
```diff
@@ -9,24 +9,15 @@ else
...
-        CROSS_COMPILE :=
-        endif
-    endif
-endif
+TOOLCHAIN_LIST := riscv-none-elf-      \
+		  riscv32-unknown-elf- \
+		  riscv64-unknown-elf- \
+		  riscv-none-embed-
+
+VALID_TOOLCHAIN := $(foreach toolchain,$(TOOLCHAIN_LIST),     \
```
jserv, line 18: Add a comment before `VALID_TOOLCHAIN`:
> TODO: add support to clang/llvm based cross compilers

mk/toolchain.mk
Risheng1128: Oops, my mistake. Thanks for reminder !

mk/toolchain.mk
```diff
@@ -9,24 +9,16 @@ else
...
+		  riscv64-unknown-elf- \
+		  riscv-none-embed-
+
+# TODO: add support to clang/llvm based cross compilers
+VALID_TOOLCHAIN := $(foreach toolchain,$(TOOLCHAIN_LIST),     \
+		   $(shell which $(toolchain)gcc > /dev/null) \
+		   $(if $(filter 0,$(.SHELLSTATUS)),$(toolchain)))
 
+# Get the first element in valid toolchain list
+CROSS_COMPILE ?= $(word 1,$(VALID_TOOLCHAIN))
```
jserv, line 23: You shall raise warning if there is no GNU toolchain for RISC-V found since we might build RISC-V Architecture Tests from source.

mk/toolchain.mk
```diff
@@ -9,24 +9,20 @@ else
...
+		  riscv64-unknown-elf- \
+		  riscv-none-embed-
+
+# TODO: add support to clang/llvm based cross compilers
+VALID_TOOLCHAIN := $(foreach toolchain,$(TOOLCHAIN_LIST),     \
+		   $(shell which $(toolchain)gcc > /dev/null) \
+		   $(if $(filter 0,$(.SHELLSTATUS)),$(toolchain)))
+
+# Get the first element in valid toolchain list
+CROSS_COMPILE = $(word 1,$(VALID_TOOLCHAIN))
```
jserv: Use `?=` for `CROSS_COMPILE` assignment.

mk/toolchain.mk
```diff
@@ -9,24 +9,20 @@ else
...
+ifeq ($(CROSS_COMPILE),)
+$(error GNU Toolchain for RISC-V is required. Please check package installation)
```
jserv: Should be `warning` instead of `error`.

PR #79, opened 2022-11-03
Decouple instruction decoding from emulation unit

Risheng1128:
To implement jit compiler in emulator, need to separate the part of decode instruction from execution because we have to fetch instructions continuously.


src/decode.c
```diff
@@ -0,0 +1,615 @@
...
+
+#define EXEC(instr)                           \
+    {                                         \
+        /* decode this instruction */         \
+        return decode_##instr(rv_insn, insn); \
+    }
+
+#define _(instr) decode_##instr : EXEC(instr);
+
+    RV32I_OPCODE_LIST
```
jserv: Add `/* clang-format off */ ` before `RV32I_OPCODE_LIST` to make `clang-format` happy.

src/decode.c
```diff
@@ -0,0 +1,615 @@
...
+#if RV32_HAS(EXT_A)
+        RV32A_OPCODE_LIST
+#endif
+#if RV32_HAS(EXT_F)
+            RV32F_OPCODE_LIST
+#endif
+#if RV32_HAS(EXT_C)
+                RV32C_OPCODE_LIST
+#endif
+                    _(unimp)
```
jserv: Ditto. Add `/* clang-format on */` after `_(unimp)`.

src/decode.h
```diff
@@ -0,0 +1,89 @@
+#pragma once
+
+#include <stdbool.h>
+#include <stdint.h>
+
+#include "riscv_private.h"
+
+#define RV32I_OPCODE_LIST \
```
jserv: Follow the manner in `src/syscall.c`. That is, rewrite as following:
```c
#define RV32_OPCODE_LIST \
    _(load)               \
    _(op_imm)             \
    _(auipc)              \
    _(store)              \
    _(op)                 \
    _(lui)                \
    _(branch)             \
    _(jalr)               \
    _(jal)                \
    _(system)             \
    IIF(RV32_HAS(EXT_A))( \
        _(amo),           \
    )
    ...
```

src/emulate.c
```diff
@@ -1884,171 +1823,44 @@ void rv_debug(struct riscv_t *rv)
...
-    TARGET(clui)
-    TARGET(cmisc_alu)
-    TARGET(ccr)
-    TARGET(cj)
-    TARGET(csw)
-    TARGET(cbeqz)
-    TARGET(cswsp)
-    TARGET(cbnez)
-#endif
+    opcode_t opcode_handler_entry[] = {
```
jserv: Don't do that! You should introduce IR while decoupling RISC-V instruction decoder from `emulate.c`. And then, execute the opcode in the form of predefined IR.

src/decode.h
```diff
@@ -0,0 +1,89 @@
...
+    _(auipc)              \
+    _(store)              \
+    _(op)                 \
+    _(lui)                \
+    _(branch)             \
+    _(jalr)               \
+    _(jal)                \
+    _(system)
+
+#define Zifencei_OPCODE_LIST _(misc_mem)
```
jserv: Drop the macros such as `Zifencei_OPCODE_LIST` and `RV32A_OPCODE_LIST` in favor of macro `IIF`.

Makefile
```diff
@@ -79,7 +79,7 @@ src/mini-gdbstub/Makefile:
 GDBSTUB_LIB := $(GDBSTUB_OUT)/libgdbstub.a
 $(GDBSTUB_LIB): src/mini-gdbstub/Makefile
 	$(MAKE) -C $(dir $<) O=$(dir $@)
-$(OUT)/emulate.o: $(GDBSTUB_LIB)
+$(OUT)/decode.o: $(GDBSTUB_LIB)
```
jserv: `src/decode.c` should have no explicit dependency with GDB stub implementation.

src/decode.c
```diff
@@ -0,0 +1,615 @@
...
+    } else { /* Compressed Extension Instruction */
+#if RV32_HAS(EXT_C)
+        insn &= 0x0000FFFF;
+        int16_t c_index = (insn & FC_FUNC3) >> 11 | (insn & FC_OPCODE);
+        rv->insn_len = INSN_16;
+        goto *rvc_jump_table[c_index];
+#endif
+    }
+
+#define EXEC(instr)                           \
```
jserv: The name `EXEC` is misleading. Here, it tends to be `DECODE`.

src/decode.c
```diff
@@ -0,0 +1,615 @@
...
+
+/* decode RISC-V instruction */
+bool rv_decode(struct riscv_t *rv, struct rv_insn_t *rv_insn, uint32_t insn)
+{
+    assert(rv && rv_insn);
+    /* store instruction */
+    rv_insn->insn = insn;
+#define DEC_UNIMP decode_unimp
+#if RV32_HAS(COMPUTED_GOTO)
+#define DEC(insn) &&decode_##insn
```
jserv: Use `OP` instead of `DEC` since we wrap the handling of each RISC-V opcode.

src/decode.c
```diff
@@ -0,0 +1,615 @@
+#include <assert.h>
+#include <stdio.h>
+
+#include "decode.h"
+
+static inline bool decode_load(struct rv_insn_t *rv_insn, uint32_t insn)
```
jserv: Use `op_load` instead of `decode_load` because we would emphasize the handling of RISC-V instruction. Later, when we introduce advanced code generators, the consistent naming scheme would benefit from them.

jserv: The expected results from RISC-V instruction decoding would be the sequence of intermediate representation (IR), which allows faster execution. Please check the relevant implementation `refactor-rv32` and `fastrv32`.

jserv: It does not make sense if you only move functions out of `emulate.c` without introducing IR. So-called decoupling must be involved with more abstractions.

src/emulate.c
Risheng1128: Understand! I will create a instruction list to represent different instrcution. Rewrite as following:
```c
enum {
    rv_insn_lui,
    rv_insn_auipc,
    rv_insn_jal,
    rv_insn_jalr,
    rv_insn_beq,
    rv_insn_bne,
    rv_insn_blt,
    rv_insn_bge,
    rv_insn_bltu,
    rv_insn_bgeu,
    ...
}
```
Thanks for your reminder.

src/decode.c
RinHizakura: I am not sure if it makes sense to give each instruction a single decode function.  Since the same type of instruction  (e.g. I-type, S-type, ...) should share the same decoding logic. I think that dealing with each instruction with an independent function introduces a lot of mess.

What you should do is define every type of decoding function, for example:

```cpp
static inline bool decode_i_type(struct rv_insn_t *rv_insn, uint32_t insn)
{
    /* I-type
     *  31       20 19   15 14    12 11   7 6      0
     * | imm[11:0] |  rs1  | funct3 |  rd  | opcode |
     */
    rv_insn->imm = dec_itype_imm(insn);
    rv_insn->rs1 = dec_rs1(insn);
    rv_insn->funct3 = dec_funct3(insn);
    rv_insn->rd = dec_rd(insn);
}
```

After that, you can fix the jump table or the label for compute-goto to jump to the correct types of decoding functions.

Risheng1128: Thanks for your reminder! I think I will try to rewrite as following for every instruction type

```c
static inline bool op_load(struct rv_insn_t *ir, uint32_t insn)
{
    /* decode I-type */
    decode_i_type(ir, insn);
    /* dispatch operation type */
    switch (ir->funct3) {
    ...
    }
    return true;
}
```

jserv: @Risheng1128, During code review, you should address the suggested improvements rather than simply responding "Thanks" as spam. That is, "show me the code" or response with something conceptual, which is ALWAYS better than nothing.

src/emulate.c
Risheng1128: On the other hand, create function `rv_emulate` to execute all implemented instructions:

For example:
```c
static bool rv_emulate(struct riscv_t *rv, struct rv_insn_t *ir)
{
    switch (ir->opcode) {
    /* RV32I Base Instruction Set*/
    case rv_insn_lui:
    case rv_insn_auipc:
    case rv_insn_jal:
    case rv_insn_jalr:
    case rv_insn_beq:
    case rv_insn_bne:
    case rv_insn_blt:
    ...
}
```

jserv: `rv_emulate` can be effectively implemented with computed-goto as what `fastrv32` does.

src/decode.c
jserv: TODO: use hashtable to map RISC-V instructions (index) to specific handlers.

Risheng1128:
riscv-arch-test result: RV32IMC and privilege situations pass.

TODO:
- Refine the commit message
- Use "computed goto" in function "rv_emulate"

PR #83, opened 2022-11-05
Update demo binaries

alanjian85:
While the most recent quake and doom update, which made the quake and doom demos automatically update the state of the cursor when the user opened/closed the menu, has been committed, update the demo binaries to keep up with the most recent progress.


jserv (changes requested):
Mention git SHA-1 hash in git commit message. See commit c8e3784b50686f373561329dfc282bf4fb425b0f

PR #86, opened 2022-11-07
Place floating point routines into a dedicated header

2011eric:


jserv (changes requested):
Remove `src/soft_float.c` since it is needed at the moment.

src/soft_float.h
```diff
@@ -0,0 +1,74 @@
...
+    FFLAG_DIV_BY_ZERO = 0b00000000000000000000000000001000,
+    FFLAG_OVERFLOW    = 0b00000000000000000000000000000100,
+    FFLAG_UNDERFLOW   = 0b00000000000000000000000000000010,
+    FFLAG_INEXACT     = 0b00000000000000000000000000000001,
+    //                    ....xxxx....xxxx....xxxx....xxxx
+    RV_NAN            = 0b01111111110000000000000000000000
+};
+
+
+/* clang-format on */
```
jserv: Move `/* clang-format on */` nearby the above `enum`.

src/emulate.c
```diff
@@ -12,6 +12,7 @@
 
 #if RV32_HAS(EXT_F)
 #include <math.h>
+#include "soft_float.h"
```
jserv, line 16: Add blank line before and after `soft_float.h`.

2011eric:
Should I also modify the commit title?

jserv:
> Should I also modify the commit title?

Go ahead.

Issue #88, opened 2022-11-07
Lower instruction decoding and dispatch overhead


jserv:
`wip/instruction-decode` branch breaks RISC-V instruction decoding and emulation into separate stage, meaning that it is feasible to incorporate further IR optimizations and JIT code generation. However, we do need additional efforts to make it practical:
1.  Executing RISC-V instructions by compiling the program a basic block at a time, thus avoiding unnecessary translation;
2. Implementing an efficient way to look in a hash map for a code block matching the current program counter as `wip/jit` does;
3. Reducing IR dispatch cost means of computed-goto or tail-call elimination (as wasm3 does).

All of the above should appear in `wip/instruction-decode` branch before its merge into `master` branch.

Risheng1128:
Known issue: Lower performance than `refactor-rv32`

When we test both of them using `dhrystone.elf`, it is discovered that the performance of the `wip/instruction-decode` branch is worse than `refactor-rv32`.
* `refactor-rv32` : 1631 DMIPS
* `rv32emu - wip/instruction-decode` : 1138 DMIPS

TODO: Use `perf` to  determine the root cause of `rv32emu`'s slowness

Risheng1128:
I use perf to find the hot spot of `rv32emu` and `refactor-rv32` respectively. Both of them execute `dhrystone.elf`.

`rv32emu`:
```diff
        │      if (!emulate(rv, block->ir + i))                                                                                                                                                            
   0.02 │        lea       (%rax,%rax,2),%rdx                                                                                                                                                              
- 10.16 │        mov       0x18(%r14),%rax                                                                                                                                                                 
   0.26 │        lea       (%rax,%rdx,4),%rbx                                                                                                                                                              
        │      emulate():                                                                                                                                                                                  
        │      rv->compressed = (ir->insn_len == INSN_16);                                                                                                                                                 
   0.26 │        movzbl    0x9(%rbx),%esi                                                                                                                                                                  
   0.09 │        cmp       $0x2,%sil                                                                                                                                                                       
-  9.74 │        sete      0x1cc(%rbp)                                                                                                                                                                     
        │      switch (ir->opcode) {                                                                                                                                                                       
        │        cmpb      $0x7a,0x7(%rbx)                                                                                                                                                                 
   0.10 │      ↓ ja        1a40                                                                                                                                                                            
-  9.02 │        movzbl    0x7(%rbx),%eax                                                                                                                                                                  
[...]

PR #91, opened 2022-11-13
Introduce basic block

Risheng1128:
This commit introduces the basic block in emulator, meaning that it makes emulator decode and execute numerous instructions at a time.

Complete the first requirement in #88 .

TODO:
1. Implementing an efficient way to look in a hash map for a code block matching the current program counter as wip/jit does;
2. Reducing IR dispatch cost means of computed-goto.

### DIFF
--- a/src/decode.c
+++ b/src/decode.c
@@ -4,9 +4,10 @@
  */
 
 #include <assert.h>
-#include <stdio.h>
+#include <stdlib.h>
 
 #include "decode.h"
+#include "riscv_private.h"
 
 /* decode rd field
  * rd = insn[11:7]
@@ -1653,19 +1654,18 @@ static inline bool op_cbnez(struct rv_insn_t *ir, const uint32_t insn)
 #define op_cfsd OP_UNIMP
 
 /* handler for all unimplemented opcodes */
-static inline bool op_unimp(struct rv_insn_t *rv_insn UNUSED,
-                            uint32_t insn UNUSED)
+static inline bool op_unimp(struct rv_insn_t *ir UNUSED, uint32_t insn UNUSED)
 {
     return false;
 }
 
 /* RV32 decode handler type */
-typedef bool (*decode_t)(struct rv_insn_t *rv_insn, uint32_t insn);
+typedef bool (*decode_t)(struct rv_insn_t *ir, uint32_t insn);
 
 /* decode RISC-V instruction */
-bool rv_decode(struct rv_insn_t *ir, uint32_t insn, uint8_t *insn_len)
+bool rv_decode(struct rv_insn_t *ir, uint32_t insn)
 {
-    assert(ir && insn_len);
+    assert(ir);
 
 #define OP_UNIMP op_unimp
 #define OP(insn) op_##insn
@@ -1704,7 +1704,7 @@ bool rv_decode(struct rv_insn_t *ir, uint32_t insn, uint8_t *insn_len)
     if ((insn & FC_OPCODE) != 3) {
         insn &= 0x0000FFFF;
         const uint16_t c_index = (insn & FC_FUNC3) >> 11 | (insn & FC_OPCODE);
-        *insn_len = INSN_16;
+        ir->insn_len = INSN_16;
 
         /* decode instruction (compressed instructions) */
         const decode_t op = rvc_jump_table[c_index];
@@ -1715,7 +1715,7 @@ bool rv_decode(struct rv_insn_t *ir, uint32_t insn, uint8_t *insn_len)
 
     /* standard uncompressed instruction */
     const uint32_t index = (insn & INSN_6_2) >> 2;
-    *insn_len = INSN_32;
+    ir->insn_len = INSN_32;
 
     /* decode instruction */
     const decode_t op = rv_jump_table[index];
@@ -1724,4 +1724,19 @@ bool rv_decode(struct rv_insn_t *ir, uint32_t insn, uint8_t *insn_len)
 
 #undef OP_UNIMP
 #undef OP
+}
+
+/* clear all block in the block map */
+void block_map_clear(struct block_map *map)
+{
+    assert(map);
+    for (uint32_t i = 0; i < map->block_capacity; i++) {
+        struct block *block = map->map[i];
+        if (block) {
+            free(block->ir);
+            free(block);
+            map->map[i] = NULL;
+        }
+    }
+    map->size = 0;
 }
--- a/src/decode.h
+++ b/src/decode.h
@@ -219,7 +219,6 @@ enum {
 /* clang-format on */
 
 enum {
-    INSN_UNKNOWN = 0,
     INSN_16 = 2,
     INSN_32 = 4,
 };
@@ -236,21 +235,36 @@ struct rv_insn_t {
 #if RV32_HAS(EXT_C)
     uint8_t shamt;
 #endif
+
+    /* instruction length */
+    uint8_t insn_len;
 };
 
-/* sign extend a 16 bit value */
-static inline uint32_t sign_extend_h(const uint32_t x)
-{
-    return (int32_t) ((int16_t) x);
-}
+/* translated basic block */
+struct block {
+    /* number of instructions encompased */
+    uint32_t insn_number;
+    /* address range of the basic block */
+    uint32_t pc_start, pc_end;
+    /* maximum of instructions encompased */
+    uint32_t insn_capacity;
+    /* block predictoin */
+    struct block *predict;
+    /* memory blocks */
+    struct rv_insn_t *ir;
+};
 
-/* sign extend an 8 bit value */
-static inline uint32_t sign_extend_b(const uint32_t x)
-{
-    return (int32_t) ((int8_t) x);
-}
+struct block_map {
+    /* max number of entries in the block map */
+    uint32_t block_capacity;
+    /* number of entries currently in the map */
+    uint32_t size;
+    /* block map */
+    struct block **map;
+};
+
+/* clear all block in the block map */
+void block_map_clear(struct block_map *map);
 
 /* decode the RISC-V instruction */
-bool rv_decode(struct rv_insn_t *rv_insn,
-               const uint32_t insn,
-               uint8_t *insn_len);
+bool rv_decode(struct rv_insn_t *ir, const uint32_t insn);
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -4,9 +4,7 @@
  */
 
 #include <assert.h>
-#include <inttypes.h>
 #include <stdbool.h>
-#include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 
@@ -52,7 +50,7 @@ enum {
 
 static void rv_exception_default_handler(struct riscv_t *rv)
 {
-    rv->csr_mepc += rv->insn_len;
+    rv->csr_mepc += rv->compressed ? INSN_16 : INSN_32;
     rv->PC = rv->csr_mepc; /* mret */
 }
 
@@ -256,8 +254,11 @@ static bool insn_is_misaligned(uint32_t pc)
     );
 }
 
-static bool rv_emulate(struct riscv_t *rv, struct rv_insn_t *ir)
+static bool emulate(struct riscv_t *rv, struct rv_insn_t *ir)
 {
+    /* check instruction is compressed or not */
+    rv->compressed = (ir->insn_len == INSN_16);
+
     switch (ir->opcode) {
     /* RV32I Base Instruction Set */
     case rv_insn_lui: /* LUI: Load Upper Immediate */
@@ -286,7 +287,7 @@ static bool rv_emulate(struct riscv_t *rv, struct rv_insn_t *ir)
         rv->PC += ir->imm;
         /* link with return address */
         if (ir->rd)
-            rv->X[ir->rd] = pc + rv->insn_len;
+            rv->X[ir->rd] = pc + ir->insn_len;
         /* check instruction misaligned */
         if (insn_is_misaligned(rv->PC)) {
             rv_except_insn_misaligned(rv, pc);
@@ -308,7 +309,7 @@ static bool rv_emulate(struct riscv_t *rv, struct rv_insn_t *ir)
         rv->PC = (rv->X[ir->rs1] + ir->imm) & ~1U;
         /* link */
         if (ir->rd)
-            rv->X[ir->rd] = pc + rv->insn_len;
+            rv->X[ir->rd] = pc + ir->insn_len;
         /* check instruction misaligned */
         if (insn_is_misaligned(rv->PC)) {
             rv_except_insn_misaligned(rv, pc);
@@ -963,7 +964,7 @@ static bool rv_emulate(struct riscv_t *rv, struct rv_insn_t *ir)
         rv->X[ir->rd] += (int16_t) ir->imm;
         break;
     case rv_insn_cjal:
-        rv->X[1] = rv->PC + rv->insn_len;
+        rv->X[1] = rv->PC + ir->insn_len;
         rv->PC += ir->imm;
         if (rv->PC & 0x1) {
             rv_except_insn_misaligned(rv, rv->PC);
@@ -1055,11 +1056,11 @@ static bool rv_emulate(struct riscv_t *rv, struct rv_insn_t *ir)
          * the value in register rs1' is zero. It expands to beq rs1', x0,
          * offset[8:1].
          */
-        rv->PC += (!rv->X[ir->rs1]) ? (uint32_t) ir->imm : rv->insn_len;
+        rv->PC += (!rv->X[ir->rs1]) ? (uint32_t) ir->imm : ir->insn_len;
         /* can branch */
         return true;
     case rv_insn_cbnez: /* C.BEQZ */
-        rv->PC += (rv->X[ir->rs1]) ? (uint32_t) ir->imm : rv->insn_len;
+        rv->PC += (rv->X[ir->rs1]) ? (uint32_t) ir->imm : ir->insn_len;
         /* can branch */
         return true;
     case rv_insn_cslli: /* C.SLLI */
@@ -1093,7 +1094,7 @@ static bool rv_emulate(struct riscv_t *rv, struct rv_insn_t *ir)
     case rv_insn_cjalr: { /* C.JALR */
         /* Unconditional jump and store PC+2 to ra */
         const int32_t jump_to = rv->X[ir->rs1];
-        rv->X[rv_reg_ra] = rv->PC + rv->insn_len;
+        rv->X[rv_reg_ra] = rv->PC + ir->insn_len;
         rv->PC = jump_to;
         if (rv->PC & 0x1) {
             rv_except_insn_misaligned(rv, rv->PC);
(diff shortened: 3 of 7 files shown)


jserv:
I merged commit 2aa7154710eade5ca0a3fc48d503dd280bbe619c and let's concentrate on basic block part.

src/emulate.c
```diff
@@ -50,41 +50,43 @@ enum {
...
-        case 0: /* DIRECT: All exceptions set PC to base */               \
-            rv->PC = base;                                                \
-            break;                                                        \
-        /* VECTORED: Asynchronous interrupts set PC to base + 4 * code */ \
-        case 1:                                                           \
-            rv->PC = base + 4 * code;                                     \
-            break;                                                        \
-        }                                                                 \
+#define EXCEPTION_HANDLER_IMPL(type, code)                                 \
+    static void rv_except_##type(struct riscv_t *rv, struct rv_insn_t *ir, \
```
jserv: Can you avoid adding `ir` a a parameter since we can look up the translated instruction from `rv`?

src/emulate.c
```diff
@@ -1125,33 +1127,102 @@ static bool rv_emulate(struct riscv_t *rv, struct rv_insn_t *ir)
...
+    return false;
+}
+
+static void rv_translate_block(struct riscv_t *rv, struct block_t *block)
+{
+    /* setup the basic block */
+    block->index = 0;
+
+    /* translate the basic block */
+    uint32_t block_limit = 1000;
```
jserv: Here `block_limit` might not fit the actual behavior of benchmark suite.

src/emulate.c
```diff
@@ -1201,6 +1272,18 @@ riscv_word_t rv_get_reg(struct riscv_t *rv, uint32_t reg)
     return ~0U;
 }
 
+/* allocate a new code block */
+struct block_t *rv_block_alloc()
```
jserv: Since we are going to perform memory allocation frequently, the technique [memory pool](https://en.wikipedia.org/wiki/Memory_pool) can be applied. Check https://github.com/jonas-fan/mempool for a simplified example. My tests:
```
# Test 001 
os malloc/free ... elapsed 246.09ms
mempool alloc/free ... elapsed 8.72ms
# Test 002 
os malloc/free ... elapsed 658.55ms
mempool alloc/free ... elapsed 24.38ms
# Test 003 
os malloc/free ... elapsed 483.08ms
mempool alloc/free ... elapsed 21.50ms
```

Risheng1128: I think that I can use queue to manage memory block. Rewrite as the following:

```c
/* a translated basic block */
struct block_t {
    /* memory blocks */
    struct rv_insn_t *ir;
    /* index of memory blocks */
    uint32_t front, rear;
    /* maximum of instructions encompased */
    uint32_t capacity;
};
```

That is, "decode" means enqueue a memory block, and "execute" means dequeue a memory block.

src/emulate.c
Risheng1128: Yes, use queue to mange memory block. At the same time, the block occurred the exception always locates on "rear".

jserv:
Instead of decoding and interpreting individual instructions in a loop, as with an interpreting emulator, an attempt is made to combine entire blocks that usually end with a branch/jump instruction. By means of the JIT frameworks such as [MIR](https://github.com/vnmakarov/mir), native instructions corresponding to the block are generated and executed at the first point in time at which a memory address is jumped to.

Jumps are not executed directly, but instead the jump target is saved and the generated function is exited with RET. This allows the runtime environment to first compile the block at the jump target and perform other parallel tasks, including interrupt, input, and peripheral emulation.

During the compilation of a program block, the number of RISC-V clock cycles required up to this point is calculated for each possible end over which the block can be exited, and this sum is added to an instruction counter during execution. By means of this counter, events that occur on the RISC-V at certain times, such as system timer, can be precisely timed despite the higher speed of the host platform. Since there may be routines in the emulated programs that are dependent on a fixed number of executed instructions in a certain period of time, the timers of the host system cannot be used without compatibility problems. Due to the block-wise execution, however, there is also the problem with the emulator presented here that interrupts or timers are only executed or updated a few clock pulses late - after the next jump.

@Risheng1128, you shall confirm that using queue-based block management meets the requirements outlined above. In particular, when an exception/interrupt occurs, the proposed emulator can jump to the specified block.

src/emulate.c
```diff
@@ -1125,33 +1126,96 @@ static bool rv_emulate(struct riscv_t *rv, struct rv_insn_t *ir)
     }
 
     /* step over instruction */
-    rv->PC += rv->insn_len;
+    rv->PC += ir->insn_len;
     return true;
 }
 
-void rv_step(struct riscv_t *rv, int32_t cycles)
+static bool rv_emulate_block(struct riscv_t *rv, struct block_t *block)
```
jserv: Don't use `rv_` prefix since this function is used internally.

src/emulate.c
```diff
@@ -1125,33 +1126,96 @@ static bool rv_emulate(struct riscv_t *rv, struct rv_insn_t *ir)
...
+#endif
+#if RV32_HAS(Zifencei)
+    case rv_insn_fencei:
+#endif
+        return true;
+    }
+    return false;
+}
+
+static void rv_translate_block(struct riscv_t *rv, struct block_t *block)
```
jserv: Ditto. Don't use `rv_` prefix since this function is used internally.

Risheng1128:
Use queue-based block management to manage the instruction decode and execution.

In decode stage, allocate a new memory block, put the decoded instruction into the block, and record the order of instructions by the member `front` in `struct block_t` until the queue is full or the latest instruction is a branch instruction. The default size of memory block in queue is 1024.

In execution stage, when the queue is not empty, emulator executes the memory block that index is `rear` in `struct block_t`, frees the memory block and increments the clock cycle finally.

In particular, when an exception/interrupt occurs, emulator will do the following steps:
1. Execute the exception/interrupt handler that resets a new program counter from the register `mtvec` and function 
`emulate` return false.
2. Enter to the decode stage again, and create new memory blocks based on the new program counter.
That is, emulator will dischard old memory blocks and create the new one from new program counter.

src/decode.h
```diff
@@ -236,6 +236,19 @@ struct rv_insn_t {
 #if RV32_HAS(EXT_C)
     uint8_t shamt;
 #endif
+
+    /* instruction length */
+    uint8_t insn_len;
+};
+
+/* a translated basic block */
+struct block_t {
```
jserv: The suffix `_t` implies the use of typedef. Thus, you shall rewrite as following:
```c
typedef struct {
   ...
} block_t;
```

src/decode.h
```diff
@@ -250,7 +263,35 @@ static inline uint32_t sign_extend_b(const uint32_t x)
...
+    return (block->front - block->rear) == block->capacity;
+}
+
+static inline bool queue_is_empty(struct block_t *block)
+{
+    assert(block);
+    return block->front == block->rear;
+}
+
+static inline struct rv_insn_t *get_rear_mem_block(struct block_t *block)
```
jserv: Apply consistent naming scheme for these helper functions.

Risheng1128: I think it could be renamed as following
```diff
- static inline struct rv_insn_t *get_rear_mem_block(struct block_t *block)
+ static inline struct rv_insn_t *queue_get_rear_block(block_t *block)

- static inline struct rv_insn_t *alloc_mem_block(block_t *block)
+ static inline struct rv_insn_t *queue_alloc_block(block_t *block)

- static inline void free_mem_block(block_t *block)
+ static inline void queue_free_block(block_t *block)
```

jserv:
TODO: use `perf` to figure out the performance hotspot between `master` branch (the "classic" one) and the proposed changes. We shall get acquainted with the causes of the slowdowns.

src/decode.h
jserv: It is a bit weird that there is an operation for "rear" but not "front" counterpart. Instead, you can hide the operation "get_rear" by explicitly mentioning its usage. We can say something like `queue_current_block`.

jserv: Both `queue_alloc_block` and `alloc_mem_block` are misleading. We can say, `block_enqueue`. Similarly, we can say, `block_dequeue` for the opposite.

src/decode.h
```diff
@@ -250,7 +263,35 @@ static inline uint32_t sign_extend_b(const uint32_t x)
     return (int32_t) ((int8_t) x);
 }
 
+static inline bool queue_is_full(struct block_t *block)
```
jserv: Since the operations are dedicated to basic blocks, the prefix `block_` should be enforced.

src/emulate.c
```diff
@@ -1201,6 +1265,19 @@ riscv_word_t rv_get_reg(struct riscv_t *rv, uint32_t reg)
...
 }
 
+/* allocate a new code block */
+static bool block_init(struct block_t *block, uint32_t bit)
+{
+    assert(block);
+
+    block->capacity = 1 << bit;
+    block->ir = malloc(sizeof(struct rv_insn_t) * block->capacity);
+    if (!block->ir)
```
jserv: Rewrite as following:
```c
    return (bool) block->ir;
```

src/decode.h
```diff
@@ -250,7 +264,60 @@ static inline uint32_t sign_extend_b(const uint32_t x)
...
+
+/* free the basic block */
+static inline void block_free(block_t *block)
+{
+    assert(block);
+    free(block->ir);
+}
+
+/* check the queue is full or not */
+static inline bool block_queue_is_full(block_t *block)
```
jserv: How about `block_is_full`?

src/decode.h
```diff
@@ -250,7 +264,60 @@ static inline uint32_t sign_extend_b(const uint32_t x)
...
+{
+    assert(block);
+    return (block->front - block->rear) == block->capacity;
+}
+
+/* check the queue is empty or not */
+static inline bool block_queue_is_empty(block_t *block)
```
jserv: Ditto. How about `block_is_empty`?

src/decode.h
```diff
@@ -250,7 +264,60 @@ static inline uint32_t sign_extend_b(const uint32_t x)
...
+{
+    assert(block);
+    return block->front == block->rear;
+}
+
+/* get the current memory block in queue */
+static inline struct rv_insn_t *queue_current_block(block_t *block)
```
jserv: How about `block_current`?

src/decode.h
```diff
@@ -250,7 +264,60 @@ static inline uint32_t sign_extend_b(const uint32_t x)
...
 }
 
+/* allocate a new basic block */
+static inline bool block_init(block_t *block, uint32_t bit)
+{
+    assert(block);
+
+    block->capacity = 1 << bit;
+    block->ir = malloc(sizeof(struct rv_insn_t) * block->capacity);
+    if (!block->ir)
```
jserv: Don't do that You can simply write as following:
```c
    return (bool) (uintptr_t) block-ir;
```

jserv: That is, you don't have to rely on if-else statement.

Risheng1128:
Let emulator executes `coremark.elf` and use perf to observe the hotspot of branch `master` and `wip/instruction-decode` respectively.

branch `master`:
```shell
Overhead    Command    Shared Object   Symbol
52.99%      rv32emu    rv32emu         [.] rv_step
9.80%       rv32emu    rv32emu         [.] memory_ifetch
8.69%       rv32emu    rv32emu         [.] on_mem_ifetch
3.88%       rv32emu    rv32emu         [.] rv_userdata
2.98%       rv32emu    rv32emu         [.] memory_write
1.98%       rv32emu    rv32emu         [.] main
1.40%       rv32emu    rv32emu         [.] memory_read_w
0.82%       rv32emu    rv32emu         [.] memory_read_s
0.73%       rv32emu    rv32emu         [.] on_mem_write_w
0.72%       rv32emu    rv32emu         [.] rv_has_halted
0.55%       rv32emu    rv32emu         [.] on_mem_read_w
0.48%       rv32emu    rv32emu         [.] on_mem_read_s
...
```

branch `wip/instruction-decode`:
```shell
Overhead    Command    Shared Object   Symbol
39.95%      rv32emu    rv32emu         [.] rv_step
10.52%      rv32emu    rv32emu         [.] rv_decode
6.59%       rv32emu    rv32emu         [.] memory_ifetch
6.09%       rv32emu    rv32emu         [.] op_op_imm
5.71%       rv32emu    rv32emu         [.] on_mem_ifetch
3.85%       rv32emu    rv32emu         [.] op_branch
2.35%       rv32emu    rv32emu         [.] rv_userdata
2.29%       rv32emu    rv32emu         [.] op_op
2.27%       rv32emu    rv32emu         [.] op_load
1.99%       rv32emu    rv32emu         [.] memory_write
0.99%       rv32emu    rv32emu         [.] memory_read_w
0.92%       rv32emu    rv32emu         [.] op_store
0.51%       rv32emu    rv32emu         [.] memory_read_s
0.51%       rv32emu    rv32emu         [.] on_mem_write_w
0.50%       rv32emu    rv32emu         [.] on_mem_read_w
...
```

src/decode.h
```diff
@@ -250,7 +264,57 @@ static inline uint32_t sign_extend_b(const uint32_t x)
...
+
+/* check the queue is empty or not */
+static inline bool block_is_empty(block_t *block)
+{
+    assert(block);
+    return block->front == block->rear;
+}
+
+/* get the current memory block in queue */
+static inline struct rv_insn_t *queue_current(block_t *block)
```
jserv: Apply the consistent naming scheme. That is, we shall have the helper function below:
* `block_is_full`, `block_is_full`
* `block_current_insn`
* `block_enqueue`
* `block_dequeue`

jserv:
> Let emulator executes `coremark.elf` and use perf to observe the hotspot of branch `master` and `wip/instruction-decode` respectively.

If a block has been translated, we would not throw it away after emulating the instructions within; instead, we'll save it for future lookups.

`wip/instruction-decode` is indeed spending more time on decoding + emulating than `master` branch. It results from the inefficient block translation which prevents block reusability from being taken.

jserv:
> `wip/instruction-decode` is indeed spending more time on decoding + emulating than `master` branch. It results from the inefficient block translation which prevents block reusability from being taken.

Check `wip/jit` and `refactor-rv32` for block prediction. We should attempt to predict the next block, resulting in known performance boost.

src/emulate.c
```diff
@@ -1213,6 +1277,10 @@ struct riscv_t *rv_create(const struct riscv_io_t *io, riscv_user_t userdata)
     /* copy over the userdata */
     rv->userdata = userdata;
 
+    /* place a new queue */
+    if (!block_init(&rv->block, 10))
+        return NULL;
```
RinHizakura: It would be better not to leak the memory `rv` if we fail to create a new queue here. (In fact, we may also have to handle the case if trying to `malloc` for `struct riscv_t` but out of memory.

src/emulate.c
```diff
@@ -1242,7 +1311,6 @@ void rv_reset(struct riscv_t *rv, riscv_word_t pc)
 
     /* set the reset address */
     rv->PC = pc;
-    rv->insn_len = INSN_UNKNOWN;
```
RinHizakura, line 1245: It looks like we can remove `INSN_UNKNOWN` in the enum if we are going to move `insn_len`  to `struct rv_insn_t` and rewriting like this.

src/decode.h
```diff
@@ -250,7 +264,57 @@ static inline uint32_t sign_extend_b(const uint32_t x)
...
+{
+    assert(block);
+    return block->ir + block->rear;
```
RinHizakura: I feel this implementation is not correct. Since you allow the capacity to be `2^n`, where n could be any valid `uint32_t` number.

Let's say we create a `block_t` with capacity 2. However, you don't wrap back `block->rear`  when it is equal to 2. Instead, `block->rear` will only be wrapped back when being 2^32. So the overflow problem will happen when doing dequeue more than 2 times.

RinHizakura: The same problem will also happen on `queue_enqueue`.

LambertWSJ:
I recommend using a hash table and block prediction. it may be more convenient to implement the control flow graph in addition to having quick access to the basic block.

Risheng1128:
Remove the queue in block because we just need to know the number of instructions encompased and the first executed instruction in block is located at first position of IR array.

Using hash table and block prediction to manage the blocks. it cans avoid to decode the duplicate block and find the next block efficiently.

Finally, using `coremark.elf` and `dhrystone.elf` to measure the performance.

old:
```shell
coremark
---------------------------------------------------------------------
2K performance run parameters for coremark.
CoreMark Size    : 666
Total ticks      : 13939747
Total time (secs): 13.939747
Iterations/Sec   : 430.423881
Iterations       : 6000
Compiler version : GCC11.1.0
Compiler flags   : -O2 -DPERFORMANCE_RUN=1  
Memory location  : Please put data memory location here
			(e.g. code in flash, data on heap etc)
seedcrc          : 0xe9f5
[0]crclist       : 0xe714
[0]crcmatrix     : 0x1fd7
[0]crcstate      : 0x8e3a
[0]crcfinal      : 0xa14c
Correct operation validated. See README.md for run and reporting rules.
CoreMark 1.0 : 430.423881 / GCC11.1.0 -O2 -DPERFORMANCE_RUN=1   / Heap

dhrystone
---------------------------------------------------------------------
Dhrystone(1.1-mc), 10000000 passes, 13554164 microseconds, 419 DMIPS
```

new:
```shell
coremark
---------------------------------------------------------------------
2K performance run parameters for coremark.
CoreMark Size    : 666
Total ticks      : 12069925
Total time (secs): 12.069925
Iterations/Sec   : 911.356119
Iterations       : 11000
Compiler version : GCC11.1.0
Compiler flags   : -O2 -DPERFORMANCE_RUN=1  
Memory location  : Please put data memory location here
			(e.g. code in flash, data on heap etc)
seedcrc          : 0xe9f5
[0]crclist       : 0xe714
[0]crcmatrix     : 0x1fd7
[0]crcstate      : 0x8e3a
[0]crcfinal      : 0x33ff

dhrystone
---------------------------------------------------------------------
Dhrystone(1.1-mc), 10000000 passes, 5904499 microseconds, 961 DMIPS
```

The new implementation is about 2 ~ 3 times faster than old one.

src/decode.h
```diff
@@ -250,7 +276,68 @@ static inline uint32_t sign_extend_b(const uint32_t x)
...
+    k ^= k >> 17;
+#if (SIZE_MAX > 0xFFFFFFFF)
+    k ^= k >> 35;
+    k ^= k >> 51;
+#endif
+    return k;
+}
+
+/* allocate a basic block */
+static inline struct block *block_alloc(const uint8_t bits)
```
jserv: `block_alloc` should not appear in `decode.h`. Make the header as simple as possible.

src/decode.h
```diff
@@ -250,7 +276,68 @@ static inline uint32_t sign_extend_b(const uint32_t x)
     return (int32_t) ((int8_t) x);
 }
 
+/* hash function is used when mapping address into the block map */
+static inline uint32_t hash(size_t k)
```
jserv: Ditto. The hash function is implementation-specific. Don't put it in the header unless it is common.

src/riscv.h
```diff
@@ -86,8 +87,8 @@ typedef void (*riscv_mem_write_b)(struct riscv_t *rv,
                                   riscv_byte_t data);
 
 /* system instruction handlers */
-typedef void (*riscv_on_ecall)(struct riscv_t *rv);
-typedef void (*riscv_on_ebreak)(struct riscv_t *rv);
+typedef void (*riscv_on_ecall)(struct riscv_t *rv, struct rv_insn_t *ir);
```
jserv: `riscv.h` is a public header, and there should be no reference to IR. Hide the implementation details.

src/riscv.h
```diff
@@ -143,10 +144,10 @@ riscv_word_t rv_get_reg(struct riscv_t *, uint32_t reg);
 void syscall_handler(struct riscv_t *rv);
 
 /* environment call handler */
-void ecall_handler(struct riscv_t *rv);
+void ecall_handler(struct riscv_t *rv, struct rv_insn_t *ir);
```
jserv: Avoid changing the function prototype of `ecall` and `ebreak` handlers.

Risheng1128: Adding a new parameter in exception handler since the exception default handler needs the length of current instruction executed. Maybe, I think retaining the `insn_len` in `struct riscv_t` to store the current instruction length cans avoid to add a new parameter in all exception handler.

jserv: Instead of naming it `insn_len`, we can call it `bool compressed`.

Risheng1128:
Retain `block_map_clear` in `decode.h` only since it is called by `emulate.c` and `riscv.c`.

src/emulate.c
jserv: Agree. @Risheng1128, can you check?

src/riscv.c
```diff
@@ -0,0 +1,131 @@
...
+        return rv->X[reg];
+
+    return ~0U;
+}
+
+struct riscv_t *rv_create(const struct riscv_io_t *io, riscv_user_t userdata)
+{
+    assert(io);
+
+    struct riscv_t *rv = (struct riscv_t *) malloc(sizeof(struct riscv_t));
```
jserv: For C programming, you don't have to cast the result (type: `void *`) of `malloc`. i.e., you can simply write as following:
```c
struct riscv_t *rv = malloc(sizeof(struct riscv_t));
```

jserv: Furthermore, use `calloc` instead of combination of `malloc` and `memset`.

src/emulate.c
```diff
@@ -1125,146 +1126,202 @@ static bool rv_emulate(struct riscv_t *rv, struct rv_insn_t *ir)
...
+    uint32_t index = hash(pc);
+    const uint32_t mask = map->block_capacity - 1;
+
+    /* find block in block map */
+    for (;; index++) {
+        struct block *block = map->map[index & mask];
+        if (!block)
+            return NULL;
+
+        if (block->pc_start == pc) {
```
jserv: Remove curly brackets (`{` and `}`) for this statement since they are not needed.

src/decode.h
```diff
@@ -250,7 +276,20 @@ static inline uint32_t sign_extend_b(const uint32_t x)
     return (int32_t) ((int8_t) x);
 }
 
+/* clear all block in the block map */
+static inline void block_map_clear(struct block_map *map)
```
jserv: We may provide similar shortcut in `src/map.h` in another pull request.

Risheng1128: > We may provide similar shortcut in `src/map.h` in another pull request.

I'm not understand about this. Can you explain more details ?

jserv: I meant that we can provide a macro or inline function to release and free a map object.

src/decode.c
```diff
@@ -1725,3 +1724,30 @@ bool rv_decode(struct rv_insn_t *ir, uint32_t insn, uint8_t *insn_len)
 #undef OP_UNIMP
 #undef OP
 }
+
+/* sign extend a 16 bit value */
```
jserv: Move `sign_extend_h` and `sign_extend_b` to `riscv_private.h` since they are intended to be implementation specific.

src/emulate.c
```diff
@@ -1125,146 +1126,201 @@ static bool rv_emulate(struct riscv_t *rv, struct rv_insn_t *ir)
...
+    block->insn_capacity = 1 << bits;
+    block->insn_number = 0;
+    block->predict = NULL;
+    block->ir = malloc(block->insn_capacity * sizeof(struct rv_insn_t));
+    return block;
 }
 
-riscv_word_t rv_get_pc(struct riscv_t *rv)
+/* insert a block into block map */
+static void block_insert(struct block_map *map, struct block *block)
```
jserv: Add modifier `const` to `block` since it would be unchanged in this function.

src/decode.c
Risheng1128: if move `sign_extend_h` and `sign_extend_b` to `riscv_private.h`, then `decode.c` needs to include `riscv_private.h`. It will make rv32emu compile failed because `riscv_private.h` includes the `gdbstub.h` as following and `gdbstub.h` is cloned until compiling `emulate.c`.
```c
#if RV32_HAS(GDBSTUB)
#include "breakpoint.h"
#include "mini-gdbstub/include/gdbstub.h"
#endif
```
Maybe, I think that `emulate.c` swaps the compiling order with `decode.c` will resolve this problem.
```diff
OBJS := \
	map.o \
	utils.o \
-       decode.o \
        emulate.o \
+       decode.o \
	riscv.o \
	io.o \
	elf.o \
	main.o \
	syscall.o \
	$(OBJS_EXT)
```

src/emulate.c
Risheng1128: I will rewrite as following because the member in `block` would be unchanged.
```diff
- static void block_insert(struct block_map *map, struct block *block)
+ static void block_insert(struct block_map *map, const struct block *block)
```
Ditto, other functions like `block_find` and `block_emulate` as well.

src/decode.c
jserv: It sounds fairly reasonable for object file reordering. Mention this within git commit messages.

PR #92, opened 2022-12-07
Shorten the time spent checking instruction length

qwe661234:
I think we can pass the value of instruction length in IR to the exception handler rather than verifying it each time the emulate function is called, because we only need the value of rv->compressed in the exception handler.

### DIFF
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -256,9 +256,6 @@ static bool insn_is_misaligned(uint32_t pc)
 
 static bool emulate(riscv_t *rv, rv_insn_t *ir)
 {
-    /* check instruction is compressed or not */
-    rv->compressed = (ir->insn_len == INSN_16);
-
     switch (ir->opcode) {
     /* RV32I Base Instruction Set */
     case rv_insn_lui: /* LUI: Load Upper Immediate */
@@ -290,6 +287,7 @@ static bool emulate(riscv_t *rv, rv_insn_t *ir)
             rv->X[ir->rd] = pc + ir->insn_len;
         /* check instruction misaligned */
         if (insn_is_misaligned(rv->PC)) {
+            rv->compressed = false;
             rv_except_insn_misaligned(rv, pc);
             return false;
         }
@@ -312,6 +310,7 @@ static bool emulate(riscv_t *rv, rv_insn_t *ir)
             rv->X[ir->rd] = pc + ir->insn_len;
         /* check instruction misaligned */
         if (insn_is_misaligned(rv->PC)) {
+            rv->compressed = false;
             rv_except_insn_misaligned(rv, pc);
             return false;
         }
@@ -324,6 +323,7 @@ static bool emulate(riscv_t *rv, rv_insn_t *ir)
             rv->PC += ir->imm;
             /* check instruction misaligned */
             if (insn_is_misaligned(rv->PC)) {
+                rv->compressed = false;
                 rv_except_insn_misaligned(rv, pc);
                 return false;
             }
@@ -338,6 +338,7 @@ static bool emulate(riscv_t *rv, rv_insn_t *ir)
             rv->PC += ir->imm;
             /* check instruction misaligned */
             if (insn_is_misaligned(rv->PC)) {
+                rv->compressed = false;
                 rv_except_insn_misaligned(rv, pc);
                 return false;
             }
@@ -352,6 +353,7 @@ static bool emulate(riscv_t *rv, rv_insn_t *ir)
             rv->PC += ir->imm;
             /* check instruction misaligned */
             if (insn_is_misaligned(rv->PC)) {
+                rv->compressed = false;
                 rv_except_insn_misaligned(rv, pc);
                 return false;
             }
@@ -366,6 +368,7 @@ static bool emulate(riscv_t *rv, rv_insn_t *ir)
             rv->PC += ir->imm;
             /* check instruction misaligned */
             if (insn_is_misaligned(rv->PC)) {
+                rv->compressed = false;
                 rv_except_insn_misaligned(rv, pc);
                 return false;
             }
@@ -380,6 +383,7 @@ static bool emulate(riscv_t *rv, rv_insn_t *ir)
             rv->PC += ir->imm;
             /* check instruction misaligned */
             if (insn_is_misaligned(rv->PC)) {
+                rv->compressed = false;
                 rv_except_insn_misaligned(rv, pc);
                 return false;
             }
@@ -394,6 +398,7 @@ static bool emulate(riscv_t *rv, rv_insn_t *ir)
             rv->PC += ir->imm;
             /* check instruction misaligned */
             if (insn_is_misaligned(rv->PC)) {
+                rv->compressed = false;
                 rv_except_insn_misaligned(rv, pc);
                 return false;
             }
@@ -409,6 +414,7 @@ static bool emulate(riscv_t *rv, rv_insn_t *ir)
     case rv_insn_lh: { /* LH: Load Halfword */
         const uint32_t addr = rv->X[ir->rs1] + ir->imm;
         if (addr & 1) {
+            rv->compressed = false;
             rv_except_load_misaligned(rv, addr);
             return false;
         }
@@ -418,6 +424,7 @@ static bool emulate(riscv_t *rv, rv_insn_t *ir)
     case rv_insn_lw: { /* LW: Load Word */
         const uint32_t addr = rv->X[ir->rs1] + ir->imm;
         if (addr & 3) {
+            rv->compressed = false;
             rv_except_load_misaligned(rv, addr);
             return false;
         }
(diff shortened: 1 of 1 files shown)


jserv:
Once we complete the instruction decoding, we can definitely know the length of instruction(s) pointed by specific address. Would it make sense?

qwe661234:
> Once we complete the instruction decoding, we can definitely know the length of instruction(s) pointed by specific address. Would it make sense?

Yes, we can know the length of instruction once instruction decoding completes. However, we should send a value to the exception handler if we want to reduce the number of times we check the instruction length.

jserv:
> Yes, we can know the length of instruction once instruction decoding completes. However, we should send a value to the exception handler if we want to reduce the number of times we check the instruction length.

If we can look up the instruction length (RV32 vs. RV32C) from a specific address, there is no need to preserve extra parameter in exception handlers. Let's keep it simple and stupid (KISS principle).

qwe661234:
> > Yes, we can know the length of instruction once instruction decoding completes. However, we should send a value to the exception handler if we want to reduce the number of times we check the instruction length.
> 
> If we can look up the instruction length (RV32 vs. RV32C) from a specific address, there is no need to preserve extra parameter in exception handlers. Let's keep it simple and stupid (KISS principle).

Sorry, I miss the important information that we get instruction length by parsing instruction opcode, instead of specific address. If we don't want to pass more parameter to exception handlers, we need to save more information in struct `riscv_t`.

jserv:
Since preliminary basic block was introduced, speculated execution can take place without frequent instruction length checks. That is, the explicit transition from 32-bit ISA to 16-bit counterpart (vice versa) can be lazily evaluated in a basic block.

qwe661234:
> Since preliminary basic block was introduced, speculated execution can take place without frequent instruction length checks. That is, the explicit transition from 32-bit ISA to 16-bit counterpart (vice versa) can be lazily evaluated in a basic block.

I think the idea above is checking instruction length in a basic block. However, there is a situation I am unsure whether it will occur or not. If a 32-bit instruction and a 16-bit compressed instruction coexist in the same basic block, it is unsafe to determine whether the instruction is compressed in basic block.

Risheng1128:
We need to know the instruction is compressed or not when the exception occurred. So, I think that `rv->compressed` can be determined until the exception occurred.

For example, rewrite instruction `jal` as following:
```diff
case rv_insn_jal: { /* JAL: Jump and Link */
        ...
        /* check instruction misaligned */
        if (insn_is_misaligned(rv->PC)) {
+           rv->compressed = false;
            rv_except_insn_misaligned(rv, pc);
            return false;
        }
        /* can branch */
        return true;
    }
```
Therefore, we don't need to pass instruction length to exception handler and avoid to check `rv->compressed` frequently.

src/emulate.c
```diff
@@ -254,10 +256,14 @@ static bool insn_is_misaligned(uint32_t pc)
...
 
+#define RV_Op(OPCODE) static bool op_##OPCODE(riscv_t *rv, rv_insn_t *ir)
+#define EMU_Op(OPCODE, RV, IR) op_##OPCODE(RV, IR)
+
+
 static bool emulate(riscv_t *rv, rv_insn_t *ir)
 {
     /* check instruction is compressed or not */
-    rv->compressed = (ir->insn_len == INSN_16);
+    // rv->compressed = (ir->insn_len == INSN_16);
```
Risheng1128: This comment should be removed.

src/emulate.c
```diff
@@ -254,10 +256,14 @@ static bool insn_is_misaligned(uint32_t pc)
     );
 }
 
+#define RV_Op(OPCODE) static bool op_##OPCODE(riscv_t *rv, rv_insn_t *ir)
+#define EMU_Op(OPCODE, RV, IR) op_##OPCODE(RV, IR)
```
Risheng1128: I don't find the other code call these macros. Maybe, I think it is redundant.

qwe661234:
> We need to know the instruction is compressed or not when the exception occurred. So, I think that `rv->compressed` can be determined until the exception occurred.
> 
> For example, rewrite instruction `jal` as following:
> 
> ```diff
> case rv_insn_jal: { /* JAL: Jump and Link */
>         ...
>         /* check instruction misaligned */
>         if (insn_is_misaligned(rv->PC)) {
> +           rv->compressed = false;
>             rv_except_insn_misaligned(rv, pc);
>             return false;
>         }
>         /* can branch */
>         return true;
>     }
> ```
> 
> Therefore, we don't need to pass instruction length to exception handler and avoid to check `rv->compressed` frequently.

Ok, I think it is better idea, I will follow this idea.

src/emulate.c
```diff
@@ -290,6 +287,7 @@ static bool emulate(riscv_t *rv, rv_insn_t *ir)
             rv->X[ir->rd] = pc + ir->insn_len;
         /* check instruction misaligned */
         if (insn_is_misaligned(rv->PC)) {
+            rv->compressed = (ir->insn_len == INSN_16);
```
Risheng1128: It should be `rv->compressed = false;` since instruction `jal` is always uncompressed. Therefore, we just assign an exact value.

src/emulate.c
```diff
@@ -312,6 +310,7 @@ static bool emulate(riscv_t *rv, rv_insn_t *ir)
...
```
Risheng1128: Ditto, rewite the code as `rv->compressed = false;`.

src/emulate.c
qwe661234: Done.

src/riscv.h
```diff
@@ -144,10 +144,10 @@ riscv_word_t rv_get_reg(riscv_t *rv, uint32_t reg);
 void syscall_handler(riscv_t *rv);
 
 /* environment call handler */
-void ecall_handler(riscv_t *rv);
+void ecall_handler(riscv_t *rv, uint8_t insn_len);
```
jserv: Avoid changing the function prototype.

src/emulate.c
```diff
@@ -1336,12 +1354,14 @@ void rv_step(riscv_t *rv, int32_t cycles)
 void ebreak_handler(riscv_t *rv)
 {
     assert(rv);
+    rv->compressed = false;
```
Risheng1128: In ebreak case, there are two cases — `ebreak` and `c.ebreak`. Both of them execute function `ebreak_handler`, but their instruction length is different. Assign the value to `rv->compressed` in ebreak case respectively.

src/emulate.c
```diff
@@ -1336,12 +1354,14 @@ void rv_step(riscv_t *rv, int32_t cycles)
...
     rv_except_breakpoint(rv, rv->PC);
 }
 
 void ecall_handler(riscv_t *rv)
 {
     assert(rv);
+    rv->compressed = false;
```
Risheng1128: For the consistency, I think we can move it to the function `emulate`.

qwe661234: OK

src/emulate.c
```diff
@@ -551,9 +551,11 @@ static bool emulate(riscv_t *rv, rv_insn_t *ir)
         rv->X[ir->rd] = rv->X[ir->rs1] & rv->X[ir->rs2];
         break;
     case rv_insn_ecall: /* ECALL: Environment Call */
+        rv->compressed = (ir->insn_len == INSN_16);
```
Risheng1128: Rewrite as `rv->compressed = false;` since `ecall` is uncompressed instruction.

src/emulate.c
```diff
@@ -551,9 +551,11 @@ static bool emulate(riscv_t *rv, rv_insn_t *ir)
...
         rv->io.on_ecall(rv);
         return true;
     case rv_insn_ebreak: /* EBREAK: Environment Break */
+        rv->compressed = (ir->insn_len == INSN_16);
```
Risheng1128: It should be `rv->compressed = false;`. In the other hand, `c.ebreak` case should add `rv->compressed = true;`

Risheng1128:
The commit `Move instruction length checking` should merge into the commit `Reduce instruction length checking`.

jserv (changes requested):
Read https://cbea.ms/git-commit/ carefully for writing informative git commit messages.
You should explain the purpose of the proposed changes as well as their advantages.

Risheng1128 (approved):

jserv (changes requested):
Again, read https://cbea.ms/git-commit/ carefully.
* [Wrap the body at 72 characters](https://cbea.ms/git-commit/#wrap-72)
* [Use the body to explain what and why vs. how](https://cbea.ms/git-commit/#why-not-how)

Last but not least, clearly state how the proposed changes would be beneficial.

jserv (changes requested):
Do you understand what [Wrap the body at 72 characters](https://cbea.ms/git-commit/#wrap-72) means?

jserv:
By the way, use [QuillBot](https://quillbot.com/) to improve your writing.

qwe661234:
> Do you understand what [Wrap the body at 72 characters](https://cbea.ms/git-commit/#wrap-72) means?

I see, I need to change line when characters in a line more than 72.

PR #95, opened 2022-12-12
Use TCO of C compiler to speed up emulation

qwe661234:
We need to refactor the function emulate to a recursive version for meeting the requirement of tail-call optimization(TCO). To achieve this, I add a variable is_tail to the struct rv_insn_t to help us determine whether the basic block is terminate or not. As a result, we can use this variable to rewrite function emulate into a self-recursive function.

Running coremark and dhrystone benchmark now produces faster results than it did previously.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ OUT ?= build
 BIN := $(OUT)/rv32emu
 
 CFLAGS = -std=gnu99 -O2 -Wall -Wextra
+CFLAGS += -Wno-unused-label
 CFLAGS += -include src/common.h
 
 # Set the default stack pointer
@@ -87,6 +88,10 @@ LDFLAGS += $(GDBSTUB_LIB)
 gdbstub-test: $(BIN)
 	$(Q)tests/gdbstub.sh && $(call notice, [OK])
 endif
+
+# For tail-call elimination, we need a specific set of build flags applied.
+# FIXME: On macOS + Apple Silicon, -fno-stack-protector might have a negative impact.
+$(OUT)/emulate.o: CFLAGS += -fomit-frame-pointer -fno-stack-check -fno-stack-protector
 
 # Clear the .DEFAULT_GOAL special variable, so that the following turns
 # to the first target after .DEFAULT_GOAL is not set.
--- a/src/common.h
+++ b/src/common.h
@@ -24,6 +24,16 @@
 #define __ALIGNED(x)
 #endif
 
+/* There is no tail-call optimization(TCO) in non-optimized builds. To work
+ * around this, we attempts to use a compiler attribute called musttail that
+ * forces the compiler to TCO even when optimizations aren't on.
+ */
+#if defined(__has_attribute) && __has_attribute(musttail)
+#define MUST_TAIL __attribute__((musttail))
+#else
+#define MUST_TAIL
+#endif
+
 /* Pattern Matching for C macros.
  * https://github.com/pfultz2/Cloak/wiki/C-Preprocessor-tricks,-tips,-and-idioms
  */
--- a/src/decode.h
+++ b/src/decode.h
@@ -8,158 +8,160 @@
 #include <stdbool.h>
 #include <stdint.h>
 
-/* RISC-V instruction list */
+#include "riscv.h"
+
+/* RISC-V instruction list in format _(instruction-name, can-branch) */
 /* clang-format off */
 #define RISCV_INSN_LIST                    \
-    _(nop)                                 \
+    _(nop, 0)                              \
     /* RV32I Base Instruction Set */       \
-    _(lui)                                 \
-    _(auipc)                               \
-    _(jal)                                 \
-    _(jalr)                                \
-    _(beq)                                 \
-    _(bne)                                 \
-    _(blt)                                 \
-    _(bge)                                 \
-    _(bltu)                                \
-    _(bgeu)                                \
-    _(lb)                                  \
-    _(lh)                                  \
-    _(lw)                                  \
-    _(lbu)                                 \
-    _(lhu)                                 \
-    _(sb)                                  \
-    _(sh)                                  \
-    _(sw)                                  \
-    _(addi)                                \
-    _(slti)                                \
-    _(sltiu)                               \
-    _(xori)                                \
-    _(ori)                                 \
-    _(andi)                                \
-    _(slli)                                \
-    _(srli)                                \
-    _(srai)                                \
-    _(add)                                 \
-    _(sub)                                 \
-    _(sll)                                 \
-    _(slt)                                 \
-    _(sltu)                                \
-    _(xor)                                 \
-    _(srl)                                 \
-    _(sra)                                 \
-    _(or)                                  \
-    _(and)                                 \
-    _(ecall)                               \
-    _(ebreak)                              \
+    _(lui, 0)                              \
+    _(auipc, 0)                            \
+    _(jal, 1)                              \
+    _(jalr, 1)                             \
+    _(beq, 1)                              \
+    _(bne, 1)                              \
+    _(blt, 1)                              \
+    _(bge, 1)                              \
+    _(bltu, 1)                             \
+    _(bgeu, 1)                             \
+    _(lb, 0)                               \
+    _(lh, 0)                               \
+    _(lw, 0)                               \
+    _(lbu, 0)                              \
+    _(lhu, 0)                              \
+    _(sb, 0)                               \
+    _(sh, 0)                               \
+    _(sw, 0)                               \
+    _(addi, 0)                             \
+    _(slti, 0)                             \
+    _(sltiu, 0)                            \
+    _(xori, 0)                             \
+    _(ori, 0)                              \
+    _(andi, 0)                             \
+    _(slli, 0)                             \
+    _(srli, 0)                             \
+    _(srai, 0)                             \
+    _(add, 0)                              \
+    _(sub, 0)                              \
+    _(sll, 0)                              \
+    _(slt, 0)                              \
+    _(sltu, 0)                             \
+    _(xor, 0)                              \
+    _(srl, 0)                              \
+    _(sra, 0)                              \
+    _(or, 0)                               \
+    _(and, 0)                              \
+    _(ecall, 1)                            \
+    _(ebreak, 1)                           \
     /* RISC-V Privileged Instruction */    \
-    _(wfi)                                 \
-    _(uret)                                \
-    _(sret)                                \
-    _(hret)                                \
-    _(mret)                                \
+    _(wfi, 0)                              \
+    _(uret, 0)                             \
+    _(sret, 0)                             \
+    _(hret, 0)                             \
+    _(mret, 1)                             \
     /* RV32 Zifencei Standard Extension */ \
     IIF(RV32_HAS(Zifencei))(               \
-        _(fencei)                          \
+        _(fencei, 0)                       \
     )                                      \
     /* RV32 Zicsr Standard Extension */    \
     IIF(RV32_HAS(Zicsr))(                  \
-        _(csrrw)                           \
-        _(csrrs)                           \
-        _(csrrc)                           \
-        _(csrrwi)                          \
-        _(csrrsi)                          \
-        _(csrrci)                          \
+        _(csrrw, 0)                        \
+        _(csrrs, 0)                        \
+        _(csrrc, 0)                        \
+        _(csrrwi, 0)                       \
+        _(csrrsi, 0)                       \
+        _(csrrci, 0)                       \
     )                                      \
     /* RV32M Standard Extension */         \
     IIF(RV32_HAS(EXT_M))(                  \
-        _(mul)                             \
-        _(mulh)                            \
-        _(mulhsu)                          \
-        _(mulhu)                           \
-        _(div)                             \
-        _(divu)                            \
-        _(rem)                             \
-        _(remu)                            \
+        _(mul, 0)                          \
+        _(mulh, 0)                         \
+        _(mulhsu, 0)                       \
+        _(mulhu, 0)                        \
+        _(div, 0)                          \
+        _(divu, 0)                         \
+        _(rem, 0)                          \
+        _(remu, 0)                         \
     )                                      \
     /* RV32A Standard Extension */         \
     IIF(RV32_HAS(EXT_A))(                  \
-        _(lrw)                             \
-        _(scw)                             \
-        _(amoswapw)                        \
-        _(amoaddw)                         \
-        _(amoxorw)                         \
-        _(amoandw)                         \
-        _(amoorw)                          \
-        _(amominw)                         \
-        _(amomaxw)                         \
-        _(amominuw)                        \
-        _(amomaxuw)                        \
+        _(lrw, 0)                          \
+        _(scw, 0)                          \
+        _(amoswapw, 0)                     \
+        _(amoaddw, 0)                      \
+        _(amoxorw, 0)                      \
+        _(amoandw, 0)                      \
+        _(amoorw, 0)                       \
+        _(amominw, 0)                      \
+        _(amomaxw, 0)                      \
+        _(amominuw, 0)                     \
+        _(amomaxuw, 0)                     \
     )                                      \
     /* RV32F Standard Extension */         \
     IIF(RV32_HAS(EXT_F))(                  \
-        _(flw)                             \
-        _(fsw)                             \
-        _(fmadds)                          \
-        _(fmsubs)                          \
-        _(fnmsubs)                         \
-        _(fnmadds)                         \
-        _(fadds)                           \
-        _(fsubs)                           \
-        _(fmuls)                           \
-        _(fdivs)                           \
-        _(fsqrts)                          \
-        _(fsgnjs)                          \
-        _(fsgnjns)                         \
-        _(fsgnjxs)                         \
-        _(fmins)                           \
-        _(fmaxs)                           \
-        _(fcvtws)                          \
-        _(fcvtwus)                         \
-        _(fmvxw)                           \
-        _(feqs)                            \
-        _(flts)                            \
-        _(fles)                            \
-        _(fclasss)                         \
-        _(fcvtsw)                          \
-        _(fcvtswu)                         \
-        _(fmvwx)                           \
+        _(flw, 0)                          \
+        _(fsw, 0)                          \
+        _(fmadds, 0)                       \
+        _(fmsubs, 0)                       \
+        _(fnmsubs, 0)                      \
+        _(fnmadds, 0)                      \
+        _(fadds, 0)                        \
+        _(fsubs, 0)                        \
+        _(fmuls, 0)                        \
+        _(fdivs, 0)                        \
+        _(fsqrts, 0)                       \
+        _(fsgnjs, 0)                       \
+        _(fsgnjns, 0)                      \
+        _(fsgnjxs, 0)                      \
+        _(fmins, 0)                        \
+        _(fmaxs, 0)                        \
+        _(fcvtws, 0)                       \
+        _(fcvtwus, 0)                      \
+        _(fmvxw, 0)                        \
+        _(feqs, 0)                         \
+        _(flts, 0)                         \
+        _(fles, 0)                         \
+        _(fclasss, 0)                      \
+        _(fcvtsw, 0)                       \
+        _(fcvtswu, 0)                      \
+        _(fmvwx, 0)                        \
     )                                      \
     /* RV32C Standard Extension */         \
     IIF(RV32_HAS(EXT_C))(                  \
-        _(caddi4spn)                       \
-        _(clw)                             \
-        _(csw)                             \
-        _(cnop)                            \
-        _(caddi)                           \
-        _(cjal)                            \
-        _(cli)                             \
-        _(caddi16sp)                       \
-        _(clui)                            \
-        _(csrli)                           \
-        _(csrai)                           \
-        _(candi)                           \
-        _(csub)                            \
-        _(cxor)                            \
-        _(cor)                             \
-        _(cand)                            \
-        _(cj)                              \
-        _(cbeqz)                           \
-        _(cbnez)                           \
-        _(cslli)                           \
-        _(clwsp)                           \
-        _(cjr)                             \
-        _(cmv)                             \
-        _(cebreak)                         \
-        _(cjalr)                           \
-        _(cadd)                            \
-        _(cswsp)                           \
+        _(caddi4spn, 0)                    \
+        _(clw, 0)                          \
+        _(csw, 0)                          \
+        _(cnop, 0)                         \
+        _(caddi, 0)                        \
+        _(cjal, 1)                         \
+        _(cli, 0)                          \
+        _(caddi16sp, 0)                    \
+        _(clui, 0)                         \
+        _(csrli, 0)                        \
+        _(csrai, 0)                        \
+        _(candi, 0)                        \
+        _(csub, 0)                         \
+        _(cxor, 0)                         \
+        _(cor, 0)                          \
+        _(cand, 0)                         \
+        _(cj, 1)                           \
+        _(cbeqz, 1)                        \
+        _(cbnez, 1)                        \
+        _(cslli, 0)                        \
+        _(clwsp, 0)                        \
+        _(cjr, 1)                          \
+        _(cmv, 0)                          \
+        _(cebreak, 1)                      \
+        _(cjalr, 1)                        \
+        _(cadd, 0)                         \
+        _(cswsp, 0)                        \
     )
 /* clang-format on */
 
 /* IR list */
 enum {
-#define _(inst) rv_insn_##inst,
+#define _(inst, can_branch) rv_insn_##inst,
     RISCV_INSN_LIST
 #undef _
 };
@@ -226,7 +228,7 @@ enum {
     INSN_32 = 4,
 };
 
-typedef struct {
+typedef struct rv_insn {
     union {
         int32_t imm;
         uint8_t rs3;
@@ -241,6 +243,22 @@ typedef struct {
 
     /* instruction length */
     uint8_t insn_len;
+
+    /* According to tail-call optimization (TCO), if a C function ends with
+     * a function call to another function or itself and simply returns that
+     * function's result, the compiler can substitute a simple jump to the
+     * other function for the 'call' and 'return' instructions . The self
+     * -recursive function can therefore use the same function stack frame.
+     *
+     * Using member tailcall, we can tell whether an IR is the final IR in
+     * a basic block. Additionally, member 'impl' allows us to invoke next
+     * instruction emulation directly without computing the jumping address.
+     * In order to enable the compiler to perform TCO, we can use these two
+     * members to rewrite all instruction emulations into a self-recursive
+     * version.
+     */
+    bool tailcall;
+    bool (*impl)(riscv_t *, const struct rv_insn *);
 } rv_insn_t;
 
 /* decode the RISC-V instruction */
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -254,1119 +254,991 @@ static inline bool insn_is_misaligned(uint32_t pc)
     );
 }
 
-/* execute a basic block */
-static bool emulate(riscv_t *rv, const block_t *block)
-{
-#if RV32_HAS(COMPUTED_GOTO)
-    static const void *dispatch_table[] = {
-#define _(inst) [rv_insn_##inst] = &&do_##inst,
-        RISCV_INSN_LIST
+/* can-branch information for each RISC-V instruction */
+enum {
+#define _(inst, can_branch) __rv_insn_##inst##_canbranch = can_branch,
+    RISCV_INSN_LIST
 #undef _
-    };
+};
 
-#define DISPATCH()              \
-    /* enforce zero register */ \
-    rv->X[rv_reg_zero] = 0;     \
-    /* current IR */            \
-    ir = block->ir + index++;   \
-    /* jump */                  \
-    goto *dispatch_table[ir->opcode];
+#define RVOP(inst, code)                                                  \
+    static bool do_##inst(riscv_t *rv UNUSED, const rv_insn_t *ir UNUSED) \
+    {                                                                     \
+        rv->X[rv_reg_zero] = 0;                                           \
+        code;                                                             \
+        if (__rv_insn_##inst##_canbranch) {                               \
+            /* can branch */                                              \
+            rv->csr_cycle++;                                              \
+            return true;                                                  \
+        }                                                                 \
+    nextop:                                                               \
+        rv->PC += ir->insn_len;                                           \
+        if (ir->tailcall)                                                 \
+            return true;                                                  \
+        const rv_insn_t *next = ir + 1;                                   \
+        MUST_TAIL return next->impl(rv, next);                            \
+    }
 
-/* clang-format off */
-#define _(inst, code)                    \
-    do_##inst: code                      \
-    /* step over instruction */          \
-    rv->PC += ir->insn_len;              \
-    /* increment the cycles CSR */       \
-    rv->csr_cycle++;                     \
-    /* all instructions have executed */ \
-    if (unlikely(index == n_insn))       \
-        return true;                     \
-    DISPATCH()
-/* clang-format on */
-#define EPILOGUE()
+/* RV32I Base Instruction Set */
 
-#else /* !RV32_HAS(COMPUTED_GOTO) */
-#define DISPATCH()                          \
-    for (uint32_t i = 0; i < n_insn; i++) { \
-        ir = block->ir + i;                 \
-        /* enforce zero register */         \
-        rv->X[rv_reg_zero] = 0;             \
-        switch (ir->opcode) {
-/* clang-format off */
-#define _(inst, code)         \
-    case rv_insn_##inst: code \
-        break;
-#define EPILOGUE()                          \
-        }                                   \
-        /* step over instruction */         \
-        rv->PC += ir->insn_len;             \
-        /* increment the cycles csr */      \
-        rv->csr_cycle++;                    \
-    }                                       \
-    return true;
-/* clang-format on */
-#endif /* RV32_HAS(COMPUTED_GOTO) */
+/* Internal */
+RVOP(nop, {/* no operation */});
 
-    const uint32_t n_insn = block->n_insn;
-    rv_insn_t *ir;
+/* LUI is used to build 32-bit constants and uses the U-type format. LUI
+ * places the U-immediate value in the top 20 bits of the destination
+ * register rd, filling in the lowest 12 bits with zeros. The 32-bit
+ * result is sign-extended to 64 bits.
+ */
+RVOP(lui, { rv->X[ir->rd] = ir->imm; })
 
-#if RV32_HAS(COMPUTED_GOTO)
-    /* current index in block */
-    uint32_t index = 0;
+/* AUIPC is used to build pc-relative addresses and uses the U-type
+ * format. AUIPC forms a 32-bit offset from the 20-bit U-immediate,
+ * filling in the lowest 12 bits with zeros, adds this offset to the
+ * address of the AUIPC instruction, then places the result in register
+ * rd.
+ */
+RVOP(auipc, { rv->X[ir->rd] = ir->imm + rv->PC; })
+
+/* JAL: Jump and Link
+ * store successor instruction address into rd.
+ * add next J imm (offset) to pc.
+ */
+RVOP(jal, {
+    const uint32_t pc = rv->PC;
+    /* Jump */
+    rv->PC += ir->imm;
+    /* link with return address */
+    if (ir->rd)
+        rv->X[ir->rd] = pc + ir->insn_len;
+    /* check instruction misaligned */
+    if (unlikely(insn_is_misaligned(rv->PC))) {
+        rv->compressed = false;
+        rv_except_insn_misaligned(rv, pc);
+        return false;
+    }
+})
+
+/*The indirect jump instruction JALR uses the I-type encoding. The
+ * target address is obtained by adding the sign-extended 12-bit
+ * I-immediate to the register rs1, then setting the least-significant
+ * bit of the result to zero. The address of the instruction following
+ * the jump (pc+4) is written to register rd. Register x0 can be used as
+ * the destination if the result is not required.
+ */
+RVOP(jalr, {
+    const uint32_t pc = rv->PC;
+    /* jump */
+    rv->PC = (rv->X[ir->rs1] + ir->imm) & ~1U;
+    /* link */
+    if (ir->rd)
+        rv->X[ir->rd] = pc + ir->insn_len;
+    /* check instruction misaligned */
+    if (unlikely(insn_is_misaligned(rv->PC))) {
+        rv->compressed = false;
+        rv_except_insn_misaligned(rv, pc);
+        return false;
+    }
+})
+
+/* BEQ: Branch if Equal */
+RVOP(beq, {
+    const uint32_t pc = rv->PC;
+    if (rv->X[ir->rs1] != rv->X[ir->rs2])
+        goto nextop;
+    rv->PC += ir->imm;
+    /* check instruction misaligned */
+    if (unlikely(insn_is_misaligned(rv->PC))) {
+        rv->compressed = false;
+        rv_except_insn_misaligned(rv, pc);
+        return false;
+    }
+})
+
+/* BNE: Branch if Not Equal */
+RVOP(bne, {
+    const uint32_t pc = rv->PC;
+    if (rv->X[ir->rs1] == rv->X[ir->rs2])
+        goto nextop;
+    rv->PC += ir->imm;
+    /* check instruction misaligned */
+    if (unlikely(insn_is_misaligned(rv->PC))) {
+        rv->compressed = false;
+        rv_except_insn_misaligned(rv, pc);
+        return false;
+    }
+})
+
+/* BLT: Branch if Less Than */
+RVOP(blt, {
+    const uint32_t pc = rv->PC;
+    if ((int32_t) rv->X[ir->rs1] >= (int32_t) rv->X[ir->rs2])
+        goto nextop;
+    rv->PC += ir->imm;
+    /* check instruction misaligned */
+    if (unlikely(insn_is_misaligned(rv->PC))) {
+        rv->compressed = false;
+        rv_except_insn_misaligned(rv, pc);
+        return false;
+    }
+})
+
+/* BGE: Branch if Greater Than */
+RVOP(bge, {
+    const uint32_t pc = rv->PC;
+    if ((int32_t) rv->X[ir->rs1] < (int32_t) rv->X[ir->rs2])
+        goto nextop;
+    rv->PC += ir->imm;
+    /* check instruction misaligned */
+    if (unlikely(insn_is_misaligned(rv->PC))) {
+        rv->compressed = false;
+        rv_except_insn_misaligned(rv, pc);
+        return false;
+    }
+})
+
+/* BLTU: Branch if Less Than Unsigned */
+RVOP(bltu, {
+    const uint32_t pc = rv->PC;
+    if (rv->X[ir->rs1] >= rv->X[ir->rs2])
+        goto nextop;
+    rv->PC += ir->imm;
+    /* check instruction misaligned */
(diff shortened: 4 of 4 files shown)


src/decode.h
```diff
@@ -240,6 +240,7 @@ typedef struct {
 
     /* instruction length */
     uint8_t insn_len;
+    uint8_t is_tail;
```
jserv: `is_tail` should be of type `bool`.

jserv (changes requested):
Run `clang-format-12 -i src/*.[ch]` to indent.
Please read https://github.com/sysprog21/rv32emu/blob/master/CONTRIBUTING.md carefully.

jserv:
> Running coremark and dhrystone benchmark now produces faster results than it did previously.

Show the numbers for reference purpose. In addition, explain explicitly.

jserv:
For clang, I need a specific set of build flags applied, otherwise the emulation would proceed much slower.
```diff
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ OUT ?= build
 BIN := $(OUT)/rv32emu
 
 CFLAGS = -std=gnu99 -O2 -Wall -Wextra
+CFLAGS += -fomit-frame-pointer -fno-stack-check -fno-stack-protector
 CFLAGS += -include src/common.h
 
 # Set the default stack pointer
```

jserv:
> We need to refactor the function emulate to a recursive version for meeting the requirement of tail-call optimization(TCO). To achieve this, I add a variable is_tail to the struct rv_insn_t to help us determine whether the basic block is terminate or not. As a result, we can use this variable to rewrite function emulate into a self-recursive function.

Quote from Wikipedia:
> code refactoring is the process of restructuring existing [computer code](https://en.wikipedia.org/wiki/Computer_code)—changing the [factoring](https://en.wikipedia.org/wiki/Decomposition_(computer_science))—without changing its external behavior. Refactoring is intended to improve the design, structure, and/or implementation of the [software](https://en.wikipedia.org/wiki/Software) (its [non-functional](https://en.wikipedia.org/wiki/Non-functional_requirement) attributes), while preserving its [functionality](https://en.wikipedia.org/wiki/Functional_requirement).

"Refactoring" should not be considered because you do change the dispatching and instruction emulation behavior.

jserv:
By the way, run `git commit --amend --author=` to specify your name in Hanyu Pinyin.

src/emulate.c
```diff
@@ -1240,27 +1244,6 @@ static block_t *block_find(const block_map_t *map, const uint32_t addr)
...
-/* execute a basic block */
-static bool block_emulate(riscv_t *rv, const block_t *block)
-{
-    const uint32_t n_insn = block->n_insn;
-    const rv_insn_t *ir = block->ir;
-
-    /* execute the block */
-    for (uint32_t i = 0; i < n_insn; i++) {
-        /* enforce zero register */
-        rv->X[rv_reg_zero] = 0;
```
Risheng1128: Register x0 should set zero before emulator executes every instruction. However, [riscv-arch-test](https://github.com/riscv-non-isa/riscv-arch-test) will be failed.

src/emulate.c
```diff
@@ -1240,27 +1244,6 @@ static block_t *block_find(const block_map_t *map, const uint32_t addr)
...
-
-        /* execute the instruction */
-        if (!emulate(rv, ir + i))
-            return false;
-
-        /* increment the cycles csr */
-        rv->csr_cycle++;
```
Risheng1128: `rv->csr_cycle` should be retain to compute the cycle.

src/decode.h
```diff
@@ -240,6 +240,7 @@ typedef struct {
 
     /* instruction length */
     uint8_t insn_len;
+    bool is_tail;
```
jserv: `is_tail` is a bit confusing. Someone might think it as "tail" in queue data structure. How about `bool tailcall`?

jserv:
You should provide performance metrics for both clang and gcc, at least running on Ubuntu Linux.

jserv:
For clang support (the major compiler on macOS), we can explicitly define MUST_TAIL in `src/common.h`
```c
#if defined(__has_attribute) && __has_attribute(musttail)
/* Clang requires a special tail recursion attribute to use tail recursion. */
#define MUST_TAIL __attribute__((musttail))
#else
#define MUST_TAIL
#endif
```

See https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/port_def.inc

jserv:
The formal name in git commit message should be "Yen-Fu Chen". See [Chinese punctuation](https://en.wikipedia.org/wiki/Chinese_punctuation) for details. Throughout every non-trivial work, you should constantly identify yourself.

jserv:
> You should provide performance metrics for both clang and gcc, at least running on Ubuntu Linux.

TODO:
1. Ask @eecheng87 for eMag (Arm-based Workstation) access.
2. We are concerned about the performance metrics of Clang versus GCC on x86-64 and AArch64 architectures, respectively.
3. `__attribute__((musttail))` must be enabled in clang-based builds, ensuring visible benefits.
4. Build generic code with `-O2` (no more aggressive optimization order is allowed at the moment). We may use `__attribute__((optimize("O3")))` for certain functions. Be aware of negative impact for debugging.

src/decode.h
jserv: Add informative comment for newly-added member.

src/emulate.c
```diff
@@ -1288,6 +1289,7 @@ static void block_translate(riscv_t *rv, block_t *block)
         if (insn_is_branch(ir->opcode))
             break;
     }
+    (block->ir + block->n_insn - 1)->is_tail = true;
```
jserv: Rewrite as `block->ir[block->n_insn - 1].is_tail = true;` which is shorter and better for bounds checks.

jserv (changes requested):
Clarify the performance metrics:
```
-----------------------------------------------------------------------
Test environment3: Ubuntu Linux 20.04 on ThunderX2
Compiler: gcc 9.4.0
Coremark test result:
Previous: 260.543173 Iterations/Sec
Now: 286.504547 Iterations/Sec
-----------------------------------------------------------------------
Test environment4: Ubuntu Linux 20.04 on ThunderX2
Compiler: gcc 9.4.0
Coremark test result:
Previous: 239.773443 Iterations/Sec
Now: 285.154751 Iterations/Sec
```

We shall check clang vs. gcc. By the way, ThunderX2 is based on older microarchitecture. Arm64-specific experiments should be carried out on [eMag](https://en.wikichip.org/wiki/ampere_computing/emag). Drop ThunderX2 related items.

src/common.h
```diff
@@ -39,3 +39,10 @@
 #define IIF_0(t, ...) __VA_ARGS__
 /* run the 1st parameter */
 #define IIF_1(t, ...) t
+
+#if defined(__has_attribute) && __has_attribute(musttail)
```
jserv: Move these line right after the part of `Alignment macro`. And, add comments for annotation.

src/decode.h
```diff
@@ -240,6 +240,8 @@ typedef struct {
 
     /* instruction length */
     uint8_t insn_len;
+    /* To determine whether the IR is last IR in a basic block */
```
jserv: More comments for the idea behind tail call optimizations.

Makefile
```diff
@@ -5,6 +5,7 @@ OUT ?= build
 BIN := $(OUT)/rv32emu
 
 CFLAGS = -std=gnu99 -O2 -Wall -Wextra
+CFLAGS += -fomit-frame-pointer -fno-stack-check -fno-stack-protector
```
jserv: Move this line right after computed goto part. That is,
```
# Whether to enable computed goto
ENABLE_COMPUTED_GOTO ?= 1
...

# For tail-call elimination, we need a specific set of build flags applied.                                                                                   
$(OUT)/emulate.o: CFLAGS += -fomit-frame-pointer -fno-stack-check -fno-stack-protector
```

We shall carefully specify the compilation options.

src/decode.h
jserv: You can take some wording in [tails](https://github.com/snej/tails) which would describe how TCO works.

src/decode.h
```diff
@@ -240,6 +240,17 @@ typedef struct {
 
     /* instruction length */
     uint8_t insn_len;
+    /* According to Tail-call optimization(TCO), if a C function ends with a
```
jserv: [QuillBot](https://quillbot.com/) can help you improve your writing!

jserv:
Performance metrics

| Microprocessor | compiler | CoreMark w/ commit 285a988dc269c43d9e9384d2119a9c70148ae952 | CoreMark w/ PR #95 | Speedup |
|:---:|:---:|:---:|:---:|:---:|
| Core i7-8700 | gcc-9 | 870.317 iter/s | 920.675 iter/s | +5.8% |
| Core i7-8700 | clang-16 | 805.702 iter/s | 849.445 iter/s | +5.4% |
| eMag 8180 | gcc-11 | 311.436 iter/s | 313.900 iter/s | +0.8% |
| eMag 8180 | clang-16 | 273.265 iter/s | 297.255 iter/s |  +8.8% |

The experiments should be amended as follows to make the comparisons more self-explanatory.
* Unify gcc versioning - use latest stable release. e.g., `gcc-12`. You can install via [Ubuntu ppa](https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test).
* Unify clang versioning - use latest stable release. e.g., `clang-15`. You can install [prebuilt packages](https://apt.llvm.org/).
* Instead of running CoreMark once, you should execute it numerous times and average the results.

Then, the explanation of the gcc-aarch64 build's lack of TCO would then be revealed. At the moment, [change the stage of this pull request](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request) to "draft" because it is not as effective as computed-goto in terms of the ratio for improvements.

jserv:
Let's investigate by validating [wasm3's statement](https://github.com/wasm3/wasm3/blob/main/docs/Interpreter.md).
> once an M3 operation is optimized, it doesn't need the regular stack (no mucking with the ebp/esp registers). This is the case for 90% of the opcodes. Branching and call operations do require stack variables. Therefore, execution can't march forward indefinitely; the stack would eventually overflow.

[Coroutine](https://www.google.com/search?q=coroutine+wikipedia) (or "fiber" in wasm3's context) is used in its execution, so that control can effortlessly switch to the "main" fiber. Since the entire function of a fiber is to save state, no explicit saving of state is required. We have to check the flow control and stack usage again in wasm3. Then, revisit the proposed changes.

qwe661234:
I adhere to the wasm3 implementation, which separates all instruction emulations. I originally had a function table with all of the instruction emulations, but after doing performance analysis, I discovered that it took a long time to calculate the function table. I therefore alter struct rv_insn_t so that we can directly assign instruction emulation to IR with adding member opfunc.

Running coremark benchmark now produces faster results than it did previously, and the test results show below.

| Microprocessor | compiler | CoreMark w/ commit f2da162 | CoreMark w/ PR sysprog21#95 | Speedup |
| -------- | -------- | -------- |-------- |-------- |
| Core i7-8700   | clang-15 |        836.4849530         |         967.195633         |  +13.5%  |
| Core i7-8700   | gcc-11   |        867.3400912         |         955.786538         |  +9.2%   |
| eMag 8180      | clang-15 |        266.5329894         |         330.816788         |  +19.4%  |
| eMag 8180      | gcc-11   |        259.6389222         |         318.623934         |  +18.5%  |

Previously, we had to calculate the jumping address using a method such as switch-case, computed-goto, or function table, but this is no longer necessary.

jserv:
> I adhere to the wasm3 implementation, which separates all instruction emulations. I originally had a function table with all of the instruction emulations, but after doing performance analysis, I discovered that it took a long time to calculate the function table. I therefore alter struct rv_insn_t so that we can directly assign instruction emulation to IR with adding member opfunc.

Rebase the latest master branch and perform minimal changes. Avoid creating `emulate.h`. Also, you can make use of X macro to eliminate code duplication.

src/emulate.h
```diff
@@ -0,0 +1,174 @@
...
+#endif
+
+#include "decode.h"
+#include "riscv.h"
+#include "riscv_private.h"
+#include "utils.h"
+
+#define RVOP(insn) bool rvop_##insn(riscv_t *rv UNUSED, rv_insn_t *ir UNUSED)
+typedef bool (*rvop_t)(riscv_t *rv, rv_insn_t *ir);
+RVOP(lui);
```
jserv: With the X macro technique, the duplication can be eliminated. See `src/emulate.c` for reference.

src/emulate.c
```diff
@@ -254,902 +226,1658 @@ static bool insn_is_misaligned(uint32_t pc)
...
-         * add next J imm (offset) to pc.
-         */
-        const uint32_t pc = rv->PC;
-        /* Jump */
+#define nextOP()                      \
+    rvop_t opfunc = (ir + 1)->opfunc; \
+    MUST_TAIL return opfunc(rv, ir + 1)
+/* RV32I Base Instruction Set */
+RVOP(lui)
+{ /* LUI: Load Upper Immediate */
```
jserv: The comment should be placed before `RVOP`. That is,
```c
/* LUI: Load Upper Immediate */
RVOP(lui)
{
    ...
}
```

jserv:
NOP was added in `master` branch, and you should regard it as the special case without invoking the function. In other words, we just advance the value of pc without taking real actions when NOP appears in IR. NOP is useful for IR level optimizations. For example, the original code sequence:
```
    sw ra, -4(sp)
    sw s0, -8(sp)
    sw s1, -12(sp)
    sw s2, -16(sp)
    sw s3, -20(sp)
    sw s4, -24(sp)
    sw s5, -28(sp)
    mv s0, a0
    mv s1, a1
    mv s2, a2
    mv s3, a3
    addi sp, sp, -64
```

The above can be replaced with super instruction `quick_push`:
```
    quick_push {ra,s0-s5}, {a0-a3}, -64
    nop
    nop
```

NOP is preserved to satisfy PC-relative addressing.

src/decode.c
```diff
@@ -407,19 +408,19 @@ static inline bool op_load(rv_insn_t *ir, const uint32_t insn)
     /* dispatch from funct3 field */
     switch (decode_funct3(insn)) {
     case 0: /* LB: Load Byte */
-        ir->opcode = rv_insn_lb;
+        ir->opfunc = (rvop_t) &rvop_lb;
```
jserv: The assignment for `opfunc` should be inside basic block generation.

src/emulate.h
```diff
@@ -0,0 +1,174 @@
...
+#if RV32_HAS(GDBSTUB)
+extern struct target_ops gdbstub_ops;
+#endif
+
+#include "decode.h"
+#include "riscv.h"
+#include "riscv_private.h"
+#include "utils.h"
+
+#define RVOP(insn) bool rvop_##insn(riscv_t *rv UNUSED, rv_insn_t *ir UNUSED)
```
jserv: For the sake of abstraction, the opcode implementations should be never visible outside `emulate.c`. Instead, you can introduce a helper function to fill the function pointers.

Makefile
```diff
@@ -65,7 +65,8 @@ endif
 $(call set-feature, COMPUTED_GOTO)
 ifeq ($(call has, COMPUTED_GOTO), 1)
 ifeq ("$(CC_IS_GCC)", "1")
-$(OUT)/emulate.o: CFLAGS += -fno-gcse -fno-crossjumping
+# For tail-call elimination, we need a specific set of build flags applied.
```
jserv: Move the flags outside `COMPUTED_GOTO` block.

jserv:
Can you eliminate the statement `rv->X[rv_reg_zero] = 0;` in each opcode implementation function since we can check in decoding phase?

src/decode.c
jserv: When used in conjunction with computed-goto, we can assign "opfunc" to a function's corresponding label rather than a function pointer. As a result, we may reuse the compiled code and even get rid of the cost of early function calls.

qwe661234: The dispatch_table and label are defined in function emulate, I have no idea to assign a function's corresponding label to "opfunc" inside basic block generation.

jserv: No problem. I will discuss [code-copying](http://www.sable.mcgill.ca/publications/papers/2008-2/paper.pdf) approach later. At present, you can proceed the minimal changes on top of the latest `master` branch. The separation among different module is vital because we are about to implement JIT compilation sooner or later.

Makefile
jserv: Don't touch COMPUTED_GOTO block.

src/common.h
jserv: Macro `MUST_TAIL` should appear before pattern matching macros.

src/decode.h
```diff
@@ -241,6 +241,18 @@ typedef struct {
...
+     * function's result, the compiler can substitute a simple jump to the
+     * other function for the 'call' and 'return' instructions . The self
+     * -recursive function can therefore use the same function stack frame.
+     *
+     * Member tailcall help us to determine whether the IR is last IR in
+     * a basic block. We can use this member to rewrite the function emulate
+     * into a self-recursive version, allowing the compiler to do TCO.
+     */
+    bool tailcall;
+    void *opfunc;
```
jserv: Declare `opfunc` with typedef.

src/emulate.c
```diff
@@ -254,100 +254,112 @@ static inline bool insn_is_misaligned(uint32_t pc)
...
-#endif /* RV32_HAS(COMPUTED_GOTO) */
-
-    const uint32_t n_insn = block->n_insn;
-    rv_insn_t *ir;
-
-#if RV32_HAS(COMPUTED_GOTO)
-    /* current index in block */
-    uint32_t index = 0;
-#endif
+#define RVOP(insn) bool rvop_##insn(riscv_t *rv UNUSED, rv_insn_t *ir UNUSED)
```
jserv: Add `static` modifier.

jserv: Rewrite `RVOP` as following:
```c
#define RVOP(inst, code) \
static bool do_##inst(riscv_t *rv UNUSED, rv_insn_t *ir UNUSED)
{
    code
    if (ir->tailcall)
        return true;
    nextOP();
}
```

`RVOP` can be redefined if we are going to introduce C2MIR as `wip/jit` branch does.

src/emulate.c
```diff
@@ -254,100 +254,112 @@ static inline bool insn_is_misaligned(uint32_t pc)
...
 
-    /* Internal */
-    _(nop, /* no operation */)
-
-    /* LUI (Load Upper Immediate) is used to build 32-bit constants and uses the
-     * U-type format. LUI places the U-immediate value in the top 20 bits of the
-     * destination register rd, filling in the lowest 12 bits with zeros. The
-     * 32-bit result is sign-extended to 64 bits.
+/* Internal */
+RVOP(nop)
```
jserv: `nop` is a special case. Think of the way with minimal dispatch overhead.

jserv (changes requested):
Measure with gcc-12, which is the latest stable GCC release.

src/emulate.c
```diff
@@ -254,100 +254,112 @@ static inline bool insn_is_misaligned(uint32_t pc)
...
+     */
+    rv->X[rv_reg_zero] = 0;
+    const uint32_t pc = rv->PC;
+    /* jump */
+    rv->PC = (rv->X[ir->rs1] + ir->imm) & ~1U;
+    /* link */
+    if (ir->rd)
+        rv->X[ir->rd] = pc + ir->insn_len;
+    /* check instruction misaligned */
+    if (insn_is_misaligned(rv->PC)) {
```
jserv: Add `unlikely` macro back.

jserv: You should check the changes line by line to ensure the previous hints to optimizer are still alive.

src/emulate.c
```diff
@@ -384,959 +394,1650 @@ static bool emulate(riscv_t *rv, const block_t *block)
...
+    rv->io.mem_write_w(rv, addr, rv->X[ir->rs2]);
+    rv->PC += ir->insn_len;
+    rv->csr_cycle++;
+    if (ir->tailcall)
+        return true;
+    nextOP();
+}
+#endif
 
+static void *dispatch_table[] = {
```
jserv: Should be `static const`.

src/emulate.c
```diff
@@ -384,959 +394,1650 @@ static bool emulate(riscv_t *rv, const block_t *block)
...
+{
+    /* nothing */
+    rv->PC += ir->insn_len;
+    rv->csr_cycle++;
+    if (ir->tailcall)
+        return true;
+    nextOP();
+}
+/* C.ADDI */
+RVOP(caddi)
```
jserv: Should be
```
/* C.ADDI adds the non-zero sign-extended 6-bit immediate to the value
 * in register rd then writes the result to rd. C.ADDI expands into addi
 * rd, rd, nzimm[5:0]. C.ADDI is only valid when rd̸=x0. The code point
 * with both rd=x0 and nzimm=0 encodes the C.NOP instruction; the
 * remaining code points with either rd=x0 or nzimm=0 encode HINTs.
 */
RVOP(caddi)
{
    ...
}
```

src/emulate.c
```diff
@@ -254,100 +254,112 @@ static inline bool insn_is_misaligned(uint32_t pc)
...
-     * 32-bit result is sign-extended to 64 bits.
+/* Internal */
+RVOP(nop)
+{
+    /* no operation */
+    if (ir->tailcall)
+        return true;
+    nextOP();
+}
+/* LUI: Load Upper Immediate */
```
jserv: Ensure that a blank line separates each listing of functions.

jserv:
Use the imperative style and third person perspective in git commit messages.

src/emulate.c
```diff
@@ -1516,8 +2218,9 @@ void rv_step(riscv_t *rv, int32_t cycles)
         /* we should have a block by now */
         assert(block);
 
+        rvop_t func = block->ir->opfunc;
         /* execute the block */
-        if (!emulate(rv, block))
+        if (!func(rv, block->ir))
```
jserv: Simply rewrite as following:
```c
    rv_insn_t *ir = block->ir;
    if (unlikely(!(ir->opfunc)(rv, ir)))
        break;
```

src/decode.h
qwe661234: The type of `opfunc` is `(*func)(riscv_t *rv, rv_insn_t *ir)`, it need the definition of `rv_insn_t`, but `opfunc` is also defined in rv_insn_t. That's the reason why I choose void type here.

jserv: Do `#include "riscv.h"` and apply the change below:
```diff
@@ -226,7 +228,7 @@ enum {
     INSN_32 = 4,
 };
 
-typedef struct {
+typedef struct rv_insn {
     union {
         int32_t imm;
         uint8_t rs3;
@@ -252,7 +254,7 @@ typedef struct {
      * into a self-recursive version, allowing the compiler to do TCO.
      */
     bool tailcall;
-    void *opfunc;
+    bool (*opfunc)(riscv_t *, const struct rv_insn *);
 } rv_insn_t;
 
 /* decode the RISC-V instruction */
```

jserv: Alternatively, you can combine `tailcall` and `opfunc` into a single member in `rv_insn_t`. See `src/map.h` and commit 7fb4ff754e48f09e21f0e21137fd28ce59361de5 for details.

Makefile
jserv: Don't touch COMPUTED_GOTO block.

src/common.h
jserv: Macro `MUST_TAIL` should appear before pattern matching macros.

jserv:
When you submit any updates, you should mark the conversation as "resolved"; otherwise, you will always leave them for tracking.

src/emulate.c
jserv: The implementation was incorrect. Fix as following:
```diff
@@ -265,6 +265,8 @@ typedef bool (*rvop_t)(riscv_t *rv, rv_insn_t *ir);
 /* Internal */
 RVOP(nop)
 {
+    rv->PC += ir->insn_len;
+    rv->csr_cycle++;
     /* no operation */
     if (ir->tailcall)
         return true;
```

src/emulate.c
```diff
@@ -254,100 +254,90 @@ static inline bool insn_is_misaligned(uint32_t pc)
...
-
-#if RV32_HAS(COMPUTED_GOTO)
-    /* current index in block */
-    uint32_t index = 0;
-#endif
+#define nextOP() \
+    ir++;        \
+    MUST_TAIL return ir->opfunc(rv, ir)
+#define RVOP(inst, code)                                              \
+    static bool rvop_##inst(riscv_t *rv UNUSED, rv_insn_t *ir UNUSED) \
```
jserv: Replace `rvop_` with `do_` because we might extend IR for several micro-ops. Functions starting with `rvop_` would potentially confuse.

src/emulate.c
```diff
@@ -254,100 +254,90 @@ static inline bool insn_is_misaligned(uint32_t pc)
...
+    {                                                                 \
+        rv->X[rv_reg_zero] = 0;                                       \
+        code rv->PC += ir->insn_len;                                  \
+        rv->csr_cycle++;                                              \
+        if (ir->tailcall)                                             \
+            return true;                                              \
+        nextOP();                                                     \
+    }
+/* RV32I Base Instruction Set */
```
jserv, line 282: Split two lines with a blank one.

src/emulate.c
```diff
@@ -254,100 +254,90 @@ static inline bool insn_is_misaligned(uint32_t pc)
...
+RVOP(auipc, { rv->X[ir->rd] = ir->imm + rv->PC; })
 
-    /* AUIPC (Add Upper Immediate to PC) is used to build pc-relative addresses
-     * and uses the U-type format. AUIPC forms a 32-bit offset from the 20-bit
-     * U-immediate, filling in the lowest 12 bits with zeros, adds this offset
-     * to the address of the AUIPC instruction, then places the result in
-     * register rd.
-     */
-    _(auipc, rv->X[ir->rd] = ir->imm + rv->PC;)
+/* JAL stores successor instruction address into rd.
```
jserv: Check the comment for JAL. Make it identical to the previous commit.

src/common.h
```diff
@@ -24,6 +24,16 @@
 #define __ALIGNED(x)
 #endif
 
+/* There is no Tail-call optimization(TCO) in non-optimized builds. To work
```
jserv: Since "tail-call" is not a proper noun, use lowercase.

src/common.h
```diff
@@ -24,6 +24,16 @@
...
+ * around this, we attempts to use a compiler attribute called musttail that
+ * forces the compiler to tail-call-optimize even when optimizations aren't on.
```
jserv: Replace "tail-call-optimize" with TCO because it frequently appears.

src/decode.h
```diff
@@ -7,6 +7,7 @@
 
 #include <stdbool.h>
 #include <stdint.h>
+#include "riscv.h"
```
jserv, line 11: Split with a blank line between `<stdint.h>` and `"riscv.h"`.

src/decode.h
```diff
@@ -241,6 +242,18 @@ typedef struct {
 
     /* instruction length */
     uint8_t insn_len;
+    /* According to Tail-call optimization (TCO), if a C function ends with
```
jserv: Ditto. Add a blank line.

src/emulate.c
jserv: Declare `rv_insn_t *ir` as `const`.

src/emulate.c
```diff
@@ -254,100 +254,90 @@ static inline bool insn_is_misaligned(uint32_t pc)
...
-
-    const uint32_t n_insn = block->n_insn;
-    rv_insn_t *ir;
-
-#if RV32_HAS(COMPUTED_GOTO)
-    /* current index in block */
-    uint32_t index = 0;
-#endif
+#define nextOP() \
+    ir++;        \
```
jserv: Use another local variable to increment. Let's say `const rv_insn_t *next = ir + 1;`.

src/emulate.c
jserv: We can modify `RISCV_INSN_LIST` in header `src/decode.h` and append branch information. i.e.,
```
/* RISC-V instruction list */
#define RISCV_INSN_LIST                    \
    _(nop)                                 \
    /* RV32I Base Instruction Set (opcode, can-branch) */       \
    _(lui, 0)                                 \
    _(auipc, 0)                               \
    _(jal, 1)                                 \
```

And then, we make use of `IIF` macro to specialize the code sequence depending on its property (can-branch or not).

Function `insn_is_branch` can be simplified with X macro trick as well.

jserv: Read [C Preprocessor tricks, tips, and idioms](https://github.com/pfultz2/Cloak/wiki/C-Preprocessor-tricks,-tips,-and-idioms) for pattern matching macros.

src/emulate.c
```diff
@@ -254,100 +254,90 @@ static inline bool insn_is_misaligned(uint32_t pc)
...
-#endif /* RV32_HAS(COMPUTED_GOTO) */
-
-    const uint32_t n_insn = block->n_insn;
-    rv_insn_t *ir;
-
-#if RV32_HAS(COMPUTED_GOTO)
-    /* current index in block */
-    uint32_t index = 0;
-#endif
+#define nextOP() \
```
jserv: Merge the body of `nextOP` into `RVOP` macro.

src/decode.h
```diff
@@ -241,6 +242,18 @@ typedef struct {
...
+     * a function call to another function or itself and simply returns that
+     * function's result, the compiler can substitute a simple jump to the
+     * other function for the 'call' and 'return' instructions . The self
+     * -recursive function can therefore use the same function stack frame.
+     *
+     * Member tailcall help us to determine whether the IR is last IR in
```
jserv: Improve the writing.

jserv (changes requested):
Refine the subject of git commit messages. We can say,
> Reduce instruction dispatch by tail-call elimination

src/decode.h
```diff
@@ -241,6 +242,18 @@ typedef struct {
...
+     * a basic block. We can use this member to rewrite the function emulate
+     * into a self-recursive version, allowing the compiler to do TCO.
+     */
+    bool tailcall;
+    bool (*opfunc)(riscv_t *, const struct rv_insn *);
```
jserv: Replace `opfunc` with `impl` because the later is self-explanatory.

src/emulate.c
```diff
@@ -384,959 +366,923 @@ static bool emulate(riscv_t *rv, const block_t *block)
...
+RVOP(sret, {
+    /* Not support */
+    return false;
+})
 
-    /* HRET: return from traps in H-mode */
-    _(hret, return false;)
+/* HRET: return from traps in H-mode */
+RVOP(hret, {
+    /* Not support */
```
jserv: Change the comment
> FIXME: Implement

src/emulate.c
```diff
@@ -384,959 +366,923 @@ static bool emulate(riscv_t *rv, const block_t *block)
...
+RVOP(uret, {
+    /* Not support */
+    return false;
+})
 
-    /* SRET: return from traps in S-mode */
-    _(sret, return false;)
+/* SRET: return from traps in S-mode */
+RVOP(sret, {
+    /* Not support */
```
jserv: Ditto. Change the comment.

src/emulate.c
```diff
@@ -384,959 +366,923 @@ static bool emulate(riscv_t *rv, const block_t *block)
...
+RVOP(wfi, {
+    /* Not support */
+    return false;
+})
 
-    /* URET: return from traps in U-mode */
-    _(uret, return false;)
+/* URET: return from traps in U-mode */
+RVOP(uret, {
+    /* Not support */
```
jserv: Ditto. Change the comment.

src/emulate.c
```diff
@@ -384,959 +366,923 @@ static bool emulate(riscv_t *rv, const block_t *block)
...
+    rv->io.on_ebreak(rv);
+    rv->csr_cycle++;
+    return true;
+})
 
-    /* WFI: Wait for Interrupt */
-    _(wfi, return false;)
+/* WFI: Wait for Interrupt */
+RVOP(wfi, {
+    /* Not support */
```
jserv: Ditto. Change the comment.

src/emulate.c
```diff
@@ -1516,8 +1463,9 @@ void rv_step(riscv_t *rv, int32_t cycles)
         /* we should have a block by now */
         assert(block);
 
+        rv_insn_t *ir = block->ir;
```
jserv: Move Line 1466 right before `if (unlikely(!(ir->opfunc)(rv, ir)))` statement.

jserv:
On Apple M1 based MacBook Air, depending on the compilers, TCO advantages can vary widely.
| compiler | TCO | CoreMark |
|:--------:|:-----:|:---------:|
| Apple clang v14.0.0 | No | 880.554045 iter/s |
| Apple clang v14.0.0 | Yes | 812.533942 iter/s |
| gcc-12 | No | 927.010950 iter/s |
| gcc-12 | Yes | 957.203684 iter/s |
|  clang v15.0.6 | No | 920.140504 iter/s |
| clang v15.0.6 | Yes |  803.948762 iter/s |

Aarch64 + clang may, in general, be a peculiar case when compared to others.

Makefile
```diff
@@ -88,6 +88,9 @@ gdbstub-test: $(BIN)
 	$(Q)tests/gdbstub.sh && $(call notice, [OK])
 endif
 
+# For tail-call elimination, we need a specific set of build flags applied.
+$(OUT)/emulate.o: CFLAGS += -fomit-frame-pointer -fno-stack-check -fno-stack-protector
```
jserv, line 94: Add a comment:
> FIXME: On macOS + Apple Silicon, -fno-stack-protector might have a negative impact.

The option `-fno-stack-protector` would therefore be omitted on Arm64-based macOS.
Quote from [-O1 implies -fomit-frame-pointer](https://reviews.llvm.org/D80828)
> Comparing GCC and Clang, it becomes apparent that GCC defaults to omitting the frame
pointer implicitly when optimizations are enabled, and Clang does not.
ie. setting -O1 (or above) implies `-fomit-frame-pointer`. Clang was
defaulting to `-fno-omit-frame-pointer` implicitly unless `-fomit-frame-pointer`
was set explicitly.

jserv: For `-fomit-frame-pointer`, see also https://github.com/godotengine/godot/issues/66296

qwe661234: I test the GCC and clang with and without flag `-fno-omit-frame-pointer`, the result shows below.
| Microprocessor | compiler | Coremark without `-fomit-frame-pointer`| Coremark with `-fomit-frame-pointer` | 
|--------|-------|-------|-------|
| Core i7-8700   | clang-15 |        966.9798478         |         971.9516670         |
| Core i7-8700   | gcc-12   |        954.0062906         |         963.3369450         |
| eMag 8180      | clang-15 |       330.0058728        |         335.728476         |
| eMag 8180      | gcc-12   |      320.5225226         |   332.5611754|

src/emulate.c
```diff
@@ -28,6 +28,7 @@ static inline int isnanf(float x)
 extern struct target_ops gdbstub_ops;
 #endif
 
+#include "common.h"
```
jserv: You don't have to include `common.h` because the build system already specifies.

src/emulate.c
```diff
@@ -254,100 +255,90 @@ static inline bool insn_is_misaligned(uint32_t pc)
...
-/* clang-format on */
-#endif /* RV32_HAS(COMPUTED_GOTO) */
-
-    const uint32_t n_insn = block->n_insn;
-    rv_insn_t *ir;
+#define RVOP(inst, code)                                                  \
+    static bool do_##inst(riscv_t *rv UNUSED, const rv_insn_t *ir UNUSED) \
+    {                                                                     \
+        rv->X[rv_reg_zero] = 0;                                           \
+        code rv->PC += ir->insn_len;                                      \
```
jserv, line 262: Use `IIF` macro here to generate branch aware code.

jserv:
We can eliminate the trailing `rv->csr_cycle++; return true;` in the implementation of the RISC-V instructions which can branch. Think of the changes below:
```diff
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ OUT ?= build
 BIN := $(OUT)/rv32emu
 
 CFLAGS = -std=gnu99 -O2 -Wall -Wextra
+CFLAGS += -Wno-unused-label
 CFLAGS += -include src/common.h
 
 # Set the default stack pointer
--- a/src/decode.h
+++ b/src/decode.h
@@ -166,6 +166,13 @@ enum {
 #undef _
 };
 
+/* can-branch information for each RISC-V instruction */
+enum {
+#define _(inst, can_branch) __rv_insn_##inst##_canbranch = can_branch,
+    RISCV_INSN_LIST
+#undef _
+};
+
 /* clang-format off */
 /* instruction decode masks */
 enum {
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -259,7 +259,14 @@ static inline bool insn_is_misaligned(uint32_t pc)
     static bool do_##inst(riscv_t *rv UNUSED, const rv_insn_t *ir UNUSED) \
     {                                                                     \
         rv->X[rv_reg_zero] = 0;                                           \
-        code rv->PC += ir->insn_len;                                      \
+        code;                                                             \
+        if (__rv_insn_##inst##_canbranch) {                               \
+            /* can branch */                                              \
+            rv->csr_cycle++;                                              \
+            return true;                                                  \
+        }                                                                 \
+    nextop:                                                               \
+        rv->PC += ir->insn_len;                                           \
         rv->csr_cycle++;                                                  \
         if (ir->tailcall)                                                 \
             return true;
```

Then, we can rewrite `BEQ` implementation for example.
```diff
 /* BEQ: Branch if Equal */
 RVOP(beq, {
     const uint32_t pc = rv->PC;
-    if (rv->X[ir->rs1] == rv->X[ir->rs2]) {
-        rv->PC += ir->imm;
-        /* check instruction misaligned */
-        if (unlikely(insn_is_misaligned(rv->PC))) {
-            rv->compressed = false;
-            rv_except_insn_misaligned(rv, pc);
-            return false;
-        }
-        /* increment the cycles csr */
-        rv->csr_cycle++;
-        /* can branch */
-        rv->csr_cycle++;
-        return true;
+    if (rv->X[ir->rs1] != rv->X[ir->rs2])
+        goto nextop;
+
+    rv->PC += ir->imm;
+    /* check instruction misaligned */
+    if (unlikely(insn_is_misaligned(rv->PC))) {
+        rv->compressed = false;
+        rv_except_insn_misaligned(rv, pc);
+        return false;
     }
 })
 
 /* BNE: Branch if Not Equal */
```

Code duplication should be avoided at all times. Each RISC-V instruction ought to see the statement `rv->csr cycle++` as hidden. If CSR is completely disabled in some configurations, we may even be able to turn off the generation later.

src/emulate.c
```diff
@@ -357,24 +348,16 @@ static bool emulate(riscv_t *rv, const block_t *block)
         /* increment the cycles csr */
```
jserv: Check if the code below is necessary. That is, consider
```diff
@@ -345,8 +345,6 @@ RVOP(beq, {
             rv_except_insn_misaligned(rv, pc);
             return false;
         }
-        /* increment the cycles csr */
-        rv->csr_cycle++;
         /* can branch */
         rv->csr_cycle++;
         return true;
@@ -364,8 +362,6 @@ RVOP(bne, {
             rv_except_insn_misaligned(rv, pc);
             return false;
         }
-        /* increment the cycles csr */
-        rv->csr_cycle++;
         /* can branch */
         rv->csr_cycle++;
         return true;
```

src/emulate.c
```diff
@@ -1517,7 +1445,8 @@ void rv_step(riscv_t *rv, int32_t cycles)
         assert(block);
 
         /* execute the block */
-        if (!emulate(rv, block))
+        const rv_insn_t *ir = block->ir;
+        if (unlikely(!(ir->impl)(rv, ir)))
```
jserv: Simplify a bit. That is,
```c
if (unlikely(!ir->impl(rv, ir)))
    break;
```

src/emulate.c
```diff
@@ -384,987 +367,931 @@ static bool emulate(riscv_t *rv, const block_t *block)
...
-    case rv_insn_cjalr:
-    case rv_insn_cbeqz:
-    case rv_insn_cbnez:
-    case rv_insn_cebreak:
-#endif
-#if RV32_HAS(Zifencei)
-    case rv_insn_fencei:
-#endif
-        return true;
+#define _(inst, can_branch) \
```
jserv: Tweak slightly.
```diff
 static bool insn_is_branch(uint8_t opcode)
 {
     switch (opcode) {
-#define _(inst, can_branch) \
-    IIF(can_branch)(case rv_insn_##inst : return true;, )
+#define _(inst, can_branch) IIF(can_branch)(case rv_insn_##inst:, )
         RISCV_INSN_LIST
 #undef _
+        return true;
     }
     return false;
 }
```

src/decode.h
```diff
@@ -8,158 +8,160 @@
 #include <stdbool.h>
 #include <stdint.h>
 
+#include "riscv.h"
+
 /* RISC-V instruction list */
 /* clang-format off */
 #define RISCV_INSN_LIST                    \
```
jserv, line 15: Add comment for the description of the instruction listing. That is, `_(instruction-name, can-branch)`

src/emulate.c
jserv: The reason why header `common.h` is not explicitly included in C source is that we have to ensure that it is always the first header to be included, otherwise we might suffer from potential macro conflicts.

jserv: Since we rely on `clang-format` for indention, the order of header files might be changed, and programmers would not be aware of it.

jserv:
According to [Ampere](https://en.wikichip.org/wiki/ampere_computing)'s product lines, [eMAG 8180](https://en.wikichip.org/wiki/ampere_computing/emag/8180) is superior than ThunderX2. Git commit message should be amended. Be case-sensitive.

jserv (changes requested):
The git commit messages were outdated. Measure based on latest code changes and rework the descriptions. In particular, potential concerns on TCO should be addressed.

src/decode.h
```diff
@@ -8,158 +8,167 @@
...
 
 /* IR list */
 enum {
-#define _(inst) rv_insn_##inst,
+#define _(inst, can_branch) rv_insn_##inst,
+    RISCV_INSN_LIST
+#undef _
+};
+
+/* can-branch information for each RISC-V instruction */
```
jserv: Move the can-branch enumeration to `src/emulate.c` since it only makes sense after decoding.

src/emulate.c
```diff
@@ -254,1119 +254,985 @@ static inline bool insn_is_misaligned(uint32_t pc)
...
+/* C.BEQZ performs conditional control transfers. The offset is
+ * sign-extended and added to the pc to form the branch target address.
+ * It can therefore target a ±256 B range. C.BEQZ takes the branch if
+ * the value in register rs1' is zero. It expands to beq rs1', x0,
+ * offset[8:1].
+ */
+RVOP(cbeqz,
+     { rv->PC += (!rv->X[ir->rs1]) ? (uint32_t) ir->imm : ir->insn_len; })
+/* C.BEQZ */
+RVOP(cbnez, { rv->PC += (rv->X[ir->rs1]) ? (uint32_t) ir->imm : ir->insn_len; })
```
jserv, line 1164: Add a new blank line between each pairing of RVOP macros.

src/emulate.c
```diff
@@ -254,1119 +254,985 @@ static inline bool insn_is_misaligned(uint32_t pc)
...
+    if (unlikely(addr & 3)) {
+        rv->compressed = true;
+        rv_except_store_misaligned(rv, addr);
+        return false;
+    }
+    rv->io.mem_write_w(rv, addr, rv->X[ir->rs2]);
+})
+
+/* C.NOP */
+RVOP(cnop,
```
jserv: Make the code more compact. Check the indention.

src/emulate.c
```diff
@@ -254,1119 +254,985 @@ static inline bool insn_is_misaligned(uint32_t pc)
...
+
+/* FSUB.S */
+RVOP(fsubs, {
+    if (isnanf(rv->F[ir->rs1]) || isnanf(rv->F[ir->rs2])) {
+        rv->F_int[ir->rd] = RV_NAN;
+    } else {
+        rv->F[ir->rd] = rv->F[ir->rs1] - rv->F[ir->rs2];
+    }
+})
+/* FMUL.S */
```
jserv, line 876: Add a new blank line between each pairing of RVOP macros.

PR #96, opened 2022-12-17
Set rv->compressed for instruction ecall and ebreak

qwe661234:
The exception handler is invoked by the instructions ecall and ebreak, thus we must set rv->compressed for these instructions.


jserv (changes requested):
Again, check your git commit message.

Risheng1128 (approved):

Issue #98, opened 2022-12-21
Fail to pass Zifencei test items


jserv:
commit ddface902243174cad2f09104bf231846a7ba329 integrates [riscv-arch-test](https://github.com/riscv-non-isa/riscv-arch-test)  (explicitly, based on [old-framework-2.x](https://github.com/riscv-non-isa/riscv-arch-test/tree/old-framework-2.x)) in CI pipeline. `Zifencei` test items failed according to the GitHub Actions report.
Reproduce:
```shell
make arch-test RISCV_DEVICE=Zifencei
```

PR #100, opened 2022-12-22
Patch for issue #98

qwe661234:
These are the patch commits for commit a304446.


jserv (changes requested):
Refine the git commit messages. Ensure both the subject and body contain insightful materials.

Append `Close #98` at the end of git commit message.

src/emulate.c
```diff
@@ -266,9 +266,8 @@ enum {
     {                                                                     \
         rv->X[rv_reg_zero] = 0;                                           \
         code;                                                             \
+        rv->csr_cycle++;                                                  \
         if (__rv_insn_##inst##_canbranch) {                               \
```
jserv, line 270: Add comment back.

PR #102, opened 2022-12-29
Refine the SDL-oriented system calls declaration

alanjian85:
Replace the type of all int parameters of system calls with unsigned, because they are usually treated as uint32_t internally by the emulator. And change the name of the parameter "screen" of the "draw_frame" system call to "base" for the sake of consistency with the "setup_queue" system call.

### DIFF
--- a/docs/syscall.md
+++ b/docs/syscall.md
@@ -103,9 +103,9 @@ These system calls are solely for the convenience of accessing the [SDL library]
 
 **system call number**: `0xBEEF`
 
-**synopsis**: `void draw_frame(void *screen, int width, int height)`
+**synopsis**: `void draw_frame(void *base, int width, int height)`
 
-If a window does not already exist, one will be created with the specified `width` and `height`. The `screen` buffer will replace the content of the framebuffer, passing a different `width` or `height` compared to the size of the window is undefined behavior. This system call additionally polls events from the SDL library, and, if necessary, update the internal input specific event queue.
+If a window does not already exist, one will be created with the specified `width` and `height`. The buffer pointed to by `base` will replace the content of the framebuffer, passing a different `width` or `height` compared to the size of the window is undefined behavior. This system call additionally polls events from the SDL library, and, if necessary, update the internal input specific event queue.
 
 The width and height are merely the virutal dimensions of the screen; they are unrelated to the window's real size. The system call would deal with resizing events internally when they occurred.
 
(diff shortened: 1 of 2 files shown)


jserv:
Please read [Unsigned Integers Are Dangerous](https://jacobegner.blogspot.com/2019/11/unsigned-integers-are-dangerous.html). We have to validate the dimension of SDL window and the corresponding ranges.

docs/syscall.md
```diff
@@ -103,7 +103,7 @@ These system calls are solely for the convenience of accessing the [SDL library]
 
 **system call number**: `0xBEEF`
 
-**synopsis**: `void draw_frame(void *screen, int width, int height)`
+**synopsis**: `void draw_frame(void *base, unsigned width, unsigned height)`
```
jserv: `screen` was mentioned in the following paragraph.

jserv:
See also
* [Almost Always Unsigned](https://graphitemaster.github.io/aau/#almost-always-unsigned)
* [Do not use unsigned for non-negativity](https://hamstergene.github.io/posts/2021-10-30-do-not-use-unsigned-for-nonnegativity/)

src/syscall_sdl.c
```diff
@@ -107,7 +107,7 @@ static void event_push(riscv_t *rv, event_t event)
     ++event_queue.end;
     event_queue.end &= queues_capacity - 1;
 
-    uint32_t count;
+    int32_t count;
```
jserv: Shall it be of `size_t` type?

src/syscall_sdl.c
```diff
@@ -218,10 +218,10 @@ void syscall_draw_frame(riscv_t *rv)
 {
     state_t *s = rv_userdata(rv); /* access userdata */
 
-    /* draw_frame(screen, width, height) */
+    /* draw_frame(base, width, height) */
     const uint32_t screen = rv_get_reg(rv, rv_reg_a0);
-    const uint32_t width = rv_get_reg(rv, rv_reg_a1);
-    const uint32_t height = rv_get_reg(rv, rv_reg_a2);
+    const int32_t width = rv_get_reg(rv, rv_reg_a1);
```
jserv: Let's follow the same type of the parameters passed to [SDL_CreateTexture](https://wiki.libsdl.org/SDL2/SDL_CreateTexture). That is, `int w, int h`.

jserv: The function prototype `check_sdl` should be adjusted accordingly.

PR #109, opened 2023-02-05
Use interger shortcut in floating point routines

2011eric:
This commit ensure that we use the F_int shortcut, instead of memcpy. 
FEQ.S/FLT.S/FLE.S are now working properly.


jserv:
Thank @2011eric for contributing!
The git commit message was amended slightly. Practice proper grammar the next time.

PR #110, opened 2023-02-05
Add definitions of rounding mode

2011eric:


src/decode.c
```diff
@@ -1030,7 +1030,7 @@ static inline bool op_op_fp(rv_insn_t *ir, const uint32_t insn)
      */
 
     /* decode R-type */
-    uint8_t funct3 = decode_funct3(insn);
+    ir->funct3 = decode_funct3(insn);
```
jserv: Rewrite as following:
```c
uint8_t funct3 = ir->funct3 = decode_funct3(insn);
```

Then, we can avoid accessing `ir->funct3` later.

src/softfloat.h
```diff
@@ -24,7 +24,14 @@ enum {
     FFLAG_UNDERFLOW   = 0b00000000000000000000000000000010,
     FFLAG_INEXACT     = 0b00000000000000000000000000000001,
     //                    ....xxxx....xxxx....xxxx....xxxx
-    RV_NAN            = 0b01111111110000000000000000000000
+    RV_NAN            = 0b01111111110000000000000000000000,
+    //                    ....xxxx....xxxx....xxxx....xxxx
```
jserv: I am not sure if the following identifiers make sense. Can you explicitly use them somewhere?

2011eric: Maybe I can use switch case to compare them directly when I need to.
Should I delete this?

jserv: You can implement the decoding for rounding mode in the proposed changes.

README.md
```diff
@@ -23,6 +23,11 @@ access to all of its features. Your target system must have a functional [SDL2 l
 * macOS: `brew install sdl2`
 * Ubuntu Linux / Debian: `sudo apt install libsdl2-dev`
 
+Install [RISCOF](https://riscof.readthedocs.io/en/stable/installation.html#install-riscof).
```
jserv: You shall submit another pull request for RISCOF.

jserv: In this pull request, let's concentrate on rounding mode.

src/softfloat.h
jserv: Any progress?

jserv (changes requested):
Use `git rebase` rather than `git merge`.

PR #114, opened 2023-02-19
Rollback partial wrong implemented code in pr #89

TheCloudlet:


TheCloudlet:
Suggestion in CWE-468
--> Always use array indexing instead of direct pointer manipulation.

I do not think there is a way to us array indexing here.

gagachang:
How about using union:
```
union {
    uint32_t csr_time_32[2];
    uint64_t csr_time_64;
};
```

TheCloudlet:
Hi @gagachang,

Maybe this new change has a better readability.
What do you think?

gagachang:
Yes, it looks better. Minor suggestion:
```
rv->csr_time[0] = t & 0xFFFFFFFF;
```

Actually, I think two dedicated variables for time is OK, since rv32emu targets RV32 system.
The spec defines two CSRs for time on RV32 system.

```
uint32_t csr_time;
uint32_t csr_timeh;
```

TheCloudlet:
Hi @gagachang 
Thank you for your code review and suggestion

PR #116, opened 2023-03-01
Remove the computed-goto option

Risheng1128:
Remove the computed-goto option. According to #95, computed-goto has been replaced by tail-call optimization (TCO). Therefore, the option about computed-goto is unnecessary.


jserv (changes requested):
Split into two pull requests.

Risheng1128:
I will retain commit "Remove the computed-goto option" in this pull request.

Issue #105, opened 2023-01-05
Manage basic block with adaptive replacement cache


jserv:
Current basic block management consumes a significant amount of memory, which leads to unnecessary waste due to frequent map allocation and release.  [Adaptive Replacement Cache](https://en.wikipedia.org/wiki/Adaptive_replacement_cache) (ARC) is a page replacement algorithm with better performance than least recently used (LRU). Better memory usage and hit rates can be achieved after the translated blocks are handled by ARC by keeping track of frequently used and recently used pages as well as a recent eviction history for both.

Quote from [ZFS Caching](https://papers.freebsd.org/2019/FOSDEM/jude-eli5_zfs_caching.files/ELI5_ZFS_ARC.pdf):
* Combines the best of LRU and LFU, plus some novel tricks
* The cache size (c) is partitioned (p) into two sections
* At the start, p = ½ c, first half of the cache is LRU, second LFU
* In addition to the two caches, there is a “ghost” list for each
* Each time an item is evicted from either cache, its key (but not its data) moves to the ghost list for that cache

Expected output:
1. Create cache.[ch] which consist of clean API. See map.[ch] for reference.
2. Implement [adaptive replacement cache](https://en.wikipedia.org/wiki/Adaptive_replacement_cache) in cache.[ch] along with configurable capacity.
3. Integrate ARC in basic block management.

Reference:
* [ZFS Caching](https://papers.freebsd.org/2019/FOSDEM/jude-eli5_zfs_caching.files/ELI5_ZFS_ARC.pdf)
* [C++ implement of Adaptive Replacement Cache](https://github.com/Xiaoccer/ARC_Cache)
* [cachehash](https://github.com/zmap/cachehash) for C API design
* [Cache replacement policies](https://en.wikipedia.org/wiki/Cache_replacement_policies)

qwe661234:
To manage basic blocks, this [commit](https://github.com/qwe661234/rv32emu/commit/5a308ef0a0fe3f245b706a48bd6d332fc09ca1b1) implements the Adaptive Replacement Cache (ARC) and basic profiling tools to show cache information. However, even if a cache hit rate of running `CoreMark` is more than 99%, the performance is worse than the previous implementation. We suppose the issue is that the overhead of cache operation is too high.

* CoreMark

|Model| Compiler | a304446 | ARC| Speedup |
|--------| -------- | -------- |-------- |-------- |
| Core i7-8700 | clang-15 |  971.951     | 895.248|-8.6%|
[...]

Issue #94, opened 2022-12-12
Integrate deterministic memory pool


jserv:
[TLSF](http://www.gii.upv.es/tlsf/) (Two-Level Segregated Fit) dynamic memory allocation algorithm is guaranteed to complete allocation and deallocation operations in constant time, suitable for real-time applications.

Reference:
* [Benchmarking Malloc with Doom 3](https://www.forrestthewoods.com/blog/benchmarking-malloc-with-doom3/)
* [rlsf](https://github.com/yvt/rlsf)
* [Documentation for TLSF implementation](https://github.com/paulhamsh/TLSF-Memory-Allocator-Overview)
* A little bit of floating point in a memory allocator — [Part 1: Background](https://brnz.org/hbr/?p=1735), [Part 2: The floating point](https://brnz.org/hbr/?p=1744)

jserv:
[0001-WIP-Integrate-deterministic-memory-pool.patch.gz](https://github.com/sysprog21/rv32emu/files/10253022/0001-WIP-Integrate-deterministic-memory-pool.patch.gz) implementation serves as proof of concept.

jserv:
Related: #105

jserv:
[HTFH-RT-Search-Cache](https://github.com/EngineersBox/HTFH-RT-Search-Cache) is a real-time search cache build on top of a Hybrid TLSF Fixed Heap allocator with DLIRS mixed-strategy caching.

See also: [DLIRS: Improving Low Inter-Reference Recency Set Cache Replacement Policy with Dynamics](https://www.systor.org/2018/pdf/systor18-4.pdf)
> Inspired by the idea of dynamic cache space partitioning from another state-of-the-art policy, Adaptive Replacement Cache (ARC), we propose a new Dynamic LIRS (DLIRS) policy. The new policy uses a simple mechanism to perform an approximated online estimation on how well IRR predicts future access behaviors, and then dynamically adapts the space allocation of low IRR blocks against high IRR blocks. Experiments are performed on traces from the UMass Trace Repository as well as a synthetic trace drawn from a stack depth distribution. While sometimes LIRS outperforms ARC with a significant margin and sometimes vice versa, the new DLIRS policy consistently performs close to the winner between ARC and LIRS in all the cases.

jserv:
[...]

PR #118, opened 2023-03-03
Add adaptive replacement cache

qwe661234:
Current basic block management consumes a significant amount of memory, which leads to unnecessary waste due to frequent map allocation and release. Adaptive Replacement Cache (ARC) is a page replacement algorithm with better performance than least recently used (LRU). After the translated blocks are handled by ARC, better memory usage and hit rates can be achieved by keeping track of frequently used and recently used pages, as well as a recent eviction history for both.

According to the cache information obtained while running CoreMark, the cache hit rate of ARC can reach over 99%.

see #105


Makefile
```diff
@@ -74,6 +74,14 @@ gdbstub-test: $(BIN)
 	$(Q)tests/gdbstub.sh && $(call notice, [OK])
 endif
 
+# Import adaptive replacement cache
+ENABLE_ARCACHE ?= 1
```
jserv: Don't add both `ENABLE_ARCACHE` and `ENABLE_ARCACHE_INFO`.
I would like to have a minimal ARC implementation along with a test program. Then, we can integrate the refined one into rv32emu core in further pull requests.

src/cache.c
```diff
@@ -0,0 +1,230 @@
+#include "cache.h"
```
jserv: Add copyright notice.

src/cache.c
```diff
@@ -0,0 +1,230 @@
...
+
+#define min(a, b) ((a < b) ? a : b)
```
jserv: Capitalize these macros. That is, `MIN` and `MAX`.

src/cache.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include "list.h"
+
+/*
+    T1: LRU List
+    T2: LFU List
+    B1: LRU Ghost List
+    B2: LFU Ghost List
+*/
+typedef enum { T1, B1, T2, B2 } arc_type_t;
```
jserv: Don't call it `arc_type_t`.
The implementation details should appear in C source rather than public headers.

src/cache.h
```diff
@@ -0,0 +1,42 @@
...
+
+struct hashtable;
+
+typedef struct cache {
+    struct list_head *list_table[4];
+    uint32_t list_size[4];
+    struct hashtable *map;
+    uint32_t c;
```
jserv: It is hard to know what `c` and `p` are.

src/cache.h
```diff
@@ -0,0 +1,42 @@
...
+    uint32_t p;
+#if RV32_HAS(ARCACHE_INFO)
```
jserv: Don't add `ARCACHE_INFO`. Minimize the code as possible.

Makefile
```diff
@@ -4,7 +4,7 @@ include mk/toolchain.mk
 OUT ?= build
 BIN := $(OUT)/rv32emu
 
-CFLAGS = -std=gnu99 -O2 -Wall -Wextra
+CFLAGS = -std=gnu99 -O2 -Wall -Wextra -lpthread
```
jserv: Why pthread?

Makefile
```diff
@@ -55,7 +55,7 @@ $(OUT)/syscall_sdl.o: CFLAGS += $(shell sdl2-config --cflags)
 LDFLAGS += $(shell sdl2-config --libs)
 endif
 
-ENABLE_GDBSTUB ?= 1
+ENABLE_GDBSTUB ?= 0
```
jserv: Don't change this.

src/cache.h
qwe661234: This flag is for showing cache information, such as cache hit rate and miss rate.

Makefile
jserv: There should no entry `ENABLE_ARCACHE`. That is, this pull patch only delivers ARC header and C implementation along with test program(s). You should tweak `Makefile` for the linkage of ARC test program(s).

src/cache.c
```diff
@@ -0,0 +1,241 @@
...
+cache_t *cache_create()
+{
+    cache_t *cache = (cache_t *) malloc(sizeof(cache_t));
+    for (int i = 0; i < 4; i++) {
+        cache->list_table[i] =
+            (struct list_head *) malloc(sizeof(struct list_head));
+        INIT_LIST_HEAD(cache->list_table[i]);
+        cache->list_size[i] = 0;
+    }
+    cache->map = (hashtable_t *) malloc(sizeof(hashtable_t));
```
jserv: Shorten the code.

src/cache.c
```diff
@@ -0,0 +1,241 @@
...
+        move_to_mru(cache, entry, LFU_list);
+    }
+
+    /* cache hit in LFU_ghost_list */
+    if (entry->arc_type == LFU_ghost_list) {
+        cache->lru_capacity = cache->lru_capacity ? cache->lru_capacity - 1 : 0;
+        replace_LRU_list(cache);
+        move_to_mru(cache, entry, LFU_list);
+    }
+#if CACHE_INFO
```
jserv: Remove all occurrences of `CACHE_INFO`.

src/cache.h
```diff
@@ -0,0 +1,44 @@
...
+#include <string.h>
+
+#include "list.h"
+#define CACHE_INFO 0
+
+typedef enum { LRU_list, LFU_list, LRU_ghost_list, LFU_ghost_list } list_type_t;
+
+struct hashtable;
+
+typedef struct cache {
```
jserv: Make use of forward declarations, so that we don't have to disclose the implementation details.

src/cache.h
```diff
@@ -0,0 +1,44 @@
...
+    struct hashtable *map;
+    uint32_t capacity;
+    uint32_t lru_capacity;
+#if CACHE_INFO
+    uint64_t get_time;
+    uint64_t hit_time;
+#endif
+} cache_t;
+
+cache_t *cache_create();
```
jserv: You MUST write down comments for each public function.

src/emulate.c
```diff
@@ -1250,7 +1250,7 @@ static bool insn_is_branch(uint8_t opcode)
     }
     return false;
 }
-
+#if !RV32_HAS(ARCACHE)
```
jserv: Don't integrate `ARCACHE` at the moment.

Makefile
qwe661234: OK, I understand.

jserv: The name `ARCACHE` sounds stupid since ARC stands for Adaptive Replacement Cache.

src/cache.h
```diff
@@ -0,0 +1,44 @@
...
+#pragma once
+
+#include <assert.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "list.h"
```
jserv: Don't expose `list.h` in this public header. Encapsulate the details as possible.

jserv: Then, it is feasible to switch among different cache replacement implementations later. Let's stick to consistent interface.

src/list.h
```diff
@@ -0,0 +1,431 @@
+/* Linux-like doubly-linked list implementation */
```
jserv: Shrink this header to keep necessary macros and functions.

jserv:
Check the build system of [linux-list](https://github.com/sysprog21/linux-list) for building test programs.

Makefile
jserv: This ARC implementation is meant be reusable for further JIT compilation. Its quality should be verified, and it is exactly why test programs should appear for validation purposes. The interface to ARC should be generic and straightforward without disclosing the details.

qwe661234:
> Check the build system of [linux-list](https://github.com/sysprog21/linux-list) for building test programs.

Does the test programs here means new test case for testing cache?

jserv:
> Does the test programs here means new test case for testing cache?

Exactly. These test programs should be built and verified before we integrate ARC into rv32emu core.

jserv (changes requested):
1. All test cases should be built from top-level Makefile.
2. Test specific file should be placed in tests directory, and there are no file names in upper case.
3. Show report of test cases and integrate into GitHub Actions.

src/cache.c
```diff
@@ -0,0 +1,231 @@
...
+    new_entry->value = value;
+    new_entry->arc_type = LRU_list;
+    list_add(&new_entry->list, cache->list_table[LRU_list]);
+    list_add(&new_entry->ht_list, &cache->map->ht_list_head[HASH(key)]);
+    cache->list_size[LRU_list]++;
+    assert_cache(cache);
+    return delete_value;
+}
+
+void cache_print_stats(cache_t *cache)
```
jserv: It is not necessary to print.

src/cache.h
```diff
@@ -0,0 +1,46 @@
...
+#pragma once
+
+#include <assert.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+typedef enum { LRU_list, LFU_list, LRU_ghost_list, LFU_ghost_list } list_type_t;
```
jserv: Avoid disclosing implementation details. This cache interface should have nothing associated with ARC.

src/cache.h
```diff
@@ -0,0 +1,46 @@
+/*
+ * rv32emu is freely redistributable under the MIT License. See the file
+ * "LICENSE" for information on usage and redistribution of this file.
+ */
+
+#pragma once
+
+#include <assert.h>
```
jserv: Clean up inclusions. Only keep the headers you need for the declarations.

src/list.h
```diff
@@ -0,0 +1,241 @@
...
+ */
+static inline void list_del(struct list_head *node)
+{
+    struct list_head *next = node->next;
+    struct list_head *prev = node->prev;
+
+    next->prev = prev;
+    prev->next = next;
+
+#ifdef LIST_POISONING
```
jserv: Remove `LIST_POISONING` since we don't tend to utilize it.

tests/cache-test/src/testCache.c
```diff
@@ -0,0 +1,64 @@
...
+{
+    char *s = strtok(str, del);
+
+    while (s != NULL) {
+        *arr++ = s;
+        s = strtok(NULL, del);
+    }
+}
+
+int main(__attribute__((unused)) int argc, char *argv[])
```
jserv: Replace `_attribute__((unused))` with `UNUSED` since `common.h` already defines.

tests/cache-test/Makefile
```diff
@@ -0,0 +1,52 @@
...
+all:$(TARGET)
+
+$(TARGET): $(OBJS)
+	$(Q) mkdir -p $(BUILDDIR)
+	$(CC) $^ -o $(BUILDDIR)/$(TARGET)
+
+$(BUILDDIR)/cache.o: $(SOURCE)/cache.c
+	$(Q)$(CC) -o $@ $(CFLAGS) -c -MMD -MF $@.d $<
+
+$(BUILDDIR)/testCache.o: src/testCache.c
```
jserv: Shorten the file names. Let's specify `tests/cache.c` for short.

tests/cache-test/Makefile
```diff
@@ -0,0 +1,52 @@
...
+CFLAGS = -std=gnu99 -O2 -Wall -Wextra
+CFLAGS += -Wno-unused-label -I$(SOURCE)
+
+OBJS := cache.o \
+        testCache.o
+
+OBJS := $(addprefix $(BUILDDIR)/, $(OBJS))
+deps := $(OBJS:%.o=%.o.d)
+
+CHECK_ELF_FILES := \
```
jserv: You may consider to check shecc for its test suite.
See [tests/driver.sh](https://github.com/jserv/shecc/blob/master/tests/driver.sh) and [Makefile](https://github.com/jserv/shecc/blob/master/Makefile).

src/cache.c
```diff
@@ -0,0 +1,231 @@
...
+}
+
+/* Rule of ARC
+ * 1. size of LRU_list + size of LFU_list <= c
+ * 2. size of LRU_list + size of LRU_ghost_list <= c
+ * 3. size of LFU_list + size of LFU_ghost_list <= 2c
+ * 4. size of LRU_list + size of LFU_list + size of LRU_ghost_list + size of
+ * LFU_ghost_list <= 2c
+ */
+void assert_cache(cache_t *cache)
```
jserv: It tends to be a private function. In addition, use macro to rework.

src/cache.c
```diff
@@ -0,0 +1,231 @@
...
+        "hits:     %12lu \n"
+        "ratio: %lf%%\n",
+        cache->get_time, cache->hit_time,
+        cache->hit_time * 100 /
+            (double) (cache->get_time ? cache->get_time : 1));
+}
+
+void cache_free(cache_t *cache, void (*release_entry)(void *))
+{
+    for (int i = 0; i < 4; i++) {
```
jserv, line 223: Avoid hardcoded `4`.

src/cache.c
```diff
@@ -0,0 +1,231 @@
...
+    struct list_head *ht_list_head;
+} hashtable_t;
+
+typedef struct cache {
+    struct list_head *list_table[4];
+    uint32_t list_size[4];
+    struct hashtable *map;
+    uint32_t capacity;
+    uint32_t lru_capacity;
+    uint64_t get_time;
```
jserv: Check if `get_time` and `hit_time` are necessary.

src/cache.c
```diff
@@ -0,0 +1,231 @@
...
+    struct list_head ht_list;
+} arc_entry_t;
+
+typedef struct hashtable {
+    struct list_head *ht_list_head;
+} hashtable_t;
+
+typedef struct cache {
+    struct list_head *list_table[4];
+    uint32_t list_size[4];
```
jserv: Reduce the memory usage.

src/cache.c
```diff
@@ -0,0 +1,231 @@
...
+ */
+
+#include "cache.h"
+#include "list.h"
+
+#define MIN(a, b) ((a < b) ? a : b)
+#define BITS 8
+#define SIZE 256
+#define GOLDEN_RATIO_32 0x61C88647
+#define HASH(val) (((val) *GOLDEN_RATIO_32) >> (32 - BITS)) % SIZE
```
jserv: Use an inline function rather than a macro for hash function.

src/cache.h
```diff
@@ -0,0 +1,46 @@
...
+#include <string.h>
+
+typedef enum { LRU_list, LFU_list, LRU_ghost_list, LFU_ghost_list } list_type_t;
+
+struct cache;
+
+/** cache_create - crate a new cache
+ * @return: a pointer points to new cache
+ */
+struct cache *cache_create();
```
jserv: How can we specify the capacity?

src/cache.c
```diff
@@ -0,0 +1,231 @@
...
+ * rv32emu is freely redistributable under the MIT License. See the file
+ * "LICENSE" for information on usage and redistribution of this file.
+ */
+
+#include "cache.h"
+#include "list.h"
+
+#define MIN(a, b) ((a < b) ? a : b)
+#define BITS 8
+#define SIZE 256
```
jserv: Change the name `SIZE` because it is too generic.

src/cache.c
```diff
@@ -0,0 +1,231 @@
+/*
+ * rv32emu is freely redistributable under the MIT License. See the file
+ * "LICENSE" for information on usage and redistribution of this file.
+ */
+
+#include "cache.h"
+#include "list.h"
```
jserv: At present, header `list.h` is only used by `src/cache.c`, and we can apply the following changes to minimize overall code size:
1. Move the definition of `container_of` into `src/common.h`
2. Move the macros and inline functions into `src/cache.c`

src/cache.c
```diff
@@ -0,0 +1,231 @@
...
+    assert(cache->list_size[LRU_list] + cache->list_size[LRU_ghost_list] <=
+           cache->capacity);
+    assert(cache->list_size[LFU_list] + cache->list_size[LFU_ghost_list] <=
+           2 * cache->capacity);
+    assert(cache->list_size[LRU_list] + cache->list_size[LRU_ghost_list] +
+               cache->list_size[LFU_list] + cache->list_size[LFU_ghost_list] <=
+           2 * cache->capacity);
+}
+
+void move_to_mru(cache_t *cache, arc_entry_t *entry, const list_type_t arc_type)
```
jserv: Do not expose the internal functions. That is, add `static`.

src/cache.c
```diff
@@ -0,0 +1,231 @@
...
+}
+
+void *cache_get(cache_t *cache, uint32_t key)
+{
+    if (cache->capacity <= 0 ||
+        list_empty(&cache->map->ht_list_head[HASH(key)]))
+        return NULL;
+
+    arc_entry_t *entry = NULL;
+    list_for_each_entry(entry, &cache->map->ht_list_head[HASH(key)], ht_list)
```
jserv: Update `.clang-format` to identify for-each macros. See https://github.com/sysprog21/lab0-c/blob/master/.clang-format

src/cache.c
jserv: Rewrite as following:
```c
#define HASH(val) (((val) * (GOLDEN_RATIO_32)) >> (32 - BITS)) % SIZE
```

Therefore, clang-format can be happier.

src/cache.c
```diff
@@ -0,0 +1,231 @@
...
+#define SIZE 256
+#define GOLDEN_RATIO_32 0x61C88647
+#define HASH(val) (((val) *GOLDEN_RATIO_32) >> (32 - BITS)) % SIZE
+
+typedef struct arc_entry {
+    void *value;
+    uint32_t key;
+    list_type_t arc_type;
+    struct list_head list;
+    struct list_head ht_list;
```
jserv: Woud you consider the use of `hlist` used in Linux kernel?

src/cache.c
```diff
@@ -0,0 +1,231 @@
...
+} arc_entry_t;
+
+typedef struct hashtable {
+    struct list_head *ht_list_head;
+} hashtable_t;
```
jserv, line 67: Ditto. You may learn from the hash table implementation in Linux kernel.

src/cache.c
```diff
@@ -0,0 +1,231 @@
...
+    struct hashtable *map;
+    uint32_t capacity;
+    uint32_t lru_capacity;
+    uint64_t get_time;
+    uint64_t hit_time;
+} cache_t;
+
+cache_t *cache_create()
+{
+    cache_t *cache = malloc(sizeof(cache_t));
```
jserv, line 199: Validate the return value of `malloc`.

tests/cache-test/src/testCache.c
jserv: You shall check `argc` in advance.

tests/cache-test/src/testCache.c
```diff
@@ -0,0 +1,64 @@
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "cache.h"
+#define TESTSIZE 1
+
+void printValue(int *val)
```
jserv: Enforce consistent style.

jserv (changes requested):
All files in directory `tests` should comply with the following naming scheme.
> `tests/cache/ITEM.TYPE`

* `ITEM` is the test item. e.g., `cache_get`
* `TYPE` is the type for given tests. e.g., `in` for input commands; `expect` for expected results, `out` for something generated .

Therefore, we would have `tests/cache/cache_get.in` .

src/cache.c
```diff
@@ -0,0 +1,369 @@
...
+typedef enum {
+    LRU_list,
+    LFU_list,
+    LRU_ghost_list,
+    LFU_ghost_list,
+    list_quantity
+} list_type_t;
+
+struct list_head {
+    struct list_head *prev;
```
jserv: Rewrite as following:
```c
    struct list_head *prev, *next;
```

src/cache.c
```diff
@@ -0,0 +1,369 @@
...
+
+int cache_size, cache_size_bits;
+
+typedef enum {
+    LRU_list,
+    LFU_list,
+    LRU_ghost_list,
+    LFU_ghost_list,
+    list_quantity
+} list_type_t;
```
jserv: `list_type_t` is a bit confusing. Someone might regard it as the types of linked list.

src/cache.c
```diff
@@ -0,0 +1,369 @@
...
+
+typedef struct cache {
+    struct list_head *list_table[list_quantity];
+    uint32_t list_size[list_quantity];
+    struct hashtable *map;
+    uint32_t capacity;
+    uint32_t lru_capacity;
+} cache_t;
+
+#if defined(__GNUC__)
```
jserv: You shall provide the generic code without `typeof`.

jserv: In addition, check clang as well. That is,
```c
```

src/cache.h
qwe661234: I make the user specify the capacity, but this implementation significantly reduces performance.

src/common.h
```diff
@@ -49,3 +49,30 @@
 #define IIF_0(t, ...) __VA_ARGS__
 /* run the 1st parameter */
 #define IIF_1(t, ...) t
+
+#if defined(__GNUC__)
```
jserv: Consolidate the checking for `typeof`.

src/cache.c
jserv: Don't touch `__LIST_HAVE_TYPEOF` since `src/common.h` already takes care of it.

src/cache.c
```diff
@@ -0,0 +1,369 @@
...
+         entry = list_entry(entry->member.next, __typeof__(*entry), member))
+#endif
+
+#define list_for_each_entry_safe(entry, safe, head, member)                \
+    for (entry = list_entry((head)->next, __typeof__(*entry), member),     \
+        safe = list_entry(entry->member.next, __typeof__(*entry), member); \
+         &entry->member != (head); entry = safe,                           \
+        safe = list_entry(safe->member.next, __typeof__(*entry), member))
+
+#undef __LIST_HAVE_TYPEOF
```
jserv: Don't touch `__LIST_HAVE_TYPEOF`. Ensure that List API works for non-gcc/clang compilers.

src/cache.h
jserv: Why?

mk/cache-test.mk
```diff
@@ -0,0 +1,42 @@
+CACHE_TEST_DIR := tests/cache
+CACHE_BUILD_DIR := build/cache
+TARGET := test_cache.out
+SOURCE := src
```
jserv: Is `SOURCE` used?

tests/cache/cache.c
```diff
@@ -0,0 +1,369 @@
...
+}
+
+/* Rule of ARC
+ * 1. size of LRU_list + size of LFU_list <= c
+ * 2. size of LRU_list + size of LRU_ghost_list <= c
+ * 3. size of LFU_list + size of LFU_ghost_list <= 2c
+ * 4. size of LRU_list + size of LFU_list + size of LRU_ghost_list + size of
+ * LFU_ghost_list <= 2c
+ */
+#define ASSERT_CACHE(cache)                                                 \
```
jserv, line 223: Rename to `CACHE_ASSERT`.

mk/cache-test.mk
```diff
@@ -0,0 +1,42 @@
...
+	cache_put \
+	cache_get \
+	cache_lru_replace \
+	cache_lfu_replace \
+	cache_lfu_ghost_replace \
+
+cache :$(TARGET)
+	$(Q)$(foreach e,$(CACHE_CHECK_ELF_FILES),\
+	    $(PRINTF) "Running $(e) ... "; \
+	    if [ "$(strip $(shell cat $(CACHE_TEST_DIR)/$(e).expect))" = "$(strip $(shell ./$(CACHE_BUILD_DIR)/$(TARGET) $(CACHE_TEST_DIR)/$(e).in))" ]; then \
```
jserv: Shouldn't you compare `.out` (the output) and `.expect` (the expected result)?

tests/cache/cache.c
```diff
@@ -0,0 +1,369 @@
...
+} arc_entry_t;
+
+typedef struct hashtable {
+    struct hlist_head *ht_list_head;
+} hashtable_t;
+
+typedef struct cache {
+    struct list_head *list_table[list_quantity];
+    uint32_t list_size[list_quantity];
+    struct hashtable *map;
```
jserv: Enforce the use of `hashtable_t`.

src/cache.h
qwe661234: I guess the reason is that the hash function, which calculating by cache size, is called frequently. So, if we use a macro to represent size, the compiler can optimize it, possibly through constant propagation.

mk/cache-test.mk
qwe661234: I campare the output (generated by printf) and .expect directly.

src/cache.h
jserv: Good point. Clarify your assumptions.

mk/cache-test.mk
jserv: How about using [cmp(1)](https://man7.org/linux/man-pages/man1/cmp.1.html)?

tests/cache/cache.c
```diff
@@ -0,0 +1,369 @@
...
+#include <stddef.h>
+#include <stdlib.h>
+#include <string.h>
+
+#define MIN(a, b) ((a < b) ? a : b)
+#define GOLDEN_RATIO_32 0x61C88647
+#define HASH(val) \
+    (((val) * (GOLDEN_RATIO_32)) >> (32 - (cache_size_bits))) % (cache_size)
+
+int cache_size, cache_size_bits;
```
jserv: Both `cache_size` and `cache_size_bits` should be private, meaning that you shall mark them as `static`.

src/cache.c
jserv: Avoid hardcoded `4`.

src/cache.h
qwe661234: I examine the assembly code on x86 architecture, if we use a macro to represent size, the hash function only needs 5 instructions. But if we use a   variable to represent size, the hash function needs 9 instructions, including the dividing instruction which has large overhead.
* MACRO
```=
imul   $0x61c88647,%esi,%eax
mov    0x30(%rcx),%rdx
mov    (%rdx),%rdx
shr    $0x18,%eax
mov    (%rdx,%rax,8),%rax
```
* variable 
```=
imul   $0x61c88647,%esi,%eax
mov    $0x20,%ecx
xor    %edx,%edx
mov    0x30(%rdi),%r9
sub    0x6a84(%rip),%ecx        # 1204c <cache_size_bits>
shr    %cl,%eax
divl   0x6a80(%rip)        # 12050 <cache_size>
mov    (%r9),%rax
mov    (%rax,%rdx,8),%rax
```

tests/cache/cache.c
```diff
@@ -0,0 +1,378 @@
...
+
+static uint32_t cache_size, cache_size_bits;
+
+typedef enum {
+    LRU_list,
+    LFU_list,
+    LRU_ghost_list,
+    LFU_ghost_list,
+    list_quantity
+} list_category_t;
```
jserv: Rename to `cache_list_t` and some comments like
> ARC improves the basic LRU strategy by splitting the cache directory into two lists, T1 and T2, for recently and frequently referenced entries. In turn, each of these is extended with a ghost list (B1 or B2), which is attached to the bottom of the two lists.

tests/cache/cache.c
```diff
@@ -0,0 +1,378 @@
...
+    (((val) * (GOLDEN_RATIO_32)) >> (32 - (cache_size_bits))) % (cache_size)
+
+static uint32_t cache_size, cache_size_bits;
+
+typedef enum {
+    LRU_list,
+    LFU_list,
+    LRU_ghost_list,
+    LFU_ghost_list,
+    list_quantity
```
jserv: Change the name `list_quantity` to `N_CACHE_LIST_TYPES` for better understanding.

tests/cache/cache.c
```diff
@@ -0,0 +1,378 @@
...
+
+typedef struct arc_entry {
+    void *value;
+    uint32_t key;
+    list_category_t arc_type;
+    struct list_head list;
+    struct hlist_node ht_list;
+} arc_entry_t;
+
+typedef struct hashtable {
```
jserv: Replace `struct hashtable` with `struct` since you are using typedef.

src/common.h
```diff
@@ -49,3 +49,28 @@
 #define IIF_0(t, ...) __VA_ARGS__
 /* run the 1st parameter */
 #define IIF_1(t, ...) t
+
+#if defined(__GNUC__) || defined(__clang__)
+#define __LIST_HAVE_TYPEOF 1
```
jserv: Rename to `__HAVE_TYPEOF`.

tests/cache/cache.c
jserv: Rename to `CACHE_ASSERT`.

src/cache.c
```diff
@@ -0,0 +1,378 @@
...
+    struct hlist_node ht_list;
+} arc_entry_t;
+
+typedef struct hashtable {
+    struct hlist_head *ht_list_head;
+} hashtable_t;
+
+typedef struct cache {
+    struct list_head *list_table[list_quantity];
+    uint32_t list_size[list_quantity];
```
jserv: What are typical ranges for the lists used in ARC? I don't think it is worthy to preserve them in `uint32_t`.

src/cache.c
```diff
@@ -0,0 +1,378 @@
...
+    if (!cache->map) {
+        free(cache);
+        free(cache->list_table);
+        return NULL;
+    }
+    cache->map->ht_list_head = malloc(cache_size * sizeof(struct hlist_head));
+    if (!cache->map->ht_list_head) {
+        free(cache);
+        free(cache->list_table);
+        free(cache->map);
```
jserv, line 219: Move the statement `free(cache);` to the last.

src/cache.c
```diff
@@ -0,0 +1,378 @@
...
+    for (int i = 0; i < list_quantity; i++) {
+        cache->list_table[i] = malloc(sizeof(struct list_head));
+        INIT_LIST_HEAD(cache->list_table[i]);
+        cache->list_size[i] = 0;
+    }
+
+    cache->map = malloc(sizeof(hashtable_t));
+    if (!cache->map) {
+        free(cache);
+        free(cache->list_table);
```
jserv: Ditto. The order of `free` invocation.

src/cache.c
```diff
@@ -0,0 +1,378 @@
...
+    struct hlist_node **pprev = n->pprev;
+
+    *pprev = next;
+    if (next)
+        next->pprev = pprev;
+}
+
+static inline void hlist_del_init(struct hlist_node *n)
+{
+    if (!hlist_unhashed(n)) {
```
jserv: Rewrite as following:
```c
    if (hlist_unhashed(n)
        return;
    hlist_del(n);
    INIT_HLIST_NODE(n);
```

tests/cache/cache.c
```diff
@@ -0,0 +1,378 @@
+/*
+ * rv32emu is freely redistributable under the MIT License. See the file
+ * "LICENSE" for information on usage and redistribution of this file.
+ */
+
+#include "cache.h"
```
jserv: Always ensure system headers prior to local ones. That is, move `cache.h` at the end of inclusion.

src/cache.c
```diff
@@ -0,0 +1,378 @@
...
+    (((val) * (GOLDEN_RATIO_32)) >> (32 - (cache_size_bits))) % (cache_size)
+
+static uint32_t cache_size, cache_size_bits;
+
+typedef enum {
+    LRU_list,
+    LFU_list,
+    LRU_ghost_list,
+    LFU_ghost_list,
+    list_quantity
```
jserv: Change the name `list_quantity` to `N_CACHE_LIST_TYPES` for better understanding.

tests/cache/cache.c
```diff
@@ -0,0 +1,378 @@
+/*
```
jserv: Why did you duplicate `cache.c`?

mk/cache-test.mk
```diff
@@ -0,0 +1,49 @@
+CACHE_TEST_DIR := tests/cache
+CACHE_BUILD_DIR := build/cache
+TARGET := test_cache.out
```
jserv: The naming scheme violates the consistency.
* `.out` is the suffix for the files generated by test programs. Typically, the text output.
* For host executables, there is no specific suffix.

Here, simply name it `test_cache`.

mk/cache-test.mk
```diff
@@ -0,0 +1,49 @@
...
+$(CACHE_OUT): $(TARGET)
+	$(Q)$(foreach e,$(CACHE_CHECK_ELF_FILES),\
+		$(CACHE_BUILD_DIR)/$(TARGET) $(CACHE_TEST_DIR)/$(e).in > $(CACHE_BUILD_DIR)/$(e).out; \
+	)
+
+$(TARGET): $(CACHE_OBJS)
+	$(VECHO) "  CC\t$@\n"
+	$(Q)$(CC) $^ -o $(CACHE_BUILD_DIR)/$(TARGET)
+	
+$(CACHE_BUILD_DIR)/cache.o: $(CACHE_TEST_DIR)/cache.c
```
jserv: Don't duplicate `cache.c`. Always reuse existing files when possible.

mk/cache-test.mk
```diff
@@ -0,0 +1,49 @@
...
+	$(VECHO) "  CC\t$@\n"
+	$(Q) mkdir -p $(CACHE_BUILD_DIR)
```
jserv: There is no space between `$(Q)` and `mkdir`.

mk/cache-test.mk
jserv: You shall honor the `CFLAGS` in top-level `Makefile`.

src/cache.c
qwe661234: The maximum of list size T1, T2, and B1 is equal to `cache_size / 2`, and the maximum of list size B2 is equal to `cache_size`. Therefore, if user set `cache_size` to more than 2^17, we would need `uint32_t` to store list_size. Otherwise, it might cause overflow.

src/common.h
```diff
@@ -49,3 +49,28 @@
...
+#ifndef container_of
+#ifdef __HAVE_TYPEOF
+#define container_of(ptr, type, member)                            \
+    __extension__({                                                \
+        const __typeof__(((type *) 0)->member) *__pmember = (ptr); \
+        (type *) ((char *) __pmember - offsetof(type, member));    \
+    })
+#else
+#define container_of(ptr, type, member) \
+    ((type *) ((char *) (ptr) -offsetof(type, member)))
```
jserv: clang-format is sometimes stupid, and you have to write as following:
```c
#define container_of(ptr, type, member) \                                                                                                                     
    ((type *) ((char *) (ptr) - (offsetof(type, member))))
```

tests/cache/test_cache.c
```diff
@@ -0,0 +1,63 @@
...
+{
+    char *s = strtok(str, del);
+
+    while (s != NULL) {
+        *arr++ = s;
+        s = strtok(NULL, del);
+    }
+}
+
+int main(int argc, char *argv[])
```
jserv: Add comments about the command sequences.

src/cache.c
```diff
@@ -0,0 +1,388 @@
...
+
+#ifdef __HAVE_TYPEOF
+#define hlist_entry_safe(ptr, type, member)                  \
+    ({                                                       \
+        typeof(ptr) ____ptr = (ptr);                         \
+        ____ptr ? hlist_entry(____ptr, type, member) : NULL; \
+    })
+#else
+#define hlist_entry_safe(ptr, type, member) \
+    ({ (ptr) ? hlist_entry(ptr, type, member) : NULL; })
```
jserv: The pair `({` and `})` is indeed a GNU extension, and we shall avoid using it for non-gcc/clang compilers. Later, rv32emu might support MSVC. Thus, we would write portable code as well.

src/cache.c
```diff
@@ -0,0 +1,388 @@
...
+    for (uint32_t i = 0; i < cache_size; i++) {
+        INIT_HLIST_HEAD(&cache->map->ht_list_head[i]);
+    }
+
+    cache->capacity = cache_size;
+    cache->lru_capacity = cache_size / 2;
+    return cache;
+}
+
+/* Rule of ARC
```
jserv: Should be "Rules"

mk/cache-test.mk
```diff
@@ -0,0 +1,49 @@
...
+		$(CACHE_BUILD_DIR)/$(TARGET) $(CACHE_TEST_DIR)/$(e).in > $(CACHE_BUILD_DIR)/$(e).out; \
+	)
+
+$(TARGET): $(CACHE_OBJS)
+	$(VECHO) "  CC\t$@\n"
+	$(Q)$(CC) $^ -o $(CACHE_BUILD_DIR)/$(TARGET)
+	
+$(CACHE_BUILD_DIR)/cache.o: src/cache.c 
+	$(VECHO) "  CC\t$@\n"
+	$(Q)mkdir -p $(CACHE_BUILD_DIR)
```
jserv: Replace this line with `$(Q)mkdir -p $(dir $@)`

src/cache.c
```diff
@@ -0,0 +1,388 @@
...
+             (cache->capacity - cache->lru_capacity))
+        move_to_mru(
+            cache,
+            list_last_entry(cache->list_table[LFU_list], arc_entry_t, list),
+            LFU_ghost_list);
+}
+
+void *cache_get(cache_t *cache, uint32_t key)
+{
+    if (cache->capacity <= 0 ||
```
jserv: The type of `capacity` is `uint32_t`, and it is always >= 0. Therefore, you shall never validate it with `<= 0`.

src/cache.c
```diff
@@ -0,0 +1,388 @@
...
+#include <stddef.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "cache.h"
+
+#define MIN(a, b) ((a < b) ? a : b)
+#define GOLDEN_RATIO_32 0x61C88647
+#define HASH(val) \
+    (((val) * (GOLDEN_RATIO_32)) >> (32 - (cache_size_bits))) % (cache_size)
```
jserv: We can specify the restriction for `cache_size`, which MUST be a power of 2. Then, `% (cache_size)` can be replaced with `& (cache_size - 1)`.

.github/workflows/main.yml
```diff
@@ -15,6 +15,8 @@ jobs:
       run: make
     - name: check
       run: make check
+    - name: cache
+      run: make cache
```
jserv, line 19: The target should not be `cache`. Instead, it would be `make tests`, which triggers the automated testing for all available tests. At the moment, cache specific items are associated.

jserv: The target should not be `cache`. Instead, it would be `make tests`, which triggers the automated testing for all available tests. At the moment, cache specific items are associated.

src/cache.c
qwe661234: I modify it to 
``` c
if (!cache->capacity)
```

.github/workflows/main.yml
qwe661234: Does this means changing the `name` column to `tests` while maintaining the `make cache` command in the `run` field? After this change, other tests would put command in the `run` field.
```
name: tests
run: make cache
     make XXX-test
     ...
```

jserv: The name of GitHub Actions would accordingly be changed to `tests`.

mk/tests.mk
```diff
@@ -0,0 +1,49 @@
...
+$(CACHE_OUT): $(TARGET)
+	$(Q)$(foreach e,$(CACHE_CHECK_ELF_FILES),\
+		$(CACHE_BUILD_DIR)/$(TARGET) $(CACHE_TEST_DIR)/$(e).in > $(CACHE_BUILD_DIR)/$(e).out; \
+	)
+
+$(TARGET): $(CACHE_OBJS)
+	$(VECHO) "  CC\t$@\n"
+	$(Q)$(CC) $^ -o $(CACHE_BUILD_DIR)/$(TARGET)
+	
+$(CACHE_BUILD_DIR)/cache.o: src/cache.c
```
jserv: You can unify the rules by the following:
```
$(CACHE_BUILD_DIR)/%.o: $(CACHE_TEST_DIR)/%.c
```

src/cache.h
```diff
@@ -0,0 +1,40 @@
...
+ * @return: the replaced entry or NULL
+ */
+void *cache_put(struct cache *cache, uint32_t key, void *value);
+
+/**
+ * cache_free - free a cache
+ * @cache: a pointer points to target cache
+ * @release_entry: a function for freeing cache entry completely
+ */
+void cache_free(struct cache *cache, void (*release_entry)(void *));
```
jserv: Rename `release_entry` to `callback`.

mk/tests.mk
```diff
@@ -0,0 +1,49 @@
+CACHE_TEST_DIR := tests/cache
+CACHE_BUILD_DIR := build/cache
+TARGET := test_cache
+
+CACHE_OBJS := cache.o \
```
jserv: Split `cache.o` into a new line.

mk/tests.mk
qwe661234: The directory of `cache.c` is `src`, but the directory of `test_cache.c` is `tests/cache`. So I cannot nuify the rules to this.

jserv: No, you don't have to build `$(CACHE_BUILD_DIR)/cache.o`. Instead, only `$(CACHE_BUILD_DIR)/test_cache.o` should be built here. The test oriented rules always depend on the generation of `$(OBJS)`.

qwe661234: I see, `cache.o` would be built by `make all`

qwe661234: However, the GitHub Action fails now, because we don't integrate the cache and it does not build `cache.o`.

jserv: The appropriate sequence would be
```
make
make tests
```

Alternatively, you can extend item `check` to perform `make tests`.

src/cache.c
```diff
@@ -0,0 +1,387 @@
...
+    struct hlist_node *first = h->first;
+    n->next = first;
+    if (first)
+        first->pprev = &n->next;
+
+    h->first = n;
+    n->pprev = &h->first;
+}
+
+static inline int hlist_unhashed(const struct hlist_node *h)
```
jserv: Change its type to `bool`.

src/cache.c
```diff
@@ -0,0 +1,387 @@
...
+    struct list_head list;
+    struct hlist_node ht_list;
+} arc_entry_t;
+
+typedef struct {
+    struct hlist_head *ht_list_head;
+} hashtable_t;
+
+typedef struct cache {
+    struct list_head *list_table[N_CACHE_LIST_TYPES];
```
jserv: Rename `list_table` to `lists` for the sake of simplicity.

src/cache.c
```diff
@@ -0,0 +1,387 @@
...
+
+struct hlist_node {
+    struct hlist_node *next, **pprev;
+};
+
+typedef struct {
+    void *value;
+    uint32_t key;
+    cache_list_t arc_type;
+    struct list_head list;
```
jserv, line 61: Is it feasible to drop the use of `list` here?

src/cache.c
```diff
@@ -0,0 +1,387 @@
...
+};
+
+struct hlist_node {
+    struct hlist_node *next, **pprev;
+};
+
+typedef struct {
+    void *value;
+    uint32_t key;
+    cache_list_t arc_type;
```
jserv: Rename `arc_type` to `type` since it is referred to a member of `cache_t` object.

src/cache.c
```diff
@@ -0,0 +1,387 @@
...
+
+#define MIN(a, b) ((a < b) ? a : b)
+#define GOLDEN_RATIO_32 0x61C88647
+#define HASH(val) \
+    (((val) * (GOLDEN_RATIO_32)) >> (32 - (cache_size_bits))) & (cache_size - 1)
+
+static uint32_t cache_size, cache_size_bits;
+
+/*
+ * ARC improves the basic LRU strategy by splitting the cache into two lists,
```
jserv: Refine the comments to reflect the meaning of each identifier enumerated.

jserv (changes requested):
For the consistency, file name would use dash (`-`) rather than underscores (`_`). One exception is to emphasize the items you want to validate. e.g., `test-cache_t`. In this pull request, we can safely enforce the replacement for all underscores.

mk/tests.mk
jserv: By the way, the next tests would be associated with #94 .

src/cache.c
qwe661234: We cannot manipulate the doubly-linked list withot `list`.

qwe661234: `list` is for four cache lists T1, T2, B1, and B2 and `ht_list` is for hash table, we should maintain 4 lists by `list` and find out the certain cache entry by `ht_list`.

jserv: Add corresponding comments along with further improvements.

.github/workflows/main.yml
```diff
@@ -14,7 +14,9 @@ jobs:
     - name: default build
       run: make
     - name: check
```
jserv: This item should be `check + tests`.

jserv (changes requested):
Resolve CI pipeline failure.

tests/cache/test-cache.c
```diff
@@ -0,0 +1,69 @@
...
+{
+    char *s = strtok(str, del);
+
+    while (s != NULL) {
+        *arr++ = s;
+        s = strtok(NULL, del);
+    }
+}
+
+/* Commands of test_cache
```
jserv: Should be `test-cache`.

PR #120, opened 2023-03-17
 Fix logical mistakes for ARC

qwe661234:
After importing more cache tests, we discovered some logical mistakes in replacing cache entries. This commit patches these mistakes.


src/cache.c
```diff
@@ -249,6 +249,20 @@ cache_t *cache_create(int size_bits)
                cache->list_size[LFU_ghost_list] <=                          \
            2 * cache->capacity);
 
+#define replace_list(op1, op2)                                                 \
```
jserv: Capitalize this macro.

Issue #122, opened 2023-04-05
Improve memory read/write


qwe661234:
After analyzing the data collected during the execution of the dhrystone benchmark, it becomes evident that the primary performance bottleneck lies in memory read and write operations. To address this issue and improve performance, we should consider modifying the implementation of these memory operations.

```
  21.15%  rv32emu  rv32emu           [.] memory_write
  17.21%  rv32emu  rv32emu           [.] do_sw
  12.82%  rv32emu  rv32emu           [.] do_lw
   6.89%  rv32emu  rv32emu           [.] memory_read_w
   6.68%  rv32emu  rv32emu           [.] on_mem_write_w

```

jserv:
[semu.c](https://github.com/jserv/semu/blob/master/semu.c) applies the technique [Duff's device](https://en.wikipedia.org/wiki/Duff%27s_device) to generalize memory operations. See `ram_load` and `ram_store`.
We may eliminate unnecessary range checks for memory chunks. Furthermore, memory chunk in `src/io.c` is not really vital for implementing RV32 emulation though.

PR #123, opened 2023-04-07
Improve memory read/write

qwe661234:
After analyzing the data collected during the execution of the Dhrystone benchmark, we found that the primary performance bottleneck lies in memory read/write operations. To address this issue, we modified the implementation of memory read/write by eliminating unnecessary checks and replacing slow operations with more efficient ones. Additionally, we integrated a simple memory pool and resolved the issue of disabled unaligned memory access, the user can add argument `--misalign` to enable misaligned memory access.

After analyzing the performance results from running Dhrystone, we observed a significant improvement following our enhancements to the memory I/O. 

| Test | 66200d0 | improvement | Speedup |
|--------| -------- | -------- |-------- |
| Dhrystone  | 815 DMIPS    |  1245 DMIPS | +52.7% |

See: #122

### DIFF
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -90,6 +90,22 @@ static void rv_exception_default_handler(riscv_t *rv)
 RV_EXCEPTION_LIST
 #undef _
 
+/* wrap load/store and insn misaligned handler
+ * @mask_or_pc: mask for load/store and pc for insn misaligned handler.
+ * @type: type of misaligned handler
+ * @compress: compressed instruction or not
+ * @IO: whether the misaligned handler is for load/store or insn.
+ */
+#define RV_EXC_MISALIGN_HANDLER(mask_or_pc, type, compress, IO)       \
+    IIF(IO)                                                           \
+    (if (!rv->io.allow_misalign && unlikely(addr & (mask_or_pc))),    \
+     if (unlikely(insn_is_misaligned(rv->PC))))                       \
+    {                                                                 \
+        rv->compressed = compress;                                    \
+        rv_except_##type##_misaligned(rv, IIF(IO)(addr, mask_or_pc)); \
+        return false;                                                 \
+    }
+
 /* Get current time in microsecnds and update csr_time register */
 static inline void update_time(riscv_t *rv)
 {
@@ -310,11 +326,7 @@ RVOP(jal, {
     if (ir->rd)
         rv->X[ir->rd] = pc + ir->insn_len;
     /* check instruction misaligned */
-    if (unlikely(insn_is_misaligned(rv->PC))) {
-        rv->compressed = false;
-        rv_except_insn_misaligned(rv, pc);
-        return false;
-    }
+    RV_EXC_MISALIGN_HANDLER(pc, insn, false, 0);
     return true;
 })
 
@@ -333,11 +345,7 @@ RVOP(jalr, {
     if (ir->rd)
         rv->X[ir->rd] = pc + ir->insn_len;
     /* check instruction misaligned */
-    if (unlikely(insn_is_misaligned(rv->PC))) {
-        rv->compressed = false;
-        rv_except_insn_misaligned(rv, pc);
-        return false;
-    }
+    RV_EXC_MISALIGN_HANDLER(pc, insn, false, 0);
     return true;
 })
 
@@ -352,11 +360,7 @@ RVOP(beq, {
     }
     rv->PC += ir->imm;
     /* check instruction misaligned */
-    if (unlikely(insn_is_misaligned(rv->PC))) {
-        rv->compressed = false;
-        rv_except_insn_misaligned(rv, pc);
-        return false;
-    }
+    RV_EXC_MISALIGN_HANDLER(pc, insn, false, 0);
     if (ir->branch_taken)
         return ir->branch_taken->impl(rv, ir->branch_taken);
     return true;
@@ -373,11 +377,7 @@ RVOP(bne, {
     }
     rv->PC += ir->imm;
     /* check instruction misaligned */
-    if (unlikely(insn_is_misaligned(rv->PC))) {
-        rv->compressed = false;
-        rv_except_insn_misaligned(rv, pc);
-        return false;
-    }
+    RV_EXC_MISALIGN_HANDLER(pc, insn, false, 0);
     if (ir->branch_taken)
         return ir->branch_taken->impl(rv, ir->branch_taken);
     return true;
@@ -394,11 +394,7 @@ RVOP(blt, {
     }
     rv->PC += ir->imm;
     /* check instruction misaligned */
-    if (unlikely(insn_is_misaligned(rv->PC))) {
-        rv->compressed = false;
-        rv_except_insn_misaligned(rv, pc);
-        return false;
-    }
+    RV_EXC_MISALIGN_HANDLER(pc, insn, false, 0);
     if (ir->branch_taken)
         return ir->branch_taken->impl(rv, ir->branch_taken);
     return true;
@@ -415,11 +411,7 @@ RVOP(bge, {
     }
     rv->PC += ir->imm;
     /* check instruction misaligned */
-    if (unlikely(insn_is_misaligned(rv->PC))) {
-        rv->compressed = false;
-        rv_except_insn_misaligned(rv, pc);
-        return false;
-    }
+    RV_EXC_MISALIGN_HANDLER(pc, insn, false, 0);
     if (ir->branch_taken)
         return ir->branch_taken->impl(rv, ir->branch_taken);
     return true;
@@ -436,11 +428,7 @@ RVOP(bltu, {
     }
     rv->PC += ir->imm;
     /* check instruction misaligned */
-    if (unlikely(insn_is_misaligned(rv->PC))) {
-        rv->compressed = false;
-        rv_except_insn_misaligned(rv, pc);
-        return false;
-    }
+    RV_EXC_MISALIGN_HANDLER(pc, insn, false, 0);
     if (ir->branch_taken)
         return ir->branch_taken->impl(rv, ir->branch_taken);
     return true;
@@ -457,11 +445,7 @@ RVOP(bgeu, {
     }
     rv->PC += ir->imm;
     /* check instruction misaligned */
-    if (unlikely(insn_is_misaligned(rv->PC))) {
-        rv->compressed = false;
-        rv_except_insn_misaligned(rv, pc);
-        return false;
-    }
+    RV_EXC_MISALIGN_HANDLER(pc, insn, false, 0);
     if (ir->branch_taken)
         return ir->branch_taken->impl(rv, ir->branch_taken);
     return true;
@@ -476,22 +460,14 @@ RVOP(lb, {
 /* LH: Load Halfword */
 RVOP(lh, {
     const uint32_t addr = rv->X[ir->rs1] + ir->imm;
-    if (unlikely(addr & 1)) {
-        rv->compressed = false;
-        rv_except_load_misaligned(rv, addr);
-        return false;
-    }
+    RV_EXC_MISALIGN_HANDLER(1, load, false, 1);
     rv->X[ir->rd] = sign_extend_h(rv->io.mem_read_s(rv, addr));
 })
 
 /* LW: Load Word */
 RVOP(lw, {
     const uint32_t addr = rv->X[ir->rs1] + ir->imm;
-    if (unlikely(addr & 3)) {
-        rv->compressed = false;
-        rv_except_load_misaligned(rv, addr);
-        return false;
-    }
+    RV_EXC_MISALIGN_HANDLER(3, load, false, 1);
     rv->X[ir->rd] = rv->io.mem_read_w(rv, addr);
 })
 
@@ -501,11 +477,7 @@ RVOP(lbu, { rv->X[ir->rd] = rv->io.mem_read_b(rv, rv->X[ir->rs1] + ir->imm); })
 /* LHU: Load Halfword Unsigned */
 RVOP(lhu, {
     const uint32_t addr = rv->X[ir->rs1] + ir->imm;
-    if (unlikely(addr & 1)) {
-        rv->compressed = false;
-        rv_except_load_misaligned(rv, addr);
-        return false;
-    }
+    RV_EXC_MISALIGN_HANDLER(1, load, false, 1);
     rv->X[ir->rd] = rv->io.mem_read_s(rv, addr);
 })
 
@@ -515,22 +487,14 @@ RVOP(sb, { rv->io.mem_write_b(rv, rv->X[ir->rs1] + ir->imm, rv->X[ir->rs2]); })
 /* SH: Store Halfword */
 RVOP(sh, {
     const uint32_t addr = rv->X[ir->rs1] + ir->imm;
-    if (unlikely(addr & 1)) {
-        rv->compressed = false;
-        rv_except_store_misaligned(rv, addr);
-        return false;
-    }
+    RV_EXC_MISALIGN_HANDLER(1, store, false, 1);
     rv->io.mem_write_s(rv, addr, rv->X[ir->rs2]);
 })
 
 /* SW: Store Word */
 RVOP(sw, {
     const uint32_t addr = rv->X[ir->rs1] + ir->imm;
-    if (unlikely(addr & 3)) {
-        rv->compressed = false;
-        rv_except_store_misaligned(rv, addr);
-        return false;
-    }
+    RV_EXC_MISALIGN_HANDLER(3, store, false, 1);
     rv->io.mem_write_w(rv, addr, rv->X[ir->rs2]);
 })
 
@@ -1088,11 +1052,7 @@ RVOP(caddi4spn, { rv->X[ir->rd] = rv->X[2] + (uint16_t) ir->imm; })
  */
 RVOP(clw, {
     const uint32_t addr = rv->X[ir->rs1] + (uint32_t) ir->imm;
-    if (unlikely(addr & 3)) {
-        rv->compressed = true;
-        rv_except_load_misaligned(rv, addr);
-        return false;
-    }
+    RV_EXC_MISALIGN_HANDLER(3, load, true, 1);
     rv->X[ir->rd] = rv->io.mem_read_w(rv, addr);
 })
 
@@ -1103,11 +1063,7 @@ RVOP(clw, {
  */
 RVOP(csw, {
     const uint32_t addr = rv->X[ir->rs1] + (uint32_t) ir->imm;
-    if (unlikely(addr & 3)) {
-        rv->compressed = true;
-        rv_except_store_misaligned(rv, addr);
-        return false;
-    }
+    RV_EXC_MISALIGN_HANDLER(3, store, true, 1);
     rv->io.mem_write_w(rv, addr, rv->X[ir->rs2]);
 })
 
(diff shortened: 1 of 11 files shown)


src/riscv_private.h
```diff
@@ -119,6 +119,8 @@ struct riscv_internal {
 
     bool compressed;       /**< current instruction is compressed or not */
     block_map_t block_map; /**< basic block map */
+
+    bool is_misaliged_enabled; /**< enable memory misaligned access */
```
jserv: Rename to `allow_ misalign`.

jserv (changes requested):
Since the misaligned memory access is optional, we don't have to check it for the internals of emulator. Instead, we can register the callback functions while initializing the emulator.

See [async.h](https://github.com/embedded2016/server-framework/blob/master/async.h) and [async.c](https://github.com/embedded2016/server-framework/blob/master/async.c) for details.
```c
/* API gateway */
struct __ASYNC_API__ Async = {
    .create = async_create,
    .signal = async_signal,
    .wait = async_wait,
    .finish = async_finish,
    .run = async_run,
};
```

Once you prepare the functions specialized for both misaligned memory and strictly-aligned memory access, the proper I/O implementations can be hooked up. Consequently, you don't have to check `allow_ misalign` all the time.

src/io.c
```diff
@@ -125,14 +109,46 @@ void memory_write(memory_t *m, uint32_t addr, const uint8_t *src, uint32_t size)
...
         c->data[p & 0xffff] = src[i];
     }
 }
 
+void memory_write_w(memory_t *m, uint32_t addr, const uint8_t *src)
+{
+    const uint32_t addr_lo = addr & mask_lo, addr_hi = addr >> 16;
+    chunk_t *c = m->chunks[addr_hi];
+    if (unlikely(!c)) {
+        c = calloc(1, sizeof(chunk_t));
```
jserv: Can you eliminate the use of `calloc` here?

src/main.c
```diff
@@ -202,7 +211,7 @@ int main(int argc, char **args)
         state->break_addr = end->st_value;
 
     /* create the RISC-V runtime */
-    riscv_t *rv = rv_create(&io, state);
+    riscv_t *rv = rv_create(&io, state, opt_misaligned);
```
jserv: Don't change the prototype of `rv_create()`. Instead, consolidate `state` or `io`.

jserv:
> | Test	| 66200d0	| Proposed |	Speedup |
> |-------|--------------|-----------|-------------|
> | Dhrystone	| 815 DMIPS	| 1200 DMIPS	| +47.2% |

What else do you consider it would be beneficial to employ in place of Spike for faster interpreting?

src/io.c
qwe661234: If we want to eliminate it, we should know the total memory size and allocate it first.

qwe661234:
> Since the misaligned memory access is optional, we don't have to check it for the internals of emulator. Instead, we can register the callback functions while initializing the emulator.
> 
> See [async.h](https://github.com/embedded2016/server-framework/blob/master/async.h) and [async.c](https://github.com/embedded2016/server-framework/blob/master/async.c) for details.
> 
> ```c
> /* API gateway */
> struct __ASYNC_API__ Async = {
>     .create = async_create,
>     .signal = async_signal,
>     .wait = async_wait,
>     .finish = async_finish,
>     .run = async_run,
> };
> ```
> 
> Once you prepare the functions specialized for both misaligned memory and strictly-aligned memory access, the proper I/O implementations can be hooked up. Consequently, you don't have to check `allow_ misalign` all the time.

src/io.c
jserv: Let's introduce memory pool within the pull request.

qwe661234:
> Since the misaligned memory access is optional, we don't have to check it for the internals of emulator. Instead, we can register the callback functions while initializing the emulator.
> 
> See [async.h](https://github.com/embedded2016/server-framework/blob/master/async.h) and [async.c](https://github.com/embedded2016/server-framework/blob/master/async.c) for details.
> 
> ```c
> /* API gateway */
> struct __ASYNC_API__ Async = {
>     .create = async_create,
>     .signal = async_signal,
>     .wait = async_wait,
>     .finish = async_finish,
>     .run = async_run,
> };
> ```
> 
> Once you prepare the functions specialized for both misaligned memory and strictly-aligned memory access, the proper I/O implementations can be hooked up. Consequently, you don't have to check `allow_ misalign` all the time.

But the memory I/O implemention for for both misaligned memory and strictly-aligned memory access is the same, the only difference is that we need to handle exception if misaligned memory access is disabled.

jserv:
> the memory I/O implemention for for both misaligned memory and strictly-aligned memory access is the same, the only difference is that we need to handle exception if misaligned memory access is disabled.

Then, you can hook exception handlers which are aware of misaligned memory access by means of runtime dispatch.

qwe661234:
> > the memory I/O implemention for for both misaligned memory and strictly-aligned memory access is the same, the only difference is that we need to handle exception if misaligned memory access is disabled.
> 
> Then, you can hook exception handlers which are aware of misaligned memory access by means of runtime dispatch.

Does this means send a callback function to memory read/ write? Because we don't need load/store handler if the memory misaligned access is enabled.

jserv:
> > Then, you can hook exception handlers which are aware of misaligned memory access by means of runtime dispatch.
> Does this means send a callback function to memory read/ write?

Sort of. In particular, we can shrink the code below:
```c
/* SW: Store Word */
RVOP(sw, {
    const uint32_t addr = rv->X[ir->rs1] + ir->imm;
    if (!rv->is_misaliged_enabled && unlikely(addr & 3)) {
        rv->compressed = false;
        rv_except_store_misaligned(rv, addr);
        return false;
    ...
```
by replacing with something like the following:
```c
/* SW: Store Word */
RVOP(sw, {
    const uint32_t addr = rv->X[ir->rs1] + ir->imm;
    RV_EXC_MISAGLIN_HANDLER(); /* some macro */
```

Meanwhile, we can hide the implementation details about `allow_ misalign` checks.

src/riscv.h
```diff
@@ -107,6 +107,9 @@ typedef struct {
     /* system */
     riscv_on_ecall on_ecall;
     riscv_on_ebreak on_ebreak;
+
+    /* enable memory misaligned access */
```
jserv: Should be "enable misaligned memory access"

src/emulate.c
```diff
@@ -90,6 +90,18 @@ static void rv_exception_default_handler(riscv_t *rv)
 RV_EXCEPTION_LIST
 #undef _
 
+#define RV_EXC_MISAGLIN_HANDLER(mask_or_pc, type, compress, IO)    \
```
jserv: Typo. Should be `RV_EXC_MISALIGN_HANDLER`.

jserv: Add some comments for this macro.

src/emulate.c
```diff
@@ -90,6 +90,19 @@ static void rv_exception_default_handler(riscv_t *rv)
 RV_EXCEPTION_LIST
 #undef _
 
+/* wrap load/store and insn misaligned handler */
```
jserv: You shall explain the parameters in details.

jserv (changes requested):
Integrate memory pool in conjunction of proposed memory operation changes.

src/emulate.c
qwe661234: Updated

src/emulate.c
```diff
@@ -90,6 +90,25 @@ static void rv_exception_default_handler(riscv_t *rv)
 RV_EXCEPTION_LIST
 #undef _
 
+/* wrap load/store and insn misaligned handler
+ * @mask_or_pc: misaligned mask for load/store misaligned handler and pc for
```
jserv: Shorten the description for `mask_or_pc`.

src/io.c
jserv: You don't have to integrate TLSF (or its variants such as xvmalloc) within this commit. Instead, an arbitrary memory pool works for this case.

qwe661234: I roughly integrate the [TLSF 2.4.6](http://www.gii.upv.es/tlsf/main/repo) to file `io.c`.

jserv (changes requested):
Get rid of GPL'd TLSF. You MUST use the BSD license re-implementation instead. See #94 .
Original TLSF is licensed under GNU GPL, which raises more restrictions over MIT License. We can integrate BSD License components in rv32emu, but it is not allowed to relicense GPL'd ones into rv32emu. That is exactly what BSD re-implementation appears.

jserv:
Also, the original TLSF suffers from several security concerns. For this issue, we only need a simplified memory pool without integrating TLSF, where #94 emphasizes.

qwe661234:
> Also, the original TLSF suffers from several security concerns. For this issue, we only need a simplified memory pool without integrating TLSF, where #94 emphasizes.

I believe that the memory pool with a fixed chunk size can fulfill our requirements, so I decided to integrate this type of memory pool into the file `io.c`.

jserv (changes requested):
Squash the commits along with comprehensive git commit messages. In addition, show the latest benchmark results.

src/emulate.c
```diff
@@ -90,6 +90,24 @@ static void rv_exception_default_handler(riscv_t *rv)
 RV_EXCEPTION_LIST
 #undef _
 
+/* wrap load/store and insn misaligned handler
+ * @mask_or_pc: mask for load/store and pc for insn misaligned handler.
+ * @type: type of misaligned handler
+ * @compress: compressed instruction or not
+ * @IO: misaligned handler is load/store or insn
```
jserv: Improve the writing for `@IO` entry.

jserv: Ask ChatGPT to proofread.

src/io.c
```diff
@@ -9,27 +9,25 @@
 #include <string.h>
 
 #include "io.h"
+#include "mpool.h"
 
 static const uint32_t mask_lo = 0xffff;
 static const uint32_t mask_hi = ~(0xffff);
 
+struct mpool *mp;
```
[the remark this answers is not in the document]
jserv: Can you make it local? That is, declare this variable as `static`.

src/io.c
```diff
@@ -125,23 +107,54 @@ void memory_write(memory_t *m, uint32_t addr, const uint8_t *src, uint32_t size)
         uint32_t x = p >> 16;
         chunk_t *c = m->chunks[x];
         if (!c) {
-            c = malloc(sizeof(chunk_t));
-            memset(c->data, 0, sizeof(c->data));
+            c = calloc(1, sizeof(chunk_t));
```
jserv: The region should be requested from memory pool.

src/mpool.c
```diff
@@ -0,0 +1,84 @@
...
+typedef struct mpool {
+    long chunk_ct;
+    long page_count;
+    size_t chunk_size;
+    char *mempage;
+    struct memchunk *free_chunk_head;
+    struct mpool *next;
+} mpool_t;
+
+static void *get_mmap(long sz)
```
jserv: Check the file `src/elf.c` for `USE_MMAP`. At present, you can figure out Microsoft Windows counterpart or stop compilation (by placing `#error`).

Portability is an important consideration.

src/mpool.h
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/mman.h>
+#include <unistd.h>
+
+struct memchunk;
```
jserv: `struct memchunk` is implementation specific. Don't appear in this header file.

src/mpool.h
```diff
@@ -0,0 +1,24 @@
+/*
+ * rv32emu is freely redistributable under the MIT License. See the file
+ * "LICENSE" for information on usage and redistribution of this file.
+ */
+
```
jserv, line 6: Add proper `#pragma once`.

src/mpool.h
```diff
@@ -0,0 +1,24 @@
...
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/mman.h>
```
jserv: Minimize the header inclusion! Only keep the necessary ones.

src/mpool.c
```diff
@@ -0,0 +1,84 @@
...
+ */
+
+#include "mpool.h"
+
+typedef struct memchunk {
+    struct memchunk *next;
+} memchunk_t;
+
+typedef struct mpool {
+    long chunk_ct;
```
jserv: `chunk_ct` is not a appropriate mnemonic.

src/mpool.c
```diff
@@ -0,0 +1,84 @@
...
+    long page_count;
+    size_t chunk_size;
+    char *mempage;
```
jserv: Rename it to `mapped`, which stands for mapped memory region.

src/mpool.c
```diff
@@ -0,0 +1,84 @@
...
+#include "mpool.h"
+
+typedef struct memchunk {
+    struct memchunk *next;
+} memchunk_t;
+
+typedef struct mpool {
+    long chunk_ct;
+    long page_count;
+    size_t chunk_size;
```
jserv, line 27: Check if `size_t` fits the range requirements for `chunk_ct` and `page_count`.

src/mpool.c
```diff
@@ -0,0 +1,84 @@
...
+    void *p =
+        mmap(0, sz, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANON, -1, 0);
+    if (p == MAP_FAILED)
+        return NULL;
+    return p;
+}
+
+mpool_t *mpool_create(size_t pool_size, size_t chunk_size)
+{
+    mpool_t *new_mp = malloc(sizeof(mpool_t));
```
jserv, line 49: Return `NULL` when `malloc` or `mmap` fails.

src/io.c
```diff
@@ -125,23 +107,54 @@ void memory_write(memory_t *m, uint32_t addr, const uint8_t *src, uint32_t size)
...
             m->chunks[x] = c;
         }
         c->data[p & 0xffff] = src[i];
     }
 }
 
+void memory_write_w(memory_t *m, uint32_t addr, const uint8_t *src)
```
jserv: Use macro to generate the function bodies of `memory_write_{w,s,b}`.

jserv (changes requested):
Append `Close #122` at the end of git commit messages.

src/mpool.h
```diff
@@ -0,0 +1,25 @@
+/*
+ * rv32emu is freely redistributable under the MIT License. See the file
+ * "LICENSE" for information on usage and redistribution of this file.
+ */
+#pragma once
+
+#include <stdlib.h>
+#include <string.h>
+#if !defined(_WIN32)
```
jserv: No, implementation specific statements should no appear in this header! Keep it simple and stupid.

src/mpool.h
```diff
@@ -0,0 +1,20 @@
+/*
+ * rv32emu is freely redistributable under the MIT License. See the file
+ * "LICENSE" for information on usage and redistribution of this file.
+ */
+#pragma once
+
+#include <stdlib.h>
```
jserv: Only `<stddef.h>` should be included here.

src/mpool.c
```diff
@@ -0,0 +1,109 @@
...
+typedef struct mpool {
+    size_t chunk_count;
+    size_t page_count;
+    size_t chunk_size;
+    char *mapped;
+    struct memchunk *free_chunk_head;
+    struct mpool *next;
+} mpool_t;
+
+static void *get_mmap(size_t sz)
```
jserv: Rename to `mem_arena` and add comments.

src/io.c
```diff
@@ -9,27 +9,25 @@
...
 static const uint32_t mask_hi = ~(0xffff);
 
+static struct mpool *mp;
+
 memory_t *memory_new()
 {
-    memory_t *m = malloc(sizeof(memory_t));
-    memset(m->chunks, 0, sizeof(m->chunks));
+    memory_t *m = calloc(1, sizeof(memory_t));
+    mp = mpool_create(sizeof(chunk_t) << 5, sizeof(chunk_t));
```
jserv: You shall explain why shift 5 is here.

src/io.c
```diff
@@ -85,63 +83,61 @@ uint32_t memory_ifetch(memory_t *m, uint32_t addr)
...
-            m->chunks[x] = c;
+            c = mpool_calloc(mp);
+            m->chunks[addr_hi] = c;
         }
-        c->data[p & 0xffff] = src[i];
+        c->data[addr_lo] = src[i];
     }
 }
 
+#define memory_write(size, type)                                             \
```
jserv: Rename to `MEM_WRITE_IMPL`.

jserv (changes requested):
Rephrase the git commit message as following:
> After analyzing the data collected during the execution of the Dhrystone benchmark, it was discovered that the primary performance bottleneck lies in memory read/write operations. To address this issue, modifications were made to the implementation of memory read/write operations, eliminating unnecessary checks and replacing slow operations with more efficient ones. Additionally, a simple memory pool was integrated. It is now possible for the user to enable misaligned memory access by launching with option "--misalign".
> 
> The enhancements made to the memory I/O resulted in a significant improvement, as observed through performance results obtained by running Dhrystone.

src/mpool.h
qwe661234: The gcc-12 compiler shows the warning when I only include `<stddef.h>`.

jserv: I don't think so.
```
gcc-12 -c mpool.h -Wall
```
It compiles without warnings. The actual header inclusion should appear in C source files.

src/mpool.h
```diff
@@ -0,0 +1,20 @@
...
+#include <string.h>
+
+struct mpool;
+
+struct mpool *mpool_create(size_t pool_size, size_t chunk_size);
```
jserv, line 17: Add informative comments for each function.

src/io.c
```diff
@@ -9,27 +9,27 @@
...
 static const uint32_t mask_hi = ~(0xffff);
 
+static struct mpool *mp;
+
 memory_t *memory_new()
 {
-    memory_t *m = malloc(sizeof(memory_t));
-    memset(m->chunks, 0, sizeof(m->chunks));
+    memory_t *m = calloc(1, sizeof(memory_t));
+    /* sizeof(chunk_t) << 5 satisfies the memory requirement for all test cases
```
jserv: If `sizeof(chunk_t) << 5` does not fit, there should be some mechanism to extend.

src/mpool.c
```diff
@@ -0,0 +1,109 @@
+/*
+ * rv32emu is freely redistributable under the MIT License. See the file
+ * "LICENSE" for information on usage and redistribution of this file.
+ */
+#if !defined(_WIN32)
+#define USE_MMAP 1
+#include <sys/mman.h>
+#include <unistd.h>
+#endif
+
```
jserv, line 12: Add proper header inclusion here.

src/emulate.c
```diff
@@ -90,6 +90,24 @@ static void rv_exception_default_handler(riscv_t *rv)
...
+ * @compress: compressed instruction or not
+ * @IO: whether the misaligned handler is for load/store or insn.
+ */
+#define RV_EXC_MISALIGN_HANDLER(mask_or_pc, type, compress, IO)    \
+    IIF(IO)                                                        \
+    (if (!rv->io.allow_misalign && unlikely(addr & (mask_or_pc))), \
+     if (unlikely(insn_is_misaligned(rv->PC))))                    \
+    {                                                              \
+        rv->compressed = compress;                                 \
+        IIF(IO)                                                    \
```
jserv: Can you simplify as following?
```c
rv_except_##type##_misaligned(rv, IIF(IO)(addr, mask_or_pc));
```

jserv (changes requested):
Append a test item which triggers option `--misalign` in GitHub Actions. That is, we shall regularly validate the case when misaligned memory access is allowed for CI pipeline.

PR #126, opened 2023-04-28
Fix the implementations of FCVT.S.W and FCVT.S.WU

alanjian85:
The implementations of FCVT.S.W and FCVT.S.WU in the commit 04ce561 are incorrect. In fact, instead of updating the bit representation of the float number directly, a correct implementation should convert the signed/unsigned number to float number while keeping the value of integer numbers. As a result, I use implicit conversion between integer and float at line 1031 and 1034 and assigned the converted value to rv->F rather than rv->F_int, which represent the float number itself and its representation, respectively.

### DIFF
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -1028,10 +1028,10 @@ RVOP(fclasss, {
 })
 
 /* FCVT.S.W */
-RVOP(fcvtsw, { rv->F_int[ir->rd] = rv->X[ir->rs1]; })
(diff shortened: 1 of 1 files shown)


qwe661234:
After examination, this patch fixes the segmentation fault error that occurs during the `make quake`.

qwe661234 (approved):

jserv:
Thank @alanjian85 for contributing!

PR #128, opened 2023-05-11
Add an option to dump registers as JSON 

long-long-float:
I added an option `--dump-registers` to dump registers as JSON.
This is useful for automated testing RISC-V binaries.


jserv (changes requested):
Squash the commits and update `README.md`.

src/emulate.c
```diff
@@ -1456,3 +1457,13 @@ void ecall_handler(riscv_t *rv)
     rv_except_ecall_M(rv, 0);
     syscall_handler(rv);
 }
+
```
RinHizakura, line 1592: I feel a little bit weird to dump the result at STDOUT and forcing to silence the output from `syscall_handler` for this. Since these two features don't depend on each other. Will providing an extra argument for option `--dump-registers` to specify where to output(it could be a file name or STDOUT, if we really need it) become better?

And if we would like to not output the message from  `syscall_handler`, we may introduce another command line option there.

jserv:
@long-long-float, Can you refine this pull as @RinHizakura suggested?

jserv (changes requested):
Rebase and resolve the conflicts.

long-long-float:
@jserv Yes, sure. 

Also I will add a silence option respecting the following comment.
> And if we would like to not output the message from syscall_handler, we may introduce another command line option there.

----

> Squash the commits and update README.md.

I will squash the commits. However how can I update `README.md` ?

jserv:
> I will squash the commits. However how can I update `README.md` ?

You can add the description about the proposed option of this emulator.

RinHizakura (approved):
Thank you! I have no problem now. Just remember to update the `README.md` as @jserv required.

jserv:
Thank @long-long-float for contributing!

PR #130, opened 2023-05-15
Consolidate CI/CD test for gdbstub

RinHizakura:
We can't get expected result when running `make gdbstub-test` under the old version of  `rv32emu`. That's because the implementation of `rv_step` changed from returning after stepping single instruction to returning after a block execution. To solve the issue, I simply take gdbstub mode as a special case for  `rv_step`, which will return directly after executing one instruction, so we can check whether the breakpoint hit after every return.

After the changed, we can add the test of gdbstub mode in CICD now. The workflow of CICD is also changed a little bit since we'll need `riscv32-gdb` for the test on gdbstub.

### DIFF
--- a/.ci/gdbstub-test.sh
+++ b/.ci/gdbstub-test.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+export PATH=`pwd`/toolchain/riscv/bin:$PATH
+
 GDB=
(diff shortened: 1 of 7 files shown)


Risheng1128 (approved):

jserv:

PR #131, opened 2023-05-18
Improve red-black tree performance

### LINKED ISSUE #29: Improve red-black tree implementation in map
`map.[ch]` is the C Implementation for C++ std::map using red-black tree. It works well, but there is room for improvements. We can take some tricks used in [Red-black Trees in Linux](https://www.kernel.org/doc/Documentation/rbtree.txt) for potential speedup. In addition, current rbtree can benefit from the introduction of indirect pointers.

TheCloudlet:
`map.[ch]` is the C Implementation for C++ std::map using red-black tree. It works well, but there is room for improvements.

Close https://github.com/sysprog21/rv32emu/issues/29

### DIFF
--- a/src/map.c
+++ b/src/map.c
@@ -1,12 +1,540 @@
-#include "map.h"
+/*
+ * rv32emu is freely redistributable under the MIT License. See the file
+ * "LICENSE" for information on usage and redistribution of this file.
+ */
+
+/* clang-format off */
+
+/*
+ * This map implementation has undergone extensive modifications, heavily
+ * relying on the rb.h header file from jemalloc.
+ * See https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/rb.h .
+ * The original rb.h file served as the foundation and source of inspiration
+ * for adapting and tailoring it specifically for this map implementation.
+ * Therefore, credit and sincere thanks are extended to jemalloc for their
+ * invaluable work.
+ */
+
+/* clang-format on */
+
+#include <assert.h>
 #include <memory.h>
 
-static map_node *map_create_node(void *key,
-                                 void *value,
-                                 size_t ksize,
-                                 size_t vsize)
+#include "map.h"
+
+/*
+ * Each node in the red-black tree consumes at least 1 byte of space (for the
+ * linkage if nothing else, so there are a maximum of sizeof(void *) << 3 rb
+ * tree nodes in any process (and thus, at most sizeof(void *) << 3 nodes in any
+ * rb tree). The choice of algorithm bounds the depth of a tree to twice the
+ * binary log of the number of elements in the tree; the following bound
+ * follows.
+ */
+#define RB_MAX_DEPTH (sizeof(void *) << 4)
+
+/* Left accessors */
+static inline map_node_t *rbnode_get_left(map_node_t *node)
 {
-    map_node *node = malloc(sizeof(map_node));
+    return node->link.left;
+}
+
+static inline void rbnode_set_left(map_node_t *node, map_node_t *left)
+{
+    node->link.left = left;
+}
+
+/* Right accessors */
+static inline map_node_t *rbnode_get_right(map_node_t *node)
+{
+    return (map_node_t *) (((uintptr_t) node->link.right_red) & ~3);
+}
+
+static inline void rbnode_set_right(map_node_t *node, map_node_t *right)
+{
+    node->link.right_red =
+        (map_node_t *) (((uintptr_t) right) |
+                        (((uintptr_t) node->link.right_red) & 1));
+}
+
+/* Color accessors */
+static inline map_color_t rbnode_get_color(const map_node_t *node)
+{
+    return (map_color_t) (((uintptr_t) node->link.right_red) & 1);
+}
+
+static inline void rbnode_set_color(map_node_t *node, map_color_t color)
+{
+    node->link.right_red =
+        (map_node_t *) (((uintptr_t) node->link.right_red & ~3) |
+                        (uintptr_t) color);
+}
+
+static inline void rbnode_set_red(map_node_t *node)
+{
+    node->link.right_red =
+        (map_node_t *) (((uintptr_t) node->link.right_red) | 1);
+}
+
+static inline void rbnode_set_black(map_node_t *node)
+{
+    node->link.right_red =
+        (map_node_t *) (((uintptr_t) node->link.right_red) & ~3);
+}
+
+/* Node initializer */
+static inline void rbnode_init(map_node_t *node)
+{
+    assert((((uintptr_t) node) & (0x1)) == RB_BLACK);
+    rbnode_set_left(node, NULL);
+    rbnode_set_right(node, NULL);
+    rbnode_set_red(node);
+}
+
+/* Internal utility macros */
+#define rbnode_rotate_left(x_type, x_field, x_node, r_node)    \
+    do {                                                       \
+        (r_node) = rbnode_get_right((x_node));                 \
+        rbnode_set_right((x_node), rbnode_get_left((r_node))); \
+        rbnode_set_left((r_node), (x_node));                   \
+    } while (0)
+
+#define rbnode_rotate_right(x_type, x_field, x_node, r_node)   \
+    do {                                                       \
+        (r_node) = rbnode_get_left((x_node));                  \
+        rbnode_set_left((x_node), rbnode_get_right((r_node))); \
+        rbnode_set_right((r_node), (x_node));                  \
+    } while (0)
+
+typedef struct {
+    map_node_t *node;
+    int cmp;
+} rb_path_entry_t;
+
+static map_node_t *rb_search(map_t rbtree, const map_node_t *keynode)
+{
+    int cmp;
+    map_node_t *ret = rbtree->root;
+    while (ret && (cmp = (rbtree->cmp)(keynode->key, ret->key)) != 0) {
+        if (cmp < 0) {
+            ret = rbnode_get_left(ret);
+        } else {
+            ret = rbnode_get_right(ret);
+        }
+    }
+    return ret;
+}
+
+static void rb_insert(map_t rbtree, map_node_t *node)
+{
+    rb_path_entry_t path[RB_MAX_DEPTH];
+    rb_path_entry_t *pathp;
+    rbnode_init(node);
+    /* Wind. */
+    path->node = rbtree->root;
+    for (pathp = path; pathp->node; pathp++) {
+        int cmp = pathp->cmp = (rbtree->cmp)(node->key, pathp->node->key);
+        if (cmp == 0) {
+            break; /* If the key matches something, don't insert */
+        }
+        if (cmp < 0) {
+            pathp[1].node = rbnode_get_left(pathp->node);
+        } else {
+            pathp[1].node = rbnode_get_right(pathp->node);
+        }
+    }
+    pathp->node = node;
+    /* A loop invariant we maintain is that all nodes with
+     * out-of-date summaries live in path[0], path[1], ..., *pathp.
+     * To maintain this, we have to summarize node, since we
+     * decrement pathp before the first iteration.
+     */
+    assert(!rbnode_get_left(node));
+    assert(!rbnode_get_right(node));
+    /* Unwind. */
+    for (pathp--; (uintptr_t) pathp >= (uintptr_t) path; pathp--) {
+        map_node_t *cnode = pathp->node;
+        if (pathp->cmp < 0) {
+            map_node_t *left = pathp[1].node;
+            rbnode_set_left(cnode, left);
+            if (rbnode_get_color(left) == RB_BLACK)
+                return;
+            map_node_t *leftleft = rbnode_get_left(left);
+            if (leftleft && (rbnode_get_color(leftleft) == RB_RED)) {
+                /* Fix up 4-node. */
+                map_node_t *tnode;
+                rbnode_set_black(leftleft);
+                rbnode_rotate_right(map_node_t, link, cnode, tnode);
+                cnode = tnode;
+            }
+        } else {
+            map_node_t *right = pathp[1].node;
+            rbnode_set_right(cnode, right);
+            if (rbnode_get_color(right) == RB_BLACK)
+                return;
+            map_node_t *left = rbnode_get_left(cnode);
+            if (left && (rbnode_get_color(left) == RB_RED)) {
+                /* Split 4-node. */
+                rbnode_set_black(left);
+                rbnode_set_black(right);
+                rbnode_set_red(cnode);
+            } else {
+                /* Lean left. */
+                map_node_t *tnode;
+                map_color_t tcolor = rbnode_get_color(cnode);
+                rbnode_rotate_left(map_node_t, link, cnode, tnode);
+                rbnode_set_color(tnode, tcolor);
+                rbnode_set_red(cnode);
+                cnode = tnode;
+            }
+        }
+        pathp->node = cnode;
+    }
+    /* Set root, and make it black. */
+    rbtree->root = path->node;
+    rbnode_set_black(rbtree->root);
+}
+
+static void rb_remove(map_t rbtree, map_node_t *node)
+{
+    rb_path_entry_t path[RB_MAX_DEPTH];
+    rb_path_entry_t *pathp = NULL;
+    rb_path_entry_t *nodep = NULL;
+    /* Wind. */
+    path->node = rbtree->root;
+    for (pathp = path; pathp->node; pathp++) {
+        int cmp = pathp->cmp = (rbtree->cmp)(node->val, pathp->node->val);
+        if (cmp < 0) {
+            pathp[1].node = rbnode_get_left(pathp->node);
+        } else {
+            pathp[1].node = rbnode_get_right(pathp->node);
+            if (cmp == 0) {
+                /* Find node's successor, in preparation for swap. */
+                pathp->cmp = 1;
+                nodep = pathp;
+                for (pathp++; pathp->node; pathp++) {
+                    pathp->cmp = -1;
+                    pathp[1].node = rbnode_get_left(pathp->node);
+                }
+                break;
+            }
+        }
+    }
+    assert(nodep->node == node);
+    pathp--;
+    if (pathp->node != node) {
+        /* Swap node with its successor. */
+        map_color_t tcolor = rbnode_get_color(pathp->node);
+        rbnode_set_color(pathp->node, rbnode_get_color(node));
+        rbnode_set_left(pathp->node, rbnode_get_left(node));
+        /* If node's successor is its right child, the following code
+         * will do the wrong thing for the right child pointer.
+         * However, it doesn't matter, because the pointer will be
+         * properly set when the successor is pruned.
+         */
+        rbnode_set_right(pathp->node, rbnode_get_right(node));
+        rbnode_set_color(node, tcolor);
+        /* The pruned leaf node's child pointers are never accessed
+         * again, so don't bother setting them to nil.
+         */
+        nodep->node = pathp->node;
+        pathp->node = node;
+        if (nodep == path) {
+            rbtree->root = nodep->node;
+        } else {
+            if (nodep[-1].cmp < 0) {
+                rbnode_set_left(nodep[-1].node, nodep->node);
+            } else {
+                rbnode_set_right(nodep[-1].node, nodep->node);
+            }
+        }
+    } else {
+        map_node_t *left = rbnode_get_left(node);
+        if (left) {
+            /* node has no successor, but it has a left child.
+             * Splice node out, without losing the left child.
+             */
+            assert(rbnode_get_color(node) == RB_BLACK);
+            assert(rbnode_get_color(left) == RB_RED);
+            rbnode_set_black(left);
+            if (pathp == path) {
+                rbtree->root = left;
+                /* Nothing to summarize -- the subtree rooted at the
+                 * node's left child hasn't changed, and it's now the
+                 * root.
+                 */
+            } else {
+                if (pathp[-1].cmp < 0) {
+                    rbnode_set_left(pathp[-1].node, left);
+                } else {
+                    rbnode_set_right(pathp[-1].node, left);
+                }
+            }
+            return;
+        } else if (pathp == path) {
+            /* The tree only contained one node. */
+            rbtree->root = NULL;
+            return;
+        }
+    }
+    /* We've now established the invariant that the node has no right
+     * child (well, morally; we didn't bother nulling it out if we
+     * swapped it with its successor), and that the only nodes with
+     * out-of-date summaries live in path[0], path[1], ..., pathp[-1].
+     */
+    if (rbnode_get_color(pathp->node) == RB_RED) {
+        /* Prune red node, which requires no fixup. */
+        assert(pathp[-1].cmp < 0);
+        rbnode_set_left(pathp[-1].node, NULL);
+        return;
+    }
+    /* The node to be pruned is black, so unwind until balance is
+     * restored.
+     */
+    pathp->node = NULL;
+    for (pathp--; (uintptr_t) pathp >= (uintptr_t) path; pathp--) {
+        assert(pathp->cmp != 0);
+        if (pathp->cmp < 0) {
+            rbnode_set_left(pathp->node, pathp[1].node);
+            if (rbnode_get_color(pathp->node) == RB_RED) {
+                map_node_t *right = rbnode_get_right(pathp->node);
+                map_node_t *rightleft = rbnode_get_left(right);
+                map_node_t *tnode;
+                if (rightleft && (rbnode_get_color(rightleft) == RB_RED)) {
+                    /* In the following diagrams, ||, //, and \\
+                     * indicate the path to the removed node.
+                     *
+                     *      ||
+                     *    pathp(r)
+                     *  //        \
+                     * (b)        (b)
+                     *           /
+                     *          (r)
+                     *
+                     */
+                    rbnode_set_black(pathp->node);
+                    rbnode_rotate_right(map_node_t, link, right, tnode);
+                    rbnode_set_right(pathp->node, tnode);
+                    rbnode_rotate_left(map_node_t, link, pathp->node, tnode);
+                } else {
+                    /*      ||
+                     *    pathp(r)
+                     *  //        \
+                     * (b)        (b)
+                     *           /
+                     *          (b)
+                     *
+                     */
+                    rbnode_rotate_left(map_node_t, link, pathp->node, tnode);
+                }
+                /* Balance restored, but rotation modified subtree
+                 * root.
+                 */
+                assert((uintptr_t) pathp > (uintptr_t) path);
+                if (pathp[-1].cmp < 0) {
+                    rbnode_set_left(pathp[-1].node, tnode);
+                } else {
+                    rbnode_set_right(pathp[-1].node, tnode);
+                }
+                return;
+            } else {
+                map_node_t *right = rbnode_get_right(pathp->node);
+                map_node_t *rightleft = rbnode_get_left(right);
+                if (rightleft && (rbnode_get_color(rightleft) == RB_RED)) {
+                    /*      ||
+                     *    pathp(b)
+                     *  //        \
+                     * (b)        (b)
+                     *           /
+                     *          (r)
+                     */
+                    map_node_t *tnode;
+                    rbnode_set_black(rightleft);
+                    rbnode_rotate_right(map_node_t, link, right, tnode);
+                    rbnode_set_right(pathp->node, tnode);
+                    rbnode_rotate_left(map_node_t, link, pathp->node, tnode);
+                    /* Balance restored, but rotation modified
+                     * subtree root, which may actually be the tree
+                     * root.
+                     */
+                    if (pathp == path) {
+                        /* Set root. */
+                        rbtree->root = tnode;
+                    } else {
+                        if (pathp[-1].cmp < 0) {
+                            rbnode_set_left(pathp[-1].node, tnode);
+                        } else {
+                            rbnode_set_right(pathp[-1].node, tnode);
+                        }
+                    }
+                    return;
+                } else {
+                    /*      ||
+                     *    pathp(b)
+                     *  //        \
+                     * (b)        (b)
+                     *           /
+                     *          (b)
+                     */
+                    map_node_t *tnode;
+                    rbnode_set_red(pathp->node);
+                    rbnode_rotate_left(map_node_t, link, pathp->node, tnode);
+                    pathp->node = tnode;
+                }
+            }
+        } else {
+            map_node_t *left;
+            rbnode_set_right(pathp->node, pathp[1].node);
+            left = rbnode_get_left(pathp->node);
+            if (rbnode_get_color(left) == RB_RED) {
+                map_node_t *tnode;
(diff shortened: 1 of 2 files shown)


jserv (changes requested):
Squash git commits and write down precise commit messages.

jserv:
> Squash git commits and write down precise commit messages.

Meanwhile, resolve the concerns raised by CodeQL.

jserv:
TODO:
* Resolve the concerns raised by CodeQL.
* Provide the comprehensive benchmark to illustrate why we shall migrate to the proposed rbtree implementation
* Implement the dedicated test suite for map.[ch].

ypaskell:
Thanks, I will work it out.

src/map.c
```diff
@@ -1,754 +1,592 @@
-/*
```
jserv: Keep this copyright notice. And mention the source was heavily modified from rb.h of jemalloc.

src/map.h
```diff
@@ -1,72 +1,87 @@
-/*
```
jserv: Keep this copyright notice.

src/map.c
```diff
@@ -1,754 +1,592 @@
...
-        memset(node->data, 0, vsize);
-    else
-        memcpy(node->data, value, vsize);
-
-    return node;
+    return (bool) (((size_t) node->link.right_red) & 1);
 }
 
-static void map_delete_node(map_t obj UNUSED, map_node_t *node)
+static inline void rbtn_color_set(map_node_t *node, bool bRed)
```
jserv: Use `enum` rather than `bool`.

src/map.c
```diff
@@ -1,754 +1,592 @@
...
-        /* Call this on the grandparent */
-        map_fix_colors(obj, grandparent);
+    assert(nodep->node == node);
+    pathp--;
+    if (pathp->node != node) {
+        /* Swap node with its successor. */
+        bool tred = rbtn_red_get(pathp->node);
+        rbtn_color_set(pathp->node, rbtn_red_get(node));
+        rbtn_left_set(pathp->node, rbtn_left_get(node));
+        /* If node's successor is its right child, the following code */
```
jserv: Use consistent comment style. That is,
```c
        /* If node's successor is its right child, the following code
         * will do the wrong thing for the right child pointer. 
         * However, it doesn't matter, because the pointer will be
         * properly set when the successor is pruned. 
         */
```

src/map.h
```diff
@@ -1,72 +1,87 @@
...
- * rv32emu is freely redistributable under the MIT License. See the file
- * "LICENSE" for information on usage and redistribution of this file.
+#pragma once
+
+#include <assert.h>
```
jserv: Don't include `<assert.h>` in public headers.

src/map.h
```diff
@@ -1,72 +1,87 @@
...
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
```
jserv: Reduce header inclusions. Drop `<stdio.h>`.

src/map.c
```diff
@@ -1,754 +1,604 @@
...
+        int cmp = pathp->cmp = (rbtree->cmp)(node, pathp->node);
+        if (cmp < 0) {
+            pathp[1].node = rbtn_left_get(pathp->node);
+        } else {
+            pathp[1].node = rbtn_right_get(pathp->node);
+            if (cmp == 0) {
+                /* Find node's successor, in preparation for swap. */
+                pathp->cmp = 1;
+                nodep = pathp;
+                for (pathp++; pathp->node; pathp++) {
```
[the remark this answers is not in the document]
TheCloudlet: Hi @jserv,
In my opinion, this is fine.
Can I disable the warning?
What do you think?

src/map.h
```diff
@@ -1,72 +1,87 @@
...
+    return (*a < *b) ? _CMP_LESS : (*a > *b) ? _CMP_GREATER : _CMP_EQUAL;
+}
 
-/*
- * Store access to the head node, as well as the first and last nodes.
- * Keep track of all aspects of the tree. All map functions require a pointer
- * to this struct.
- */
-typedef struct map_internal *map_t;
+static inline void cb(map_node_t *node, void *UNUSED)
```
jserv: Is it necessary to declare `cb` here?

src/map.c
```diff
@@ -1,754 +1,592 @@
...
+        rbtn_left_set((x_node), rbtn_right_get((r_node))); \
+        rbtn_right_set((r_node), (x_node));                \
+    } while (0)
+
+typedef struct {
+    map_node_t *node;
+    int cmp;
+} internal_map_path_entry_t;
+
+static void internal_map_new(map_internal_t *rbtree)
```
jserv: Why does this function exist? It looks trivial.

src/map.c
```diff
@@ -1,754 +1,592 @@
...
-    grandparent = rb_parent(parent);
-    uncle =
-        (grandparent->left == parent) ? grandparent->right : grandparent->left;
-
-    /* Apply right-right case */
-    map_r_r(obj, node, parent, grandparent, uncle);
+    internal_map_path_entry_t path[RB_MAX_DEPTH];
+    internal_map_path_entry_t *pathp;
+    rbt_node_init(node);
+    /* Wind. */
```
jserv, line 132: Improve the comment.

src/map.h
```diff
@@ -1,72 +1,87 @@
...
+#pragma once
+
+#include <assert.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+/* C macro implementation of left-leaning 2-3 red-black trees.  Parent
```
jserv: It is no longer a matter of macros. You should refine the comments and mention the origin (jemalloc).

src/map.c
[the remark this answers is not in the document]
jserv: Honor CodeQL if possible.

src/map.c
```diff
@@ -1,754 +1,592 @@
...
+        map_node_t *left = rbtn_left_get(node);
+        if (left) {
+            /* node has no successor, but it has a left child.        */
+            /* Splice node out, without losing the left child.        */
+            assert(!rbtn_red_get(node));
+            assert(rbtn_red_get(left));
+            rbtn_black_set(left);
+            if (pathp == path) {
+                rbtree->root = left;
+                /* Nothing to summarize -- the subtree rooted at the  */
```
jserv: Shall we care about the summary originally required by jemalloc's rb.h? Here, we don't really need to summarize anything. Consolidate this implementation with fewer code.

src/map.c
```diff
@@ -1,754 +1,592 @@
...
+    rbtree->root = path->node;
+    assert(!rbtn_red_get(rbtree->root));
 }
 
-/*
- * Recursive wrapper for deleting nodes in a graph at an accelerated pace.
- * Skips rotations. Just aggressively goes through all nodes and deletes.
- */
-static void map_clear_nested(map_t obj, map_node_t *node)
+static void internal_map_destroy_recurse(map_internal_t *rbtree,
```
jserv: The callback `cb` can be removed since we always release via `free`.

src/map.c
```diff
@@ -1,754 +1,612 @@
...
+        if (left) {
+            /* node has no successor, but it has a left child.
+             * Splice node out, without losing the left child.
+             */
+            assert(!rbtn_red_get(node));
+            assert(rbtn_red_get(left));
+            rbtn_black_set(left);
+            if (pathp == path) {
+                rbtree->root = left;
+                /* Nothing to summarize -- the subtree rooted at the
```
jserv, line 261: See if "Nothing to summarize" makes sense or not. We shall ignore the handling of summary.

src/map.c
```diff
@@ -1,754 +1,612 @@
...
- * "LICENSE" for information on usage and redistribution of this file.
- */
-
-#include <assert.h>
-#include <stdint.h>
-#include <stdlib.h>
-#include <string.h>
-
 #include "map.h"
+#include <memory.h>
```
jserv, line 21: Include `<assert.h>` here.

jserv: The order of header inclusion is
* System-wide headers appear first;
* Implementation-specific ones come later.

src/map.c
```diff
@@ -1,754 +1,628 @@
 /*
  * rv32emu is freely redistributable under the MIT License. See the file
  * "LICENSE" for information on usage and redistribution of this file.
+ * This map implementation is heavily modified from rb.h of jemalloc.
```
jserv: Drop this line since we would mention in next comment.

src/map.c
```diff
@@ -1,754 +1,628 @@
...
-};
-
-typedef enum { RB_RED = 0, RB_BLACK } map_color_t;
-
 /*
- * Get parent of node
- * @node: pointer to the rb node
- *
- * Return: rb parent node of @node
+ * This map implementation is heavily modified based on the rb.h header file of
```
jserv: Proofread the comment as following:
> This map implementation has undergone extensive modifications, heavily relying on the rb.h header file from jemalloc. See https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/rb.h .
The original rb.h file served as the foundation and source of inspiration for adapting and tailoring it specifically for this map implementation. Therefore, credit and sincere thanks are extended to jemalloc for their invaluable work.

src/map.h
```diff
@@ -4,69 +4,82 @@
  */
 
 /*
- * C Implementation for C++ std::map using red-black tree.
```
jserv: Keep the original comments since it is actually written by us.

src/map.h
```diff
@@ -4,69 +4,82 @@
...
- *
- * Any data type can be stored in a map, just like std::map.
- * A map instance requires the specification of two file types:
- *   1. the key;
- *   2. what data type the tree node will store;
- *
- * It will also require a comparison function to sort the tree.
+ * This map implementation is heavily modified based on the rb.h header file of
```
jserv: jemalloc should be only mentioned in file `map.c`.

src/map.h
```diff
@@ -4,69 +4,82 @@
...
  */
 
 #pragma once
 
 #include <stdbool.h>
 #include <stddef.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+#define RB_MAX_DEPTH (sizeof(void *) << 4)
```
jserv: This definition should appear in file `map.c`.

src/map.h
```diff
@@ -4,69 +4,82 @@
...
+
+/* Node structure */
+#define rb_node(x_type)           \
```
jserv: Drop the macro `rb_node` since you can expand it later.

src/map.h
```diff
@@ -4,69 +4,82 @@
...
+struct map_node {
+    void *key;
+    void *val;
+    rb_node(map_node_t) link;
+};
+
+typedef struct rb_tree {
+    map_node_t *root;
+    int (*cmp)(const map_node_t *, const map_node_t *);
+    size_t key_size;
```
jserv, line 36: I am not sure if `key_size` and `val_size` make sense because we do know the actual size of keys and values.

src/map.h
```diff
@@ -4,69 +4,82 @@
...
+    size_t count;
+} map_iter_t;
+
+#define map_iter_value(it, type) (*(type *) (it)->node->val)
 
 enum { _CMP_LESS = -1, _CMP_EQUAL = 0, _CMP_GREATER = 1 };
 
 /* Integer comparison */
-static inline int map_cmp_int(const void *arg0, const void *arg1)
+static inline int map_cmp_int(const map_node_t *arg0, const map_node_t *arg1)
```
jserv: Can you check if the use of `map_node_t` violates the casting rules? That is, we shall rely on the casting from `void *`.

src/map.c
```diff
@@ -3,752 +3,630 @@
  * "LICENSE" for information on usage and redistribution of this file.
  */
 
+/*
+ * This map implementation has undergone extensive modifications, heavily
+ * relying on the rb.h header file from jemalloc. See
```
jserv: Split the words starting with "See " into a new line.

jserv: That is,
```c
 * See https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/rb.h .
```

jserv:
> Issue #29

Show the benchmark numbers and append your analysis in the top message.

src/map.c
```diff
@@ -3,752 +3,630 @@
...
+#include <memory.h>
 
 #include "map.h"
 
-struct map_internal {
-    struct map_node *head;
-
-    /* Properties */
-    size_t key_size, element_size, size;
+#define RB_MAX_DEPTH (sizeof(void *) << 4)
```
jserv, line 33: Add some comments here.

src/map.c
```diff
@@ -3,752 +3,630 @@
...
-
-/*
- * Get parent of node
- * @node: pointer to the rb node
- *
- * Return: rb parent node of @node
- */
-static inline map_node_t *rb_parent(const map_node_t *node)
+/* Left accessors */
+static inline map_node_t *rbtn_left_get(map_node_t *node)
```
jserv: I would like to proposed the following simplifications:
* `rbtn` -> `rbnode` (rbtree node)
* `rbtree` -> `rb` (not necessary to use fully qualified name)

src/map.c
```diff
@@ -3,752 +3,630 @@
...
- * Set parent of node
- * @node: pointer to the rb node
- * @parent: pointer to the new parent node
- */
-static inline void rb_set_parent(map_node_t *node, map_node_t *parent)
+/* Right accessors */
+static inline map_node_t *rbtn_right_get(map_node_t *node)
 {
-    node->parent_color = (unsigned long) parent | (node->parent_color & 1LU);
+    return (map_node_t *) (((size_t) node->link.right_red) & ~3);
```
jserv: Replace `size_t` with `uintptr_t`.

src/map.c
```diff
@@ -3,752 +3,630 @@
...
- * Set color of node
- * @node: pointer to the rb node
- * @color: new color of the node
- */
-static inline void rb_set_color(map_node_t *node, map_color_t color)
+static inline void rbtn_right_set(map_node_t *node, map_node_t *right)
 {
-    node->parent_color = (node->parent_color & ~1LU) | color;
+    node->link.right_red =
+        (map_node_t *) (((size_t) right) |
```
jserv: Ditto. Always use `uintptr_t` rather than `size_t`.

src/map.c
```diff
@@ -3,752 +3,630 @@
...
- * Set parent and color of node
- * @node: pointer to the rb node
- * @parent: pointer to the new parent node
- * @color: new color of the node
- */
-static inline void rb_set_parent_color(map_node_t *node,
-                                       map_node_t *parent,
-                                       map_color_t color)
+/* Color accessors */
+static inline bool rbtn_red_get(map_node_t *node)
```
jserv: Can you make the naming scheme more consistent? That is, something like `rbnode_get_red`.

src/map.c
jserv: Ditto. How about `rbnode_set_color`?

src/map.c
```diff
@@ -3,752 +3,630 @@
...
-        memset(node->data, 0, vsize);
-    else
-        memcpy(node->data, value, vsize);
-
-    return node;
+    node->link.right_red = (map_node_t *) (((size_t) node->link.right_red) | 1);
 }
 
-static void map_delete_node(map_t obj UNUSED, map_node_t *node)
+static inline void rbtn_black_set(map_node_t *node)
```
jserv: Shall it be `rbnode_set_black`?

jserv:
> TODO

Append "Close #29" at the end of the git commit message.

src/map.c
```diff
@@ -3,752 +3,630 @@
...
+    do {                                                   \
+        (r_node) = rbtn_left_get((x_node));                \
+        rbtn_left_set((x_node), rbtn_right_get((r_node))); \
+        rbtn_right_set((r_node), (x_node));                \
+    } while (0)
+
+typedef struct {
+    map_node_t *node;
+    int cmp;
+} internal_map_path_entry_t;
```
jserv: Think of possible micro-optimizations: encode the value of `cmd` into the following bit representations:
* less : $11_2$
* equal : $00_2$
* greater :  $01_2$

Then, we can even embed the bits of encoded `cmp` into `node` because of ABI alignment. Please evaluate the feasibility and see if we can benefit from such micro-optimization.

jserv: The prefix `internal_` is a bit annoying. Avoid it if possible.

src/map.h
```diff
@@ -3,70 +3,70 @@
...
+typedef map_internal_t *map_t;
+
+typedef struct {
+    map_node_t *prev, *node;
+    size_t count;
+} map_iter_t;
+
+#define map_iter_value(it, type) (*(type *) (it)->node->val)
 
 enum { _CMP_LESS = -1, _CMP_EQUAL = 0, _CMP_GREATER = 1 };
```
jserv, line 49: We shall make use of these enumerations: `_CMP_LESS`, `_CMP_EQUAL`, and `_CMP_GREATER`.

src/map.c
jserv: In the proposed naming scheme, the above function name would be adjusted as `rbnode_get_left` where `rbnode_` is the prefix followed by certain verbs such as `get` and `set`.

src/map.c
```diff
@@ -3,752 +3,630 @@
...
-    it->prev = tmp.node;
+    map_node_t tmp_node;
+    tmp_node.key = key;
+    it->node = internal_map_search(obj, &tmp_node);
 }
 
 bool map_empty(map_t obj)
 {
-    return (obj->size == 0);
+    return (NULL == obj->root);
```
jserv: Shorten as the following:
```
    return !obj->root;
```

Also, consider to mark it as "static inline" in public header.

src/map.c
```diff
@@ -3,752 +3,630 @@
...
 }
 
-/* Return true if at the the end of the map */
-bool map_at_end(map_t obj UNUSED, map_iter_t *it)
+/* Iteration */
+bool map_at_end(map_t UNUSED, map_iter_t *it)
 {
-    return (it->node == NULL);
+    return (NULL == it->node);
```
jserv: Ditto. Shorten the code and consider to mark it as "static inline" in public header.

src/map.c
```diff
@@ -3,752 +3,630 @@
...
 {
-    map_node_t *x, *y;
-    map_node_t *node = it->node, *target, *double_blk, *x_parent;
-
-    /* If it is the head, and the size is 1, just delete it. */
-    if (obj->size == 1 && node == obj->head) {
-        map_delete_node(obj, node);
-        obj->head = NULL;
-        obj->size--;
+    if (NULL == it->node)
```
jserv: Shorten it below:
```c
    if (!it->node)
        return;
```

src/map.c
TheCloudlet: If we can handle enum in compare properly,  we can use the technique from the following example.
Still working it out.
```
#include <stdio.h>

enum Comparison {
    _CMP_LESS = -1,
    _CMP_EQUAL = 0,
    _CMP_GREATER = 1
};

struct Data {
    unsigned int value : 2;
};

int main() {
    struct Data data;

    // Assigning a value to the bit field
    enum Comparison cmp = _CMP_LESS;
    data.value = cmp & 0x3;

    // Accessing and printing the value of the bit field
    printf("Value: %d\n", (int)data.value);

    return 0;
}

```

src/map.c
```diff
@@ -3,752 +3,619 @@
...
+        int cmp = pathp->cmp = (rbtree->cmp)(node, pathp->node);
+        if (cmp < 0) {
+            pathp[1].node = rbtnode_get_left(pathp->node);
+        } else {
+            pathp[1].node = rbtnode_get_right(pathp->node);
+            if (cmp == 0) {
+                /* Find node's successor, in preparation for swap. */
+                pathp->cmp = 1;
+                nodep = pathp;
+                for (pathp++; pathp->node; pathp++) {
```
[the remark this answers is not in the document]
TheCloudlet: Hi @jserv,

I came up with a solution, but it might cause some slowdown.
What do you think?
```
for (int i = 0; path[i].node; i++) {
        pathp = &path[i];
        int cmp = pathp->cmp = (rbtree->cmp)(node, pathp->node);
        if (cmp < 0) {
            path[i + 1].node = rbtnode_get_left(pathp->node);
        } else {
            path[i + 1].node = rbtnode_get_right(pathp->node);
            if (cmp == 0) {
                /* Find node's successor, in preparation for swap. */
                pathp->cmp = 1;
                nodep = pathp;
                for (int j = i + 1; path[j].node; j++) {
                    path[j].cmp = -1;
                    path[j + 1].node = rbtnode_get_left(path[j].node);
                }
                break;
            }
        }
    }
   
```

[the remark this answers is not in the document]
jserv: Use switch-case since we do know the value of comparisons.

src/map.c
jserv: Don't declare `enum Comparison` because `map.h` already provides.

src/map.c
jserv: Rewrite as following:
```c
    return !it->node;
```

src/map.c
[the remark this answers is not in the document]
TheCloudlet: Should I use switch case with goto to exit the for loop?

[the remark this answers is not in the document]
jserv: > Should I use switch case with goto to exit the for loop?

Yes, it should be intentionally applied.

src/map.h
```diff
@@ -18,52 +18,60 @@
...
+        map_node_t *left, *right_red;
+    } link;
+};
+
+typedef struct rb_tree {
+    map_node_t *root;
+    int (*cmp)(const map_node_t *, const map_node_t *);
+    size_t key_size;
+    size_t val_size;
+} map_internal_t;
```
jserv: I didn't get the intention why you introduced `map_internal_t`. Can you write more consistent and simpler code?

src/map.h
```diff
@@ -18,52 +18,60 @@
...
-    size_t count;
-} map_iter_t;
-
-/*
- * Store access to the head node, as well as the first and last nodes.
- * Keep track of all aspects of the tree. All map functions require a pointer
- * to this struct.
- */
-typedef struct map_internal *map_t;
+static inline int map_cmp_sizet(const void *arg0, const void *arg1)
```
jserv: Why did `map_cmp_sizet` exist?

src/map.c
```diff
@@ -3,752 +3,624 @@
...
-{
-    node->parent_color = (unsigned long) parent | (node->parent_color & 1LU);
-}
+#include "map.h"
 
 /*
- * Set color of node
- * @node: pointer to the rb node
- * @color: new color of the node
+ * Each node in the RB tree consumes at least 1 byte of space (for the linkage
```
jserv: s/RB tree/red-black tree/

src/map.c
```diff
@@ -3,752 +3,624 @@
...
- */
+/* Empty map */
 void map_clear(map_t obj)
 {
-    if (obj->head) /* Aggressively delete by recursion */
-        map_clear_nested(obj, obj->head);
-
-    obj->size = 0;
-    obj->head = NULL;
+    rb_destroy(obj);
```
jserv: Drop `rb_destroy` and move its function body here.

src/map.c
```diff
@@ -3,752 +3,624 @@
...
- *        / \              / \
- *       A   C     =>     B   D
- *            \          /
- *             D        A
- *
- * Returns the new node pointing in the spot of the original node.
- */
-static map_node_t *map_rotate_left(map_t obj, map_node_t *node)
+/* Color accessors */
+static inline bool rbtnode_get_red(map_node_t *node)
```
jserv: I would like to propose the following change:
1. Keep `typedef enum { RB_BLACK = 0, RB_RED } map_color_t;`
2. Reimplement `rbtnode_get_red` as following:
```c
static inline map_color_t rbnode_get_color(const map_node_t *node)
{
    return (map_color_t) (((uintptr_t) node->link.right_red) & 1);
}
```
3. Replace all occurrences  `rbtnode_get_red()` with `rbnode_get_color() == RB_RED`

Then, the readability should be improved.

src/map.h
```diff
@@ -18,52 +18,60 @@
...
+typedef struct {
+    map_node_t *prev, *node;
+    size_t count;
+} map_iter_t;
+
+#define map_iter_value(it, type) (*(type *) (it)->node->val)
 
 enum { _CMP_LESS = -1, _CMP_EQUAL = 0, _CMP_GREATER = 1 };
 
+enum RbtnodeColor { BLACK = 0, RED = 1 };
```
jserv: Don't do that. Follow the original naming scheme. i.e.,
```c
typedef enum { RB_BLACK = 0, RB_RED } map_color_t;
```

src/map.c
```diff
@@ -3,752 +3,623 @@
...
-
 /*
- * Get parent of node
- * @node: pointer to the rb node
- *
- * Return: rb parent node of @node
+ * This map implementation has undergone extensive modifications, heavily
+ * relying on the rb.h header file from jemalloc. See
+ * https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/rb.h
+ * . The original rb.h file served as the foundation and source of inspiration
```
jserv: The dot (`.`) should not appear right before `The`.

src/map.c
```diff
@@ -3,752 +3,623 @@
...
+#include "map.h"
 
 /*
- * Set color of node
- * @node: pointer to the rb node
- * @color: new color of the node
+ * Each node in the red-black tree consumes at least 1 byte of space (for the
+ * linkage if nothing else, so there are a maximum of sizeof(void *) << 3 rb
+ * tree nodes in any process (and thus, at most sizeof(void *) << 3 nodes in any
+ * rb tree). The choice of algorithm bounds the depth of a tree to twice the
```
jserv, line 29: Improve the English writing.

src/map.c
```diff
@@ -3,752 +3,623 @@
...
- * Set parent and color of node
- * @node: pointer to the rb node
- * @parent: pointer to the new parent node
- * @color: new color of the node
- */
-static inline void rb_set_parent_color(map_node_t *node,
-                                       map_node_t *parent,
-                                       map_color_t color)
+/* Left accessors */
+static inline map_node_t *rbtnode_get_left(map_node_t *node)
```
jserv: Enforce consistent naming scheme: prefix `rbnode_`. That is, `rbnode_get_left`.

src/map.c
```diff
@@ -3,752 +3,625 @@
...
-typedef enum { RB_RED = 0, RB_BLACK } map_color_t;
-
 /*
- * Get parent of node
- * @node: pointer to the rb node
- *
- * Return: rb parent node of @node
+ * This map implementation has undergone extensive modifications, heavily
+ * relying on the rb.h header file from jemalloc.
+ * See
```
jserv: It should be
```c
 * See https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/rb.h .
 ```

Comment (TheCloudlet) [reply src/map.c]: clang-format will make it weird every round.
How can I bypass that?

Comment (jserv) [reply src/map.c]: Try `/* clang-format off */` and `/* clang-format on */`.

File: src/map.h (hunk 38 above)
Comment (jserv) [reply src/map.h]: Why?

File: src/map.c [hunk 47]
```diff
@@ -3,752 +3,625 @@
...
-        it->node = NULL;
-        return;
-    }
-    it->node = *indirect;
-
-    /* Generate a "prev" as well */
-    map_iter_t tmp = *it;
-    map_prev(obj, &tmp);
-    it->prev = tmp.node;
+    map_node_t tmp_node;
```
jserv: Rewrite as following:
```c
    map_node_t tmp_node = {.key = key};
```

src/map.c
```diff
@@ -3,752 +3,629 @@
...
 {
-    /* Rotate to the left according to grandparent */
-    grandparent = map_rotate_left(obj, grandparent);
-
-    /* Swap grandparent and uncle's colors */
-    map_color_t c1 = rb_color(grandparent), c2 = rb_color(grandparent->left);
-
-    rb_set_color(grandparent, c2);
-    rb_set_color(grandparent->left, c1);
+    assert((((uintptr_t) node) & (0x1)) == 0);
```
jserv: Replace `0` with `RB_BLACK` for readability.

src/map.h
TheCloudlet: It should not be here. Removed.

jserv (changes requested):
Prepare test program(s) for map.[ch] which appear in directory `tests/map`.
Example: https://github.com/sysprog21/rv32emu/tree/master/tests/cache

jserv (changes requested):
Prepare test program(s) for map.[ch] which appear in directory `tests/map`.
Example: https://github.com/sysprog21/rv32emu/tree/master/tests/cache

src/map.c
```diff
@@ -3,752 +3,629 @@
...
+        (map_node_t *) (((uintptr_t) node->link.right_red) & ~3);
 }
 
-static void map_r_r(map_t obj,
-                    map_node_t *node UNUSED,
-                    map_node_t *parent UNUSED,
-                    map_node_t *grandparent,
-                    map_node_t *uncle UNUSED)
+/* Node initializer */
+static inline void rbt_node_init(map_node_t *node)
```
jserv: Enforce consistent prefix `rbnode_`. That is, `rbnode_init`.

src/map.c
```diff
@@ -3,752 +3,625 @@
...
-    node = parent->right;
-    grandparent = rb_parent(parent);
-    uncle =
-        (grandparent->left == parent) ? grandparent->right : grandparent->left;
-
-    /* Apply right-right case */
-    map_r_r(obj, node, parent, grandparent, uncle);
+    int cmp;
+    map_node_t *ret = rbtree->root;
+    while (ret && (cmp = (rbtree->cmp)(keynode->key, ret->key)) != 0) {
```
jserv, line 117: Replace `0` with `_CMP_EQUAL` for readability.

src/map.c
```diff
@@ -3,752 +3,625 @@
...
-        rb_set_color(node, RB_BLACK);
-        return;
+    rb_path_entry_t path[RB_MAX_DEPTH];
+    rb_path_entry_t *pathp;
+    rbnode_init(node);
+    /* Wind. */
+    path->node = rbtree->root;
+    for (pathp = path; pathp->node; pathp++) {
+        int cmp = pathp->cmp = (rbtree->cmp)(node->key, pathp->node->key);
+        if (cmp == 0) {
```
jserv, line 136: I would propose the following:
```c
    switch (cmp) {
    case _CMP_LESS:
        pathp[1].node = rbnode_get_left(pathp->node);
        break;
    case _CMP_GREATER:
        pathp[1].node = rbnode_get_right(pathp->node);
        break;
    default: /* _CMP_EQUAL */
        break;
    }
```

src/map.c
```diff
@@ -3,752 +3,625 @@
...
-        for (it->node = it->node->left; it->node->right;
-             it->node = it->node->right)
-            it->prev = it->node;
-        return;
-    }
-
-    /* If there is no left child, keep going up until there is a left child */
-    it->prev = it->node;
-    it->node = rb_parent(it->node);
+    map_t tree = (map_t) malloc(sizeof(map_head_t));
```
jserv, line 572: No cast is required. That is,
```c
map_t tree = malloc(sizeof(map_head_t));
```

jserv:
Close in favor of incoming revised pull request for reviewing.

PR #141, opened 2023-06-04
Add test case for map API

TheCloudlet:
In preparation for new map implementation

### DIFF
--- a/mk/tests.mk
+++ b/mk/tests.mk
@@ -1,15 +1,26 @@
 CACHE_TEST_DIR := tests/cache
 CACHE_BUILD_DIR := build/cache
-TARGET := test-cache
+CACHE_TARGET := test-cache
+
+MAP_TEST_DIR := tests/map
+MAP_BUILD_DIR:= build/map
+MAP_TARGET := test-map
 
 CACHE_OBJS := \
 	test-cache.o
+
+MAP_OBJS := \
+	test-map.o
 
 CACHE_OBJS := $(addprefix $(CACHE_BUILD_DIR)/, $(CACHE_OBJS))
 OBJS += $(CACHE_OBJS)
 deps += $(CACHE_OBJS:%.o=%.o.d)
 
-# Check adaptive replacement cache policy is enabled or not, default is LRU 
+MAP_OBJS := $(addprefix $(MAP_BUILD_DIR)/, $(MAP_OBJS))
+OBJS += $(MAP_OBJS)
+deps += $(MAP_OBJS:%.o=%.o.d)
+
+# Check adaptive replacement cache policy is enabled or not, default is LRU
 ifeq ($(ENABLE_ARC), 1)
 CACHE_CHECK_ELF_FILES := \
 	arc/cache-new \
@@ -18,18 +29,21 @@ CACHE_CHECK_ELF_FILES := \
 	arc/cache-lru-replace \
 	arc/cache-lfu-replace \
 	arc/cache-lru-ghost-replace \
-	arc/cache-lfu-ghost-replace 
+	arc/cache-lfu-ghost-replace
 else
 CACHE_CHECK_ELF_FILES := \
 	lfu/cache-new \
 	lfu/cache-put \
 	lfu/cache-get \
-	lfu/cache-lfu-replace 
+	lfu/cache-lfu-replace
 endif
 
 CACHE_OUT = $(addprefix $(CACHE_BUILD_DIR)/, $(CACHE_CHECK_ELF_FILES:%=%.out))
+MAP_OUT = $(addprefix $(MAP_BUILD_DIR)/$(MAP_TARGET), ".out")
 
-tests : $(CACHE_OUT)
+tests : run_cache_test run_map_test
+
+run_cache_test : $(CACHE_OUT)
 	$(Q)$(foreach e,$(CACHE_CHECK_ELF_FILES),\
 		$(PRINTF) "Running $(e) ... "; \
 		if cmp $(CACHE_TEST_DIR)/$(e).expect $(CACHE_BUILD_DIR)/$(e).out; then \
@@ -40,16 +54,34 @@ tests : $(CACHE_OUT)
 		fi; \
 	)
 
-$(CACHE_OUT): $(TARGET)
+run_map_test: $(MAP_TARGET)
+
+$(CACHE_OUT): $(CACHE_TARGET)
 	$(Q)$(foreach e,$(CACHE_CHECK_ELF_FILES),\
-		$(CACHE_BUILD_DIR)/$(TARGET) $(CACHE_TEST_DIR)/$(e).in > $(CACHE_BUILD_DIR)/$(e).out; \
+		$(CACHE_BUILD_DIR)/$(CACHE_TARGET) $(CACHE_TEST_DIR)/$(e).in > $(CACHE_BUILD_DIR)/$(e).out; \
 	)
 
-$(TARGET): $(CACHE_OBJS)
+$(CACHE_TARGET): $(CACHE_OBJS)
 	$(VECHO) "  CC\t$@\n"
-	$(Q)$(CC) $^ build/cache.o build/mpool.o -o $(CACHE_BUILD_DIR)/$(TARGET)
-	
-$(CACHE_BUILD_DIR)/%.o: $(CACHE_TEST_DIR)/%.c 
+	$(Q)$(CC) $^ build/cache.o build/mpool.o -o $(CACHE_BUILD_DIR)/$(CACHE_TARGET)
+
+$(CACHE_BUILD_DIR)/%.o: $(CACHE_TEST_DIR)/%.c
 	$(VECHO) "  CC\t$@\n"
 	$(Q)mkdir -p $(dir $@)/arc $(dir $@)/lfu
+	$(Q)$(CC) -o $@ $(CFLAGS) -I./src -c -MMD -MF $@.d $<
+
+$(MAP_TARGET): $(MAP_OBJS)
+	$(VECHO) "  CC\t$@\n"
+	$(Q)$(CC) $^ build/map.o -o $(MAP_BUILD_DIR)/$(MAP_TARGET)
+	$(Q)$(MAP_BUILD_DIR)/$(MAP_TARGET)
+	$(Q)$(PRINTF) "Running test-map ... "; \
+	if [ $$? -eq 0 ]; then \
+	$(call notice, [OK]); \
+	else \
+	$(PRINTF) "Failed.\n"; \
+	fi;
+
+$(MAP_BUILD_DIR)/%.o: $(MAP_TEST_DIR)/%.c
+	$(VECHO) "  CC\t$@\n"
+	$(Q)mkdir -p $(dir $@)
 	$(Q)$(CC) -o $@ $(CFLAGS) -I./src -c -MMD -MF $@.d $<
--- a/tests/map/test-map.c
+++ b/tests/map/test-map.c
@@ -1,0 +1,113 @@
+#include <assert.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <time.h>
+
+#include "map.h"
+
+static void swap(int *x, int *y)
+{
+    int tmp = *x;
+    *x = *y;
+    *y = tmp;
+}
+
+enum { N_NODES = 10000 };
+
+/* return 0 on success; non-zero values on failure */
+static int test_map_mixed_operations()
+{
+    int ret = 0;
+    map_t tree = map_init(int, int, map_cmp_uint);
+
+    int key[N_NODES], val[N_NODES];
+
+    /*
+     *  Generate data for insertion
+     */
+    for (int i = 0; i < N_NODES; i++) {
+        key[i] = i;
+        val[i] = i + 1;
+    }
+
+    /* TODO: This is not a reconmended way to randomize stuff, just a simple
+     * test. Using MT19937 might be better
+     */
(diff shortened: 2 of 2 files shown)


tests/map/test-map.c
```diff
@@ -0,0 +1,111 @@
...
+#include "map.h"
+
+void swap(int *x, int *y)
+{
+    int tmp = *x;
+    *x = *y;
+    *y = tmp;
+}
+
+enum { NNODES = 10000 };
```
jserv: Rename to `N_NODES` for consistency.

tests/map/test-map.c
```diff
@@ -0,0 +1,111 @@
...
+
+bool test_mix_insert()
+{
+    bool failed = false;
+
+    map_t tree = map_init(int, int, map_cmp_uint);
+
+    int key[NNODES];
```
jserv: Merge two similar declarations. i.e.,
```c
int key[N_NODES], val[N_NODES];
```

tests/map/test-map.c
```diff
@@ -0,0 +1,111 @@
...
+    /*
+     *  Generate data for insertion
+     */
+    for (int i = 0; i < NNODES; i++) {
+        key[i] = i;
+        val[i] = i + 1;
+    }
+
+    /* This is not a reconmended way to randomize stuff, just a simple test */
+    /* I know MT19937 might be better */
```
jserv: Change the comment to "FIXME" or "TODO".

tests/map/test-map.c
```diff
@@ -0,0 +1,111 @@
...
+
+    /* add first 1/2 items */
+    for (int i = 0; i < NNODES / 2; i++) {
+        map_iter_t my_it;
+        map_insert(tree, key + i, val + i);
+        map_find(tree, &my_it, key + i);
+        if (NULL == my_it.node) {
+            failed = true;
+        }
+        // assert((*(int*) (my_it.node->data)) == val[i]);
```
jserv: Don't leave the dead code. For testing purpose, more assertions are pretty well.

tests/map/test-map.c
```diff
@@ -0,0 +1,111 @@
...
+
+    return !failed;
+}
+
+int main(int argc, char *argv[])
+{
+    (void) argc;
+    (void) argv;
+    bool passed = test_mix_insert();
+    printf("%s", passed ? "PASS" : "FAILED");
```
jserv: Instead of printing something, return `0` on success; `1` (or other values) on failure.

tests/map/test-map.c
```diff
@@ -0,0 +1,112 @@
...
+
+enum { N_NODES = 10000 };
+
+bool test_mix_insert()
+{
+    bool failed = false;
+
+    map_t tree = map_init(int, int, map_cmp_uint);
+
+    int key[N_NODES];
```
jserv: Merge two similar declarations. i.e.,
```c
int key[N_NODES], val[N_NODES];
```

tests/map/test-map.c
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /* remove first 1/4 items */
+    for (int i = 0; i < N_NODES / 4; i++) {
+        map_iter_t my_it;
+        map_find(tree, &my_it, key + i);
+        if (!map_at_end(tree, &my_it)) {
+            map_erase(tree, &my_it);
+            map_find(tree, &my_it, key + i);
+            if (NULL != my_it.node) {
```
jserv: Simply rewrite as
```c
if (my_it.node)) {
..
}
```

tests/map/test-map.c
```diff
@@ -0,0 +1,111 @@
...
+                assert(1);
```
jserv: Why `assert(1)`?

jserv (changes requested):
Check the assertions.

tests/map/test-map.c
TheCloudlet: my_it.node should be NULL.
if it is not NULL, means the map_erase fails

TheCloudlet: Or maybe I should write assert(0 && "map_erase error")

tests/map/test-map.c
```diff
@@ -0,0 +1,111 @@
...
+        swap(&key[pos_a], &key[pos_b]);
+        swap(&val[pos_a], &val[pos_b]);
+    }
+
+    /* add first 1/2 items */
+    for (int i = 0; i < N_NODES / 2; i++) {
+        map_iter_t my_it;
+        map_insert(tree, key + i, val + i);
+        map_find(tree, &my_it, key + i);
+        if (NULL == my_it.node) {
```
jserv: Ditto. Simply rewrite as
```c
if (!my_it.node)
    failed = true;
```

tests/map/test-map.c
```diff
@@ -0,0 +1,111 @@
...
+            failed = true;
+        }
+        assert((*(int *) (my_it.node->data)) == val[i]);
+    }
+
+    /* remove first 1/4 items */
+    for (int i = 0; i < N_NODES / 4; i++) {
+        map_iter_t my_it;
+        map_find(tree, &my_it, key + i);
+        if (!map_at_end(tree, &my_it)) {
```
jserv: Rewrite as following:
```c
if (map_at_end(tree, &my_it))
    continue;

map_erase(tree, &my_it);
map_find(tree, &my_it, key + i);
...
```

Check `CONTRIBUTING.md` : Avoid unnecessary nesting levels

tests/map/test-map.c
```diff
@@ -0,0 +1,111 @@
...
+    /* remove 2nd quarter of items */
+    for (int i = N_NODES / 4 + 1; i < N_NODES / 2; i++) {
+        map_iter_t my_it;
+        map_find(tree, &my_it, key + i);
+        if (!map_at_end(tree, &my_it)) {
+            map_erase(tree, &my_it);
+            map_find(tree, &my_it, key + i);
+            if (NULL != my_it.node) {
+                assert(1);
+                failed = true;
```
jserv: If it failed, shall we continue the loop?

tests/map/test-map.c
```diff
@@ -0,0 +1,111 @@
+#include <assert.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <time.h>
+
+#include "map.h"
+
+void swap(int *x, int *y)
```
jserv: Declare this internal function as a `static` one to avoid potential symbol conflicts.

tests/map/test-map.c
jserv: Why not assert on `my_it.node` directly?

tests/map/test-map.c
```diff
@@ -0,0 +1,111 @@
...
+
+    return !failed;
+}
+
+int main(int argc, char *argv[])
+{
+    (void) argc;
+    (void) argv;
+    bool passed = test_mix_insert();
+    printf("%d", passed);
```
jserv: If not passed, return non-zero value.

tests/map/test-map.c
TheCloudlet: static inline ?

tests/map/test-map.c
jserv: You should not click "resolve" sooner than reviewers.

tests/map/test-map.c
TheCloudlet: What is the purpose of this?

Pass --> 1 
!Pass --> 999

tests/map/test-map.c
jserv: No, simply use `static` since we don't care about further optimizations for a simple program.

tests/map/test-map.c
jserv: No, `0` on success; `1` (or other non-zero values) on failure. It is a convention of UNIX.

tests/map/test-map.c
```diff
@@ -0,0 +1,101 @@
...
+    map_delete(tree);
+
+    return failed;
+}
+
+int main(int argc, char *argv[])
+{
+    (void) argc;
+    (void) argv;
+    bool failed = test_mix_insert();
```
jserv: Simply rewrite as following:
```c
return test_mix_insert();
```

And add comment prior to the implementation of `test_mix_insert`.
```c
/* return 0 on success; non-zero values on failure */
```

The function prototype of `test_mix_insert` should be
```c
static int test_mix_insert()
```

tests/map/test-map.c
```diff
@@ -0,0 +1,101 @@
...
+static void swap(int *x, int *y)
+{
+    int tmp = *x;
+    *x = *y;
+    *y = tmp;
+}
+
+enum { N_NODES = 10000 };
+
+bool test_mix_insert()
```
jserv: Rename to `test_mixed_operations`.

tests/map/test-map.c
```diff
@@ -0,0 +1,101 @@
...
+     */
+    for (int i = 0; i < N_NODES; i++) {
+        key[i] = i;
+        val[i] = i + 1;
+    }
+
+    /* TODO: This is not a reconmended way to randomize stuff, just a simple
+     * test. Using MT19937 might be better
+     */
+    srand((unsigned) time(NULL));
```
jserv: `srand` should appear in `main` rather than here.

tests/map/test-map.c
```diff
@@ -0,0 +1,101 @@
...
+    int tmp = *x;
+    *x = *y;
+    *y = tmp;
+}
+
+enum { N_NODES = 10000 };
+
+bool test_mix_insert()
+{
+    bool failed = false;
```
jserv: The variable `failed` should be updated accordingly.
When the program is compiled with `NDEBUG` set, macro `assert` would be no operation, and we do need the result of this function for validation.

jserv (changes requested):
Besides the correctness verification, add execution time measurements, so that we can 
distinguish from each implementation.

jserv:
You may check https://github.com/martinus/map_benchmark/tree/master/src/benchmarks for figuring out the available benchmarking on `std::map`.

TheCloudlet:
I wonder whether it is possible to checkin the master first.
Then, I'll implement time measurements for newer features later.

tests/map/test-map.c
```diff
@@ -0,0 +1,108 @@
...
+        swap(&val[pos_a], &val[pos_b]);
+    }
+
+    /* add first 1/2 items */
+    for (int i = 0; i < N_NODES / 2; i++) {
+        map_iter_t my_it;
+        map_insert(tree, key + i, val + i);
+        map_find(tree, &my_it, key + i);
+        if (!my_it.node)
+            failed = 1;
```
jserv: Shall you return back to main function rather than pass through?

jserv: Alternatively, using `goto` may be an ordinary approach.

tests/map/test-map.c
```diff
@@ -0,0 +1,108 @@
...
+    *x = *y;
+    *y = tmp;
+}
+
+enum { N_NODES = 10000 };
+
+/* return 0 on success; non-zero values on failure */
+static int test_map_mixed_operations()
+{
+    int failed = 0;
```
jserv: Rename to `ret` for general usage.

tests/map/test-map.c
```diff
@@ -0,0 +1,108 @@
+#include <assert.h>
+#include <stdbool.h>
```
jserv: This inclusion is not necessary.

tests/map/test-map.c
TheCloudlet: I think a have a better approach, can you check at the newest update :)

mk/tests.mk
```diff
@@ -40,16 +54,34 @@ tests : $(CACHE_OUT)
...
+$(MAP_TARGET): $(MAP_OBJS)
+	$(VECHO) "  CC\t$@\n"
+	$(Q)$(CC) $^ build/map.o -o $(MAP_BUILD_DIR)/$(MAP_TARGET)
+	$(Q)$(MAP_BUILD_DIR)/$(MAP_TARGET)
+	$(Q)$(PRINTF) "Running test-map ... "; \
+	if [ $$? -eq 0 ]; then \
+	$(call notice, [OK]); \
+	else \
+	$(PRINTF) "Failed.\n"; \
+	fi; \
```
jserv: Replace `fi; \` with `fi;` since the latter is final.

tests/map/test-map.c
```diff
@@ -0,0 +1,102 @@
...
+        swap(&val[pos_a], &val[pos_b]);
+    }
+
+    /* add first 1/2 items */
+    for (int i = 0; i < N_NODES / 2; i++) {
+        map_iter_t my_it;
+        map_insert(tree, key + i, val + i);
+        map_find(tree, &my_it, key + i);
+        if (!my_it.node)
+            return 1; /* test failed */
```
jserv: Don't return directly. Instead, you shall release the allocated resources. Using `goto` may be reasonably acceptable. See https://hackmd.io/@sysprog/c-control-flow

tests/map/test-map.c
```diff
@@ -0,0 +1,102 @@
...
+    /* remove first 1/4 items */
+    for (int i = 0; i < N_NODES / 4; i++) {
+        map_iter_t my_it;
+        map_find(tree, &my_it, key + i);
+        if (map_at_end(tree, &my_it))
+            continue;
+        map_erase(tree, &my_it);
+        map_find(tree, &my_it, key + i);
+        if (my_it.node)
+            return 1; /* test failed */
```
jserv: Ditto. Release the allocated resources on failure.

jserv:
Thank @ypaskell for contributing!

PR #143, opened 2023-06-05
tests: Add script to extract CoreMark results

qwe661234:
The script calculates standard deviation and filters out outliers.

### DIFF
--- a/tools/coremark.py
+++ b/tools/coremark.py
@@ -1,0 +1,39 @@
+#!/usr/bin/env python3
+import subprocess
+import re
+import numpy
+import os
+
+iter = 10
+res = []
+file_exist = os.path.exists('build/rv32emu')
+if not file_exist:
+    print("Please compile before running test")
+    exit(1)
+print("Start Test CoreMark benchmark")
+comp_proc = subprocess.check_output(
+    'build/rv32emu build/coremark.elf', shell=True).decode("utf-8")
+if not comp_proc or comp_proc.find("Error") != -1:
+    print("Test Error")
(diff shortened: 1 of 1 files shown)


tools/coremark.py
```diff
@@ -0,0 +1,24 @@
+import subprocess
+import re
+import numpy
+
+iter = 10
+res = []
+for i in range(iter):
+    print("Running CoreMark benchmark - Run #{}".format(i + 1))
+    comp_proc = subprocess.check_output(
+        'build/rv32emu build/coremark.elf', shell=True).decode("utf-8")
```
jserv: You shall check the existence of `build/rv32emu` and validate its results for the first time.

tools/coremark.py
```diff
@@ -0,0 +1,24 @@
+import subprocess
```
jserv, line 2: Add proper [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)).

tools/coremark.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python3
+import subprocess
+import re
+import numpy
+
+iter = 10
+res = []
+print("Start Test CoreMark benchmark")
+comp_proc = subprocess.check_output(
+    'build/rv32emu build/coremark.elf', shell=True).decode("utf-8")
```
jserv, line 15: You have to validate the existence of `build/rv32emu` before executing the command `build/rv32emu build/coremark.elf`, otherwise you might run into uncertain states.

PR #146, opened 2023-06-17
Refactor code cache

qwe661234:
According to the MIR author statement, MIR does not generate position-independent code. This means that the code can only be executed in the location where it was originally generated. Since it is not possible to move the generated code to a specific memory location, we need to refactor a code cache that solely stores the memory address of the generated code.

The following statistics demonstrate the performance improvement observed when importing JIT (Just-in-Time) compilation on various host machines.

* Core i7-8700

|Metric| rv32emu (interpreter) |rv32emu (JIT) |Speedup|
|--------| -------- | -------- | -------- | 
| CoreMark | 1155.174 (iter/s)| 1932.496 (iter/s) |+67.2％|
| dhrystone  | 1282 DMIPS |2495.11 DMIPS |+94.6%|

* eMAG 8180

|Metric| rv32emu (interpreter) |rv32emu (baseline JIT) |Speedup|
|--------| -------- | -------- | -------- | 
| CoreMark | 438.073 (iter/s)| 644.426 (iter/s) |+47.1％|
| dhrystone  | 354.60 DMIPS |867.00 DMIPS |+144%|

### DIFF
--- a/src/riscv_private.h
+++ b/src/riscv_private.h
@@ -12,7 +12,9 @@
 #endif
 #include "decode.h"
 #include "riscv.h"
-
+#if RV32_HAS(JIT)
+#include "cache.h"
+#endif
 /* CSRs */
 enum {
     /* floating point */
@@ -59,8 +61,12 @@ typedef struct block {
     uint32_t insn_capacity;    /**< maximum of instructions encompased */
     struct block *predict;     /**< block prediction */
     rv_insn_t *ir;             /**< IR as memory blocks */
+#if RV32_HAS(JIT)
+    bool hot; /**< Determine the block is hotspot or not */
+#endif
 } block_t;
 
+#if !RV32_HAS(JIT)
 typedef struct {
     uint32_t block_capacity; /**< max number of entries in the block map */
     uint32_t size;           /**< number of entries currently in the map */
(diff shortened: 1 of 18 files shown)


jserv (changes requested):
Read [code refactoring](https://en.wikipedia.org/wiki/Code_refactoring) carefully.
> In computer programming and software design, code refactoring is the process of restructuring existing computer code — changing the factoring — without changing its external behavior.

However, the proposed pull request already changes the external behavior. Thus, don't use the term "refactor." You shall refine the subject.

src/riscv_private.h
```diff
@@ -131,6 +131,7 @@ struct riscv_internal {
     block_map_t block_map; /**< basic block map */
 #else
     struct cache *cache;
+    struct cache *code_cache;
```
jserv, line 134: The variable name is misleading. Refine the two instances of `struct cache`.

jserv:
Known issue: JIT works on Linux/Arm64, but it fails on macOS/Arm64 (Apple M1).
```
$ make ENABLE_JIT=1 check
Running hello.elf ... [OK]
Running puzzle.elf ... Failed.
make: *** [check] Error 1
```

jserv:
Known issue: misalign test case fails.
```shell
$ make ENABLE_JIT=1 misalign 
Segmentation fault
Running aes.elf ... Failed.
```

jserv:
I plan to merge this pull request as earlier as possible, so that we can rebase the recent `master` branch.
@qwe661234, it is acceptable to support x86-64 only at the moment. Refine the proposed changes as I mentioned.

PR #148, opened 2023-06-27
Introduce baseline JIT compiler

qwe661234:
The baseline JIT compiler design involves tracing an EBB when its usage frequency exceeds a predetermined threshold. We then utilize a code generator to convert the instruction sequence into C code. The C code is subsequently compiled using mir, and the resulting target machine code is stored in the code cache for future utilization.

The primary objective of introducing the baseline JIT compiler is to enhance the execution speed of RISC-V instructions with minimal modification. This implementation requires three additional components: a code generator, mir compiler, and code cache, while maintaining the original design of the interpreter. Furthermore, this baseline JIT compiler serves as the foundational target for future improvements.

In addition, we have developed a Python script that effectively traces code templates and automatically generates JIT code templates. This approach eliminates the need for manually writing duplicated codey.

* Core i7-8700

| Metric    |     master      |  baseline JIT   |Speedup |
|-----------|-----------------|-----------------|--------|
| CoreMark  |1352.843 (iter/s)|1932.496 (iter/s)| +42.8% |
| Dhrystone |1282 DMIPS       |2495.11 DMIPS    | +94.6% |

* eMAG

| Metric    |     master      |  baseline JIT   |Speedup |
|-----------|-----------------|----------------|--------|
| CoreMark  | 438.073 (iter/s)| 644.426 (iter/s)| +47.1% |
| Dhrystone | 354.60 DMIPS    | 867.00 DMIPS    |  +144% |

Known issue: this implementation does not correctly work on Apple M1 MacOS.


jserv (changes requested):
You shall rework the commit "Decouple RV32 instructions as implementation templates" and then rebase. i.e., at least two commits.

jserv (changes requested):
Explicitly mention commit 54024f74d7f1b86bddae0b011bd0469e1ba378db while comparing with JIT compilation in git commit message.

src/emulate.c
```diff
@@ -310,971 +171,10 @@
...
-     * is misaligned or if the memory chunk does not exist.
-     */
-    RV_EXC_MISALIGN_HANDLER(3, load, false, 1);
-    rv->X[fuse[0].rd] = rv->io.mem_read_w(rv, addr);
-    for (int i = 1; i < ir->imm2; i++) {
-        addr = rv->X[fuse[i].rs1] + fuse[i].imm;
-        rv->X[fuse[i].rd] = rv->io.mem_read_w(rv, addr);
-    }
-})
+/* FIXME: Add JIT-based execution path */
```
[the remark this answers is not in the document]
jserv: Remove this comment.

src/rv32_template.c
```diff
@@ -0,0 +1,946 @@
...
+RVOP(fcvtwus, { rv->X[ir->rd] = (uint32_t) rv->F[ir->rs1]; })
+
+/* FMV.X.W */
+RVOP(fmvxw, { rv->X[ir->rd] = rv->F_int[ir->rs1]; })
+
+/* FEQ.S performs a quiet comparison: it only sets the invalid operation
+ * exception flag if either input is a signaling NaN.
+ */
+RVOP(feqs, {
+    rv->X[ir->rd] = (rv->F[ir->rs1] == rv->F[ir->rs2]) ? 1 : 0;
```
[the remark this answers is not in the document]
jserv: @qwe661234, let's create another pull request for fixing this.

src/cache.h
```diff
@@ -37,4 +38,8 @@ void *cache_put(struct cache *cache, uint32_t key, void *value);
  * @cache: a pointer points to target cache
  * @callback: a function for freeing cache entry completely
  */
-void cache_free(struct cache *cache, void (*callback)(void *));
\ No newline at end of file
+void cache_free(struct cache *cache, void (*callback)(void *));
+
+#if RV32_HAS(JIT)
+bool cache_hot(struct cache *cache, uint32_t key);
```
jserv, line 50: Add comments here.

src/compile.c
```diff
@@ -0,0 +1,411 @@
...
+    uint8_t debug_level;
+    uint8_t optimize_level;
+} riscv_jit_t;
+
+typedef struct {
+    char *name;
+    void *func;
+} func_obj_t;
+
+#define SET_SIZE_BITS 10
```
jserv, line 40: Add some comments here, illustrating set data structure.

src/compile.c
```diff
@@ -0,0 +1,411 @@
...
+        if (!strcmp(name, func_list[i].name))
+            return func_list[i].func;
+    }
+    return NULL;
+}
+
+static riscv_jit_t *jit = NULL;
+static jit_item_t *jit_ptr = NULL;
+
+/* TODO: fix error when running on ARM architecture */
```
jserv: You shall describe the problem statement in details.

src/compile.c
```diff
@@ -0,0 +1,430 @@
...
+#include "mir-gen.h"
+#include "mir.h"
+#include "riscv_private.h"
+#include "utils.h"
+
+typedef struct {
+    char *code;
+    size_t code_size;
+    size_t curr;
+} jit_item_t;
```
jserv: I am afraid that `jit_item_t` is a bit misleading. Did you mean that it is a matter of code cache?

qwe661234: This is a data structure used by c2mir to retrieve the code string.

qwe661234: Maybe we can rename it to `jit_gencode_t`?

jserv: How about `code_string_t`? By the way, you shall provide more relevant comments.

src/riscv_private.h
```diff
@@ -12,7 +12,9 @@
 #endif
 #include "decode.h"
 #include "riscv.h"
-
+#if RV32_HAS(JIT)
+#include "cache.h"
+#endif
```
jserv, line 17: Append a new line right after `#endif`.

src/riscv_private.h
```diff
@@ -119,8 +126,13 @@ struct riscv_internal {
     uint32_t csr_mip;
     uint32_t csr_mbadaddr;
 
-    bool compressed;       /**< current instruction is compressed or not */
+    bool compressed; /**< current instruction is compressed or not */
+#if !RV32_HAS(JIT)
     block_map_t block_map; /**< basic block map */
+#else
+    struct cache *cache;
```
jserv: The naming of this object `cache` is misleading. Can you clarify?

src/riscv_private.h
```diff
@@ -59,8 +61,12 @@ typedef struct block {
     uint32_t insn_capacity;    /**< maximum of instructions encompased */
     struct block *predict;     /**< block prediction */
     rv_insn_t *ir;             /**< IR as memory blocks */
+#if RV32_HAS(JIT)
+    bool hot; /**< Determine the block is hotspot or not */
+#endif
 } block_t;
 
+#if !RV32_HAS(JIT)
```
jserv: Drop this preprocessor directive since it is a well-defined declaration. Keeping it doesn't hurt.

jserv: Add comments to annotate the functionality of this struct.

src/riscv_private.h
```diff
@@ -59,8 +62,13 @@ typedef struct block {
     uint32_t insn_capacity;    /**< maximum of instructions encompased */
     struct block *predict;     /**< block prediction */
     rv_insn_t *ir;             /**< IR as memory blocks */
+#if RV32_HAS(JIT)
+    bool hot; /**< Determine the block is hotspot or not */
+#endif
 } block_t;
 
+/* map the translated basic block to its program counter  to accelerate
+ * searching. */
```
jserv: Mention that this struct is used only in interpreter at the moment.

src/compile.c
jserv: ?

src/compile.c
```diff
@@ -0,0 +1,435 @@
...
+    strcat(gencode, PROLOGUE);
+    set_reset(&set);
+    block_t *block = cache_get(rv->block_cache, rv->PC);
+    trace_ebb(rv, gencode, block->ir, &set);
+    strcat(gencode, EPILOGUE);
+}
+
+static int get_string_func(void *data)
+{
+    code_string_t *jit_code_string = data;
```
jserv: Replace `jit_code_string` with `codestr` for short.

src/compile.c
```diff
@@ -0,0 +1,435 @@
...
+#include "mir-gen.h"
+#include "mir.h"
+#include "riscv_private.h"
+#include "utils.h"
+
+/* Provide for c2mir to retrieve the generated code string. */
+typedef struct {
+    char *code;
+    size_t code_size;
+    size_t curr;
```
jserv: You shall explain the functionality of `curr`.

src/compile.c
```diff
@@ -0,0 +1,435 @@
...
+
+#define SET_SIZE_BITS 10
+#define SET_SIZE 1 << SET_SIZE_BITS
+#define HASH(val) \
+    (((val) * (GOLDEN_RATIO_32)) >> (32 - SET_SIZE_BITS)) & ((SET_SIZE) -1)
+#define sys_icache_invalidate(addr, size) \
+    __builtin___clear_cache((char *) (addr), (char *) (addr) + (size));
+
+/*
+ *  The set consists of SET_SIZE buckets, with each bucket containing 32 slots.
```
jserv: Why 32 slots?

src/compile.c
```diff
@@ -0,0 +1,435 @@
...
+}
+
+/**
+ * set_has - check whether the element exist in the set or not
+ * @set: a pointer points to target set
+ * @key: the key of the inserted entry
+ */
+static bool set_has(set_t *set, uint32_t key)
+{
+    uint32_t index = HASH(key);
```
jserv: Add `const` qualifier for `index`.

src/compile.c
```diff
@@ -0,0 +1,435 @@
...
+    uint8_t count = 0;
+    while (set->table[index][count]) {
```
jserv: Rewrite as following:
```c
    for (uint8_t count = 0; set->table[index][count]; count++) {
```

src/compile.c
```diff
@@ -0,0 +1,435 @@
...
+    if (!jit) {
+        jit = calloc(1, sizeof(riscv_jit_t));
+        jit->options = calloc(1, sizeof(struct c2mir_options));
+        jit->ctx = MIR_init();
+        jit->optimize_level = 1;
+    }
+
+    if (!jit_code_string) {
+        jit_code_string = malloc(sizeof(code_string_t));
+        jit_code_string->curr = 0;
```
jserv: Move the statement `jit_code_string->curr = 0;` out of if-else block.

src/compile.c
qwe661234: I just set a sufficiently large number to prevent the bucket from becoming full.

src/compile.c
qwe661234: I add comments of set structure below.

src/compile.c
jserv: Add more comments and mark it as configurable. We shall always distinguish between magic numbers and configurations.

src/emulate.c
```diff
@@ -1313,6 +211,7 @@ static inline bool insn_is_unconditional_branch(uint8_t opcode)
     return false;
 }
 
+#if !RV32_HAS(JIT)
```
jserv, line 215: Is it possible to unify hash function implementation? i.e., unify macro `HASH` and function `hash`.

qwe661234: It is possible, but it may impact the performance of interpreter.

jserv: Add FIXME/TODO comments within this pull request. Then, submit an issue on GitHub to address code consolidation.

PR #151, opened 2023-07-06
Consolidate memory R/W operations

qwe661234:
In the current memory design, we did not allocate any memory initially. As a result, we had to check whether memory was allocated before reading or writing data, leading to a deterioration in the performance of memory I/O instructions. To overcome this limitation, we have implemented a new memory design that directly maps a memory size of $2^{32}$ bytes. With this enhancement, every memory read or write operation can directly access memory without the need for any checks.

The following statistics demonstrate the performance improvements achieved by the proposed memory I/O operations.

|  Metric  | original |  proposed | Speedup |
|----------|----------------|----------------|-------|
| Dhrystone |     1151.44    |    1374.20     | +19.3%|

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -5,65 +5,59 @@
 
 #include <assert.h>
 #include <stdint.h>
+#include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#if !defined(_WIN32)
+#define USE_MMAP 1
+#include <sys/mman.h>
+#include <unistd.h>
+#endif
 
 #include "io.h"
-#include "mpool.h"
 
-static const uint32_t mask_lo = 0xffff;
-static const uint32_t mask_hi = ~(0xffff);
-
-static struct mpool *mp;
+static uint8_t *data_memory_base;
+/* set memory size to 2^32 bytes */
+#define MEM_SIZE 0x100000000ULL
 
 memory_t *memory_new()
 {
-    memory_t *m = calloc(1, sizeof(memory_t));
-    /* Initialize the mpool size to sizeof(chunk_t) << 2, and it will extend
-     * automatically if needed */
-    mp = mpool_create(sizeof(chunk_t) << 2, sizeof(chunk_t));
-    return m;
+    memory_t *mem = malloc(sizeof(memory_t));
+#if defined(USE_MMAP)
+    data_memory_base = mmap(NULL, MEM_SIZE, PROT_READ | PROT_WRITE,
+                            MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
+    if (data_memory_base == MAP_FAILED)
+        return NULL;
+#else
+    data_memory_base = malloc(MEM_SIZE);
+    if (data_memory_base)
+        return NULL;
+#endif
+    mem->mem_base = data_memory_base;
+    mem->mem_size = MEM_SIZE;
+    return mem;
 }
 
-void memory_delete(memory_t *m)
+void memory_delete(memory_t *mem)
 {
-    if (!m)
-        return;
-    mpool_destory(mp);
-    free(m);
+    munmap(mem->mem_base, MEM_SIZE);
 }
 
-void memory_read(memory_t *m, uint8_t *dst, uint32_t addr, uint32_t size)
+void memory_read(memory_t *mem, uint8_t *dst, uint32_t addr, uint32_t size)
 {
-    /* test if this read is entirely within one chunk */
-    if ((addr & mask_hi) == ((addr + size) & mask_hi)) {
-        chunk_t *c;
-        if ((c = m->chunks[addr >> 16])) {
-            /* get the subchunk pointer */
-            const uint32_t p = (addr & mask_lo);
-
-            /* copy over the data */
-            memcpy(dst, c->data + p, size);
-        } else {
-            memset(dst, 0, size);
-        }
-    } else {
-        /* naive copy */
-        for (uint32_t i = 0; i < size; ++i) {
-            uint32_t p = addr + i;
-            chunk_t *c = m->chunks[p >> 16];
-            dst[i] = c ? c->data[p & 0xffff] : 0;
-        }
-    }
+    memcpy(dst, mem->mem_base + addr, size);
 }
 
-uint32_t memory_read_str(memory_t *m, uint8_t *dst, uint32_t addr, uint32_t max)
+uint32_t memory_read_str(memory_t *mem,
+                         uint8_t *dst,
+                         uint32_t addr,
+                         uint32_t max)
 {
     uint32_t len = 0;
     const uint8_t *end = dst + max;
     for (;; ++len, ++dst) {
         uint8_t ch = 0;
-        memory_read(m, &ch, addr + len, 1);
+        memory_read(mem, &ch, addr + len, 1);
         if (dst < end)
             *dst = ch;
         if (!ch)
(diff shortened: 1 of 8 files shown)


src/elf.c
```diff
@@ -342,13 +342,12 @@ bool elf_load(elf_t *e, riscv_t *rv, memory_t *mem)
         /* memcpy required range */
         const int to_copy = min(phdr->p_memsz, phdr->p_filesz);
         if (to_copy)
-            memory_write(mem, phdr->p_vaddr, e->raw_data + phdr->p_offset,
-                         to_copy);
+            memory_write(phdr->p_vaddr, e->raw_data + phdr->p_offset, to_copy);
```
jserv: Don't do that. For public C functions like `memory_write`, we have to ensure the encapsulation (or information hiding) of internal states. This means that we may create several RISC-V virtual machine instances simultaneously, and there must be appropriate responses for the corresponding calls. Keep this design principle in mind.

jserv:
Utilize [uftrace](https://github.com/namhyung/uftrace) tool to analyze function tracing.

Apply the following changes:
```diff
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ BIN := $(OUT)/rv32emu
 CFLAGS = -std=gnu99 -O2 -Wall -Wextra
 CFLAGS += -Wno-unused-label
 CFLAGS += -include src/common.h
+CFLAGS += -g -pg
 
 # Set the default stack pointer
 CFLAGS += -D DEFAULT_STACK_ADDR=0xFFFFF000
@@ -80,7 +81,7 @@ endif
 
 # For tail-call elimination, we need a specific set of build flags applied.
 # FIXME: On macOS + Apple Silicon, -fno-stack-protector might have a negative impact.
-$(OUT)/emulate.o: CFLAGS += -fomit-frame-pointer -fno-stack-check -fno-stack-protector
+$(OUT)/emulate.o: CFLAGS += -fno-stack-check -fno-stack-protector
 
 # Clear the .DEFAULT_GOAL special variable, so that the following turns
 # to the first target after .DEFAULT_GOAL is not set.
```

And then `make clean all`.

Run
```shell
$ uftrace build/rv32emu build/hello.elf 
```

The reference output:
```
            [233860] |   memory_new() {
  13.068 us [233860] |     calloc();
            [233860] |     mpool_create() {
   0.503 us [233860] |       malloc();
   0.212 us [233860] |       getpagesize();
   3.541 us [233860] |       mmap();
  17.226 us [233860] |     } /* mpool_create */
  31.291 us [233860] |   } /* memory_new */
```

You shall show the corresponding difference based on function tracing.

src/elf.c
qwe661234: Does this mean we still need to pass the virtual machine instances for all read and write functions, including read and write instructions? Due to we may create several RISC-V virtual machine instances.

jserv: Standalone functions such as `memory_write` should come with a parameter as context. However, the function pointers inside a structure can omit the context.

qwe661234:
> Utilize [uftrace](https://github.com/namhyung/uftrace) tool to analyze function tracing.

Based on the function tracing results from the [uftrace](https://github.com/namhyung/uftrace) tool, in the original memory design, `do_lw` takes approximately 9 us to allocate data memory and read data. Even when it doesn't require allocating data memory, it still takes around 1.7 us to read data. However, with the proposed design, we only need about 1 us to read data without allocating any data memory. Hence, the same applies to `do_sw`.

## sw
### original memory design
```
            [3659971] |    do_sw() {
            [3659971] |        on_mem_write_w() {
   0.255 us [3659971] |            rv_userdata();
   0.220 us [3659971] |            memory_write_w();
   9.141 us [3659971] |        } /* on_mem_write_w */
   ...
            [3659971] |    do_sw() {
            [3659971] |        on_mem_write_w() {
   0.265 us [3659971] |            rv_userdata();
   0.225 us [3659971] |            memory_write_w();
   1.795 us [3659971] |        } /* on_mem_write_w */
```
### proposed memory design
```
            [3658503] |    do_sw() {
            [3658503] |        on_mem_write_w() {
   0.245 us [3658503] |            memory_write_w();
   1.055 us [3658503] |        } /* on_mem_write_w */
```
### lw
### original memory design
```
            [3659971] |    do_lw() {
            [3659971] |        on_mem_read_w() {
   0.265 us [3659971] |            rv_userdata();
   0.255 us [3659971] |            memory_read_w();
   9.626 us [3659971] |         } /* on_mem_read_w */
   ...
            [3659971] |    do_lw() {
            [3659971] |        on_mem_read_w() {
   0.260 us [3659971] |            rv_userdata();
   0.235 us [3659971] |            memory_read_w();
   1.485 us [3659971] |        } /* on_mem_read_w */
```
### proposed memory design
```
       
            [3658503] |    do_lw() {
            [3658503] |        on_mem_read_w() {
   0.265 us [3658503] |            memory_read_w();
   1.025 us [3658503] |        } /* on_mem_read_w */
```

jserv (changes requested):
Refine the git commit message by mentioning uftrace comparisons.

src/io.c
```diff
@@ -5,65 +5,59 @@
...
 memory_t *memory_new()
 {
-    memory_t *m = calloc(1, sizeof(memory_t));
-    /* Initialize the mpool size to sizeof(chunk_t) << 2, and it will extend
-     * automatically if needed */
-    mp = mpool_create(sizeof(chunk_t) << 2, sizeof(chunk_t));
-    return m;
+    memory_t *mem = malloc(sizeof(memory_t));
+#if defined(USE_MMAP)
+    data_memory_base = (uint8_t *) mmap(NULL, MEM_SIZE, PROT_READ | PROT_WRITE,
```
jserv: The casting is not necessary.

src/io.c
```diff
@@ -5,65 +5,59 @@
...
+                                        MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
+    if (data_memory_base == MAP_FAILED)
+        fprintf(stderr, "warning: %s:%d %s\n", __FILE__, __LINE__, "MMAP FAIL");
+#else
+    data_memory_base = malloc(MEM_SIZE);
+    if (data_memory_base)
+        fprintf(stderr, "warning: %s:%d %s\n", __FILE__, __LINE__, "MMAP FAIL");
```
jserv: The message was incorrect.

src/io.c
```diff
@@ -72,76 +66,40 @@ uint32_t memory_read_str(memory_t *m, uint8_t *dst, uint32_t addr, uint32_t max)
...
+    {                                                              \
+        *(type *) (data_memory_base + addr) = *(const type *) src; \
     }
 
 MEM_WRITE_IMPL(w, uint32_t);
 MEM_WRITE_IMPL(s, uint16_t);
 MEM_WRITE_IMPL(b, uint8_t);
 
-void memory_fill(memory_t *m, uint32_t addr, uint32_t size, uint8_t val)
+void memory_fill(memory_t *mem, uint32_t addr, uint32_t size, uint8_t val)
```
jserv, line 102: I am not sure if it still makes sense to keep `memory_fill` as a public function. Can you check?

jserv: Alternatively, you may define it as a static inline function.

qwe661234: `memory_fill` is a public function for `elf_load`.

src/io.c
```diff
@@ -72,76 +66,40 @@ uint32_t memory_read_str(memory_t *m, uint8_t *dst, uint32_t addr, uint32_t max)
...
-    const uint32_t addr_lo = addr & mask_lo;
-    assert((addr_lo & 1) == 0);
-
-    chunk_t *c = m->chunks[addr >> 16];
-    assert(c);
-    return *(const uint32_t *) (c->data + addr_lo);
-}
-
-uint32_t memory_read_w(memory_t *m, uint32_t addr)
+uint32_t memory_ifetch(uint32_t addr)
```
jserv, line 69: Let's inline short functions like `memory_ifetch` in header files.

src/io.c
```diff
@@ -72,76 +66,40 @@ uint32_t memory_read_str(memory_t *m, uint8_t *dst, uint32_t addr, uint32_t max)
...
-{
-    chunk_t *c = m->chunks[addr >> 16];
-    return c->data[addr & mask_lo];
-}
+MEM_READ_IMPL(w, uint32_t);
+MEM_READ_IMPL(s, uint16_t);
+MEM_READ_IMPL(b, uint8_t);
 
-void memory_write(memory_t *m, uint32_t addr, const uint8_t *src, uint32_t size)
+void memory_write(memory_t *mem,
```
jserv, line 84: Ditto. Define it as a static inline function.

src/io.c
```diff
@@ -72,76 +66,40 @@ uint32_t memory_read_str(memory_t *m, uint8_t *dst, uint32_t addr, uint32_t max)
...
+    return *(const uint32_t *) (data_memory_base + addr);
 }
 
-uint16_t memory_read_s(memory_t *m, uint32_t addr)
-{
-    const uint32_t addr_lo = addr & mask_lo;
-    chunk_t *c = m->chunks[addr >> 16];
-    return *(const uint16_t *) (c->data + addr_lo);
-}
+#define MEM_READ_IMPL(size, type)                   \
```
jserv, line 74: Since they are frequently invoked, it is worthy to aid compiler optimizations by explicit static inline function definitions.

src/io.h
```diff
@@ -5,14 +5,12 @@
 
 #pragma once
 
+#include <bits/stdint-uintn.h>
```
jserv: Don't include internal headers. i.e., header filenames starting with `bits/`.

jserv:
Tested under Apple M1:

[ original ]
Average DMIPS : 1246.44

[ proposed ]
Average DMIPS : 1638.12

That is, 31% speedup.

src/io.c
qwe661234: we cannot define it as inline function, because we assign it as function pointer to struct `riscv_io_t`.

src/io.c
qwe661234: Ditto.

src/io.c
jserv: Confirmed, the compiler cannot inline a function unless it can determine that only one possible version of the function will be called. When calling through a pointer, it is not trivially obvious that this is the case.

However, we may still look for the approaches to lower function invocation cost later.

jserv: Append TODO in git commit messages for the possibility to inline the essential memory operations.

jserv:
You shall refer the operating system by Apple to [macOS](https://en.wikipedia.org/wiki/MacOS) rather than MacOS. It is a matter of brand identity.

src/io.h
```diff
@@ -8,11 +8,8 @@
 #include <stdint.h>
 
 typedef struct {
```
jserv, line 14: Add some comments to describe the interaction between RISC-V core and the memory subsystem.

jserv:
Always use the pair "original" vs. "proposed" while you are about to propose non-trivial changes by comprehensive comparisons.

jserv (changes requested):
Refine the messages.

Both "memory I/O" and "memory read/write" are commonly used terms in the field of computer architecture and memory management. 

If you want to specifically focus on the operations of reading from and writing to memory, "memory read/write" is a more specific term. It highlights the fundamental actions of retrieving data from memory (read) and storing data in memory (write).

In this case, consider the context to use "memory read/write" (or "memory R/W" for short) which term aligns better.

jserv:
@qwe661234, The commit messages have been amended. For future pull requests, it is recommended to use more professional commit messages. General statements should avoid the use of 'we'.

PR #156, opened 2023-07-14
Fix incorrect EBB

qwe661234:
1. When block cache is full, we need to clear the EBB information if the block has been replaced.
2. BB should not be extended if the exception hanlder is invoked.
3. Based on running benchmark SciMark2, frequently compilation cause error in MIR, so the threashold should be adjusted to pass the benchmark.

### DIFF
--- a/src/cache.c
+++ b/src/cache.c
@@ -20,7 +20,7 @@
 /* THRESHOLD is set to identify hot spots. Once the frequency of use for a block
  * exceeds the THRESHOLD, the JIT compiler flow is triggered.
  */
-#define THRESHOLD 1000
+#define THRESHOLD 32768
 
 #if RV32_HAS(JIT)
 #define sys_icache_invalidate(addr, size) \
--- a/src/rv32_template.c
+++ b/src/rv32_template.c
@@ -56,24 +56,36 @@ RVOP(jalr, {
 })
 
 /* clang-format off */
-#define BRANCH_FUNC(type, cond)                                  \
-    const uint32_t pc = rv->PC;                                  \
-    if ((type) rv->X[ir->rs1] cond (type) rv->X[ir->rs2]) {      \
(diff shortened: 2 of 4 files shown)


src/cache.c
```diff
@@ -20,7 +20,7 @@
 /* THRESHOLD is set to identify hot spots. Once the frequency of use for a block
  * exceeds the THRESHOLD, the JIT compiler flow is triggered.
  */
-#define THRESHOLD 1000
+#define THRESHOLD 100000
```
jserv: Do you employ any scientific methods or criteria to determine or adjust the value of the threshold?

qwe661234: I just simply increase threshold to the value that can pass the Sci Mark2.

jserv: > I just simply increase threshold to the value that can pass the Sci Mark2.

Additionally, it is important to provide a clear explanation of the reasoning behind such adjustments and reference the intended target (SciMark2) accordingly.

Always refer to it as "[SciMark2](https://math.nist.gov/scimark2/)" as it ought to be.

src/rv32_template.c
```diff
@@ -56,25 +56,55 @@ RVOP(jalr, {
...
-    branch_taken = true;                                         \
-    rv->PC += ir->imm;                                           \
-    /* check instruction misaligned */                           \
-    RV_EXC_MISALIGN_HANDLER(pc, insn, false, 0);                 \
-    if (ir->branch_taken) {                                      \
-        last_pc = rv->PC;                                        \
-        return ir->branch_taken->impl(rv, ir->branch_taken);     \
-    }                                                            \
+#if RV32_HAS(JIT)
+#define BRANCH_FUNC(type, cond)                                   \
```
jserv: Can you shorten the definition of macro `BRANCH_FUNC`?

PR #157, opened 2023-07-16
jit: Evaluate constant values prior to code generation

qwe661234:
Modify the script to translate the known IR fields into constant values.


jserv:
@qwe661234, You shall always write down meaningful git commit messages.

Issue #158, opened 2023-07-17
jit: format warnings during code generation


jserv:
clang raises the following warnings:
```
src/jit_template.c:317:57: warning: flag ' ' results in undefined behavior with 'u' conversion specifier [-Wformat]
GEN("    rv->X[%u] = !udivisor ? udividend : udividend % udivisor;\n", ir->rd);
                                                       ~^~
```

For the modulo (`%`) operation, `%%` should be used in `GEN` rather than `%`.

jserv:
Fixed in #160

PR #160, opened 2023-07-17
Substitute modulo operation % with %%

qwe661234:
Solve: #158


jserv:
If you close a pull request or an issue, there must be a reason.

qwe661234:
> If you close a pull request or an issue, there must be a reason.

I close it for testing more test cases before merge.

tools/gen-jit-template.py
```diff
@@ -216,6 +216,7 @@ def remove_comment(str):
         output = output + "RVOP(" + op[i] + ", {\n"
         substr = impl[i].split("\n")
         for str in substr:
+            str = re.sub("%", "%%", str)
```
jserv: The substitution should only apply outside of program comments, where it can retain its meaningful context.

Issue #169, opened 2023-07-19
jit: Fail to successfully execute maj2random


jserv:
[maj2random](https://github.com/michaeljclark/maj2random) s a simplified floating-point hash function derived from SHA-2. The JIT compiler fails to successfully execute maj2random.
Reproduce:
```
$ build/rv32emu build/maj2random.elf
```

The output:
```
maj_random (NROUNDS=4)                 count     mean(x)     mean(y) variance(x) variance(y)  std-dev(x)  std-dev(y)
(             0, (0 - 1K)/8K )    999999,057     0.49849     0.48933     0.08351     0.07539     0.28897     0.27457
(             0, (0 - 8K)/8K )    999999,057     0.50179     0.49853     0.08464     0.08309     0.29094     0.28825
( (0 - 1K)/8K ),           0 )    999999,057     0.50079     0.50445     0.08196     0.08217     0.28629     0.28665
( (0 - 8K)/8K ),           0 )    999999,057     0.50482     0.50053     0.08331     0.08368     0.28863     0.28928
( (0 - 1K)/8K ), (0 - 1K)/8K )    999999,057     0.50820     0.47531     0.08091     0.08333     0.28445     0.28867
Segmentation fault (core dumped)
```

jserv:
Close via commit [cf3b433](https://github.com/sysprog21/rv32emu/commit/cf3b433813031bf1a6507373e9acf67706282e2f)

PR #172, opened 2023-07-21
jit: Fix float-point instruction

qwe661234:
1. Trace the file src/softfloat.h in gen-jit script to include an enumeration for floating point instructions.
2. To simplify the definition of JIT gencode, we adjust the data structure of riscv_internal by placing the fields needed by JIT at the top.

Solve: https://github.com/sysprog21/rv32emu/issues/169

### DIFF
--- a/src/compile.c
+++ b/src/compile.c
@@ -176,24 +176,24 @@ RVOP(bgeu, { BRNACH_FUNC(uint32_t, >=); })
 
 RVOP(lb, {
     GEN("  addr = rv->X[%u] + %u;\n", ir->rs1, ir->imm);
-    GEN("  rv->X[%u] = sign_extend_b(* (const uint8_t *) (m->mem_base + "
-        "addr));\n",
+    GEN("  rv->X[%u] = sign_extend_b(*((const uint8_t *) (m->mem_base + "
+        "addr)));\n",
         ir->rd);
 })
 
 RVOP(lh, {
     GEN("  addr = rv->X[%u] + %u;\n", ir->rs1, ir->imm);
-    GEN("  rv->X[%u] = sign_extend_h(* (const uint16_t *) (m->mem_base + "
-        "addr));\n",
+    GEN("  rv->X[%u] = sign_extend_h(*((const uint16_t *) (m->mem_base + "
+        "addr)));\n",
         ir->rd);
 })
 
-#define MEMORY_FUNC(type, IO)                                                \
-    GEN("  addr = rv->X[%u] + %u;\n", ir->rs1, ir->imm);                     \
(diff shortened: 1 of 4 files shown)


src/softfloat.h
```diff
@@ -11,19 +11,19 @@
 
 /* clang-format off */
 enum {
-    //                    ....xxxx....xxxx....xxxx....xxxx
+    /*                    ....xxxx....xxxx....xxxx....xxxx */
```
jserv: Don't do this. Minimize the necessary changes.

jserv: Python script should be aware of C++ style comments.

src/riscv_private.h
```diff
@@ -130,6 +116,21 @@ struct riscv_internal {
     uint32_t csr_mbadaddr;
 
     bool compressed; /**< current instruction is compressed or not */
+
+#if RV32_HAS(GDBSTUB)
```
jserv, line 120: Why did you make such change?

qwe661234: If we place the fields needed by JIT at the top of the data structure, we could reduce the definition in gencode. Otherwise, we need to define a data structure for GDB in gencode.

jserv (changes requested):
Improve the git commit messages. ChatGPT can help refine.

src/compile.c
```diff
@@ -208,6 +208,20 @@ RVOP(sh, {MEMORY_FUNC(uint16_t, 0)})
 
 RVOP(sw, {MEMORY_FUNC(uint32_t, 0)})
 
+#if RV32_HAS(EXT_F)
+RVOP(flw, {
+    GEN("  addr = rv->X[%u] + %u;\n", ir->rs1, ir->imm);
+    GEN("  rv->F_int[%u] = * (const uint32_t *) (m->mem_base + addr);\n",
```
jserv: Should be `*((const uint32_t *) (m->mem_base + addr));`.

src/compile.c
```diff
@@ -208,6 +208,20 @@ RVOP(sh, {MEMORY_FUNC(uint16_t, 0)})
...
+        ir->rd);
+})
+
+/* FSW */
+RVOP(fsw, {
+    GEN("  addr = rv->X[%u] + %u;\n", ir->rs1, ir->imm);
+    GEN("  *(uint32_t *) (m->mem_base + addr) = rv->F_int[%u];\n", ir->rs2);
```
jserv: Ditto. Use `*((uint32_t *) (m->mem_base + addr))`.

src/riscv_private.h
jserv: Explain in git commit messages.

PR #174, opened 2023-07-22
Fix incorrectly generated config file

2011eric:
According to the documentation,
the letter 'F' must occur before letter 'C' in ISA string.

### DIFF
--- a/tests/arch-test-target/setup.py
+++ b/tests/arch-test-target/setup.py
@@ -15,12 +15,12 @@ def setup_testlist(riscv_device):
     if 'M' in riscv_device:
         misa |= constants.misa_M
         ISA += 'M'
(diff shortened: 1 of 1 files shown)


2011eric:
The original `setup.py` lead to this message when using `make arch-test RISCV_DEVICE=IMFCZicsrZifencei`:
```
ERROR | Error in /home/eric2011/src/working/rv32emu/tests/arch-test-target/rv32emu/rv32emu_isa.yaml.
ERROR |  ISA:
ERROR |   - Alphabet 'C' should occur after 'F
```

Here is the documentaion about config file: https://riscv-config.readthedocs.io/en/stable/yaml-specs.html#isa

Risheng1128 (approved):

Issue #173, opened 2023-07-21
jit: Fail to execute Doom


jserv:
While running Doom with JIT-enabled rv32emu, the SDL window ever appeared, but it soon got into assertion failure.
```
player 1 of 1 (1 nodes)
S_Init: Setting up sound.
HU_Init: Setting up heads up display.
ST_Init: Init status bar.
Assertion failed: (prev), function rv_step, file src/emulate.c, line 500.
/bin/sh: line 1: 71215 Abort trap: 6           ../build/rv32emu doom.elf
```

Quake was similar for the failure.

jserv:
Close via commit [e9c955b](https://github.com/sysprog21/rv32emu/commit/e9c955b715a9e31d5ccd9fff0f38ab6050aa6e37).

PR #175, opened 2023-07-23
jit: Handle EBB information correctly

qwe661234:
1. When BB is replaced by cache, the EBB information recorded in other BB is not updated. This could result in outdated EBB information, and the replaced BB might have been created in a different memory location. To address this issue, we should verify whether the EBB information is outdated or not and update it accordingly each time we use EBB information.
2. Some non-brnach instruction can be the end of a BB when a BB is fill, so we modify insn_is_unconditional_branch into insn_is_conditional_branch that we can only extend BB when its end is conditional branch instruction.

Solve: #173

### DIFF
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -195,20 +195,18 @@ static inline bool insn_is_branch(uint8_t opcode)
     return false;
 }
 
-static inline bool insn_is_unconditional_branch(uint8_t opcode)
+static inline bool insn_is_conditional_branch(uint8_t opcode)
 {
     switch (opcode) {
-    case rv_insn_ecall:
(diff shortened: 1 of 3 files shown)


src/emulate.c
```diff
@@ -195,20 +195,18 @@ static inline bool insn_is_branch(uint8_t opcode)
     return false;
 }
 
-static inline bool insn_is_unconditional_branch(uint8_t opcode)
+static inline bool insn_is_conditional_branch(uint8_t opcode)
```
jserv, line 198: You shall explain the motivation and considerations for such modifications.

PR #176, opened 2023-07-25
WIP: Prefetch next operation inside execution loop

jserv:
Data operations loops and function call loops have a difference: data accesses can be predicted due to modern microprocessors implementing a prefetch mechanism. This patch adds an explicit prefetch to the next operation before executing the current one. By hinting to the processor about an upcoming jump, it can pre-load the L1 code cache and potentially save clock cycles when the next function is called.

Both Dhrystone and Richards benchmarks benefit from the commit, showing a speedup of up to 13% on Ryzen Threadripper 2990WX.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -51,6 +51,13 @@
 #define MUST_TAIL
 #endif
 
+/* built-in prefetch intrinsic */
+#if defined(__GNUC__) || defined(__clang__)
+#define PREFETCH(x) __builtin_prefetch(x, 0)
+#else
+#define PREFETCH(x)
+#endif
+
 /* Pattern Matching for C macros.
  * https://github.com/pfultz2/Cloak/wiki/C-Preprocessor-tricks,-tips,-and-idioms
  */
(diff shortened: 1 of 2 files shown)


jserv:
@qwe661234, Please help with the following tasks:
1. Apply this patch and check on x86-64 and Arm64 machines to verify the validity of the speedup;
3. Perform additional benchmarking using existing suites.

qwe661234:
After running three different benchmarks (Drystone, CoreMark, Richards) on x86-64 and arm64 architectures, the following statistics demonstrate the performance decline resulting from this commit.
* Intel Core i7-11700

|   Metric |Speedup|
| -------- |  -------- | 
| Dhrystone  |  -8% |
| CoreMark   | -5%|
|Richards|-6% |

* eMAG (Arm64)

|   Metric   |Speedup|
| -------- |  -------- | 
| Dhrystone  | -5% |
| CoreMark   | -6%|
|Richards|-5% |

jserv:
This change might conflict with modern compiler optimizations, resulting in unpredictable behavior.

PR #181, opened 2023-08-01
Fix return NULL when malloc failed

ChinYikMing:


qwe661234 (approved):

jserv:

Issue #161, opened 2023-07-17
Non-portable memory allocation


jserv:
There is a non-portable issue in `src/io.c`:
```c
/* set memory size to 2^32 bytes */
#define MEM_SIZE 0x100000000ULL
```

If we are building with [Emscripten](https://emscripten.org/) (see also #75), 32-bit target is [set by default](https://github.com/emscripten-core/emscripten/issues/19833), and the clang would complain as following:
```
src/io.c:27:35: warning: implicit conversion from 'unsigned long long' to 'size_t' (aka 'unsigned long') changes value from 4294967296 to 0 [-Wconstant-conversion]
   27 |     data_memory_base = mmap(NULL, MEM_SIZE, PROT_READ | PROT_WRITE,
      |                        ~~~~       ^~~~~~~~
src/io.c:21:18: note: expanded from macro 'MEM_SIZE'
   21 | #define MEM_SIZE 0x100000000ULL
      |                  ^~~~~~~~~~~~~~
src/io.c:43:27: warning: implicit conversion from 'unsigned long long' to 'size_t' (aka 'unsigned long') changes value from 4294967296 to 0 [-Wconstant-conversion]
   43 |     munmap(mem->mem_base, MEM_SIZE);
      |     ~~~~~~                ^~~~~~~~
src/io.c:21:18: note: expanded from macro 'MEM_SIZE'
   21 | #define MEM_SIZE 0x100000000ULL
      |                  ^~~~~~~~~~~~~~
```

rv32emu should be portable for both 32-bit and 64-bit environments. Therefore, we need to address the improper memory allocation issue that has been raised on 32-bit targets.

jserv:
There is another non-portable issue in `src/io.c`:
```c
/* set memory size to 2^32 bytes */
#define MEM_SIZE 0x100000000ULL
```

If we are building with [Emscripten](https://emscripten.org/), 32-bit target is [set by default](https://github.com/emscripten-core/emscripten/issues/19833), and the clang would complain as following:
```
src/io.c:27:35: warning: implicit conversion from 'unsigned long long' to 'size_t' (aka 'unsigned long') changes value from 4294967296 to 0 [-Wconstant-conversion]
   27 |     data_memory_base = mmap(NULL, MEM_SIZE, PROT_READ | PROT_WRITE,
      |                        ~~~~       ^~~~~~~~
src/io.c:21:18: note: expanded from macro 'MEM_SIZE'
[...]

PR #182, opened 2023-08-01
Call appropriate release function

ChinYikMing:
mmap should be paired with munmap and malloc should be paired with free


jserv:
@ChinYikMing, by the way, feel free to submit another pull request to close #161 .

qwe661234 (approved):

PR #183, opened 2023-08-01
Make memory size portable on both 32/64-bit target

### LINKED ISSUE #161: Non-portable memory allocation

ChinYikMing:
Close #161

### DIFF
--- a/src/decode.h
+++ b/src/decode.h
@@ -163,7 +163,8 @@
     _(fuse1, 0)                            \
     _(fuse2, 0)                            \
     _(fuse3, 0)                            \
-    _(fuse4, 0)
+    _(fuse4, 0)                            \
+    _(fuse5, 0)
 /* clang-format on */
 
 /* IR list */
@@ -253,7 +254,7 @@ typedef struct rv_insn {
     uint8_t shamt;
 #endif
     /* fuse operation */
-    int16_t imm2;
+    int32_t imm2;
     opcode_fuse_t *fuse;
 
     /* instruction length */
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -1251,15 +1251,38 @@ RVOP(cswsp, {
 #endif
 
 /* auipc + addi */
-RVOP(fuse1, { rv->X[ir->rd] = (int32_t) (rv->PC + ir->imm + ir->imm2); })
+static bool do_fuse1(riscv_t *rv, const rv_insn_t *ir)
+{
+    rv->X[rv_reg_zero] = 0;
+    rv->csr_cycle += 2;
+    rv->X[ir->rd] = rv->PC + ir->imm;
+    rv->X[ir->rs1] = rv->X[ir->rd] + ir->imm2;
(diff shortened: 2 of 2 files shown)


jserv:
@ChinYikMing, Quick tip for submitting pull requests: when creating branches for your changes, consider using branch names other than `master`. This practice allows you to keep the master branch up to date and enables the proposed branches associated with pull requests to be easily rebased when needed.

jserv:
@qwe661234, An empirical approach is necessary to specify the limits, including memory size in this case. Can you do that?

qwe661234:
> @qwe661234, An empirical approach is necessary to specify the limits, including memory size in this case. Can you do that?

In my opinion, we can modify memory size to `0xFFFFFFFF` without checking data model. This memory size works in both 32-bit and 64-bit machine. Actually, the data memory usage of most test cases would not reach `0xFFFFFFFF`.

jserv:
@ChinYikMing, If you would like to close an issue, there must be an evident reason.

ChinYikMing:
My bad, I accidentally fetch and sync the upstream thus make this PR closed. I think the opinion of @qwe661234 is good since all test cases would not reach there for now. Shall I push another PR for closing this issue?

jserv:
> My bad, I accidentally fetch and sync the upstream thus make this PR closed. I think the opinion of @qwe661234 is good since all test cases would not reach there for now. Shall I push another PR for closing this issue?

Up to you. You can either reopen this pull request or create a new one. Next time, be aware of git history (the DAG).

PR #185, opened 2023-08-06
Support passing argc and argv to target program

### LINKED ISSUE #150: Handle argc/argv/envp properly
On Linux, after argc, argv, and envp, there is auxv, which serves as a key-value store for binary data. However, current rv32emu fails to set argc, argc, and envp properly. They should be specified during the setting of default stack pointer.
```c
void rv_reset(riscv_t *rv, riscv_word_t pc)
{
    ...
    rv->X[rv_reg_sp] = DEFAULT_STACK_ADDR;
    ...
}
```

Reference:
* [pk.c](https://github.com/riscv-software-src/riscv-pk/blob/master/pk/pk.c): Check the comment "place argc, argv, envp, auxp on stack"
* [Where do argc and argv come from?](https://briancallahan.net/blog/20200808.html)

ChinYikMing:
Half close #150, I left envp in TODO

I assumed all emulator's option are passed before the target program and its args thus I modify the parse_args function consequently

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,9 @@ CFLAGS += -Wno-unused-label
 CFLAGS += -include src/common.h
 
 # Set the default stack pointer
-CFLAGS += -D DEFAULT_STACK_ADDR=0xFFFFF000
+CFLAGS += -D DEFAULT_STACK_ADDR=0xFFFE000
+# Set the default args starting address
+CFLAGS += -D DEFAULT_ARGS_ADDR=0xFFFF000
 
 OBJS_EXT :=
 
@@ -140,7 +142,7 @@ check: $(BIN)
 EXPECTED_aes = Dec 15 2022 16:35:12 Test results AES-128 ECB encryption: PASSED! AES-128 ECB decryption: PASSED! AES-128 CBC encryption: PASSED! AES-128 CBC decryption: PASSED! AES-128 CFB encryption: PASSED! AES-128 CFB decryption: PASSED! AES-128 OFB encryption: PASSED! AES-128 OFB decryption: PASSED! AES-128 CTR encryption: PASSED! AES-128 CTR decryption: PASSED! AES-128 XTS encryption: PASSED! AES-128 XTS decryption: PASSED! AES-128 validate CMAC : PASSED! AES-128 Poly-1305 mac : PASSED! AES-128 GCM encryption: PASSED! AES-128 GCM decryption: PASSED! AES-128 CCM encryption: PASSED! AES-128 CCM decryption: PASSED! AES-128 OCB encryption: PASSED! AES-128 OCB decryption: PASSED! AES-128 SIV encryption: PASSED! AES-128 SIV decryption: PASSED! AES-128 GCMSIV encrypt: PASSED! AES-128 GCMSIV decrypt: PASSED! AES-128 EAX encryption: PASSED! AES-128 EAX decryption: PASSED! AES-128 key wrapping  : PASSED! AES-128 key unwrapping: PASSED! AES-128 FF1 encryption: PASSED! AES-128 FPE decryption: PASSED! +-> Let's do some extra tests AES-128 OCB encryption: PASSED! AES-128 OCB decryption: PASSED! AES-128 GCMSIV encrypt: PASSED! AES-128 GCMSIV decrypt: PASSED! AES-128 GCMSIV encrypt: PASSED! AES-128 GCMSIV decrypt: PASSED! AES-128 SIV encryption: PASSED! AES-128 SIV decryption: PASSED! AES-128 SIV encryption: PASSED! AES-128 SIV decryption: PASSED! AES-128 EAX encryption: PASSED! AES-128 EAX decryption: PASSED! AES-128 EAX encryption: PASSED! AES-128 EAX decryption: PASSED! AES-128 Poly-1305 mac : PASSED! inferior exit code 0
 misalign: $(BIN)
 	$(Q)$(PRINTF) "Running aes.elf ... "; 
-	$(Q)if [ "$(shell $(BIN) --misalign $(OUT)/aes.elf | uniq)" = "$(EXPECTED_aes)" ]; then \
+	$(Q)if [ "$(shell $(BIN) -m $(OUT)/aes.elf | uniq)" = "$(EXPECTED_aes)" ]; then \
         $(call notice, [OK]); \
 	else	\
         $(PRINTF) "Failed.\n"; \
--- a/Makefile
+++ b/Makefile
@@ -9,9 +9,9 @@ CFLAGS += -Wno-unused-label
 CFLAGS += -include src/common.h
 
 # Set the default stack pointer
-CFLAGS += -D DEFAULT_STACK_ADDR=0xFFFE000
+CFLAGS += -D DEFAULT_STACK_ADDR=0xFFFFE000
 # Set the default args starting address
-CFLAGS += -D DEFAULT_ARGS_ADDR=0xFFFF000
+CFLAGS += -D DEFAULT_ARGS_ADDR=0xFFFFF000
 
 OBJS_EXT :=
 
--- a/src/main.c
+++ b/src/main.c
@@ -6,6 +6,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <unistd.h>
 
 #include "elf.h"
 #include "state.h"
@@ -31,6 +32,11 @@ static bool opt_quiet_outputs = false;
 
 /* target executable */
 static const char *opt_prog_name = "a.out";
+
+/* target argc and argv */
+static int prog_argc;
+static char **prog_args;
+static const char *optstr = "tgqmhd:a:";
 
 /* enable misaligned memory access */
 static bool opt_misaligned = false;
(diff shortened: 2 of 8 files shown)


Makefile
```diff
@@ -9,7 +9,7 @@ CFLAGS += -Wno-unused-label
 CFLAGS += -include src/common.h
 
 # Set the default stack pointer
-CFLAGS += -D DEFAULT_STACK_ADDR=0xFFFFF000
+CFLAGS += -D DEFAULT_STACK_ADDR=0xFFFE000 -D DEFAULT_ARGS_ADDR=0xFFFF000
```
jserv: Split into a new line. i.e.,
```
CFLAGS += -D DEFAULT_STACK_ADDR=0xFFFE000
CFLAGS += -D DEFAULT_ARGS_ADDR=0xFFFF000
```

src/main.c
```diff
@@ -103,11 +103,11 @@ static void print_usage(const char *filename)
             filename);
 }
 
-static bool parse_args(int argc, char **args)
+static bool parse_args(int *argc, char ***args)
```
jserv: It is not elegant to have `char ***arg` here? Can you simplify?

jserv (changes requested):
Resolve the concerns raised by CodeQL.

src/main.c
```diff
@@ -103,11 +103,11 @@ static void print_usage(const char *filename)
...
 {
     /* parse each argument in turn */
-    for (int i = 1; i < argc; ++i) {
-        const char *arg = args[i];
+    for (int i = 1; i < *argc; ++i) {
```
jserv: You might consider to utilize [getopt(3)](https://man7.org/linux/man-pages/man3/getopt.3.html).
Example: https://github.com/taviso/123elf/blob/main/main.c#L167

src/riscv.c
```diff
@@ -124,17 +127,109 @@ void rv_delete(riscv_t *rv)
...
+    uintptr_t stack_top = stack_bottom - stack_size;
+    stack_top &= -16;
+
+    /* argc */
+    uintptr_t *sp = (uintptr_t *) stack_top;
+    memory_write(s->mem, (uintptr_t) sp,
+                 (void *) (s->mem->mem_base + (uintptr_t) args_p), sizeof(int));
+    args_p++;
+    sp = (uintptr_t *) ((uint32_t *) sp + 1); /* keep argc and args[0] within
+                                                 one word due to RISCV32 ABI */
```
jserv: It is meant to be "RV32 ABI" instead of "RISCV32 ABI."

tests/args.c
```diff
@@ -0,0 +1,10 @@
+#include <stdio.h>
```
jserv: Skip this test program. Let's consolidate `tests/readelf/readelf.c` instead.

src/main.c
ChinYikMing: I think `getopt ` is a cleaner way, but it does not support long option such as --quiet. If we want to parse long option, we might use `getopt_long`  but it is only compliance to GNU ( might having porting problem ), is it still recommended?

jserv: > I think `getopt ` is a cleaner way, but it does not support long option such as --quiet. If we want to parse long option, we might use `getopt_long` but it is only compliance to GNU ( might having porting problem ), is it still recommended?

It is acceptable to handle `--quiet` with GNU extensions since this option is not part of core functionalities. You still have to denote the usage though.

ChinYikMing: Is it also acceptable for other options? Or we could introduce a short option ( such as -q means --quiet ) to replace long option, if so we can use `getopt` seamlessly

jserv: The latter sounds good.

src/riscv.c
```diff
@@ -124,17 +127,109 @@ void rv_delete(riscv_t *rv)
...
 
+    state_t *s = rv_userdata(rv);
+
     /* set the default stack pointer */
     rv->X[rv_reg_sp] = DEFAULT_STACK_ADDR;
 
+    /* set emulating program's argc and args */
+    /*
+     * store argc, args to state->mem
+     * so that, we can use offset technique to compatible
```
jserv: Make the comments grammatically correct.

src/riscv.c
```diff
@@ -135,6 +138,98 @@ void rv_reset(riscv_t *rv, riscv_word_t pc)
...
+    size_t args_space_idx = 0;
+    size_t args_len;
+    size_t args_len_total = 0;
+    for (i = 0; i < argc; i++) {
+        const char *arg = args[i];
+        args_len = strlen(arg);
+        memory_write(s->mem, (uintptr_t) args_p, (void *) arg,
+                     (args_len + 1) * sizeof(char));
+        args_space[args_space_idx++] = args_len + 1;
+        args_p = (uintptr_t *) (((char *) args_p) + args_len + 1);
```
jserv: Rewrite as following:
```c
    args_p = (uintptr_t *) ((uintptr_t) args_p + args_len + 1);
```

src/riscv.c
```diff
@@ -135,6 +138,98 @@ void rv_reset(riscv_t *rv, riscv_word_t pc)
...
+        args_len_total += args_len + 1;
+    }
+    args_p = (uintptr_t *) (((char *) args_p) - args_len_total);
```
jserv: Ditto. Instead of casting to `char *`, specify `uintptr_t`.

qwe661234:
In file `src/risc.c`, the `memory_write` has been invoked twice to wrtie argc or args to memory (according to your comments). Why we need to maintain two copies of argc or args in virtual memory?

ChinYikMing:
> In file `src/risc.c`, the `memory_write` has been invoked twice to wrtie argc or args to memory (according to your comments). Why we need to maintain two copies of argc or args in virtual memory?

If we are running emulator on 64-bit machine, the virtual address would be 64-bit unless it is compile with -m32 option. Since we are RV32 emulator, so the address space should be < 4GB, from this point we cannot directly access 64-bit address from emulator. Thus, first call of `memory write` store the argc and argv to < 4GB address, and second call of `memory write` for RV32 ABI.

qwe661234:
> If we are running emulator on 64-bit machine, the virtual address would be 64-bit unless it is compile with -m32 option. Since we are RV32 emulator, so the address space should be < 4GB, from this point we cannot directly access 64-bit address from emulator. Thus, first call of `memory write` store the argc and argv to < 4GB address, and second call of `memory write` for RV32 ABI.

Ok, but I still confuse. I guess these two copies are for different purposes. Could you explain the different purpose between store the argc and argv to < 4GB address and `memory write` for RV32 ABI?(or maybe leave some comments) Thx.

ChinYikMing:
> Ok, but I still confuse. I guess these two copies are for different purposes. Could you explain the different purpose between store the argc and argv to < 4GB address and `memory write` for RV32 ABI?(or maybe leave some comments) Thx.

For example, we want access the args[0], the instruction might look like this: `lw a0, 4(sp)`. In mapping, emulator will call `memory_read_w`. From implementation of `memory_read_w`, it is base  memory + "offset",  this "offset" should within the memory size otherwise we get segmentation fault. Thus, value stored in `sp+4` should be the offset to the args[0]. The first copy of argc and argv mainly for maintaining the offset and making sure it is < 4GB.

qwe661234:
> For example, we want access the args[0], the instruction might look like this: `lw a0, 4(sp)`. In mapping, emulator will call `memory_read_w`. From implementation of `memory_read_w`, it is base memory + "offset", this "offset" should within the memory size otherwise we get segmentation fault. Thus, value stored in `sp+4` should be the offset to the args[0]. The first copy of argc and argv mainly for maintaining the offset and making sure it is < 4GB.

Does you mean the instruction `lw a0, 4(sp)` get the offset of memory address of args[0], and the data of args[0] store in this memory address(mem_base + offset)?

ChinYikMing:
> Does you mean the instruction `lw a0, 4(sp)` get the offset of memory address of args[0], and the data of args[0] store in this memory `address(mem_base + offset)`?

yes

ChinYikMing:
> Resolve the concerns raised by CodeQL.

I think I have solved it.

jserv:
I defer to @qwe661234 for confirmation.

Makefile
```diff
@@ -9,7 +9,9 @@ CFLAGS += -Wno-unused-label
 CFLAGS += -include src/common.h
 
 # Set the default stack pointer
-CFLAGS += -D DEFAULT_STACK_ADDR=0xFFFFF000
+CFLAGS += -D DEFAULT_STACK_ADDR=0xFFFE000
```
qwe661234: Why you remove the highest 4 bit?

ChinYikMing: > Why you remove the highest 4 bit?

To left some space(4096) for list of arguments. According to the [ARG_MAX](https://man7.org/linux/man-pages/man3/sysconf.3.html), this size of the list of arguments must not be less than _POSIX_ARG_MAX (4096). Since we are emulating target program, so we can decide our own size instead and I think 4096 is decent for now.

qwe661234: Your reply seems like you change `F000` to `E000`, but my question is why you change `FFFFF000` to `FFFF000`. Maybe I misunderstand your reply, could you explain it more detail.

ChinYikMing: > Your reply seems like you change `F000` to `E000`, but my question is why you change `FFFFF000` to `FFFF000`. Maybe I misunderstand your reply, could you explain it more detail.

My bad, I will change it back to 'FFFFF000'

qwe661234 (approved):

jserv:

PR #187, opened 2023-08-12
Implement static instruction usage histogram

### LINKED ISSUE #34: Implement instruction usage histogram
With the ability to record and print histograms, we can observe instruction frequency and print. Sample output:
```
instruction usage histogram
~~~~~~~~~~~~~~~~~~~~~~~~~~~
    1. lw         16.37% [843055543] #######################################
    2. xor        13.85% [713031972] ################################
    3. add        13.69% [704643870] ################################
    4. slli       13.20% [679477645] ###############################
    5. srliw      10.75% [553648296] #########################
    6. andi        9.94% [511705332] #######################
    7. srli        3.05% [157286473] #######
    8. lbu         2.93% [150995155] ######
    9. addi        2.69% [138412722] ######
   10. addiw       2.44% [125829177] #####
   11. sd          1.79% [92275184 ] ####
   12. sb          1.47% [75497501 ] ###
   13. jal         1.14% [58720476 ] ##
   14. beq         1.06% [54526059 ] ##
   15. ld          0.98% [50332182 ] ##
   16. and         0.98% [50331715 ] ##
   17. slliw       0.98% [50331682 ] ##
   18. bne         0.90% [46137534 ] ##
   19. or          0.65% [33554442 ] #
   20. auipc       0.41% [20971539 ] 
   21. lui         0.24% [12583002 ] 
   22. mulw        0.16% [8388608  ] 
   23. lwu         0.16% [8388608  ] 
   24. jalr        0.08% [4194443  ] 
   25. sraiw       0.08% [4194314  ] 
   26. sw          0.00% [213      ] 
   27. bltu        0.00% [78       ] 
   28. bge         0.00% [39       ] 
   29. blt         0.00% [33       ] 
   30. bgeu        0.00% [33       ] 
   31. sub         0.00% [29       ] 
...
```

Reference:
* [rv8](https://github.com/michaeljclark/rv8) : Generate instructions Histogram via "rv-bin histogram"

ChinYikMing:
Copy and modify the elf.h/c from src directory for the tool to not mess up with the main  elf.h/c in src directory(code size is bigger but cleaner).

Close #34

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,8 +118,6 @@ $(BIN): $(OBJS)
 # Tools
 include mk/tools.mk
 
-deps += $(HIST_MAIN_OBJ:%.o=%.o.d)
-
 # RISC-V Architecture Tests
 include mk/riscv-arch-test.mk
 include mk/tests.mk
@@ -166,7 +164,7 @@ endif
 endif
 
 clean:
-	$(RM) $(BIN) $(OBJS) $(HIST_BIN) $(HIST_MAIN_OBJ) $(deps) $(CACHE_OUT)
+	$(RM) $(BIN) $(OBJS) $(HIST_BIN) $(HIST_OBJS) $(deps) $(CACHE_OUT)
 distclean: clean
 	-$(RM) $(DOOM_DATA) $(QUAKE_DATA)
 	$(RM) -r $(OUT)/id1
--- a/Makefile
+++ b/Makefile
@@ -115,6 +115,11 @@ $(BIN): $(OBJS)
 	$(VECHO) "  LD\t$@\n"
 	$(Q)$(CC) -o $@ $^ $(LDFLAGS)
 
+# Tools
+include mk/tools.mk
+
+deps += $(HIST_MAIN_OBJ:%.o=%.o.d)
+
 # RISC-V Architecture Tests
 include mk/riscv-arch-test.mk
 include mk/tests.mk
@@ -161,7 +166,7 @@ endif
 endif
 
 clean:
-	$(RM) $(BIN) $(OBJS) $(deps) $(CACHE_OUT)
+	$(RM) $(BIN) $(OBJS) $(HIST_BIN) $(HIST_MAIN_OBJ) $(deps) $(CACHE_OUT)
 distclean: clean
 	-$(RM) $(DOOM_DATA) $(QUAKE_DATA)
 	$(RM) -r $(OUT)/id1
--- a/mk/tools.mk
+++ b/mk/tools.mk
@@ -5,19 +5,19 @@ HIST_OBJS := \
 	riscv.o \
 	map.o \
 	elf.o \
-	decode.o
-HIST_OBJS := $(addprefix $(OUT)/, $(HIST_OBJS))
+	decode.o \
+	rv_histogram.o
 
-HIST_MAIN_OBJ := rv_histogram.o
-HIST_MAIN_OBJ := $(addprefix $(OUT)/, $(HIST_MAIN_OBJ))
+HIST_OBJS := $(addprefix $(OUT)/, $(HIST_OBJS))
+deps += $(HIST_OBJS:%.o=%.o.d)
 
 $(OUT)/%.o: tools/%.c
 	$(VECHO) "  CC\t$@\n"
 	$(Q)$(CC) -o $@ $(CFLAGS) -Isrc -c -MMD -MF $@.d $<
 
 # GDBSTUB is diabled for excluding the mini-gdb during compilation
-$(HIST_BIN): $(HIST_OBJS) $(HIST_MAIN_OBJ)
+$(HIST_BIN): $(HIST_OBJS)
 	$(VECHO) "  LD\t$@\n"
-	$(Q)$(CC) -o $@ $^ $(LDFLAGS) -D RV32_FEATURE_GDBSTUB=0
+	$(Q)$(CC) -o $@ -D RV32_FEATURE_GDBSTUB=0 $^ $(LDFLAGS) 
 
 tool: $(HIST_BIN)
--- a/mk/tools.mk
+++ b/mk/tools.mk
@@ -1,0 +1,23 @@
+HIST_BIN := $(OUT)/rv_histogram
+
+# FIXME: riscv.o and map.o are dependency of elf.o not the tool
+HIST_OBJS := \
+	riscv.o \
+	map.o \
+	elf.o \
+	decode.o
+HIST_OBJS := $(addprefix $(OUT)/, $(HIST_OBJS))
+
+HIST_MAIN_OBJ := rv_histogram.o
+HIST_MAIN_OBJ := $(addprefix $(OUT)/, $(HIST_MAIN_OBJ))
+
+$(OUT)/%.o: tools/%.c
+	$(VECHO) "  CC\t$@\n"
+	$(Q)$(CC) -o $@ $(CFLAGS) -Isrc -c -MMD -MF $@.d $<
+
+# GDBSTUB is diabled for excluding the mini-gdb during compilation
+$(HIST_BIN): $(HIST_OBJS) $(HIST_MAIN_OBJ)
+	$(VECHO) "  LD\t$@\n"
+	$(Q)$(CC) -o $@ $^ $(LDFLAGS) -D RV32_FEATURE_GDBSTUB=0
+
+tool: $(HIST_BIN)
--- a/src/elf.c
+++ b/src/elf.c
@@ -50,71 +50,7 @@ enum {
     STT_TLS = 6,
 };
 
-enum {
-    SHT_NULL = 0,     /* Section header table entry unused */
-    SHT_PROGBITS = 1, /* Program data */
-    SHT_SYMTAB = 2,   /* Symbol table */
-    SHT_STRTAB = 3,   /* String table */
-    SHT_RELA = 4,     /* Relocation entries with addends */
-    SHT_HASH = 5,     /* Symbol hash table */
-    SHT_DYNAMIC = 6,  /* Dynamic linking information */
-    SHT_NOTE = 7,     /* Notes */
-    SHT_NOBITS = 8,   /* Program space with no data (bss) */
-    SHT_REL = 9,      /* Relocation entries, no addends */
-    SHT_SHLIB = 10,   /* Reserved */
-    SHT_DYNSYM = 11,  /* Dynamic linker symbol table */
-    SHT_NUM = 12,
-    SHT_LOPROC = 0x70000000, /* Start of processor-specific */
-    SHT_HIPROC = 0x7fffffff, /* End of processor-specific */
-    SHT_LOUSER = 0x80000000, /* Start of application-specific */
-    SHT_HIUSER = 0xffffffff, /* End of application-specific */
-};
-
 #define ELF_ST_TYPE(x) (((unsigned int) x) & 0xf)
-
-/* Elf32 header */
-struct Elf32_Ehdr {
-    uint8_t e_ident[EI_NIDENT];
-    Elf32_Half e_type;      /* Object file type */
-    Elf32_Half e_machine;   /* Architecture */
-    Elf32_Word e_version;   /* Object file version */
-    Elf32_Addr e_entry;     /* Entry point virtual address */
-    Elf32_Off e_phoff;      /* Program header table file offset */
-    Elf32_Off e_shoff;      /* Section header table file offset */
-    Elf32_Word e_flags;     /* Processor-specific flags */
-    Elf32_Half e_ehsize;    /* ELF header size in bytes */
-    Elf32_Half e_phentsize; /* Program header table entry size */
-    Elf32_Half e_phnum;     /* Program header table entry count */
-    Elf32_Half e_shentsize; /* Section header table entry size */
-    Elf32_Half e_shnum;     /* Section header table entry count */
-    Elf32_Half e_shstrndx;  /* Section header string table index */
-};
-
-/* Elf32 program header table */
-struct Elf32_Phdr {
-    Elf32_Word p_type;   /* Type, a combination of ELF_PROGRAM_TYPE_ */
-    Elf32_Off p_offset;  /* Offset in the file of the program image */
-    Elf32_Addr p_vaddr;  /* Virtual address in memory */
-    Elf32_Addr p_paddr;  /* Optional physical address in memory */
-    Elf32_Word p_filesz; /* Size of the image in the file */
-    Elf32_Word p_memsz;  /* Size of the image in memory */
-    Elf32_Word p_flags;  /* Type-specific flags */
-    Elf32_Word p_align;  /* Memory alignment in bytes */
-};
-
-/* Elf32 section header table */
-struct Elf32_Shdr {
-    Elf32_Word sh_name;      /* Section name */
-    Elf32_Word sh_type;      /* Section type */
-    Elf32_Word sh_flags;     /* Section flags */
-    Elf32_Addr sh_addr;      /* Section virtual addr at execution */
-    Elf32_Off sh_offset;     /* Section file offset */
-    Elf32_Word sh_size;      /* Section size in bytes */
-    Elf32_Word sh_link;      /* Link to another section */
-    Elf32_Word sh_info;      /* Additional section information */
-    Elf32_Word sh_addralign; /* Section alignment */
-    Elf32_Word sh_entsize;   /* Entry size if section holds table */
-};
 
 struct elf_internal {
     const struct Elf32_Ehdr *hdr;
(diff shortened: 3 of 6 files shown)


tools/Makefile
```diff
@@ -0,0 +1,71 @@
+CC ? = gcc OUT = $(CURDIR) BIN
```
jserv: Indent this file for the sake of consistency.

jserv (changes requested):
Reuse existing src/elf.[ch] as much as possible.

src/decode.h
```diff
@@ -170,7 +170,7 @@
 /* IR list */
 enum {
```
jserv: Rewrite as following:
```c
enum {
#define _(inst, can_branch) rv_insn_##inst,
    RISCV_INSN_LIST
#undef _
N_RISCV_INSN_LIST
};
```

tools/Makefile
```diff
@@ -0,0 +1,65 @@
+CC ?= gcc
```
jserv: Move the targets for generating tools in `mk/tools.mk` and integrate it into top-level `Makefile`.

tools/insn_freq_explorer.c
```diff
@@ -0,0 +1,265 @@
...
+    }
+    uint8_t *elf_first_byte = (uint8_t *) e->raw_data;
+    size_t total_insn = 0;
+    rv_insn_t ir;
+    bool res;
+
+    for (int i = 0; i < hdr->e_shnum; i++) {
+        struct Elf32_Shdr *shdr = (struct Elf32_Shdr *) shdrs + i;
+
+        if (!strcmp(get_sh_type(shdr->sh_type), "PROGBITS") &&
```
jserv: Avoid comparing strings. Instead, check the values.

jserv: Rewrite as following:
```c
    if ( ... COND_TO_SKIP_THIS_ITERATION ...)
        continue;
     uint32_t *exec_start_addr =
                (uint32_t *) ((uint8_t *) elf_first_byte + shdr->sh_offset);
    uint32_t *exec_end_addr =
    (uint32_t *) ((uint8_t *) exec_start_addr + shdr->sh_size);
    ...
```

jserv: See https://github.com/sysprog21/rv32emu/blob/master/CONTRIBUTING.md (Avoid redundant else)

tools/insn_freq_explorer.c
```diff
@@ -0,0 +1,265 @@
...
+
+static char *gen_hist_bar(char *hist_bar,
+                          size_t insn_freq,
+                          size_t max_insn_freq,
+                          unsigned short max_col,
+                          unsigned short used_col)
+{
+    int hist_bar_len = insn_freq * (max_col - used_col) / max_insn_freq;
+    for (int i = 0; i < hist_bar_len; i++) {
+        hist_bar[i] = '*';
```
jserv: Use the text rendering technique as `_mi_make_bar` (See https://github.com/microsoft/mimalloc/pull/529/files ).

tools/insn_freq_explorer.c
```diff
@@ -0,0 +1,265 @@
...
+
+    for (int i = 0; i < hdr->e_shnum; i++) {
+        struct Elf32_Shdr *shdr = (struct Elf32_Shdr *) shdrs + i;
+
+        if (!strcmp(get_sh_type(shdr->sh_type), "PROGBITS") &&
+            strstr(get_sh_flags(shdr->sh_flags), "X")) {
+            uint32_t *exec_start_addr =
+                (uint32_t *) ((uint8_t *) elf_first_byte + shdr->sh_offset);
+            uint32_t *exec_end_addr =
+                (uint32_t *) ((uint8_t *) exec_start_addr + shdr->sh_size);
```
jserv: Use `uintptr_t` when possible.

jserv:
The result of command `tools/insn_freq_explorer build/hello.elf `
```
+---------------------------------------------+
| RV32 Target Instruction Frequency Histogram |
+---------------------------------------------+
  1. addi       0.50% [9         ] ***************************************************************************************************************************
  2. unknown    0.28% [5         ] ********************************************************************
  3. jal        0.17% [3         ] *****************************************
  4. ecall      0.11% [2         ] ***************************
  5. nop        0.06% [1         ] *************
  6. auipc      0.06% [1         ] *************
  7. beq        0.06% [1         ] *************
  8. cnop       0.06% [1         ] *************
  9. lui        0.00% [0         ] 
 10. jalr       0.00% [0         ] 
 ...
123. cjalr      0.00% [0         ] 
124. cadd       0.00% [0         ] 
125. cswsp      0.00% [0         ] 
```

We shall skip unused instructions. i.e., the one with 0%. In addition, the second `unknown` is weird.

ChinYikMing:
> Reuse existing src/elf.[ch] as much as possible.

I will get "incomplete definition of type" error without moving the definition of `struct Elf32_Ehdr`, `struct Elf32_Shdr` to **elf.h** header file. To preserve the encapsulation of these structures, is it suggest to implement the logic of the tool in **elf.c** file? This is the main reason that I separate new elf.[ch].

jserv:
> I will get "incomplete definition of type" error without moving the definition of `struct Elf32_Ehdr`, `struct Elf32_Shdr` to **elf.h** header file. To preserve the encapsulation of these structures, is it suggest to implement the logic of the tool in **elf.c** file? This is the main reason that I separate new elf.[ch].

Extend the existing src/elf.[ch] to incorporate the necessary functionalities required for implementing ELF analysis.

ChinYikMing:
> > I will get "incomplete definition of type" error without moving the definition of `struct Elf32_Ehdr`, `struct Elf32_Shdr` to **elf.h** header file. To preserve the encapsulation of these structures, is it suggest to implement the logic of the tool in **elf.c** file? This is the main reason that I separate new elf.[ch].
> 
> Extend the existing src/elf.[ch] to incorporate the necessary functionalities required for implementing ELF analysis.

The trivial solution is moving `struct Elf32_Ehdr`, `struct Elf32_Shdr` to **elf.h**, is it applicable?(lost encapsulation)

jserv:
> The trivial solution is moving `struct Elf32_Ehdr`, `struct Elf32_Shdr` to **elf.h**, is it applicable?(lost encapsulation)

`src/elf.h` in not part of public headers, so the proposed change looks fine.

qwe661234:
This work looks like a static analysis, there is a more simple way to achieve this. You can just utilize a python script to parse the result of `objdump`. Because we want to keep the source code tidy and short.

jserv:
> This work looks like a static analysis, there is a more simple way to achieve this. You can just utilize a python script to parse the result of `objdump`. Because we want to keep the source code tidy and short.

In essence, the proposed tool is a standalone program. Given its non-essential nature to core functionalities, it is acceptable to develop a simple program that mimics the tasks of `objdump` from the GNU toolchain, eliminating the need for a runtime dependency on the GNU toolchain.

jserv:
@qwe661234 , By the way, how can one determine the most frequent instruction sequence during execution (in particular, CoreMark and Dhrystone), rather than focusing on a specific instruction?

qwe661234:
For runtime analysis, we need to insert profiling code into emulation loop and analyze the profiling data after emulation.

jserv:
> For runtime analysis, we need to insert profiling code into emulation loop and analyze the profiling data after emulation.

Create an issue to track runtime analysis and profiling. There should be an explicit profiling mechanism when JIT compiler (in particular, its mixture of interpreter and/or tier compilation) .

jserv (changes requested):
Refine the git commit messages to emphasize on static instruction analysis.

tools/insn_freq_explorer.c
```diff
@@ -0,0 +1,265 @@
+/*
```
jserv: Rename the file name to `tools/rv-histogram.c`. At the moment, only instruction usage is implemented. Later, we may extend for register usage.

ChinYikMing:
> In addition, the second `unknown` is weird.

The tool works well for other programs. I think maybe there is something wrong during the build of `hello.elf`.

tools/rv_histogram.c
```diff
@@ -0,0 +1,276 @@
+#include <errno.h>
```
jserv: Add copyright notice.

tools/rv_histogram.c
```diff
@@ -0,0 +1,276 @@
...
+static int cmp_asc(const void *a, const void *b)
+{
+    return ((rv_insn_hist_t *) a)->freq - ((rv_insn_hist_t *) b)->freq;
+}
+
+/* used to adjust the length of histogram bar */
+static unsigned short get_win_max_col()
+{
+    struct winsize terminal_size;
+    ioctl(STDOUT_FILENO, TIOCGWINSZ, &terminal_size);
```
jserv, line 54: This tool is supposed to be portable, and you shall consider the environments lacking of full POSIX compliance. e.g., MS-Windows.

tools/rv_histogram.c
```diff
@@ -0,0 +1,276 @@
...
+        default:
+            return false;
+        }
+    }
+
+    elf_prog = args[optind];
+
+    return true;
+}
+
```
jserv, line 136: Remove this blank line.

tools/rv_histogram.c
```diff
@@ -0,0 +1,276 @@
...
+#if RV32_HAS(EXT_C)
+    {"caddi4spn", 0}, {"clw", 0},      {"csw", 0},      {"cnop", 0},
+    {"caddi", 0},     {"cjal", 0},     {"cli", 0},      {"caddi16sp", 0},
+    {"clui", 0},      {"csrli", 0},    {"csrai", 0},    {"candi", 0},
+    {"csub", 0},      {"cxor", 0},     {"cor", 0},      {"cand", 0},
+    {"cj", 0},        {"cbeqz", 0},    {"cbnez", 0},    {"cslli", 0},
+    {"clwsp", 0},     {"cjr", 0},      {"cmv", 0},      {"cebreak", 0},
+    {"cjalr", 0},     {"cadd", 0},     {"cswsp", 0},
+#endif
+    {"unknown", 0}};
```
jserv: Add a comma (`,`) right after `0}`, so that clang-format can perform well.

tools/rv_histogram.c
```diff
@@ -0,0 +1,276 @@
...
+static bool ascending_order = false;
+static const char *elf_prog = NULL;
+static const char *optstr = "a";
+
+typedef struct {
+    char insn_str[16];
+    size_t freq;
+} rv_insn_hist_t;
+
+static rv_insn_hist_t rv_insn_str_and_freq_tbl[] = {
```
jserv: Rename this table to `rv_insn_stats`, which means the statistics of RV32 instructions including frequency and the mnemonic.

tools/rv_histogram.c
```diff
@@ -0,0 +1,276 @@
...
+    {"nop", 0},       {"lui", 0},      {"auipc", 0},    {"jal", 0},
```
jserv: Use [X-macro](https://en.wikipedia.org/wiki/X_macro) trick and [Stringizing](https://gcc.gnu.org/onlinedocs/cpp/Stringizing.html) to generate this table. Of course, you have to skip fuse instructions when manipulations.

tools/rv_histogram.c
```diff
@@ -0,0 +1,276 @@
...
+#include <sys/ioctl.h>
+#include <unistd.h>
+
+#include "../src/decode.h"
+#include "../src/elf.h"
+
+#define ARRAYS_SIZE(arr) (sizeof(arr) / sizeof(arr[0]))
+
+/* 5 are fuse instructions */
+#define RISCV_INSN_ACTUAL_SIZE N_RISCV_INSN_LIST - 5
```
jserv: Rename to `RV_INSN_AVAIL_SIZE`, which emphasizes on the availability.

jserv:
> > the second `unknown` is weird.
> The tool works well for other programs. I think maybe there is something wrong during the build of `hello.elf`.

Create an issue to track unknown instruction appearing in statistics after this pull request is merged.

tools/rv_histogram.c
```diff
@@ -0,0 +1,276 @@
+#include <errno.h>
+#include <getopt.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/ioctl.h>
+#include <unistd.h>
+
+#include "../src/decode.h"
```
jserv: Don't specify `../src`. Enforce `CFLAGS` instead.

tools/rv_histogram.c
```diff
@@ -0,0 +1,276 @@
...
+#include "../src/elf.h"
+
+#define ARRAYS_SIZE(arr) (sizeof(arr) / sizeof(arr[0]))
```
jserv: Move the definition of `ARRAYS_SIZE` to `src/common.h`.

mk/tools.mk
```diff
@@ -0,0 +1,23 @@
+OUT := build
```
jserv: You don't have to specify `OUT`.

tools/rv_histogram.c
```diff
@@ -0,0 +1,376 @@
+#include <stdio.h>
```
jserv, line 7: Add copyright notice at the beginning of this file.

tools/rv_histogram.c
```diff
@@ -0,0 +1,376 @@
...
+static bool ascending_order = false;
+static const char *elf_prog = NULL;
+static const char *optstr = "a";
+
+typedef struct {
+    char insn_str[16];
+    size_t freq;
+} rv_insn_hist_t;
+
+#define RISCV_INSN_STATS           \
```
jserv: Do do this! You can simply use `RISCV_INSN_LIST`.

tools/rv_histogram.c
```diff
@@ -0,0 +1,376 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <sys/ioctl.h>
```
jserv: Be aware of conditional build for `_WIN32`. Both `sys/ioctl.h` and `unistd.h` might be absent.

tools/rv_histogram.c
ChinYikMing: But `RISCV_INSN_LIST` is concatenate the string not stringizing the string

Makefile
```diff
@@ -61,7 +61,11 @@ $(OUT)/syscall_sdl.o: CFLAGS += $(shell sdl2-config --cflags)
 LDFLAGS += $(shell sdl2-config --libs)
 endif
 
-ENABLE_GDBSTUB ?= 1
+ifeq ($(filter tool,$(MAKECMDGOALS)),tool)
```
jserv: Don't do this. Minimize the necessary changes. It does not harm if gdb-stub is built with `tool` target.

src/elf.c
```diff
@@ -3,6 +3,7 @@
  * "LICENSE" for information on usage and redistribution of this file.
  */
 
+#include <stdio.h>
```
jserv: This inclusion of `<stdio.h>` is not required.

tools/rv_histogram.c
jserv: Check `src/common.h` for catenation tricks.

tools/rv_histogram.c
```diff
@@ -0,0 +1,384 @@
...
+        X(cmv)                     \
+        X(cebreak)                 \
+        X(cjalr)                   \
+        X(cadd)                    \
+        X(cswsp)                   \
+    )                              \
+    X(unknown)
+/* clang-format on */
+
+static rv_insn_hist_t rv_insn_stats[] = {
```
jserv, line 29: Remove the proposed `RISCV_INSN_STATS` macro and rewrite as following:
```c
static rv_insn_hist_t rv_insn_stats[] = {
#define _(inst, can_branch) {#inst, 0},
    RISCV_INSN_LIST
#undef _                                                                                                                                                      
};
```

Be mindful of the notation: "insn" is employed for mnemonic identifiers, while "inst" is used for parameters.

tools/rv_histogram.c
```diff
@@ -0,0 +1,384 @@
...
+
+#include "elf.h"
+#include "decode.h"
+
+/* 5 are fuse instructions */
+#define RISCV_INSN_AVAIL_SIZE N_RISCV_INSN_LIST - 5 
+
+static bool ascending_order = false;
+static const char *elf_prog = NULL;
+static const char *optstr = "a";
```
jserv: Reduce the scope of `optstr`. i.e., move it to function `parse_args`.

jserv (changes requested):
Rebase the latest `master` branch where macro-op fusion is moved out of decoder header.

mk/tools.mk
```diff
@@ -0,0 +1,25 @@
...
+	riscv.o \
+	map.o \
+	elf.o \
+	decode.o
+HIST_OBJS := $(addprefix $(OUT)/, $(HIST_OBJS))
+
+HIST_MAIN_OBJ := rv_histogram.o
+HIST_MAIN_OBJ := $(addprefix $(OUT)/, $(HIST_MAIN_OBJ))
+
+CFLAGS += -Isrc
```
jserv: Remove this line and specify `-Isrc` in the rules associated with `tools/%.c`.

tools/rv_histogram.c
```diff
@@ -0,0 +1,384 @@
...
+#else
+#include <unistd.h>
+#include <sys/ioctl.h>
+#endif
+
+#include "elf.h"
+#include "decode.h"
+
+/* 5 are fuse instructions */
+#define RISCV_INSN_AVAIL_SIZE N_RISCV_INSN_LIST - 5
```
jserv: After rebasing the latest master branch, this macro no longer makes sense.

tools/rv_histogram.c
ChinYikMing: Great technique for reusing `RISCV_INSN_LIST` !

jserv:
Consider the changes below:
```diff
diff --git a/Makefile b/Makefile
index 78b0890..52a24b9 100644
--- a/Makefile
+++ b/Makefile
@@ -118,8 +118,6 @@ $(BIN): $(OBJS)
 # Tools
 include mk/tools.mk
 
-deps += $(HIST_MAIN_OBJ:%.o=%.o.d)
-
 # RISC-V Architecture Tests
 include mk/riscv-arch-test.mk
 include mk/tests.mk
@@ -166,7 +164,7 @@ endif
 endif
 
 clean:
-       $(RM) $(BIN) $(OBJS) $(HIST_BIN) $(HIST_MAIN_OBJ) $(deps) $(CACHE_OUT)
+       $(RM) $(BIN) $(OBJS) $(HIST_BIN) $(HIST_OBJS) $(deps) $(CACHE_OUT)
 distclean: clean
        -$(RM) $(DOOM_DATA) $(QUAKE_DATA)
        $(RM) -r $(OUT)/id1
diff --git a/mk/tools.mk b/mk/tools.mk
index 41f6040..11a3ac0 100644
--- a/mk/tools.mk
+++ b/mk/tools.mk
@@ -5,19 +5,19 @@ HIST_OBJS := \
        riscv.o \
        map.o \
        elf.o \
-       decode.o
-HIST_OBJS := $(addprefix $(OUT)/, $(HIST_OBJS))
+       decode.o \
+       rv_histogram.o
 
-HIST_MAIN_OBJ := rv_histogram.o
-HIST_MAIN_OBJ := $(addprefix $(OUT)/, $(HIST_MAIN_OBJ))
+HIST_OBJS := $(addprefix $(OUT)/, $(HIST_OBJS))
+deps += $(HIST_OBJS:%.o=%.o.d)
 
 $(OUT)/%.o: tools/%.c
        $(VECHO) "  CC\t$@\n"
        $(Q)$(CC) -o $@ $(CFLAGS) -Isrc -c -MMD -MF $@.d $<
 
 # GDBSTUB is diabled for excluding the mini-gdb during compilation
-$(HIST_BIN): $(HIST_OBJS) $(HIST_MAIN_OBJ)
+$(HIST_BIN): $(HIST_OBJS)
        $(VECHO) "  LD\t$@\n"
-       $(Q)$(CC) -o $@ $^ $(LDFLAGS) -D RV32_FEATURE_GDBSTUB=0
+       $(Q)$(CC) -o $@ -D RV32_FEATURE_GDBSTUB=0 $^ $(LDFLAGS)
 
 tool: $(HIST_BIN)
```

jserv:

PR #195, opened 2023-08-16
Fix total frequency calculation and update example screenshots

ChinYikMing:


tools/rv_histogram.c
```diff
@@ -193,14 +193,22 @@ void reg_hist_incr(rv_insn_t *ir)
 
     uint8_t reg_mask = rv_insn_stats[ir->opcode].reg_mask;
 
-    if (reg_mask & 0x1)
+    if (reg_mask & 0x1) {
```
jserv: Use `F_rs1`, `F_rs2`, `F_rs3`, and `F_rd`.

jserv:

Issue #153, opened 2023-07-07
Fast-path execution through memcpy/memset Injection


jserv:
Special super-instructions, such as `memcpy` and `memset` provided by newlib, have the capability to substitute specific functions implemented in RISC-V machine code.

By utilizing these purpose-built super-instructions, which are specifically designed for efficient memory copying and setting operations, it is anticipated that the overall speed of emulation will be enhanced. This proposed solution entails replacing the original RISC-V machine code functions with the optimized super-instructions, leading to improved performance while maintaining full functionality.

See [MEMZERO and MEMCOPY instructions proposal](https://github.com/AndyGlew/Ri5-stuff/wiki/MEMZERO-and-MEMCOPY-instructions-proposal) for reference.

Proposed changes:
1. Implement super-instructions `MEMZERO` and `MEMCOPY` within the `src/emulation.c` file.
2. Verify whether the input ELF file includes the symbols for `memcpy` or `memset`.
3. If these symbols are present, employ the existing macro operation fusion mechanism to substitute the function calls to `memcpy` and/or `memset` with the aforementioned super-instructions.

jserv:
`strcpy`, `memset` and `memcpy` built with newlib are disassembled below:
```c
000461f8 <strcpy>:
   461f8:       00050793                mv      a5,a0
   461fc:       0005c703                lbu     a4,0(a1)
   46200:       00178793                addi    a5,a5,1
   46204:       00158593                addi    a1,a1,1
   46208:       fee78fa3                sb      a4,-1(a5)
   4620c:       fe0718e3                bnez    a4,461fc <strcpy+0x4>
   46210:       00008067                ret

00044cf8 <memcpy>:
   44cf8:       00050313                mv      t1,a0
   44cfc:       00060e63                beqz    a2,44d18 <memcpy+0x20>
   44d00:       00058383                lb      t2,0(a1)
   44d04:       00730023                sb      t2,0(t1)
   44d08:       fff60613                addi    a2,a2,-1
   44d0c:       00130313                addi    t1,t1,1
   44d10:       00158593                addi    a1,a1,1
   44d14:       fe0616e3                bnez    a2,44d00 <memcpy+0x8>
   44d18:       00008067                ret

00044d1c <memset>:
   44d1c:       00050313                mv      t1,a0
   44d20:       00060a63                beqz    a2,44d34 <memset+0x18>
   44d24:       00b30023                sb      a1,0(t1)
[...]

PR #196, opened 2023-08-24
Substitute memset/memcpy calls with faster alternatives

qwe661234:
1. Because RISC-V use jal instruction as function call, so we remove recursive jump translation to detect function call.
2. To identify memset/memcpy, we analyze the pattern of the first basic block of memset/memcpy for comparison. If the basic block aligns with the pattern, we extract the instruction sequence starting from the initial PC of the basic block and then compare it with the pre-recorded memset/memcpy instruction sequence. Upon detecting a memset/memcpy, we substitute the basic block with the standard library function memset/memcpy.

Based on performance results below, we gain 2% performance improvement when running coreMark and dhrysone. The performance impact is resulted in the number of invoking memset/memcpy times.

* Intel Core i7-11700

|   Metric   | origin  | proposed |Speedup|
|------------| ---------| ---------| -------|
| Dhrystone  | 1413.11 | 1447.11  |  +2%  |
| CoreMark   | 1497.35 | 1525.31  |  +2%  |

See: #153

### DIFF
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -550,19 +550,8 @@ static void block_translate(riscv_t *rv, block_t *block)
         block->pc_end += ir->insn_len;
         block->n_insn++;
         /* stop on branch */
-        if (insn_is_branch(ir->opcode)) {
-            /* recursive jump translation */
-            if (ir->opcode == rv_insn_jal
-#if RV32_HAS(EXT_C)
-                || ir->opcode == rv_insn_cj || ir->opcode == rv_insn_cjal
-#endif
-            ) {
-                block->pc_end = block->pc_end - ir->insn_len + ir->imm;
-                ir->branch_taken = ir + 1;
-                continue;
-            }
+        if (insn_is_branch(ir->opcode))
             break;
-        }
     }
     block->ir[block->n_insn - 1].tailcall = true;
 }
@@ -744,6 +733,14 @@ void rv_step(riscv_t *rv, int32_t cycles)
                     last_ir->branch_taken = block->ir;
                 else if (!last_ir->branch_untaken)
                     last_ir->branch_untaken = block->ir;
+            } else if (last_ir->opcode == rv_insn_jal
+#if RV32_HAS(EXT_C)
+                       || last_ir->opcode == rv_insn_cj ||
+                       last_ir->opcode == rv_insn_cjal
+#endif
+            ) {
+                if (!last_ir->branch_taken)
+                    last_ir->branch_taken = block->ir;
             }
         }
         last_pc = rv->PC;
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -325,7 +325,9 @@ static uint32_t last_pc = 0;
     _(fuse2)           \
     _(fuse3)           \
     _(fuse4)           \
-    _(fuse5)
+    _(fuse5)           \
+    _(fuse6)           \
+    _(fuse7)
 
 enum {
     rv_insn_fuse0 = N_RV_INSNS,
@@ -421,6 +423,36 @@ static bool do_fuse5(riscv_t *rv, const rv_insn_t *ir)
     if (unlikely(RVOP_NO_NEXT(ir)))
         return true;
     const rv_insn_t *next = ir + 2;
+    MUST_TAIL return next->impl(rv, next);
+}
+
+/* memset */
+static bool do_fuse6(riscv_t *rv, const rv_insn_t *ir)
+{
+    rv->X[rv_reg_zero] = 0;
+    rv->csr_cycle += 2;
+    memory_t *m = ((state_t *) rv->userdata)->mem;
+    memset((char *) m->mem_base + rv->X[rv_reg_a0], rv->X[rv_reg_a1],
+           rv->X[rv_reg_a2]);
+    rv->PC = rv->X[rv_reg_ra] & ~1U;
+    if (unlikely(RVOP_NO_NEXT(ir)))
+        return true;
+    const rv_insn_t *next = ir + 1;
+    MUST_TAIL return next->impl(rv, next);
+}
+
+/* memcpy */
+static bool do_fuse7(riscv_t *rv, const rv_insn_t *ir)
+{
+    rv->X[rv_reg_zero] = 0;
+    rv->csr_cycle += 2;
+    memory_t *m = ((state_t *) rv->userdata)->mem;
+    memcpy((char *) m->mem_base + rv->X[rv_reg_a0],
+           (char *) m->mem_base + rv->X[rv_reg_a1], rv->X[rv_reg_a2]);
+    rv->PC = rv->X[rv_reg_ra] & ~1U;
+    if (unlikely(RVOP_NO_NEXT(ir)))
+        return true;
+    const rv_insn_t *next = ir + 1;
     MUST_TAIL return next->impl(rv, next);
 }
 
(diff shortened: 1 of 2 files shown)


src/emulate.c
```diff
@@ -593,18 +597,57 @@ static void block_translate(riscv_t *rv, block_t *block)
         ir->tailcall = next_ir->tailcall;                                  \
     }
 
+bool parse_memset(riscv_t *rv)
+{
+    uint32_t memset_insn[] = {
```
jserv: Append the corresponding assembly as inlined comments for reference purpose.

jserv: Add `static const` qualifier for `memset_insn`.

src/emulate.c
```diff
@@ -593,18 +597,57 @@ static void block_translate(riscv_t *rv, block_t *block)
         ir->tailcall = next_ir->tailcall;                                  \
     }
 
+bool parse_memset(riscv_t *rv)
```
jserv: Add `static` qualifier.

src/emulate.c
```diff
@@ -593,18 +597,57 @@ static void block_translate(riscv_t *rv, block_t *block)
...
+        0x00b72623, 0x01070713, 0xfed766e3, 0x00061463, 0x00008067, 0x40c306b3,
+        0x00269693, 0x00000297, 0x005686b3, 0x00c68067, 0x00b70723, 0x00b706a3,
+        0x00b70623, 0x00b705a3, 0x00b70523, 0x00b704a3, 0x00b70423, 0x00b703a3,
+        0x00b70323, 0x00b702a3, 0x00b70223, 0x00b701a3, 0x00b70123, 0x00b700a3,
+        0x00b70023, 0x00008067, 0x0ff5f593, 0x00859693, 0x00d5e5b3, 0x01059693,
+        0x00d5e5b3, 0xf6dff06f, 0x00279693, 0x00000297, 0x005686b3, 0x00008293,
+        0xfa0680e7, 0x00028093, 0xff078793, 0x40f70733, 0x00f60633, 0xf6c378e3,
+        0xf3dff06f};
+    uint32_t tmp_pc = rv->PC;
+    for (int i = 0; i < 55; i++) {
```
jserv, line 614: It seems that both prologue and epilogue are included in the above sequence, right?

jserv: `memcmp` provided by glibc or other C runtimes is usually optimized for modern microprocessors, and it can be much faster than simple iterations. Do you have any idea to utilize `memcmp`?

jserv (changes requested):
Add memcpy substitution as well. Applications such as Doom can benefit.

src/emulate.c
```diff
@@ -593,18 +597,57 @@ static void block_translate(riscv_t *rv, block_t *block)
...
  */
-static void match_pattern(block_t *block)
+static void match_pattern(riscv_t *rv, block_t *block)
 {
     for (uint32_t i = 0; i < block->n_insn - 1; i++) {
         rv_insn_t *ir = block->ir + i, *next_ir = NULL;
         int32_t count = 0, sign = 1;
         switch (ir->opcode) {
+        case rv_insn_addi:
+            if (ir->imm == 15 && ir->rd == 6 && ir->rs1 == 0) {
```
jserv: Add more comments.

src/emulate.c
```diff
@@ -593,18 +613,255 @@ static void block_translate(riscv_t *rv, block_t *block)
         ir->tailcall = next_ir->tailcall;                                  \
     }
 
+static bool parse_memset(riscv_t *rv)
```
jserv: The function name should be `detect_memset`.

src/emulate.c
```diff
@@ -593,18 +613,255 @@ static void block_translate(riscv_t *rv, block_t *block)
...
+    for (int i = 0; i < 55; i++) {
+        const uint32_t insn = rv->io.mem_ifetch(tmp_pc);
+        if (insn != memset_insn[i])
+            return false;
+        tmp_pc += 4;
+    }
+    return true;
+}
+
+static bool parse_memcpy(riscv_t *rv)
```
jserv: Ditto. Should be `detect_memcpy`.

jserv:
If a certain region is recognized as containing a sequence of known `memset` or `memcpy` instructions, is it possible to replace the invocation of such a region? In other words, instead of replacing the region with a faster alternative, can we directly modify the instructions on the caller's side?

jserv:
Rather than identifying `memcpy` / `memset` during the execution phase, I am considering whether it might be more feasible to detect them during the decoding phase. This approach would involve examining the initial portion of the instruction sequence in the buffer and using `memcmp` to validate, resulting in faster and deterministic instruction substitution.

src/emulate.c
```diff
@@ -593,18 +613,314 @@ static void block_translate(riscv_t *rv, block_t *block)
...
+    for (int i = 0; i < 55; i++) {
+        const uint32_t insn = rv->io.mem_ifetch(tmp_pc);
+        if (insn != memset_insn[i])
+            return false;
+        tmp_pc += 4;
+    }
+    return true;
+}
+
+static bool parse_memset2(riscv_t *rv)
```
jserv: Should be `detect_memset2`. Of course, it would be better if one routine can manipulate several patterns.

src/emulate.c
```diff
@@ -593,18 +613,314 @@ static void block_translate(riscv_t *rv, block_t *block)
...
+        0x00050313, /* mv	t1,a0 */
+        0x00060a63, /* beqz	a2, 0x18 */
+        0x00b30023, /* sb	a1,0(t1) */
+        0xfff60613, /* add	a2,a2,-1 */
+        0x00130313, /* add	t1,t1,1 */
+        0xfe061ae3, /* bnez	a2, 0x8 */
+        0x00008067, /* ret */
+    };
+    uint32_t tmp_pc = rv->PC;
+    for (int i = 0; i < 7; i++) {
```
jserv: Replace the hardcoded `7` with `ARRAYS_SIZE` macro.

src/emulate.c
```diff
@@ -593,18 +613,314 @@ static void block_translate(riscv_t *rv, block_t *block)
...
+        0xfee882e3, /* beq	a7,a4,0x21520 */
+        0x0005c783, /* lbu	a5,0(a1) */
+        0x00170713, /* add	a4,a4,1 */
+        0x00158593, /* add	a1,a1,1 */
+        0xfef70fa3, /* sb	a5,-1(a4) */
+        0xfce89ee3, /* bne	a7,a4,0x2152c */
+        0xfcdff06f, /* j	0x21520 */
+    };
+    uint32_t tmp_pc = rv->PC;
+    for (int i = 0; i < 105; i++) {
```
jserv: Ditto. Utilize `ARRAYS_SIZE` macro.

src/emulate.c
```diff
@@ -35,6 +35,8 @@ extern struct target_ops gdbstub_ops;
 #include "state.h"
 #include "utils.h"
 
+#define ARRAY_SIZE(arr) (sizeof(arr) / sizeof((arr)[0]))
```
jserv: Don't do that! The macro `ARRAYS_SIZE` is already defined in `src/common.h`.

ChinYikMing:
To maintain the consistency of accessing riscv's register, I think the code below which `X[1]`, `X[10]`, `X[11]` and `X[12]` can be rewrite with `X[rv_reg_ra]`, `X[rv_reg_a0]`, `X[rv_reg_a1]` and `X[rv_reg_a2]` respectively.

```c
memset((char *) m->mem_base + rv->X[10], rv->X[11], rv->X[12]);
rv->PC = rv->X[1] & ~1U;
```

PR #197, opened 2023-08-24
Introduce sound related system calls

### LINKED ISSUE #152: Introduce sound related system calls
For the purpose of engaging in video game emulation, sound is of utmost importance. Unfortunately, there is a deficiency in the emulator's support for sound-related system calls, which could be effectively addressed through the integration of SDL2.

Reference:
* [FPGRARS](https://github.com/LeoRiether/FPGRARS): Fast Pretty Good RISC-V Assembly Rendering System

ChinYikMing:
`mus2midi` utility is referenced by [Mus2Midi](https://github.com/fawtytoo/Mus2Midi) . 

Tested in Ubuntu 20.04.

Close #152


Makefile
```diff
@@ -96,6 +101,7 @@ OBJS := \
 	decode.o \
 	io.o \
 	syscall.o \
+	mus2midi.o \
```
jserv: Append `mus2midi.o` to `OBJS_EXT` instead. That is,
```
OBJS_EXT += mus2midi.o syscall_sdl.o
```

src/syscall_sdl.c
```diff
@@ -276,3 +337,162 @@ void syscall_submit_queue(riscv_t *rv)
...
+}
+
+void syscall_shutdown_sound(riscv_t *rv)
+{
+    syscall_stop_music(rv);
+    Mix_HaltChannel(-1);
+    Mix_CloseAudio();
+    Mix_Quit();
+
+    if (mid_data)
```
jserv: Simply rewrite as following:
```c
free(mid_data);
```

Quote from [free(3)](https://man7.org/linux/man-pages/man3/free.3p.html):
> If ptr is a null pointer, no action shall occur.

src/syscall_sdl.c
```diff
@@ -276,3 +337,162 @@ void syscall_submit_queue(riscv_t *rv)
...
+    /*
+     * hack of offset as DOOM is 32-bit but
+     * Emulator is 64-bit when accessing the musicinfo_t
+     */
+    uint32_t m_data_offset = *(uint32_t *) ((uint8_t *) &m + 8);
+    uint32_t m_data_size = *(uint32_t *) ((uint8_t *) &m + 12);
+    uint8_t m_data[MUSIC_MAX_SIZE];
+    memory_read(s->mem, m_data, m_data_offset, m_data_size);
+
+    sound_t music;
```
jserv: Use C99 [designated initializers](https://gcc.gnu.org/onlinedocs/gcc/Designated-Inits.html) instead.

src/mus2midi.h
```diff
@@ -0,0 +1,13 @@
+/*
+ * rv32emu is freely redistributable under the MIT License. See the file
+ * "LICENSE" for information on usage and redistribution of this file.
+ */
+
+#ifndef __MUS2MIDI_H__
```
jserv: It is not worthy to have a header containing only one function prototype. Declare in `syscall_sdl.c` directly.

src/mus2midi.h
```diff
@@ -0,0 +1,13 @@
...
+#define __MUS2MIDI_H__
+
+typedef unsigned char BYTE;
```
jserv: Avoid pre-C99 stuff such as `BYTE`. Always use types such as `uint8_t`.

src/syscall_sdl.c
```diff
@@ -276,3 +337,162 @@ void syscall_submit_queue(riscv_t *rv)
         }
     }
 }
+
+void *sfx_handler(void *arg)
```
jserv: Add `static` qualifier.

src/syscall_sdl.c
```diff
@@ -276,3 +337,162 @@ void syscall_submit_queue(riscv_t *rv)
...
+
+    /* multiplied by 8 because sfx->volume's max is 15 */
+    Mix_Volume(chan, sfx->volume * 8);
+
+    Mix_Playing(chan);
+
+    return NULL;
+}
+
+void *music_handler(void *arg)
```
jserv: Ditto. `static` qualifer.

src/mus2midi.c
```diff
@@ -0,0 +1,206 @@
...
+ *
+ */
+
+#include <stddef.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "mus2midi.h"
+
+typedef unsigned short WORD;
```
jserv: Utilize the types defined in `<stdint.h>`.

src/mus2midi.c
```diff
@@ -0,0 +1,206 @@
...
+
+typedef struct {
+    char id[4];
+    WORD scoreLen;
+    WORD scoreStart;
+} __attribute__((packed)) HDR_MUS;
```
jserv: Check [CONTRIBUTING.md](https://github.com/sysprog21/rv32emu/blob/master/CONTRIBUTING.md):
> Use [snakecase](https://en.wikipedia.org/wiki/Snake_case). Do not use "camelcase".

src/mus2midi.c
```diff
@@ -0,0 +1,206 @@
...
+    WORD ticks;
+} __attribute__((packed)) HDR_MID;
+
+char magicMus[4] = {'M', 'U', 'S', 0x1a};
+char magicMid[4] = {'M', 'T', 'h', 'd'};
+char magicTrk[4] = {'M', 'T', 'r', 'k'};
+BYTE magicEOT[4] = {0x00, 0xff, 0x2f, 0x00};
+
+int controllerMap[16] = {-1, 0,  1,   7,   10,  11,  91,  93,
+                         64, 67, 120, 123, 126, 127, 121, -1};
```
jserv: Add an extra comma before `}`, so that clang-format can behave better. i.e.,
```
    ..., 27, 121, -1,
};
```

src/mus2midi.c
```diff
@@ -0,0 +1,206 @@
...
+
+typedef struct {
+    char id[4];
+    int length;
+    WORD type;
+    WORD ntracks;
+    WORD ticks;
+} __attribute__((packed)) HDR_MID;
+
+char magicMus[4] = {'M', 'U', 'S', 0x1a};
```
jserv: Hide the internal symbols as possible as you can. That is, add `static` qualifers.

src/mus2midi.c
```diff
@@ -0,0 +1,206 @@
...
+BYTE *midData;
+int midSize;
+
+BYTE *musPos;
+int musEOT;
+
+BYTE deltaBytes[4];
+int deltaCount;
+
+// maintain a list of channel volume
```
jserv: Use C-style comments.

jserv (changes requested):
Update `docs/syscall.md` and provide a test program in `tests` directory.

ChinYikMing:
> Provide a test program in `tests` directory.

Sound version of `doom.elf` is come with the commit, simply run `make doom` in `sound` branch

src/syscall.c
```diff
@@ -18,21 +18,27 @@
...
+    _(read,                 63)      \
+    _(write,                64)      \
+    _(fstat,                80)      \
+    _(exit,                 93)      \
+    _(gettimeofday,         169)     \
+    _(brk,                  214)     \
+    _(clock_gettime,        403)     \
+    _(open,                 1024)    \
+    IIF(RV32_HAS(SDL))(              \
+        _(play_sfx,         0xBABE)  \
```
jserv: You proposed 6 system calls for audio/sound. i.e.,
* play_sfx
* play_music
* stop_music
* set_music_volume
* setup_sound
* shutdown_sound

Can you use fewer ones? We may discuss on #152 before code reviewing.

src/syscall_sdl.c
```diff
@@ -276,3 +337,162 @@ void syscall_submit_queue(riscv_t *rv)
...
+    int volume = rv_get_reg(rv, rv_reg_a2);
+
+    musicinfo_t m;
+    memory_read(s->mem, (uint8_t *) &m, musicinfo_addr, sizeof(musicinfo_t));
+
+    /*
+     * hack of offset as DOOM is 32-bit but
+     * Emulator is 64-bit when accessing the musicinfo_t
+     */
+    uint32_t m_data_offset = *(uint32_t *) ((uint8_t *) &m + 8);
```
jserv: The prefix `m_` is useless. Older compilers and development environments were stupid, and old programmers manually made use of such hacky prefix. Now, drop them.

jserv:
> Sound version of `doom.elf` is come with the commit, simply run `make doom` in `sound` branch

No, let's minimize the necessary changes. You should provide a minimal C program that demonstrates the use of the proposed system calls instead of making extensive modifications to Doom. Once this pull request is merged, we can proceed with the Doom and Quake pull requests in their respective repositories.

jserv:
[Simple SDL2 Audio](https://github.com/jakebesworth/Simple-SDL2-Audio) operates independently of SDL2 Mixer and is designed to play a single music file at a time. It also supports an unlimited number of sounds, achieving audio mixing natively without the need for an external mixer. It might be useful.

ChinYikMing:
> Update `docs/syscall.md` and provide a test program in `tests` directory.

You may test the result of integration using the test program in `tests/sound` directory, use `=h` option to check the usage. 

To test all at once, simply run `../../build/rv32emu sound.elf =m d_e1m1.mus =s dspistol =upmv =srep` in `tests/sound` directory.

README.md
```diff
@@ -35,9 +35,10 @@ Features:
 ## Build and Verify
 
 `rv32emu` relies on certain third-party packages for full functionality and access to all its features.
-To ensure proper operation, the target system should have the [SDL2 library](https://www.libsdl.org/) installed.
-* macOS: `brew install sdl2`
-* Ubuntu Linux / Debian: `sudo apt install libsdl2-dev`
+To ensure proper operation, the target system should have the [SDL2 library](https://www.libsdl.org/) 
+and [SDL2_Mixer library](https://wiki.libsdl.org/SDL2_mixer/FrontPage) installed.
```
jserv: Use shorter URL. i.e., `https://wiki.libsdl.org/SDL2_mixer`

jserv (changes requested):

docs/syscall.md
```diff
@@ -95,9 +112,9 @@ Currently only a couple system calls are supported via semihosting, which enable
 
 Any other system calls will fail with an "unknown syscall" error.
 
-## Experimental Display and Event System Calls
+## Experimental Display, Event, and Sound System Calls
 
-These system calls are solely for the convenience of accessing the [SDL library](https://www.libsdl.org/) and are only intended for the presentation of RISC-V graphics applications. They are not present in the ABI interface of POSIX or Linux.
+These system calls are solely for the convenience of accessing the [SDL library](https://www.libsdl.org/) and [SDL2_Mixer](https://wiki.libsdl.org/SDL2_mixer/FrontPage) and are only intended for the presentation of RISC-V graphics applications. They are not present in the ABI interface of POSIX or Linux.
```
jserv: Ditto. Always use shorter URL.

docs/syscall.md
```diff
@@ -138,3 +155,31 @@ To inform the emulator that a batch of submissions should be processed, the appl
 The submission entry is structured similarly to an event entry, with a 32-bit type field and an associated dynamic-sized value buffer whose width depends on the type of submission.
 
 * `RELATIVE_MODE_SUBMISSION`: Enable or disable the mouse relative mode. If the mouse relative mode is enabled, the mouse cursor is wrapped within the window border, it's associated with an 8-bit wide boolean value that indicates whether the relative mouse mode should be enbled.
+
+### `control_sound` - control the behavior of music and sound effect(sfx)
```
jserv: Follow the consistent manner described above.

src/mus2midi.c
```diff
@@ -0,0 +1,224 @@
...
+#include <stddef.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+
+typedef struct {
+    char id[4];
+    uint16_t score_len;
+    uint16_t score_start;
+} __attribute__((packed)) hdr_mus_t;
```
jserv: Why not `music_header_t`?

src/mus2midi.c
```diff
@@ -0,0 +1,224 @@
...
+
+typedef struct {
+    char id[4];
+    int length;
+    uint16_t type;
+    uint16_t ntracks;
+    uint16_t ticks;
+} __attribute__((packed)) hdr_mid_t;
```
jserv: Why no `midi_header_t`?

src/mus2midi.c
```diff
@@ -0,0 +1,224 @@
...
+
+static char magic_mus[4] = {
```
jserv: It is a bit confusing to specify abbreviations such as `mus`, `mid`, `trk`, etc. Use the fully qualified ones.

src/mus2midi.c
```diff
@@ -0,0 +1,224 @@
...
+    0xff,
+    0x2f,
+    0x00,
+};
+
+static int controller_map[16] = {
+    -1, 0, 1, 7, 10, 11, 91, 93, 64, 67, 120, 123, 126, 127, 121, -1,
+};
+
+uint8_t *_mid_data;
```
jserv: Declare the internal symbols as `static`.

src/syscall.c
```diff
@@ -18,21 +18,23 @@
...
+    _(read,                 63)      \
+    _(write,                64)      \
+    _(fstat,                80)      \
+    _(exit,                 93)      \
+    _(gettimeofday,         169)     \
+    _(brk,                  214)     \
+    _(clock_gettime,        403)     \
+    _(open,                 1024)    \
+    IIF(RV32_HAS(SDL))(              \
+        _(control_sound,    0xBABE)  \
```
jserv: Move the sound specific ones to the end.

docs/syscall.md
jserv: Don't mix sound and audio. Instead, distinguish them properly.

src/mus2midi.c
ChinYikMing: `mus` is the type of file format before conversion so I think stick with `mus_header_t` is OK

src/mus2midi.c
ChinYikMing: `mus` is the type of file format before conversion so I think stick with `mus` is OK but I will rename `mid` to `midi` and `trk` to `track` instead

docs/syscall.md
```diff
@@ -155,3 +155,39 @@ To inform the emulator that a batch of submissions should be processed, the appl
 The submission entry is structured similarly to an event entry, with a 32-bit type field and an associated dynamic-sized value buffer whose width depends on the type of submission.
 
 * `RELATIVE_MODE_SUBMISSION`: Enable or disable the mouse relative mode. If the mouse relative mode is enabled, the mouse cursor is wrapped within the window border, it's associated with an 8-bit wide boolean value that indicates whether the relative mouse mode should be enbled.
+
+### `control_sound` - control the behavior of music
```
jserv: The description is confusing. Don't mix the term sound and music.

src/mus2midi.c
jserv: Enforce `static const` qualifiers to objects with prefix `magic_`.

src/mus2midi.c
```diff
@@ -0,0 +1,224 @@
...
+static uint8_t *mus_pos;
+static int mus_end_of_track;
+
+static uint8_t delta_bytes[4];
+static int delta_cnt;
+
+/* maintain a list of channel volume */
+static uint8_t mus_channel[16];
+
+static void convert()
```
jserv: Rewrite as `static void convert(void)`.

jserv (changes requested):
Move the content of `src/mus2midi.c` to `src/syscall_sdl.c` since the former is short and dedicated to the latter.

docs/syscall.md
```diff
@@ -155,3 +155,39 @@ To inform the emulator that a batch of submissions should be processed, the appl
...
+
+**synopsis**: `void control_sound(int sound_request)`
+
+To request the emulator to do some sound actions, the application should prepare the sound data then pass the address of the sound data to the registers(which register depends on the request type). The request will be processed as soon as possible. There are four types of request:
+* `PLAY_MUSIC`: Play the music. If there is a music is playing, it will be replaced by new music. The music is played by a new thread to prevent blocking on main thread.
+* `STOP_MUSIC`: Halt the music.
+* `SET_MUSIC_VOLUME`: Update the music volume at some point if needed.
+
+#### Music
+Music data is defined in a structure called `musicinfo_t` which is from Doom application. The main data in the structure is `data` and `size`. Emulator use `data` and `size` to play music. Other fields are not used by now so the structure should be more generic to support other applications.
```
jserv: What about Quake porting?

docs/syscall.md
```diff
@@ -155,3 +155,39 @@ To inform the emulator that a batch of submissions should be processed, the appl
...
+
+**system call number**: `0xBABE`
+
+**synopsis**: `void control_sound(int sound_request)`
+
+The request is similar to the control of music which described earlier and it support one type of request:
+* `PLAY_SFX`: Play the sfx, note that too many sfx at the same time could cause channel to run out and the sfx cannot be played. The sfx is played by a new thread and the reason is same as `PLAY_MUSIC`
+
+#### Sound Effect(sfx)
+Sfx data is defined in a structure called `sfxinfo_t` which is also from Doom application. The main data in the structure is `data`, `size`, and `volume`. Emulator use `data`, `size`, and `volume` to play sfx. Other fields are not used by now so the structure should be more generic to support other applications.
```
jserv: Improve the English writing.

jserv: [QuillBot](https://quillbot.com/) is your friend.

src/syscall.c
```diff
@@ -18,21 +18,23 @@
...
+    _(read,                 63)      \
+    _(write,                64)      \
+    _(fstat,                80)      \
+    _(exit,                 93)      \
+    _(gettimeofday,         169)     \
+    _(brk,                  214)     \
+    _(clock_gettime,        403)     \
+    _(open,                 1024)    \
+    IIF(RV32_HAS(SDL))(              \
+        _(setup_audio,      0xD00D)  \
```
jserv: Again, don't mix audio and sound in mnemonic identifiers. `setup_audio` and `control_audio` should appear at the end of syscall listing.

docs/syscall.md
ChinYikMing: I introduce an ABI such that first two fields of structure  `musicinfo_t` and `sfxinfo_t` in any applications must be sound data and follows the sound data size, such that the sound system of emulator can compatible to any applications including Quake. A simple example is `tests/sound/sounds.c`.

 i.e.,
```c
typedef stuct {
    void *data;
    int size;
    ...
} musicinfo_t;
```

Simply ignore other fields of the structure for now.

ChinYikMing:
> Move the content of `src/mus2midi.c` to `src/syscall_sdl.c` since the former is short and dedicated to the latter.

tests/sound/sound.c requires `src/mus2midi.c`, I think we should leave it to be.

src/syscall.c
```diff
@@ -18,21 +18,23 @@
...
+    _(exit,                 93)      \
+    _(gettimeofday,         169)     \
+    _(brk,                  214)     \
+    _(clock_gettime,        403)     \
+    _(open,                 1024)    \
+    IIF(RV32_HAS(SDL))(              \
+        _(draw_frame,       0xBEEF)  \
+        _(setup_queue,      0xC0DE)  \
+        _(submit_queue,     0xFEED)  \
+        _(setup_audio,      0xD00D)  \
```
jserv: We may swap the magic numbers.
The term 'dood' is an informal and playful variation of the word "dude," which is a slang term that originated in American English. "Dude" is often used to refer to a person, typically a man or a friend.

The setup of audio is called "baby," and then the control of audio is called "dood," which sounds reasonable.

tests/sound/sound.c
```diff
@@ -0,0 +1,278 @@
...
+    SET_MUSIC_VOLUME,
+    STOP_MUSIC,
+};
+
+typedef struct {
+    void *data;
+    int size;
+} musicinfo_t;
+
+typedef struct sfxinfo_struct {
```
jserv: Simply remove `sfxinfo_struct` as an anonymous structure.

jserv: To avoid code duplication, let's move the structure definitions and the function prototype to the header, `src/mus2midi.h`.

src/mus2midi.c
```diff
@@ -0,0 +1,223 @@
+/*
+ * rv32emu is freely redistributable under the MIT License. See the file
+ * "LICENSE" for information on usage and redistribution of this file.
+ */
+
+/*
+ * Reference: https://github.com/fawtytoo/Mus2Midi
```
jserv: Add more comments. Take the example below:
> This is a basic MUS to MIDI converter designed for applications such as DOOM.

jserv: Of course, it would be better to mention Quake adoption.

jserv: Remove the hyperlink of Mus2Midi and put the license notice below:
```c
/* Copyright (C) 2021-2022 Steve Clark
 * 
 * This software is provided 'as-is', without any express or implied
 * warranty.  In no event will the authors be held liable for any damages
 * arising from the use of this software.
 *
 * Permission is granted to anyone to use this software for any purpose,
 * including commercial applications, and to alter it and redistribute it
 * freely, subject to the following restrictions:
 *
 * 1. The origin of this software must not be misrepresented; you must not
 *  claim that you wrote the original software. If you use this software
 *  in a product, an acknowledgment in the product documentation would 
 *  be
 *  appreciated but is not required. 
 * 2. Altered source versions must be plainly marked as such, and must not be
 *  misrepresented as being the original software.
 * 3. This notice may not be removed or altered from any source distribution.
*/
```

jserv: See https://github.com/fawtytoo/Mus2Midi/blob/main/Licence

src/mus2midi.c
```diff
@@ -0,0 +1,223 @@
...
+static uint8_t *mus_pos;
+static int mus_end_of_track;
+
+static uint8_t delta_bytes[4];
+static int delta_cnt;
+
+/* maintain a list of channel volume */
+static uint8_t mus_channel[16];
+
+static void convert(void)
```
jserv: Attempt to add comments.

src/mus2midi.c
```diff
@@ -0,0 +1,223 @@
...
+    }
+}
+
+uint8_t *mus2midi(uint8_t *data, int *length)
+{
+    mus_header_t *mus_hdr = (mus_header_t *) data;
+    midi_header_t midi_hdr;
+    uint8_t *mid_track_len;
+    int track_len;
+    int i;
```
jserv: Reduce the scope of `i`. i.e., move it into the `for` iteration below.

src/mus2midi.c
jserv: `rv32emu` is licensed under an MIT-style license, which is compatible with the zlib license used by the `MusMidi` project. However, since the license specifies that "This notice may not be removed or altered from any source distribution," we are obligated to comply.

src/mus2midi.c
```diff
@@ -0,0 +1,223 @@
...
+
+    if (strncmp(mus_hdr->id, magic_mus, 4) != 0)
```
jserv: Rewrite as following:
```c
if (!strncmp(mus_hdr->id, magic_mus, 4))
```

src/syscall_sdl.c
```diff
@@ -7,13 +7,76 @@
...
 #include "state.h"
 
+extern uint8_t *mus2midi(uint8_t *, int *);
+
+/*
+ * In DOOM1.WAD only DSITMBK sound effect use sample rate 22050
+ * but normally only head in multiplayer
+ * since we are single player, stick with 11025 is OK
+ *
+ * In QUAKE, most sound effects with sample rate 11025
```
jserv: Always refer it as [Quake](https://en.wikipedia.org/wiki/Quake_(video_game)).

src/syscall_sdl.c
```diff
@@ -276,3 +339,222 @@ void syscall_submit_queue(riscv_t *rv)
...
+
+    return NULL;
+}
+
+static void *music_handler(void *arg)
+{
+    sound_t *music = (sound_t *) arg;
+    int looping = music->looping ? -1 : 1;
+
+    if (mid_data)
```
jserv: Simply rewrite as following:
```c
    free(mid_data).
```

src/syscall_sdl.c
```diff
@@ -7,13 +7,76 @@
...
+typedef struct sound {
+    uint8_t *data;
+    size_t size;
+    int looping;
+    int volume;
+} sound_t;
+
+/* SDL-mixer-related and music-related variables */
+static pthread_t music_thr;
+static uint8_t *mid_data;
```
jserv: Rename it to `midi_data`. The original term `mid_data` might lead to confusion, as it could be interpreted as either "middle data" or "data in the middle."

src/syscall_sdl.c
```diff
@@ -7,13 +7,76 @@
...
+
+typedef struct sound {
+    uint8_t *data;
+    size_t size;
+    int looping;
+    int volume;
+} sound_t;
+
+/* SDL-mixer-related and music-related variables */
+static pthread_t music_thr;
```
jserv: Rename it to `music_thread`.

src/syscall_sdl.c
```diff
@@ -7,13 +7,76 @@
...
+static uint8_t *mid_data;
+static Mix_Music *mid;
+
+/* SDL-mixer-related and sfx-related variables */
+static pthread_t sfx_thr;
```
jserv: Ditto. Use formal term "thread" rather than `thr`.

src/syscall_sdl.c
```diff
@@ -7,13 +7,76 @@
...
+static Mix_Chunk *sfx_chunk;
+static uint8_t *sfx_samples;
+static uint32_t sfx_sample_nums;
```
jserv: Rename to `nr_sfx_samples`, following Linux kernel naming scheme.
See https://stackoverflow.com/questions/8306202/what-does-nr-stand-for-in-system-call-number-that-is-usually-used-as-suffix

src/syscall_sdl.c
```diff
@@ -276,3 +339,222 @@ void syscall_submit_queue(riscv_t *rv)
...
+void syscall_setup_audio(riscv_t *rv)
+{
+    /* setup_audio(request) */
+    const int request = rv_get_reg(rv, rv_reg_a0);
+
+    switch (request) {
+    case INIT_AUDIO:
+        init_audio();
+        break;
+    case SHUTDOWN_AUDIO:
```
jserv, line 763: I am unsure whether it is meaningful to offer operations like initialization and shutdown for users, considering that the emulator manages the actual lifecycle of the audio instance. Consider the framebuffer management system call, `draw_frame`: the existence of a single system call is due to the fact that we only enable users to update the available screen, without providing functionality for initializing or closing the window.

tests/sound/sound.c
```diff
@@ -0,0 +1,278 @@
...
+static void load_sfx()
+{
+    sfx = malloc(sizeof(sfxinfo_t));
+    assert(sfx != NULL);
+    sfx->data = malloc(sizeof(uint8_t) * SFX_MAX_SIZE);
+    assert(sfx->data != NULL);
+    sfx->size = 5669; /* FIXME: hardcoded size */
+
+    FILE *_sfx = fopen(sfx_src, "rb");
+    assert(_sfx != NULL);
```
jserv: Rewrite as `assert(_sfx)`.

tests/sound/sound.c
```diff
@@ -0,0 +1,278 @@
...
+ */
+static void load_sfx()
+{
+    sfx = malloc(sizeof(sfxinfo_t));
+    assert(sfx != NULL);
+    sfx->data = malloc(sizeof(uint8_t) * SFX_MAX_SIZE);
+    assert(sfx->data != NULL);
+    sfx->size = 5669; /* FIXME: hardcoded size */
+
+    FILE *_sfx = fopen(sfx_src, "rb");
```
jserv: Avoid the naming scheme starting with underscore.

Here is what the C standard states (section 7.1.3):
* All identifiers that start with an underscore followed by an uppercase letter or another underscore are always reserved for any purpose.
* All identifiers that start with an underscore are always reserved to be used as identifiers with file scope in both the ordinary and tag name spaces.

(The section proceeds to enumerate specific identifiers and groups of identifiers that are reserved by certain standard headers.)

The implication is that, for example, the implementation (whether it's the compiler or a standard header) can utilize the name `__FOO` for any purpose it deems fit.

In light of this, it's best to abstain from defining any such identifiers in your own code (unless your code is part of a C implementation -- and if you are uncertain, it likely is not). There's no real necessity to define these identifiers anyway; there is an abundance of unreserved identifiers at your disposal.

You can employ an identifier like `_foo` as long as it is confined to a local definition (not at file scope). However, I personally find it much simpler to avoid the use of leading underscores entirely.

tests/sound/sound.c
```diff
@@ -0,0 +1,278 @@
...
+{
+    free(sfx->data);
+    free(sfx);
+}
+
+/* when emulating the Doom, music is in lump */
+static void load_music()
+{
+    music = malloc(sizeof(musicinfo_t));
+    assert(music != NULL);
```
jserv: Ditto. Always write shorter and simpler code.

Makefile
```diff
@@ -53,12 +53,17 @@ ifeq (, $(shell which sdl2-config))
...
+override ENABLE_SDL := 0
+endif
 endif
 $(call set-feature, SDL)
 ifeq ($(call has, SDL), 1)
-OBJS_EXT += syscall_sdl.o
+OBJS_EXT += mus2midi.o syscall_sdl.o
 $(OUT)/syscall_sdl.o: CFLAGS += $(shell sdl2-config --cflags)
 LDFLAGS += $(shell sdl2-config --libs)
+LDFLAGS += -lSDL2_mixer
```
jserv: Rewrite as following:
```
LDFLAGS += $(shell pkg-config --libs SDL2_mixer)
```

Makefile
```diff
@@ -53,12 +53,17 @@ ifeq (, $(shell which sdl2-config))
 $(warning No sdl2-config in $$PATH. Check SDL2 installation in advance)
 override ENABLE_SDL := 0
 endif
+ifeq (0, $(shell pkg-config --exists SDL2_mixer && echo 1 || echo 0))
```
jserv: Replace this line with the following:
```
ifeq (1, $(shell pkg-config --exists SDL2_mixer; echo $$?))
```

ChinYikMing:
> > Move the content of `src/mus2midi.c` to `src/syscall_sdl.c` since the former is short and dedicated to the latter.
> 
> tests/sound/sound.c requires `src/mus2midi.c`, I think we should leave it to be.

I made  a mistake which `mus2midi` is only used in emulator since it control the sound, I will adopt the required change.

src/syscall_sdl.c
ChinYikMing: I rename it as `music_midi_data` to avoid naming collision with `midi_data` that used by `mus2mid` converter since I have moved the `src/mus2midi.c` to `src/syscall_sdl.c`

src/syscall_sdl.c
ChinYikMing: Most games like Doom and Quake, they will maintain their own sound system or server so there should be a pair of initialization and destroy semantic. In order to keep the semantic clear, I think we should provide operations like initialization and shutdown for users. Otherwise, the emulator will keep checking if the audio is enabled or not when to play the music or sound effects.

jserv: Document the above considerations in `docs/` directory.

src/syscall_sdl.c
```diff
@@ -7,13 +7,311 @@
...
 #include <stdint.h>
 #include <stdio.h>
+#include <unistd.h>
 
 #include <SDL.h>
+#include <SDL_mixer.h>
 
 #include "state.h"
 
+/* Copyright (C) 2021-2022 Steve Clark
```
jserv: s/Copyright/Portions Copyright/

src/syscall_sdl.c
```diff
@@ -7,13 +7,311 @@
...
+ * warranty.  In no event will the authors be held liable for any damages
+ * arising from the use of this software.
+ *
+ * Permission is granted to anyone to use this software for any purpose,
+ * including commercial applications, and to alter it and redistribute it
+ * freely, subject to the following restrictions:
+ *
+ * 1. The origin of this software must not be misrepresented; you must not
+ *  claim that you wrote the original software. If you use this software
+ *  in a product, an acknowledgment in the product documentation would
```
jserv: Indent the notice.

src/syscall_sdl.c
jserv: Rearrange these newly-added lines immediately preceding the function `sfx_handler` for enhanced readability.

jserv (changes requested):
Improve git commit messages.

src/syscall_sdl.c
```diff
@@ -276,3 +337,457 @@ void syscall_submit_queue(riscv_t *rv)
         }
     }
 }
+
+/* Copyright (C) 2021-2022 Steve Clark
```
jserv: Replace "Copyright" with "Portions Copyright"

src/syscall_sdl.c
```diff
@@ -276,3 +337,457 @@ void syscall_submit_queue(riscv_t *rv)
...
+ *    be appreciated but is not required.
+ * 2. Altered source versions must be plainly marked as such, and must not be
+ *    misrepresented as being the original software.
+ * 3. This notice may not be removed or altered from any source distribution.
+ */
+
+/*
+ * This is a basic MUS to MIDI converter designed for applications such as DOOM.
+ *
+ * For Quake, it only have sound effects which are in WAV format so sfx_handler
```
jserv: Improve the English writing.

docs/syscall.md
```diff
@@ -155,3 +155,41 @@ To inform the emulator that a batch of submissions should be processed, the appl
 The submission entry is structured similarly to an event entry, with a 32-bit type field and an associated dynamic-sized value buffer whose width depends on the type of submission.
 
 * `RELATIVE_MODE_SUBMISSION`: Enable or disable the mouse relative mode. If the mouse relative mode is enabled, the mouse cursor is wrapped within the window border, it's associated with an 8-bit wide boolean value that indicates whether the relative mouse mode should be enbled.
+
+### `control_audio` - control the behavior of music
```
jserv: Refine the following documentation.

ChinYikMing: My bad, I have swap `D00D` and `BABE`

src/syscall_sdl.c
```diff
@@ -276,3 +337,458 @@ void syscall_submit_queue(riscv_t *rv)
...
+ *
+ * Sfx_handler can handle Quake's sound effects because they are all in WAV
+ * format.
+ */
+
+typedef struct {
+    char id[4];
+    uint16_t score_len;
+    uint16_t score_start;
+} __attribute__((packed)) mus_header_t;
```
jserv: Wrap `__attribute__((packed))` via the macro defined in `src/common.h. Consider Win32/MSVC compatibility.

docs/syscall.md
```diff
@@ -155,3 +155,41 @@ To inform the emulator that a batch of submissions should be processed, the appl
...
+
+**system call number**: `0xBABE`
```
jserv: Check the magic number.

docs/syscall.md
```diff
@@ -155,3 +155,41 @@ To inform the emulator that a batch of submissions should be processed, the appl
...
+
+The application must prepare the sound data and then give the address of the sound data to register `a1` in order to ask the emulator to perform some sound operations. The request will be processed as soon as possible. Three different sorts of requests exist:
+* `PLAY_MUSIC`: Play the music. If any music is currently playing, it will be changed to something new. In order to avoid blocking on the main thread, the music is played by a new thread.
+* `STOP_MUSIC`: Halt the music.
+* `SET_MUSIC_VOLUME`: If necessary, adjust the music level at some point.
+
+#### Music
+Music data is defined in a structure called `musicinfo_t`. The SDL2_mixer library is used by the emulator to play music using the fields `data` and `size` in the structure.
+
+### `control_audio` - control the behavior of sound effect(sfx)
```
jserv: Merge the paragraph into the previous one.

src/syscall_sdl.c
```diff
@@ -276,3 +337,460 @@ void syscall_submit_queue(riscv_t *rv)
...
+
+/*
+ * This is a straightforward MUS to MIDI converter made for programs
+ * like DOOM that use MIDI to store sound.
+ *
+ * Sfx_handler can handle Quake's sound effects because they are all in WAV
+ * format.
+ */
+
+__PACKED(typedef struct {
```
jserv: Rewrite as following:
```c
typedef __PACKED(struct {
    char id[4];
    uint16_t score_len;
    uint16_t score_start;
}) mus_header_t;
```

src/syscall_sdl.c
```diff
@@ -276,3 +337,460 @@ void syscall_submit_queue(riscv_t *rv)
...
+        if (SDL_Init(SDL_INIT_AUDIO) != 0) {
+            fprintf(stderr, "Failed to call SDL_Init()\n");
+            exit(1);
+        }
+    }
+
+    /* sfx samples buffer */
+    sfx_samples = malloc(SFX_SAMPLE_SIZE);
+
+    // Initialize SDL2 Mixer
```
jserv: Enforce C-style comments.

tests/sound/sound.c
```diff
@@ -0,0 +1,278 @@
...
+                if (has_sfx_repeat)
+                    play_sfx();
+            }
+        }
+    }
+}
+
+static void do_play_sound()
+{
+    if (has_music && has_sfx) {
```
jserv: Don't do this! Employ bitwise tricks. See https://github.com/jserv/ir/blob/dev/tools/ir.c
e.g.,
```c
    if (dump & IR_DUMP_DUMP) {
        ir_dump(ctx, f);
    }
```

src/syscall_sdl.c
ChinYikMing: I did this before but I think it will not work in MSVC

src/common.h
```diff
@@ -24,10 +24,30 @@
 #define __ALIGNED(x) __attribute__((aligned(x)))
 #elif defined(_MSC_VER)
 #define __ALIGNED(x) __declspec(align(x))
-#else /* unspported compilers */
+#else /* unsupported compilers */
 #define __ALIGNED(x)
 #endif
 
+/* Packed macro */
+#if defined(__GNUC__) || defined(__clang__)
```
jserv, line 32: Rewrite as following:
```c
#define PACKED_STRUCT(name) struct __attribute__((packed)) name
#elif defined(_MSC_VER)
#define PACKED_STRUCT(name) \
    __pragma(pack(push, 1)) struct name __pragma(pack(pop))
#endif
```

Then this can be used like so:
```c
typedef PACKED_STRUCT(/*anonymous*/) { short a; int b } my_struct_t;
```

src/syscall_sdl.c
jserv: Rework with macro `PACKED_STRUCT` mentioned earlier.

ChinYikMing: My mistake, the rewrite suggestion previously is the one really work.

```c
struct x {
  char;
  double;
  short;
  char;
};
```
print result of  offset of the structure `x` compiled using MSVC:

no_pack_struct:
char = 0
double = 8
short = 16
char = 18

pack_struct:
char = 0
double = 1
short = 9
char = 11

tests/sound/sound.c
```diff
@@ -0,0 +1,281 @@
...
+musicinfo_t *music;
+
+static char *sfx_src;
+static char *music_src;
+static int music_delay = 30;
+static int music_looping = 0; /* FIXME: parsing looping CLI argument */
+static int music_volume = 3;
+static int sfx_volume = 15;
+
+#define MUSIC 1
```
jserv: Use `enum` instead of macros.

tests/sound/sound.c
```diff
@@ -0,0 +1,281 @@
...
+                if (flag & SFX)
+                    play_sfx();
+            }
+        }
+    }
+}
+
+static void do_play_sound()
+{
+    if (flag & MUSIC && flag & SFX) {
```
jserv: Can you rework as following?
```c
    if (flag & MUSIC)
        load_music();
    if (flag & SFX)
        load_sfx();
    ...

```

jserv:

Issue #142, opened 2023-06-05
Utilize dominators for constructing extended basic blocks


jserv:
Quoted from [Basic Blocks and CFG](http://www.cse.iitm.ac.in/~krishna/courses/2012/odd-cs3300/pm-lecture1.pdf), the definition of extended basic block (EBB):
* Extended basic block a maximal sequence of instructions beginning with a leader that contains no join nodes other than its first node.
* Has a single entry, but possible multiple exit points.
* Some optimizations are more effective on extended basic blocks.

<img width="283" src="https://github.com/sysprog21/rv32emu/assets/478921/ba5cbdbc-a101-43f0-981e-10812b8211f1">

We can identify loops by using [dominators](https://en.wikipedia.org/wiki/Dominator_(graph_theory)):
* a node A in the flowgraph dominates a node B if every path from entry node to B includes A.
* This relations is antisymmetric, reflexive, and transitive.

<img width="325" src="https://github.com/sysprog21/rv32emu/assets/478921/90510206-819a-4529-b389-982a4eadf985">

back edge: An edge in the flow graph, whose head dominates its tail (example - edge from B6 to B4).

A loop consists of all nodes dominated by its entry node (head of the back edge) and having exactly one back edge in it.

[Intercept](https://github.com/LensPlaysGames/Intercept) contains an effective [dominator](https://en.wikipedia.org/wiki/Dominator_(graph_theory)) implementation. See
* [src/ir/dom.c](https://github.com/LensPlaysGames/Intercept/blob/main/src/ir/dom.c)
* [src/codegen/opt/opt.c](https://github.com/LensPlaysGames/Intercept/blob/main/src/codegen/opt/opt.c)

Usage:
```c
void codegen_optimise(CodegenContext *ctx) {
  opt_inline_global_vars(ctx);
  opt_analyse_functions(ctx);

  /// Optimise each function individually.
  do {
    foreach_ptr (IRFunction*, f, ctx->functions) {
      if (f->is_extern) continue;

      DominatorInfo dom = {0};
      do {
        build_dominator_tree(f, &dom, true);
        opt_reorder_blocks(f, &dom);
      } while (
          opt_const_folding_and_strengh_reduction(f) ||
          opt_dce(f) ||
          opt_mem2reg(f) ||
          opt_jump_threading(f, &dom) ||
          opt_tail_call_elim(f)
      );
      free_dominator_info(&dom);
    }
  }

  /// Cross-function optimisations.
[...]

PR #201, opened 2023-08-30
Avoid recursive jump for better loop detection

qwe661234:
Previously, we employed recursive jump translation to implement an extended basic block. Nevertheless, this approach makes it challenging to detect loop paths since we position the loop's entry block inside the block, and its using frequency would not be updated.

Based on this observation, it becomes necessary to eliminate the recursive jump translation. By doing so, we can accurately update the using frequency of the loop's entry block.

As shown in the performance results below, we gain 4% performance improvement when running coreMark and 37% when running SciMark2, but we lost 1% performance when running dhrysone.

* Intel Core i7-11700

|   Metric   | origin  | proposed |Speedup|
|--------- |--------- |--------- |--------- |
| CoreMark   | 2193.28 | 2289.26  |  +4%  |
| SciMark2   |   13.45 |   18.48  |  +37%  |
| Dhrystone  | 1413.11 | 1447.11  |  -1%  |

See: #142

### DIFF
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -405,19 +405,8 @@ static void block_translate(riscv_t *rv, block_t *block)
         block->pc_end += ir->insn_len;
         block->n_insn++;
         /* stop on branch */
-        if (insn_is_branch(ir->opcode)) {
-            /* recursive jump translation */
-            if (ir->opcode == rv_insn_jal
-#if RV32_HAS(EXT_C)
-                || ir->opcode == rv_insn_cj || ir->opcode == rv_insn_cjal
-#endif
-            ) {
-                block->pc_end = block->pc_end - ir->insn_len + ir->imm;
-                ir->branch_taken = ir + 1;
-                continue;
-            }
+        if (insn_is_branch(ir->opcode))
             break;
-        }
     }
     block->ir[block->n_insn - 1].tailcall = true;
 }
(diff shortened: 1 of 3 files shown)


src/emulate.c
```diff
@@ -405,19 +405,8 @@ static void block_translate(riscv_t *rv, block_t *block)
...
-            if (ir->opcode == rv_insn_jal
-#if RV32_HAS(EXT_C)
-                || ir->opcode == rv_insn_cj || ir->opcode == rv_insn_cjal
-#endif
-            ) {
-                block->pc_end = block->pc_end - ir->insn_len + ir->imm;
-                ir->branch_taken = ir + 1;
-                continue;
-            }
+        if (insn_is_branch(ir->opcode))
```
jserv, line 408: Shall we change `master` branch accordingly?

qwe661234: Yes, this change is also required for detecting memset and memcpy.

jserv:
> Metric	origin	proposed	Speedup
> SciMark2	13.45	18.48	+4%

Can you check the numbers of SciMark2?
(18.48 - 13.45) / 13.45 = 37.39%

qwe661234:
> > Metric	origin	proposed	Speedup
> > SciMark2	13.45	18.48	+4%
> 
> Can you check the numbers of SciMark2? (18.48 - 13.45) / 13.45 = 37.39%

I Update this in git log.

PR #203, opened 2023-08-31
Fix typo

visitorckw:
The function mpool_destory was incorrectly named, and this commit corrects the function name to mpool_destroy.


jserv:

PR #205, opened 2023-09-01
Let read/write reuse the preallocated memory

qwe661234:
When running nyancat.elf, syscall read/write create about 200 KiB memory. This modification enable syscall read/write to reuse the pre-allocated memory.

### DIFF
--- a/src/syscall.c
+++ b/src/syscall.c
@@ -3,6 +3,7 @@
  * "LICENSE" for information on usage and redistribution of this file.
  */
 
+#include <assert.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -11,6 +12,7 @@
 #include "state.h"
 #include "utils.h"
 
+#define PREALLOC_SIZE 4096
 /* newlib is a portable (not RISC-V specific) C library, which implements
  * printf(3) and other functions described in C standards. Some system calls
  * should be provided in conjunction with newlib.
@@ -75,6 +77,7 @@ static const char *get_mode_str(uint32_t flags, uint32_t mode UNUSED)
     }
 }
 
+static uint8_t tmp[PREALLOC_SIZE];
 static void syscall_write(riscv_t *rv)
 {
     state_t *s = rv_userdata(rv); /* access userdata */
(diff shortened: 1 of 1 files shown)


src/syscall.c
```diff
@@ -85,7 +86,25 @@ static void syscall_write(riscv_t *rv)
     riscv_word_t count = rv_get_reg(rv, rv_reg_a2);
 
     /* read the string being printed */
-    uint8_t *tmp = malloc(count);
+
+    while (count > 4096) {
```
jserv: Avoid hardcoding `4096`.

src/syscall.c
```diff
@@ -10,7 +10,7 @@
 
 #include "state.h"
 #include "utils.h"
-
+#define PRE_ALLOCATED_SIZE 4096
```
jserv: Rename to `PREALLOC_SIZE` for short.

src/syscall.c
```diff
@@ -85,7 +86,26 @@ static void syscall_write(riscv_t *rv)
...
+        if (!map_at_end(s->fd_map, &it)) {
+            /* write out the data */
+            size_t written =
+                fwrite(tmp, 1, PREALLOC_SIZE, map_iter_value(&it, FILE *));
+
+            /* return number of bytes written */
+            rv_set_reg(rv, rv_reg_a0, written);
+        } else {
+            /* error */
+            rv_set_reg(rv, rv_reg_a0, -1);
```
jserv: If an error occurs, shall we escape from loop?

src/syscall.c
```diff
@@ -85,7 +86,26 @@ static void syscall_write(riscv_t *rv)
...
+        /* lookup the file descriptor */
+        map_iter_t it;
+        map_find(s->fd_map, &it, &fd);
+        if (!map_at_end(s->fd_map, &it)) {
+            /* write out the data */
+            size_t written =
+                fwrite(tmp, 1, PREALLOC_SIZE, map_iter_value(&it, FILE *));
+
+            /* return number of bytes written */
+            rv_set_reg(rv, rv_reg_a0, written);
```
jserv: Check if the accumulated byte is correct during multiple iterations.

src/syscall.c
```diff
@@ -10,7 +10,7 @@
 
 #include "state.h"
 #include "utils.h"
-
+#define PREALLOC_SIZE 4096
```
jserv, line 15: Add a blank line prior to `#define`.

src/syscall.c
```diff
@@ -85,7 +88,27 @@ static void syscall_write(riscv_t *rv)
...
+        if (!map_at_end(s->fd_map, &it)) {
+            /* write out the data */
+            size_t written =
+                fwrite(tmp, 1, PREALLOC_SIZE, map_iter_value(&it, FILE *));
+
+            /* return number of bytes written */
+            rv_set_reg(rv, rv_reg_a0, written);
+            total_write += written;
+        } else
+            /* error */
```
jserv: This comment is not meaningful. Remove.

src/syscall.c
```diff
@@ -97,12 +120,13 @@ static void syscall_write(riscv_t *rv)
 
         /* return number of bytes written */
         rv_set_reg(rv, rv_reg_a0, written);
-    } else {
+        total_write += written;
+    } else
```
jserv, line 111: If we can return earlier, there is no need to place `else` statement here.

PR #206, opened 2023-09-01
Rebuilt quake for sound support

ChinYikMing:
The prebuilt Quake binary has been updated to reflect the recent addition of sound related system calls.


jserv (changes requested):
Make the git commit message informative.
See https://github.com/sysprog21/rv32emu/commit/7ab4b6c51ed803bb479d80979a7a024463d1dba2

PR #207, opened 2023-09-04
Improve calc_fclass() with binary decision tree

visitorckw:
I have optimized the calc_fclass function by changing it into a binary decision tree. This optimization significantly improves the performance of the function, resulting in a speedup of approximately 69% in my test. The test I wrote to evaluate the performance: https://github.com/visitorckw/rv32emu-calc_fclass-optimization/

### DIFF
--- a/src/softfloat.h
+++ b/src/softfloat.h
@@ -35,29 +35,45 @@ static inline uint32_t calc_fclass(uint32_t f)
     const uint32_t expn = f & FMASK_EXPN;
     const uint32_t frac = f & FMASK_FRAC;
 
-    /* TODO: optimize with a binary decision tree */
-
     uint32_t out = 0;
-    /* 0x001    rs1 is -INF */
-    out |= (f == 0xff800000) ? 0x001 : 0;
-    /* 0x002    rs1 is negative normal */
-    out |= (expn && (expn != FMASK_EXPN) && sign) ? 0x002 : 0;
-    /* 0x004    rs1 is negative subnormal */
-    out |= (!expn && frac && sign) ? 0x004 : 0;
-    /* 0x008    rs1 is -0 */
-    out |= (f == 0x80000000) ? 0x008 : 0;
-    /* 0x010    rs1 is +0 */
-    out |= (f == 0x00000000) ? 0x010 : 0;
-    /* 0x020    rs1 is positive subnormal */
-    out |= (!expn && frac && !sign) ? 0x020 : 0;
-    /* 0x040    rs1 is positive normal */
-    out |= (expn && (expn != FMASK_EXPN) && !sign) ? 0x040 : 0;
-    /* 0x080    rs1 is +INF */
-    out |= (expn == FMASK_EXPN && !frac && !sign) ? 0x080 : 0;
-    /* 0x100    rs1 is a signaling NaN */
-    out |= (expn == FMASK_EXPN && frac && !(frac & FMASK_QNAN)) ? 0x100 : 0;
-    /* 0x200    rs1 is a quiet NaN */
-    out |= (expn == FMASK_EXPN && (frac & FMASK_QNAN)) ? 0x200 : 0;
+
+    /*
+     * 0x001    rs1 is -INF
+     * 0x002    rs1 is negative normal
+     * 0x004    rs1 is negative subnormal
+     * 0x008    rs1 is -0
+     * 0x010    rs1 is +0
+     * 0x020    rs1 is positive subnormal
+     * 0x040    rs1 is positive normal
+     * 0x080    rs1 is +INF
+     * 0x100    rs1 is a signaling NaN
+     * 0x200    rs1 is a quiet NaN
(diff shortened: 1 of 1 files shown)


src/softfloat.h
```diff
@@ -35,29 +35,44 @@ static inline uint32_t calc_fclass(uint32_t f)
...
     /* 0x040    rs1 is positive normal */
-    out |= (expn && (expn != FMASK_EXPN) && !sign) ? 0x040 : 0;
     /* 0x080    rs1 is +INF */
-    out |= (expn == FMASK_EXPN && !frac && !sign) ? 0x080 : 0;
     /* 0x100    rs1 is a signaling NaN */
-    out |= (expn == FMASK_EXPN && frac && !(frac & FMASK_QNAN)) ? 0x100 : 0;
     /* 0x200    rs1 is a quiet NaN */
-    out |= (expn == FMASK_EXPN && (frac & FMASK_QNAN)) ? 0x200 : 0;
+
+    // Check if it is special value -INF
```
jserv: Use C-style comments. That is, `/* Check if it is special value -INF */`.

src/softfloat.h
```diff
@@ -35,29 +35,44 @@ static inline uint32_t calc_fclass(uint32_t f)
...
+    out |= (f == 0xff800000) ? 0x001 : 0;
```
2011eric: `-INF` could be integrated into the if-else statement

2011eric: I rewrote the decision tree as follow, and gained more speedup:
 
```
/* Check the exponent bits */
    if (expn) {
        if (expn != FMASK_EXPN) {
            /* Check if it is negative normal or positive normal */
            out = sign ? 0x002 : 0x040;
        } else {
            /* Check if it is NaN */
            if (frac) {
                out = frac & FMASK_QNAN ? 0x200 : 0x100;
            } else if (!sign) {
                /* Check if it is +INF */
                out = 0x080;
            } else {
                /* Check if it is -INF */
                out = 0x001;
            }
        }
    } else if (frac) {
        /* Check if it is negative or positive subnormal */
        out = sign ? 0x004 : 0x020;
    } else{
        /* Check if it is +0 or -0 */
        out = sign ? 0x008 : 0x010;
    }
 ```

Comment (visitorckw) [reply src/softfloat.h]: @2011eric 
Thanks for your review and suggestions.
In my tests, the current version's execution efficiency is now approximately 69% faster than the original version.

Comment (2011eric) [reply src/softfloat.h]: Are you able to run arch-test with DEVICE=F?
Just to make sure this implementation is correct.

File: src/softfloat.h [hunk 3]
```diff
@@ -35,29 +35,45 @@ static inline uint32_t calc_fclass(uint32_t f)
...
+
+    /* Check the exponent bits */
+    if (expn) {
+        if (expn != FMASK_EXPN) {
+            /* Check if it is negative normal or positive normal */
+            out = sign ? 0x002 : 0x040;
+        } else {
+            /* Check if it is NaN */
+            if (frac) {
+                out = frac & FMASK_QNAN ? 0x200 : 0x100;
```
2011eric, line 61: I forgot to change this line:
```
out = FMASK_QNAN ? 0x200 : 0x100;
```

src/softfloat.h
visitorckw: When I attempted to execute the command ```make arch-test RISCV_DEVICE=FZicsr``` as your request, I consistently encountered the following error message, regardless of using the new function or the original one:
```
inferior exit code 0
inferior exit code 0  
ERROR | Segmentation fault (core dumped)
```
I can successfully execute other tests, such as ```RISCV_DEVICE=IMC```, without encountering any errors. I'm not sure about the cause of the error.

visitorckw: However, I used a loop to iterate through the entire range of uint32_t values as input and compared the output differences between the original function and the new function. The experimental results showed that for all inputs, both functions would return the same value.

Here is the code used for testing:
```
uint32_t i = 0;
while(1) {
    if(calc_fclass(i) != calc_fclass_new(i)) {
        fprintf(stderr, "  Error:    %x %x %x", i, calc_fclass(i), calc_fclass_new(i));
        return 1;
    }
    if(i == 0xffffffff) {
        break;
    }
    i++;
}
printf("  All tests passed.\n");
return 0;
```

src/softfloat.h
visitorckw: I believe the current implementation is correct. Changing line 61 to ```out = FMASK_QNAN ? 0x200 : 0x100;``` would cause different behavior for certain inputs compared to the original function. For example, when the input is 0x7f800001, the original function would return 0x100, but applying this modification would make it return 0x200.

2011eric: Yes, you're right.
Seems that the current implementation is correct.

jserv (changes requested):
Squash git commits and refine commit messages.

jserv:

PR #210, opened 2023-09-07
Fix round_pow2 function for edge cases

visitorckw:
The round_pow2 function was returning incorrect values when the input x was less than or equal to 1. Specifically, it was returning 2 for x=1 when compiled with GCC or Clang, and 0 for x=0 when compiled with other compilers. According to the documentation in syscall.md, both of these cases should return 1. This commit addresses the issue by explicitly handling these edge cases and returning 1.


jserv:

PR #217, opened 2023-09-12
Add comments about CSRs

TheCloudlet:


src/riscv_private.h
```diff
@@ -71,7 +71,7 @@ typedef struct {
 void block_map_clear(block_map_t *map);
 
 struct riscv_internal {
-    bool halt;
+    bool halt; /* current emualtors core is halted or not */
```
jserv: emualtors?

src/riscv_private.h
```diff
@@ -71,7 +71,7 @@ typedef struct {
 void block_map_clear(block_map_t *map);
 
 struct riscv_internal {
-    bool halt;
+    bool halt; /* current emualtor core is halted or not */
```
jserv: emualtor ?

jserv: Rephrase using Grammarly or similar tools.

PR #222, opened 2023-09-17
Fix memory leak

visitorckw:
In the io.c, memory allocated for mem was not properly freed. This change resolves the memory leak problem.


jserv:

PR #224, opened 2023-09-24
Link the register jump

### LINKED ISSUE #223: Link register jump
In the RISC-V architecture, if the jump offset is greater than or equal to +- 1 MiB, the compiler should select a register jump. Based on our observations, there are a significant number of register jumps in an executable file if it frequently invokes standard libraries or functions in different modules. However, the interpreter would exit the execution loop if the instruction is a register jump, which is a critical issue for performance. We should figure out some approaches to handle this.

* Analysis

|	Metric|dynamic instruction count of jalr |
| -------- | -------- |
|coremark	|87868155|
|stream	|350555389|

qwe661234:
Originally, the interpreter would exit the execution loop if the instruction was a register jump, which posed a critical performance issue. To address this problem, we now search for the next block at the end of the execution of a register jump. If we find the next block, we link the register jump to it.

As shown in below analysis, this modification improves the overall performance by 1~2%.

* Core i7-11700

| Metric  | Original       |    Proposed    | Speedup |
|---------|----------------|----------------|-------|
| CoreMark| 1931.047 iter/s| 1956.64 iter/s | +1.33%|
| stream  |   79.156 sec   |   77.123 sec   | +2.57%|

Close: #223

### DIFF
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -265,6 +265,67 @@ void rv_debug(riscv_t *rv)
 }
 #endif /* RV32_HAS(GDBSTUB) */
 
+/* hash function is used when mapping address into the block map */
+static inline uint32_t hash(size_t k)
+{
+    k ^= k << 21;
+    k ^= k >> 17;
+#if (SIZE_MAX > 0xFFFFFFFF)
+    k ^= k >> 35;
+    k ^= k >> 51;
+#endif
+    return k;
+}
+
+/* allocate a basic block */
+static block_t *block_alloc(const uint8_t bits)
+{
+    block_t *block = malloc(sizeof(struct block));
+    assert(block);
+    block->insn_capacity = 1 << bits;
+    block->n_insn = 0;
+    block->predict = NULL;
+    block->ir = malloc(block->insn_capacity * sizeof(rv_insn_t));
+    assert(block->ir);
+    return block;
+}
+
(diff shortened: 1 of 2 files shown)


src/emulate.c
```diff
@@ -265,6 +265,65 @@ void rv_debug(riscv_t *rv)
...
+}
+
+/* allocate a basic block */
+static block_t *block_alloc(const uint8_t bits)
+{
+    block_t *block = malloc(sizeof(struct block));
+    block->insn_capacity = 1 << bits;
+    block->n_insn = 0;
+    block->predict = NULL;
+    block->ir = malloc(block->insn_capacity * sizeof(rv_insn_t));
```
jserv, line 288: Add assertion on the return value of `malloc`.

src/emulate.c
```diff
@@ -265,6 +265,65 @@ void rv_debug(riscv_t *rv)
...
+            break;
+        }
+    }
+    map->size++;
+}
+
+/* try to locate an already translated block in the block map */
+static block_t *block_find(const block_map_t *map, const uint32_t addr)
+{
+    assert(map);
```
jserv, line 313: Assert on `addr` as well.

qwe661234: It is possible that the addr is zero, so we can add `assert(addr)` here.

PR #227, opened 2023-09-24
Add a new fusion instruction and remove unnecessary ones 

### LINKED ISSUE #226: Constant propagation and constant folding
Based on our observation, RISC-V architecture need two instruction `auipc + addi` to create a constant. If we apply constant propagation and constant folding, we could replace them with a simple instruction `lui`. Moreover, these two optimizations can apply in other scenarioes. For example, any instruction uses the constant created by `auipc + addi` can be optimized.

If you want to move 0xffffffff to a register, all the assembly code you need is:
```
LUI x2, 0xfffff000
ADDI x2, x2, 0xfff
```

qwe661234:
After constant propagation and constant folding, we can optimize some ALU or brnach instructions into simple instructions.

Close: #226

### DIFF
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -386,8 +386,7 @@ static uint32_t last_pc = 0;
     _(fuse3)           \
     _(fuse4)           \
     _(fuse5)           \
-    _(fuse6)           \
-    _(fuse7)
+    _(fuse6)
 
 enum {
     rv_insn_fuse0 = N_RV_INSNS,
@@ -396,24 +395,26 @@ enum {
 #undef _
 };
 
-/* AUIPC + ADDI */
+/* multiple lui */
 static bool do_fuse1(riscv_t *rv, const rv_insn_t *ir)
 {
-    rv->csr_cycle += 2;
-    rv->X[ir->rd] = rv->PC + ir->imm;
-    rv->X[ir->rs1] = rv->X[ir->rd] + ir->imm2;
-    rv->PC += 2 * ir->insn_len;
+    rv->csr_cycle += ir->imm2;
+    for (int i = 0; i < ir->imm2; i++) {
+        const rv_insn_t *cur_ir = ir + i;
+        rv->X[cur_ir->rd] = cur_ir->imm;
+    }
+    rv->PC += ir->imm2 * ir->insn_len;
     if (unlikely(RVOP_NO_NEXT(ir)))
         return true;
-    const rv_insn_t *next = ir + 2;
+    const rv_insn_t *next = ir + ir->imm2;
     MUST_TAIL return next->impl(rv, next);
 }
 
-/* AUIPC + ADD */
+/* LUI + ADD */
 static bool do_fuse2(riscv_t *rv, const rv_insn_t *ir)
 {
     rv->csr_cycle += 2;
-    rv->X[ir->rd] = rv->PC + ir->imm;
+    rv->X[ir->rd] = ir->imm;
     rv->X[ir->rs2] = rv->X[ir->rd] + rv->X[ir->rs1];
     rv->PC += 2 * ir->insn_len;
     if (unlikely(RVOP_NO_NEXT(ir)))
@@ -468,21 +469,8 @@ static bool do_fuse4(riscv_t *rv, const rv_insn_t *ir)
     MUST_TAIL return next->impl(rv, next);
 }
 
-/* LUI + ADDI */
-static bool do_fuse5(riscv_t *rv, const rv_insn_t *ir)
-{
-    rv->csr_cycle += 2;
-    rv->X[ir->rd] = ir->imm;
-    rv->X[ir->rs1] = ir->imm + ir->imm2;
-    rv->PC += 2 * ir->insn_len;
-    if (unlikely(RVOP_NO_NEXT(ir)))
-        return true;
-    const rv_insn_t *next = ir + 2;
-    MUST_TAIL return next->impl(rv, next);
-}
-
 /* memset */
-static bool do_fuse6(riscv_t *rv, const rv_insn_t *ir)
+static bool do_fuse5(riscv_t *rv, const rv_insn_t *ir)
 {
     rv->csr_cycle += 2;
     memory_t *m = ((state_t *) rv->userdata)->mem;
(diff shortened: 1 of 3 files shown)


src/emulate.c
```diff
@@ -834,102 +823,102 @@ static bool detect_memcpy(riscv_t *rv, int lib)
     return true;
 }
 
+static bool stdlib_injection(riscv_t *rv, block_t *block)
```
jserv: Rename to `libc_substitute`.

jserv (changes requested):
Consistently employ the prefix `constopt_` rather than `const_opt` to prevent potential confusion.

src/const_opt.c
```diff
@@ -0,0 +1,1009 @@
+/* RV32I Base Instruction Set */
```
jserv: Rename to `src/rv32_constopt.c`.

src/emulate.c
```diff
@@ -957,25 +961,46 @@ static void match_pattern(riscv_t *rv, block_t *block)
...
     }
 }
 
+
+typedef struct const_opt_info {
+    bool is_constant[32];
+    uint32_t const_val[32];
+} const_opt_info_t;
+
+#define CONSTOPT(inst, code)                                              \
```
jserv: Do `#undef CONSTOPT` when the code generation completes.

jserv:
Both constant folding and constant propagation are related compiler optimization techniques. Constant propagation replaces the bound variable with the constant expression it is bound to. Conversely, constant folding evaluates the expression with all compile-time inputs. I am a bit confused about the proposed changes, as I cannot ensure that a fully functional implementation exists for either constant folding or constant propagation. Can you verify this?

src/emulate.c
```diff
@@ -957,25 +961,46 @@ static void match_pattern(riscv_t *rv, block_t *block)
...
+#include "const_opt.c"
+/* clang-format off */
+static const void *const_opt_table[] = {
+    /* RV32 instructions */
+#define _(inst, can_branch, reg_mask) [rv_insn_##inst] = const_opt_##inst,
+    RV_INSN_LIST
+#undef _
+};
+/* clang-format on */
+typedef void (*opt_func)(rv_insn_t *, const_opt_info_t *);
```
jserv: Should be `constopt_func_t`.

src/emulate.c
```diff
@@ -957,25 +961,46 @@ static void match_pattern(riscv_t *rv, block_t *block)
...
+static void constant_opt(block_t *block)
```
jserv: Rename to `optimize_constant` for consistency.

src/emulate.c
```diff
@@ -957,25 +961,46 @@ static void match_pattern(riscv_t *rv, block_t *block)
...
-            break;
             /* TODO: mixture of SW and LW */
             /* TODO: reorder insturction to match pattern */
         }
     }
 }
 
+
+typedef struct const_opt_info {
+    bool is_constant[32];
```
jserv: Why hard-coded `32`?

src/emulate.c
```diff
@@ -834,102 +823,102 @@ static bool detect_memcpy(riscv_t *rv, int lib)
...
+                            ir->impl = dispatch_table[ir->opcode];
+                            ir->tailcall = true;
+                            return true;
+                        };
+                    }
+                }
+            }
+        }
+        break;
+        /* TODO: inject other frequently-used standard library */
```
jserv: Rephrase as following:
> Inject other frequently used function calls from the C standard library.

src/emulate.c
```diff
@@ -957,25 +961,48 @@ static void match_pattern(riscv_t *rv, block_t *block)
...
-            }
-            break;
             /* TODO: mixture of SW and LW */
             /* TODO: reorder insturction to match pattern */
         }
     }
 }
 
+
+typedef struct constopt_info {
```
jserv: Drop the use of `constopt_info`. i.e., use anonymous structure instead.

qwe661234:
> Both constant folding and constant propagation are related compiler optimization techniques. Constant propagation replaces the bound variable with the constant expression it is bound to. Conversely, constant folding evaluates the expression with all compile-time inputs. I am a bit confused about the proposed changes, as I cannot ensure that a fully functional implementation exists for either constant folding or constant propagation. Can you verify this?

There is a simple example in `./build/dhrystone.elf` can be optimized by constant folding and constant propagation.
* Before
```
   0x100f0:	auipc	gp,0x18
   0x100f4:	add	gp,gp,864 
   0x100f8:	add	a0,gp,-1996
   0x100fc:	auipc	a2,0x1a
   0x10100:	add	a2,a2,1384 #
   0x10104:	sub	a2,a2,a0
```
* After
```
   0x100f0:	lui	gp,  0x10108        
   0x100f4:	lui	gp,  0x10468
   0x100f8:	lui	a0, 0xFC9C
   0x100fc:	lui	a2,0x10116
   0x10100:	lui	a2, 0x1067E 
   0x10104:	lui    a2, 0x9e2
```
Moreover, these 6 `lui` instructions can be fused into a fused instruction.

jserv:
> There is a simple example in `build/dhrystone.elf` can be optimized by constant folding and constant propagation.
[...]
> Moreover, these 6 `lui` instructions can be fused into a fused instruction.

Describe this within git commit messages.

src/emulate.c
```diff
@@ -957,25 +962,48 @@ static void match_pattern(riscv_t *rv, block_t *block)
...
-            break;
             /* TODO: mixture of SW and LW */
             /* TODO: reorder insturction to match pattern */
         }
     }
 }
 
+
+typedef struct {
+    bool is_constant[N_RV_REGS];
```
jserv, line 973: Does it make sense to check x0, which is always zero?

src/emulate.c
```diff
@@ -957,25 +962,48 @@ static void match_pattern(riscv_t *rv, block_t *block)
...
+#undef _
+};
+#undef CONSTOPT
+
+/* clang-format on */
+typedef void (*constopt_func_t)(rv_insn_t *, constopt_info_t *);
+static void optimize_constant(block_t *block)
+{
+    constopt_info_t constopt_info;
+    memset(&constopt_info, 0, sizeof(constopt_info));
```
jserv: The C standards guarantee that even if an integer array is located on automatic storage and if there are fewer initializers in a brace-enclosed list then the uninitialized elements must be initialized to 0.

C99 Standard 6.7.8.21
> If there are fewer initializers in a brace-enclosed list than there are elements or members of an aggregate, or fewer characters in a string literal used to initialize an array of known size than there are elements in the array, the remainder of the aggregate shall be initialized implicitly the same as objects that have static storage duration.

You can safely rewrite as following without the invocation of `memset`.
```c
    constopt_info_t info = {0};
```

src/emulate.c
qwe661234: Added

PR #231, opened 2023-09-26
Fix tailcall updating in fused instructions

### LINKED ISSUE #228: Attempt to add new fusion instruction leads to error
The PR #227 introduced some bugs. Since the CI/CD for the compliance tests didn't run on the commit. This annoys contributors like me because they may not realize that the error for passing the compliance test doesn't come from themselves at first glance.

image

The problem can be reproduced by running the compliance tests with `.ci/riscv-tests.sh` on the local side. As you can see, there's a segmentation fault which didn't happen on the commit before #227.  Also, on this action job(https://github.com/sysprog21/rv32emu/actions/runs/6290355606/job/17077685558) I make sure that the PR does have some problems, but it still gets merged to the master branch.

image

qwe661234:
Close: #228

### DIFF
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -926,6 +926,7 @@ static void match_pattern(block_t *block)
                 ir->rs2 = next_ir->rd;
                 ir->rs1 = next_ir->rs1;
                 ir->impl = dispatch_table[ir->opcode];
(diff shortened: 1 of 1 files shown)


qwe661234:
There are still some issues unsolved.

PR #233, opened 2023-09-26
Fix constant optimization and fused instruction

### LINKED ISSUE #228: Attempt to add new fusion instruction leads to error

qwe661234:
1. Fix tailcall updating in fused instructions
2. Fix errors in constant optimization

Close: #228

### DIFF
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -918,41 +918,37 @@ static void match_pattern(block_t *block)
         switch (ir->opcode) {
         case rv_insn_lui:
             next_ir = ir + 1;
-            if (next_ir->opcode == rv_insn_add && ir->rd == next_ir->rs2) {
-                /* The destination register of the LUI instruction is the
-                 * same as the source register 2 of the next instruction ADD.
-                 */
-                ir->opcode = rv_insn_fuse2;
-                ir->rs2 = next_ir->rd;
-                ir->rs1 = next_ir->rs1;
-                ir->impl = dispatch_table[ir->opcode];
-            } else if (next_ir->opcode == rv_insn_add &&
-                       ir->rd == next_ir->rs1) {
-                /* The destination register of the LUI instruction is the
-                 * same as the source register 1 of the next instruction ADD.
-                 */
-                ir->opcode = rv_insn_fuse2;
-                ir->rs2 = next_ir->rd;
-                ir->rs1 = next_ir->rs2;
-                ir->impl = dispatch_table[ir->opcode];
-            } else {
+            switch (next_ir->opcode) {
+            case rv_insn_add:
+                if (ir->rd == next_ir->rs2 || ir->rd == next_ir->rs1) {
+                    ir->opcode = rv_insn_fuse2;
+                    ir->rs2 = next_ir->rd;
+                    if (ir->rd == next_ir->rs2)
+                        ir->rs1 = next_ir->rs1;
+                    else
+                        ir->rs1 = next_ir->rs2;
+                    ir->impl = dispatch_table[ir->opcode];
+                    ir->tailcall = next_ir->tailcall;
+                }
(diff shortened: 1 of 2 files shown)


src/emulate.c
```diff
@@ -935,21 +936,22 @@ static void match_pattern(block_t *block)
                 ir->rs2 = next_ir->rd;
                 ir->rs1 = next_ir->rs2;
                 ir->impl = dispatch_table[ir->opcode];
-            } else {
+                ir->tailcall = next_ir->tailcall;
+            } else if (next_ir->opcode == rv_insn_lui) {
```
jserv: Is it possible to rewrite with switch-case statements for readability?

src/emulate.c
```diff
@@ -918,41 +918,40 @@ static void match_pattern(block_t *block)
...
-                 * same as the source register 1 of the next instruction ADD.
-                 */
-                ir->opcode = rv_insn_fuse2;
-                ir->rs2 = next_ir->rd;
-                ir->rs1 = next_ir->rs2;
-                ir->impl = dispatch_table[ir->opcode];
-            } else {
+            switch (next_ir->opcode) {
+            case rv_insn_add:
+                if (ir->rd == next_ir->rs2) {
```
jserv: Can you avoid duplicating code in L923 to L934?

src/rv32_constopt.c
```diff
@@ -757,7 +756,7 @@ CONSTOPT(csw, {})
 CONSTOPT(caddi, {
     if (constopt_info->is_constant[ir->rd]) {
         constopt_info->is_constant[ir->rd] = true;
```
RinHizakura: It may not relate directly in this PR, but I'm Not sure why should we assign `true` to a variable that we have already known as true in the if statement. For example:

 ```cpp
 if (constopt_info->is_constant[ir->rd]) {
        constopt_info->is_constant[ir->rd] = true;
```

It just looks like `constopt_info->is_constant[ir->rd] = true` is a redundant operation. Or do I miss something?

Comment (qwe661234) [reply src/rv32_constopt.c]: Yes, it is redundant, I'll remove it.

File: src/rv32_constopt.c [hunk 4]
```diff
@@ -815,9 +812,9 @@ CONSTOPT(clui, {
...
 CONSTOPT(csrli, {
     if (constopt_info->is_constant[ir->rs1]) {
-        constopt_info->is_constant[ir->rs1] = true;
         ir->imm = constopt_info->const_val[ir->rs1] >> ir->shamt;
-        constopt_info->const_val[ir->rd] = ir->imm;
+        constopt_info->const_val[ir->rs1] = ir->imm;
+        ir->rd = ir->rs1;
         ir->opcode = rv_insn_clui;
         ir->impl = dispatch_table[ir->opcode];
     } else
```
RinHizakura: How about those else cases here? A statement like `constopt_info->is_constant[ir->rs1] = false` here should also be redundant.

RinHizakura:
Since it now looks great on the architectural test, I am fine with the change now. Thanks!

src/rv32_constopt.c
qwe661234: Yes, it is redundant, I'll remove it.

PR #232, opened 2023-09-26
Reduce memory usage for instruction block

RinHizakura:
The original strategy for the allocation of instruction blocks is waste of memory. For every single block, we always create a space that can contain (1 << 10) `rv_insn_t`. However, we usually don't have that much instruction in one block in most of the case.

We can simply know the heap usage by using valgrind. We can see that 20,306,989 bytes are allocated on the run of `puzzle.elf` for the old design.
```
$ valgrind ./build/rv32emu ./build/puzzle.elf
==62803== Memcheck, a memory error detector
==62803== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==62803== Using Valgrind-3.18.1 and LibVEX; rerun with -h for copyright info
==62803== Command: ./build/rv32emu ./build/puzzle.elf
==62803== 
==62803== Warning: set address range perms: large range [0x59c87000, 0x159c87000) (defined)
success in 2005 trials
inferior exit code 0
==62803== Warning: set address range perms: large range [0x59c87000, 0x159c87000) (noaccess)
==62803== 
==62803== HEAP SUMMARY:
==62803==     in use at exit: 39,124 bytes in 259 blocks
==62803==   total heap usage: 1,281 allocs, 1,022 frees, 20,306,989 bytes allocated
==62803== 
==62803== LEAK SUMMARY:
==62803==    definitely lost: 0 bytes in 0 blocks
==62803==    indirectly lost: 0 bytes in 0 blocks
==62803==      possibly lost: 0 bytes in 0 blocks
==62803==    still reachable: 37,108 bytes in 238 blocks
==62803==         suppressed: 0 bytes in 0 blocks
==62803== Rerun with --leak-check=full to see details of leaked memory
==62803== 
==62803== For lists of detected and suppressed errors, rerun with: -s
==62803== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

To address the issue, we can simply maintain a pool of `rv_insn_t`, and take only the required numbers of  `rv_insn_t` space from it. This ensures heap allocation only happens when the pool is out of `rv_insn_t`. By using the parameter `BLOCK_POOL_SIZE`, we have the flexibility to get the balance between the numbers of `calloc` calls and the memory usage. 

In this way, we have great improvement for the heap memory allocation. As the following result, only 313,461 bytes are allocated on the `puzzle.elf` example.
```
$ valgrind ./build/rv32emu ./build/puzzle.elf
==62927== Memcheck, a memory error detector
[...]

PR #234, opened 2023-09-27
Remove useless IR when fusing instructions

qwe661234:
Originally, memory usage sharply increased during the frequent allocation and freeing of basic blocks and IR arrays. To reduce memory usage, we introduced a block memory pool and an IR array memory pool.

As shown in the analysis below, we experienced about a 10% performance loss due to the reduction in memory usage. However, we are currently utilizing KiB of memory instead of MiB.

* Performance

|  Metric  |     Original    |     Proposed    |
|----------|-----------------|-----------------|
| Dhrystone|    2359 DMIPS   |   2085.5 DMIPS  |
| CoreMark | 1967.382 iter/s | 1823.333 iter/s |

* Memory usage

|  Metric  | Original | Proposed |
|----------|----------|----------|
| Dhrystone|   99 MiB | 169 KiB  |
| CoreMark |  2.4 MiB | 176 KiB  |

### DIFF
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -398,29 +398,18 @@ enum {
 #undef _
 };
 
-/* FIXME: This will simply find the n-th instruction by iterating
- * the linked list linearly, we may want to find better approach. */
-FORCE_INLINE rv_insn_t *next_nth_insn(rv_insn_t *ir, int32_t n)
-{
-    rv_insn_t *tmp = ir;
-    for (int32_t iter = 0; iter < n; iter++)
-        tmp = tmp->next;
-    return tmp;
-}
-
 /* multiple lui */
 static bool do_fuse1(riscv_t *rv, rv_insn_t *ir)
 {
     rv->csr_cycle += ir->imm2;
-    int i;
-    rv_insn_t *cur_ir;
-    for (i = 0, cur_ir = ir; i < ir->imm2; i++, cur_ir = cur_ir->next) {
-        rv->X[cur_ir->rd] = cur_ir->imm;
+    opcode_fuse_t *fuse = ir->fuse;
+    for (int i = 0; i < ir->imm2; i++) {
+        rv->X[fuse[i].rd] = fuse[i].imm;
     }
     rv->PC += ir->imm2 * ir->insn_len;
     if (unlikely(RVOP_NO_NEXT(ir)))
         return true;
-    const rv_insn_t *next = next_nth_insn(ir, ir->imm2);
+    const rv_insn_t *next = ir->next;
     MUST_TAIL return next->impl(rv, next);
 }
 
@@ -433,7 +422,7 @@ static bool do_fuse2(riscv_t *rv, rv_insn_t *ir)
     rv->PC += 2 * ir->insn_len;
     if (unlikely(RVOP_NO_NEXT(ir)))
         return true;
-    const rv_insn_t *next = next_nth_insn(ir, 2);
+    const rv_insn_t *next = ir->next;
     MUST_TAIL return next->impl(rv, next);
 }
 
@@ -456,7 +445,7 @@ static bool do_fuse3(riscv_t *rv, rv_insn_t *ir)
     rv->PC += ir->imm2 * ir->insn_len;
     if (unlikely(RVOP_NO_NEXT(ir)))
         return true;
-    const rv_insn_t *next = next_nth_insn(ir, ir->imm2);
+    const rv_insn_t *next = ir->next;
     MUST_TAIL return next->impl(rv, next);
 }
 
@@ -479,7 +468,7 @@ static bool do_fuse4(riscv_t *rv, rv_insn_t *ir)
     rv->PC += ir->imm2 * ir->insn_len;
     if (unlikely(RVOP_NO_NEXT(ir)))
         return true;
-    const rv_insn_t *next = next_nth_insn(ir, ir->imm2);
+    const rv_insn_t *next = ir->next;
     MUST_TAIL return next->impl(rv, next);
 }
 
@@ -622,7 +611,7 @@ static void block_translate(riscv_t *rv, block_map_t *map, block_t *block)
         for (int j = 1; j < count; j++, next_ir = next_ir->next) {         \
             memcpy(ir->fuse + j, next_ir, sizeof(opcode_fuse_t));          \
         }                                                                  \
-        ir->tailcall = next_ir->tailcall;                                  \
+        remove_next_nth_ir(rv, ir, block, count - 1);                      \
     }
 
 static bool detect_memset(riscv_t *rv, int lib)
@@ -851,6 +840,23 @@ static bool detect_memcpy(riscv_t *rv, int lib)
     return true;
 }
 
+FORCE_INLINE void remove_next_nth_ir(riscv_t *rv,
+                                     rv_insn_t *ir,
+                                     block_t *block,
+                                     uint8_t n)
+{
+    for (uint8_t i = 0; i < n; i++) {
+        rv_insn_t *next = ir->next;
+        ir->next = ir->next->next;
+        mpool_free(rv->block_map.block_ir_mp, next);
+    }
+    if (!ir->next) {
+        block->ir_tail = ir;
+        ir->tailcall = true;
+    }
+    block->n_insn -= n;
+}
+
 static bool libc_substitute(riscv_t *rv, block_t *block)
 {
     rv_insn_t *ir = block->ir_head, *next_ir = NULL;
@@ -872,7 +878,7 @@ static bool libc_substitute(riscv_t *rv, block_t *block)
                     if (detect_memset(rv, 1)) {
                         ir->opcode = rv_insn_fuse5;
                         ir->impl = dispatch_table[ir->opcode];
-                        ir->tailcall = true;
+                        remove_next_nth_ir(rv, ir, block, 2);
                         return true;
                     };
                 }
@@ -885,12 +891,12 @@ static bool libc_substitute(riscv_t *rv, block_t *block)
                 if (next_ir->imm == 20 && detect_memset(rv, 2)) {
                     ir->opcode = rv_insn_fuse5;
                     ir->impl = dispatch_table[ir->opcode];
-                    ir->tailcall = true;
+                    remove_next_nth_ir(rv, ir, block, 2);
                     return true;
                 } else if (next_ir->imm == 28 && detect_memcpy(rv, 2)) {
                     ir->opcode = rv_insn_fuse6;
                     ir->impl = dispatch_table[ir->opcode];
-                    ir->tailcall = true;
+                    remove_next_nth_ir(rv, ir, block, 2);
                     return true;
                 };
             }
@@ -918,7 +924,7 @@ static bool libc_substitute(riscv_t *rv, block_t *block)
                         if (detect_memcpy(rv, 1)) {
                             ir->opcode = rv_insn_fuse6;
                             ir->impl = dispatch_table[ir->opcode];
-                            ir->tailcall = true;
+                            remove_next_nth_ir(rv, ir, block, 3);
                             return true;
                         };
                     }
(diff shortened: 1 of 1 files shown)


qwe661234:
Compared to PR #232, introducing a memory pool can effectively reduce memory usage, as valgrind only records the memory you actually use. For example, we allocate 2^32 bytes of memory using `mmap` for virtual data memory, but the total memory usage is less than 2^32.

Here, I compare the performance based on Benchmark / Performance regression check and memory usage based on host machine.

* Performance

|  Metric  |     PR #232    |     This PR    |
|----------|-----------------|-----------------|
| Dhrystone|   1058.12 DMIPS   |  1084.62 DMIPS  |
| CoreMark | 922.773 iter/s | 1010.928 iter/s |

* Memory usage

|  Metric  | Original | Proposed |
|----------|----------|----------|
| Dhrystone|   382,269 B | 169,197 B  |
| CoreMark |  835,597 B | 176,605 B  |

RinHizakura:
I believe that the benefit of memory reduction mostly comes from `mmap`. When we undefined the `USE_MMAP`, you can still see the terrible memory usage.  The memory pool can do less since we only have to do `mpool_free` when `block_map_clear` happens.

If we think supporting a platform without `mmap` is not important, I would feel good to apply this, however.

jserv:
> I believe that the benefit of memory reduction mostly comes from `mmap`. When we undefined the `USE_MMAP`, you can still see the terrible memory usage. The memory pool can do less since we only have to do `mpool_free` when `block_map_clear` happens.
> If we think supporting a platform without `mmap` is not important, I would feel good to apply this, however.

Good point.

In the context of utilizing the [WebAssembly System Interface](https://wasi.dev/) (WASI), it is apparent that there is no equivalent functionality to `mmap`. [An issue](https://github.com/WebAssembly/WASI/issues/304) has been raised on the WASI GitHub repository, highlighting the absence of mmap support within the WebAssembly environment. Consequently, if rv32emu were to be ported to WebAssembly, it should be noted that the mmap functionality would not be available.

@qwe661234, you should take into consideration the scenario where mmap may not be feasibly usable on certain platforms.

qwe661234:
Even if we cannot use `mmap`, the total memory usage for blocks and IR arrays has been limited to `sizeof(block) * block_capacity + sizeof(rv_insn_t) * block_capacity * 1024` by the memory pool. We should note that the total memory usage would exceed 2^32 bytes if we could not use `mmap`, due to virtual data memory. Therefore, I believe the memory usage of blocks is not a critical issue when your total memory usage has already exceeded 4 GiB.

I believe that using a linked list to implement IR is still a more flexible approach if we want to remove any IR nodes from this linked list during instruction fusion. However, it consumes more memory than `qemu` when running certain benchmarks.

src/riscv.h
```diff
@@ -11,6 +11,9 @@
 extern "C" {
 #endif
 
+#define BLOCK_MAP_CAPACITY_BITS 10
```
jserv: Avoid defining implementation-specific constants in `riscv.h` which is supposed to be a public header.

src/emulate.c
```diff
@@ -1009,12 +1010,12 @@ static block_t *block_find_or_translate(riscv_t *rv)
 
     if (!next) {
         if (map->size * 1.25 > map->block_capacity) {
```
jserv: We may check if `1.25` makes sense still.

RinHizakura:
> Even if we cannot use mmap, the total memory usage for blocks and IR arrays has been limited to sizeof(block) * block_capacity + sizeof(rv_insn_t) * block_capacity * 1024 by the memory pool. We should note that the total memory usage would exceed 2^32 bytes if we could not use mmap, due to virtual data memory. Therefore, I believe the memory usage of blocks is not a critical issue when your total memory usage has already exceeded 4 GiB.

I'm not sure if it is safe to say something like" Although we are not good, at least we aren't that bad at all compared to others". In other words, the 2^32 bytes requirement of data memory may not be something that isn't able to do better. So I am not sure if we have to be restricted for this.

> I believe that using a linked list to implement IR is still a more flexible approach if we want to remove any IR nodes from this linked list during instruction fusion. However, it consumes more memory than qemu when running certain benchmarks.

If `mmap` is the key to reducing the consumption of memory, it is worth noting that we can also apply `mmap` for #232, which may save more memory. We should use `mmap` on #232 for fair comparison if qemu also make use of `mmap` .

RinHizakura:
Another idea here. What if we try to combine both solutions' advantages? Specifically, we still use the memory pool but allocate the chunk of memory in a single ` sizeof(rv_insn_t)` unit but not a `1 << 10 * sizeof(rv_insn_t)` unit. This allows us to still get the benefit of the memory pool, and also the benefit of `mmap` (if enabled). If `mmap` is disabled for the platform unluckily, the small chunk of memory could still reduce the arena we should prepare for mpool, since we have less chance to do `mpool_expand`.

The possible issue for this solution could be the cost for `mpool_alloc`, since allocated with smaller chunks means frequent calls of this API.  The `mpool_free` should not be a big problem because the allocation of `rv_insn_t` is special, which will free all of the allocated objects at once at `block_map_clear`, and this shall have room to be optimized.

qwe661234:
> Another idea here. What if we try to combine both solutions' advantages? Specifically, we still use the memory pool but allocate the chunk of memory in a single ` sizeof(rv_insn_t)` unit but not a `1 << 10 * sizeof(rv_insn_t)` unit. This allows us to still get the benefit of the memory pool, and also the benefit of `mmap` (if enabled). If `mmap` is disabled for the platform unluckily, the small chunk of memory could still reduce the arena we should prepare for mpool, since we have less chance to do `mpool_expand`.
> 
> The possible issue for this solution could be the cost for `mpool_alloc`, since allocated with smaller chunks means frequent calls of this API. The `mpool_free` should not be a big problem because the allocation of `rv_insn_t` is special, which will free all of the allocated objects at once at `block_map_clear`, and this shall have room to be optimized.

To maintain compactness, we must reduce our memory usage to be lower than that of QEMU. If this approach can achieve this goal while maintaining acceptable performance, it would be preferable for this PR.

RinHizakura:
> To maintain compactness, we must reduce our memory usage to be lower than that of QEMU. If this approach can achieve this goal while maintaining acceptable performance, it would be preferable for this PR.

Sure, [196cffbe](https://github.com/RinHizakura/rv32emu/commit/196cffbe33c9478fa34c095155e5af361d911e3b) shows the implementation of my new idea.

[196cffbe](https://github.com/RinHizakura/rv32emu/commit/196cffbe33c9478fa34c095155e5af361d911e3b) with `mmap`:
```
$ ./build/rv32emu ./build/dhrystone.elf 
==11644==   total heap usage: 338 allocs, 338 frees, 15,204 bytes allocated
```

#234 with `mmap`
```
valgrind ./build/rv32emu ./build/dhrystone.elf 
==12367==   total heap usage: 336 allocs, 336 frees, 15,172 bytes allocated
```

[196cffbe](https://github.com/RinHizakura/rv32emu/commit/196cffbe33c9478fa34c095155e5af361d911e3b) w/o `mmap`:

```
valgrind ./build/rv32emu ./build/dhrystone.elf 
==13356==   total heap usage: 342 allocs, 342 frees, 252,772 bytes allocated
```
#234 w/o `mmap`
```
valgrind ./build/rv32emu ./build/dhrystone.elf 
==14116==   total heap usage: 338 allocs, 338 frees, 58,768,196 bytes allocated
```

I haven't carefully experimented with the performance and other benchmarks yet. but it looks like some advantage could be brought for this new solution:
* Compare to #232, less changes are needed because we can rely on existed `mpool` mechanism
* Compare to #234, similar memory usage with `mmap` enabled. If `mmap` is disabled, however, we still don't need to allocate terrible amounts of memory.

qwe661234:
> > To maintain compactness, we must reduce our memory usage to be lower than that of QEMU. If this approach can achieve this goal while maintaining acceptable performance, it would be preferable for this PR.
> 
> Sure, [196cffbe](https://github.com/RinHizakura/rv32emu/commit/196cffbe33c9478fa34c095155e5af361d911e3b) shows the implementation of my new idea.
> 
> [196cffbe](https://github.com/RinHizakura/rv32emu/commit/196cffbe33c9478fa34c095155e5af361d911e3b) with `mmap`:
> 
> ```
> $ ./build/rv32emu ./build/dhrystone.elf 
> ==11644==   total heap usage: 338 allocs, 338 frees, 15,204 bytes allocated
> ```
> 
> #234 with `mmap`
> 
> ```
> valgrind ./build/rv32emu ./build/dhrystone.elf 
> ==12367==   total heap usage: 336 allocs, 336 frees, 15,172 bytes allocated
> ```
> 
> [196cffbe](https://github.com/RinHizakura/rv32emu/commit/196cffbe33c9478fa34c095155e5af361d911e3b) w/o `mmap`:
> 
> ```
> valgrind ./build/rv32emu ./build/dhrystone.elf 
> ==13356==   total heap usage: 342 allocs, 342 frees, 252,772 bytes allocated
> ```
> 
> #234 w/o `mmap`
> 
> ```
> valgrind ./build/rv32emu ./build/dhrystone.elf 
> ==14116==   total heap usage: 338 allocs, 338 frees, 58,768,196 bytes allocated
> ```
> 
> I haven't carefully experimented with the performance and other benchmarks yet. but it looks like some advantage could be brought for this new solution:
> 
> * Compare to [Reduce memory usage for instruction block #232](https://github.com/sysprog21/rv32emu/pull/232), less changes are needed because we can rely on existed `mpool` mechanism
> * Compare to [Limit the memory usage of block and block IR #234](https://github.com/sysprog21/rv32emu/pull/234), similar memory usage with `mmap` enabled. If `mmap` is disabled, however, we still don't need to allocate terrible amounts of memory.

@jserv, I think this idea is better than using array to store IRs. With this modification, we can easily remove useless IR directly, instead of skipping it.

jserv:
> @jserv, I think  @RinHizakura's proposal is better than using array to store IRs. With this modification, we can easily remove useless IR directly, instead of skipping it.

Based on the feedback, I would like to merge pull request #232 first and then request improvements in IR manipulation by effectively removing unneeded instructions.
@qwe661234, create new pull requests for IR manipulation and take action on this pull request, either by refining it or marking it as obsolete.

RinHizakura:
The new implementation is merged to #232 now! We could discuss over there if we have a consensus to use that solution.

jserv (changes requested):
Take further actions.

src/emulate.c
```diff
@@ -622,7 +611,9 @@ static void block_translate(riscv_t *rv, block_map_t *map, block_t *block)
         for (int j = 1; j < count; j++, next_ir = next_ir->next) {         \
             memcpy(ir->fuse + j, next_ir, sizeof(opcode_fuse_t));          \
         }                                                                  \
-        ir->tailcall = next_ir->tailcall;                                  \
+        for (int j = 1; j < count; j++) {                                  \
+            remove_next_ir(rv, ir, block);                                 \
```
RinHizakura, line 615: Maybe we can just merge this loop with its previous one, which has a similar for-loop condition?

src/emulate.c
```diff
@@ -851,6 +842,18 @@ static bool detect_memcpy(riscv_t *rv, int lib)
     return true;
 }
 
+FORCE_INLINE void remove_next_ir(riscv_t *rv, rv_insn_t *ir, block_t *block)
```
RinHizakura: How about we let the function become `remove_next_n_ir`, which can accept a number for how many IR in a row should be removed? In this way, we can avoid to introduce explicit for-loop to do `remove_next_ir` elsewhere. Some small optimizations could also be improved with this. For example,  `block->n_insn` can be updated at once, and we can also link the two IR directly.

jserv (changes requested):
Squash the git commit and refine the commit message to address the benefit by use case.

jserv:
I defer to @RinHizakura for confirmation.

RinHizakura (approved):
@jserv Now it looks great to me!

PR #244, opened 2023-10-11
Manipulate csr_cycle and PC by registers. Drop insn_len

### LINKED ISSUE #241: Interpreter should manipulate pc in registers when possible
The current interpreter is empowered by tail-call optimization (TCO), which keeps most operations in registers. Slow paths are implemented as simple jumps and do not impact register allocation. Each opcode handler, defined using `RVOP` macro, includes the statement `rv->PC += ir->insn_len`. The program counter (`pc`) can be manipulated using [continuation-passing style](https://en.wikipedia.org/wiki/Continuation-passing_style), taking advantage of TCO, instead of frequent memory access.

Check [hawk/src/vm.c](https://github.com/djwatson/hawk/blob/main/src/vm.c) for reference.

qwe661234:
To eliminate the overhead of memory operation on frequently used rv
fields and ir fields, we pass csr_cycle and PC as parameter and assigin
insn_len as constant value.

* Before
```
000000000000b1f0 <do_nop>:
    b1f0: addq $0x1,0x1a8(%rdi)
    b1f8: movl $0x0,0x58(%rdi)
    b1ff: movzbl 0x1c(%rsi),%eax
    b203: add %eax,0xd8(%rdi)
    b209: movzbl 0x1d(%rsi),%eax
    b20d: or 0x110(%rdi),%al
    b213: jne b220 <do_nop+0x30>
    b215: mov 0x38(%rsi),%rsi
    b219: jmpq *0x20(%rsi)
    b21c: nopl 0x0(%rax)
    b220: retq
    b221: nopw %cs:0x0(%rax,%rax,1)
    b22c: nopl 0x0(%rax)
```
* After
```
000000000000a6d0 <do_nop>:
    a6d0: movl $0x0,0x58(%rdi)
    a6d7: movzbl 0x1d(%rsi),%eax
    a6db: add $0x1,%rdx
    a6df: add $0x4,%ecx
    a6e2: or 0x110(%rdi),%al
    a6e8: jne a6f8 <do_nop+0x28>
    a6ea: mov 0x38(%rsi),%rsi
    a6ee: jmp *0x20(%rsi)
    a6f1: nopl 0x0(%rax)
    a6f8: mov %rdx,0x1a8(%rdi)
    a6ff: mov %ecx,0xd8(%rdi)
    a705: ret
    a706: cs nopw 0x0(%rax,%rax,1)
```
The operands of two x86 add instructions for csr_cycle, PC and insn_len
are all registers, instead of memory location.

Close: #241

### DIFF
--- a/src/decode.h
+++ b/src/decode.h
@@ -152,7 +152,7 @@ enum op_field {
         _(caddi4spn, 0, 2, ENC(rd))                 \
         _(clw, 0, 2, ENC(rs1, rd))                  \
         _(csw, 0, 2, ENC(rs1, rs2))                 \
-        /* cnop is mapped to nop */                 \
+        _(cnop, 0, 2, ENC())                        \
         _(caddi, 0, 2, ENC(rd))                     \
         _(cjal, 1, 2, ENC())                        \
         _(cli, 0, 2, ENC(rd))                       \
--- a/src/decode.h
+++ b/src/decode.h
@@ -28,160 +28,161 @@ enum op_field {
 #define ENC_GEN(X, A) ENCN(X, A)
 #define ENC(...) ENC_GEN(ENC, COUNT_VARARGS(__VA_ARGS__))(__VA_ARGS__)
 
-/* RISC-V instruction list in format _(instruction-name, can-branch, reg-mask)
+/* RISC-V instruction list in format _(instruction-name, can-branch, insn_len,
+ * reg-mask)
  */
 /* clang-format off */
-#define RV_INSN_LIST                             \
-    _(nop, 0, ENC(rs1, rd))                      \
-    /* RV32I Base Instruction Set */             \
-    _(lui, 0, ENC(rd))                           \
-    _(auipc, 0, ENC(rd))                         \
-    _(jal, 1, ENC(rd))                           \
-    _(jalr, 1, ENC(rs1, rd))                     \
-    _(beq, 1, ENC(rs1, rs2))                     \
-    _(bne, 1, ENC(rs1, rs2))                     \
-    _(blt, 1, ENC(rs1, rs2))                     \
-    _(bge, 1, ENC(rs1, rs2))                     \
-    _(bltu, 1, ENC(rs1, rs2))                    \
-    _(bgeu, 1, ENC(rs1, rs2))                    \
-    _(lb, 0, ENC(rs1, rd))                       \
-    _(lh, 0, ENC(rs1, rd))                       \
-    _(lw, 0, ENC(rs1, rd))                       \
-    _(lbu, 0, ENC(rs1, rd))                      \
-    _(lhu, 0, ENC(rs1, rd))                      \
-    _(sb, 0, ENC(rs1, rs2))                      \
-    _(sh, 0, ENC(rs1, rs2))                      \
-    _(sw, 0, ENC(rs1, rs2))                      \
-    _(addi, 0, ENC(rs1, rd))                     \
-    _(slti, 0, ENC(rs1, rd))                     \
-    _(sltiu, 0, ENC(rs1, rd))                    \
-    _(xori, 0, ENC(rs1, rd))                     \
-    _(ori, 0, ENC(rs1, rd))                      \
-    _(andi, 0, ENC(rs1, rd))                     \
-    _(slli, 0, ENC(rs1, rd))                     \
-    _(srli, 0, ENC(rs1, rd))                     \
-    _(srai, 0, ENC(rs1, rd))                     \
-    _(add, 0, ENC(rs1, rs2, rd))                 \
-    _(sub, 0, ENC(rs1, rs2, rd))                 \
-    _(sll, 0, ENC(rs1, rs2, rd))                 \
-    _(slt, 0, ENC(rs1, rs2, rd))                 \
-    _(sltu, 0, ENC(rs1, rs2, rd))                \
-    _(xor, 0, ENC(rs1, rs2, rd))                 \
-    _(srl, 0, ENC(rs1, rs2, rd))                 \
-    _(sra, 0, ENC(rs1, rs2, rd))                 \
-    _(or, 0, ENC(rs1, rs2, rd))                  \
-    _(and, 0, ENC(rs1, rs2, rd))                 \
-    _(ecall, 1, ENC(rs1, rd))                    \
-    _(ebreak, 1, ENC(rs1, rd))                   \
-    /* RISC-V Privileged Instruction */          \
-    _(wfi, 0, ENC(rs1, rd))                      \
-    _(uret, 0, ENC(rs1, rd))                     \
-    _(sret, 0, ENC(rs1, rd))                     \
-    _(hret, 0, ENC(rs1, rd))                     \
-    _(mret, 1, ENC(rs1, rd))                     \
-    /* RV32 Zifencei Standard Extension */       \
-    IIF(RV32_HAS(Zifencei))(                     \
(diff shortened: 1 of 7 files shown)


jserv (dismissed):
Rebase the latest `master` branch where I made some stupid mistakes and reverted.

src/emulate.c
```diff
@@ -368,17 +370,21 @@ static bool branch_taken = false;
...
-    {                                                       \
-        rv->csr_cycle++;                                    \
-        code;                                               \
-    nextop:                                                 \
-        rv->PC += ir->insn_len;                             \
-        if (unlikely(RVOP_NO_NEXT(ir)))                     \
-            return true;                                    \
-        const rv_insn_t *next = ir->next;                   \
-        MUST_TAIL return next->impl(rv, next);              \
+#define RVOP(inst, compressed, code)                                        \
```
jserv: I don't think it is a good idea to specify `compress` in parameter list. Can you avoid it?

qwe661234: Maybe we can hardcode PC += 4 and place PC -= 2 in all C-extension instruction.

jserv: > Maybe we can hardcode PC += 4 and place PC -= 2 in all C-extension instruction.

No, the offset for the next PC should be determined within each instruction. Whether it's 4 or 2 bytes, this value should be predictable and consistent. Additionally, the support of RV32C is build-time configurable, which means that we should handle the PC in a general way.

qwe661234: Because we cannot get the information about the instruction is compressed or not, I only figure two ideas, the first is adding this information in parameter list, and the second is handling PC += 2 or 4 per function.

jserv: > Because we cannot get the information about the instruction is compressed or not, I only figure two ideas, the first is adding this information in parameter list, and the second is handling PC += 2 or 4 per function.

How about extending `src/decode.h` by addressing compressed instructions?

qwe661234: However, the RVOP(...) marco is unrelated to `src/decode.h`.

jserv: > However, the RVOP(...) marco is unrelated to `src/decode.h`.

We can utilize the X-macro trick to enumerate the length of each instruction. For example, `RV_INSN_LEN_addi = 4`. Then, the macro RVOP can directly employ concatenation in the C preprocessor to manage PC increment.

src/decode.h
```diff
@@ -31,157 +31,157 @@ enum op_field {
...
-        _(cbnez, 1, ENC(rs1))                    \
-        _(cslli, 0, ENC(rd))                     \
-        _(clwsp, 0, ENC(rd))                     \
-        _(cjr, 1, ENC(rs1, rs2, rd))             \
-        _(cmv, 0, ENC(rs1, rs2, rd))             \
-        _(cebreak, 1, ENC(rs1, rs2, rd))         \
-        _(cjalr, 1, ENC(rs1, rs2, rd))           \
-        _(cadd, 0, ENC(rs1, rs2, rd))            \
-        _(cswsp, 0, ENC(rs2))                    \
+#define RV_INSN_LIST                                \
```
jserv, line 35: Refine the above comment for correct format description.

src/emulate.c
```diff
@@ -350,7 +352,15 @@ FORCE_INLINE bool insn_is_misaligned(uint32_t pc)
 
 /* can-branch information for each RISC-V instruction */
 enum {
-#define _(inst, can_branch, reg_mask) __rv_insn_##inst##_canbranch = can_branch,
+#define _(inst, can_branch, insn_len, reg_mask) \
+    __rv_insn_##inst##_insn_len = insn_len,
```
jserv: Simply rewrite as `__rv_insn_##inst##_len` to get rid of repeated `insn`.

src/emulate.c
```diff
@@ -350,7 +352,15 @@ FORCE_INLINE bool insn_is_misaligned(uint32_t pc)
...
+    RV_INSN_LIST
```
jserv, line 357: Drop `RV_INSN_LIST` here.

jserv:
Can you review all instances of `insn_len` (e.g., `src/decode.[ch]`) and determine if it still serves a meaningful purpose? It is possible that we can remove it to simplify the code.

src/emulate.c
```diff
@@ -348,9 +350,11 @@ FORCE_INLINE bool insn_is_misaligned(uint32_t pc)
...
 }
 
-/* can-branch information for each RISC-V instruction */
+/* can-branch and insn_len information for each RISC-V instruction */
 enum {
-#define _(inst, can_branch, reg_mask) __rv_insn_##inst##_canbranch = can_branch,
+#define _(inst, can_branch, insn_len, reg_mask) \
+    __rv_insn_##inst##_len = insn_len,          \
+    __rv_insn_##inst##_canbranch = can_branch,
     RV_INSN_LIST
```
jserv, line 365: This modification of single `enum` would provide `RV_INSN_LIST` an unexpected value. Instead, you must use two 'enum'. The term `RV_INSN_LIST` only appears once.

qwe661234: However, `enum` is outside the define,
```c
enum {
#define _(inst, can_branch, insn_len, reg_mask) \
    __rv_insn_##inst##_len = insn_len,
    RV_INSN_LIST
#undef _
};
```
Do you means a `enum` inside a `enum`?

jserv: Ouch. I misinterpreted the code. No visible concern here.

src/emulate.c
```diff
@@ -582,14 +628,14 @@ static void block_translate(riscv_t *rv, block_map_t *map, block_t *block)
 
         /* decode the instruction */
         if (!rv_decode(ir, insn)) {
-            rv->compressed = (ir->insn_len == INSN_16);
+            rv->compressed = ((insn & FC_OPCODE) != 3);
```
jserv: The enumeration of `INSN_16` and `INSN_32` can be removed from `src/decode.h`.

jserv: Add an inline helper function in `src/decode.h` or `src/riscv_private.h` to perform `insn & FC_OPCODE) != 3`.

src/decode.h
```diff
@@ -307,3 +300,5 @@ typedef struct rv_insn {
 
 /* decode the RISC-V instruction */
 bool rv_decode(rv_insn_t *ir, const uint32_t insn);
+
+#define CHECK_COMPRESSED_INSN ((insn & FC_OPCODE) != 3)
```
jserv: Rewrite as an inline function, so that compiler can check the types. In addition, write the corresponding comment for the detection of RV32C instructions.

src/decode.h
```diff
@@ -307,3 +300,9 @@ typedef struct rv_insn {
 
 /* decode the RISC-V instruction */
 bool rv_decode(rv_insn_t *ir, const uint32_t insn);
+
+/* Detect the instruction is RV32C or not */
+FORCE_INLINE bool is_compressed(uint32_t insn)
```
jserv: This type of helper function is mostly concerned with implementation details and is best off not being placed in `src/decode.h` public header.

qwe661234: We can move it to src/riscv_private.h.

jserv (changes requested):

src/emulate.c
```diff
@@ -348,9 +350,18 @@ FORCE_INLINE bool insn_is_misaligned(uint32_t pc)
     );
 }
 
+/* insn_len information for each RISC-V instruction */
```
jserv: Replace `insn_len` with instruction length in the comment.

PR #245, opened 2023-10-11
Remove tailcall member in rv_insn_t

RinHizakura:
As we introduce the member 'next' for struct rv_insn to link IRs which are in sequence, the other member 'tailcall' becomes a redundant one.

### DIFF
--- a/src/decode.h
+++ b/src/decode.h
@@ -283,14 +283,14 @@ typedef struct rv_insn {
      * optimization enables the self-recursive function to reuse the same
      * function stack frame.
      *
-     * The @tailcall member indicates whether an intermediate representation
-     * (IR) is the final instruction in a basic block. The @impl member
-     * facilitates the direct invocation of the next instruction emulation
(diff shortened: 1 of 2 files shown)


src/decode.h
```diff
@@ -283,14 +283,14 @@ typedef struct rv_insn {
      * optimization enables the self-recursive function to reuse the same
      * function stack frame.
      *
-     * The @tailcall member indicates whether an intermediate representation
-     * (IR) is the final instruction in a basic block. The @impl member
-     * facilitates the direct invocation of the next instruction emulation
-     * without the need to compute the jump address. By utilizing these two
-     * members, all instruction emulations can be rewritten into a
+     * The @next member will indicate the next intermediate representation
```
jserv: Why do you use "will indicate"?

RinHizakura: Sorry for the bad grammar. It is fixed now.

qwe661234 (approved):

PR #246, opened 2023-10-12
Move block and IRs memory pool to rv

qwe661234:
Because the block map isn't used in JIT mode, we need to move the block and IRs memory pool to rv structure.

### DIFF
--- a/src/riscv.c
+++ b/src/riscv.c
@@ -20,17 +20,12 @@ static void block_map_init(block_map_t *map, const uint8_t bits)
     map->block_capacity = 1 << bits;
     map->size = 0;
     map->map = calloc(map->block_capacity, sizeof(struct block *));
-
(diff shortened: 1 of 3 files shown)


jserv (changes requested):
Refine the Git commit messages, eliminating any grammar mistakes.

src/riscv.c
```diff
@@ -20,19 +20,14 @@ static void block_map_init(block_map_t *map, const uint8_t bits)
...
 }
 
 /* clear all block in the block map */
-void block_map_clear(block_map_t *map)
+void block_map_clear(riscv_t *rv)
 {
-    assert(map);
-    for (uint32_t i = 0; i < map->block_capacity; i++) {
-        block_t *block = map->map[i];
+    block_map_t map = rv->block_map;
```
RinHizakura: I think this should be a direct reference(`block_map_t *map = &rv->block_map`) instead of having another instance on the stack?

RinHizakura (approved):

PR #249, opened 2023-10-16
Unify the hash function of cache and map

### LINKED ISSUE #149: Unify the hash function of cache and map
Now, we implement macro `HASH` for cache and function `hash` for map. We need to unify them for code consolidation

qwe661234:
Based on the below performance analysis, the golden ratio hash is a little bit better than map hash. Therefore, we unify the hash function to golden ratio hash.

|Metric    | map hash |	golden ratio hash |
|----------|----------|-------------------|
|Dhrystone |	2.06 s|            2.07 s |
|Nqueens   |	4.35 s|	           4.25 s |
|Stream	   |    76.7 s|  	  75.59 s |

Close: #149

### DIFF
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -277,17 +277,8 @@ void rv_debug(riscv_t *rv)
 }
 #endif /* RV32_HAS(GDBSTUB) */
 
-/* hash function is used when mapping address into the block map */
-static inline uint32_t hash(size_t k)
-{
-    k ^= k << 21;
-    k ^= k >> 17;
-#if (SIZE_MAX > 0xFFFFFFFF)
-    k ^= k >> 35;
-    k ^= k >> 51;
-#endif
-    return k;
-}
+/* hash function for the block map */
+HASH_FUNC_IMPL(map_hash, BLOCK_MAP_CAPACITY_BITS, 1 << BLOCK_MAP_CAPACITY_BITS);
 
 /* allocate a basic block */
 static block_t *block_alloc(riscv_t *rv)
@@ -304,7 +295,7 @@ static void block_insert(block_map_t *map, const block_t *block)
 {
     assert(map && block);
     const uint32_t mask = map->block_capacity - 1;
-    uint32_t index = hash(block->pc_start);
+    uint32_t index = map_hash(block->pc_start);
 
     /* insert into the block map */
     for (;; index++) {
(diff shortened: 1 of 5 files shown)


src/emulate.c
```diff
@@ -277,17 +277,9 @@ void rv_debug(riscv_t *rv)
...
-{
-    k ^= k << 21;
-    k ^= k >> 17;
-#if (SIZE_MAX > 0xFFFFFFFF)
-    k ^= k >> 35;
-    k ^= k >> 51;
-#endif
-    return k;
-}
+#define HASH(val)                                                     \
```
jserv: Could you create a macro that generates a hash function based on specific inputs like `cache_size_bits` and `BLOCK_MAP_CAPACITY_BITS`? By doing so, we can consolidate the `HASH` macros used in both `src/cache.c` and `src/emulate.c`. Furthermore, because the generated function body is in the form of a static inline function, the compiler can perform type checks.

qwe661234: I make this modification, but it caused the performance lost. For example, the performance lost about 40 %(0.13s -> 0.22 s) when running the aes.elf. Due to the performance impact, I think marco is better even if it is not clean.

qwe661234: The hash function is very critical to performance, because it would be invoked everytime we find the block.

jserv: How about using `FORCE_INLINE`?

qwe661234: Ok, it works.

src/utils.h
```diff
@@ -10,3 +10,10 @@ void rv_gettimeofday(struct timeval *tv);
 
 /* Retrieve the value used by a clock which is specified by clock_id. */
 void rv_clock_gettime(struct timespec *tp);
+
+/* Detect the instruction is RV32C or not */
```
jserv: Revise the comments and reference to Linux kernel source file for hash function.

jserv (changes requested):
Consider the following macro:
```c
/* This hashing routine is adapted from Linux kernel.
 * See
 * https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/include/linux/hash.h                                                                 
 */
#define HASH_FUNC_IMPL(name, size_bits, size)                       \
    FORCE_INLINE uint32_t name(uint32_t val)                        \
    {                                                               \
        /* 0x61C88647 is 32-bit golden ratio */                     \
        return (val * 0x61C88647 >> (32 - size_bits)) & (size - 1); \
    }
```

It should appear in `src/utils.h` and be used in C source files. For example,
```c
HASH_FUNC_IMPL(hash, cache_size_bis, cache_size)
```

PR #250, opened 2023-10-16
Fix rb_remove() key comparison

### LINKED ISSUE #168: jit: Fail to successfully execute the JIT assembler
[xkon](https://github.com/ezaki-k/xkon_beta) was tested with interpreter. However, JIT compiler fails to execute it successfully.
The way to reproduce:
```
$ build/rvemu build/jit-bf.elf
```

Its output:
```
Execute 5 times with JIT precompile using xkon.

+0x2a = <.L0B>
+0x6e = <.L0E>
+0x118 = <.L1B>
+0x15c = <.L1E>
Address OPcode  ------- Instruction --------------------------------------------
...
??;(f0?Iaddi sp,sp,-32	# c.addi sp,-32
Segmentation fault (core dumped)

```

visitorckw:
The origin code was erroneously comparing ```node->data``` when searching
for the node to remove, instead of using the ```node->key```. This PR
corrects the comparison to use the ```node->key```, ensuring the proper
removal of nodes. Also, to ensure that the map functions do not rely on
node data values when comparing and performing operations, this PR
modifies the unit tests for map. It replaces the sequential data values
with random values obtained from ```mt19937_extract()```.

Close: #168

### DIFF
--- a/src/map.c
+++ b/src/map.c
@@ -221,7 +221,7 @@ static void rb_remove(map_t rb, map_node_t *node)
     pathp = path;
     while (pathp->node) {
         map_cmp_t cmp = pathp->cmp =
-            (rb->comparator)(node->data, pathp->node->data);
+            (rb->comparator)(node->key, pathp->node->key);
         if (cmp == _CMP_LESS) {
             pathp[1].node = rb_node_get_left(pathp->node);
         } else {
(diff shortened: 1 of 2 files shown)


TheCloudlet (approved):

jserv:

PR #251, opened 2023-10-18
Fix cache tests fail after adding newline at the end of files

### LINKED ISSUE #155: Cache tests fail after the enforcement of newline at the end of files
Issue #154 enforces [a newline at the end of files](https://medium.com/@alexey.inkin/how-to-force-newline-at-end-of-files-and-why-you-should-do-it-fdf76d1d090e), with the exception of cache tests, which are excluded due to failures.

The way to reproduce:
```shell
$ git ls-files -z |  while IFS= read -rd '' f; do if file --mime-encoding "$f" | grep -qv binary; then tail -c1 < "$f" | read -r _ || echo >> "$f"; fi; done
$ make tests
```

Then, the error would be caught.
```
Running lfu/cache-new ... cmp: EOF on build/cache/lfu/cache-new.out after byte 10, line 1
Failed.
make: *** [run-test-cache] Error 1
```

visitorckw:
The current test code fails to accurately match the "FREE" instruction on the final line of each input file, after introducing newline characters at the end of the test input files. This pull request addresses this issue by modifying the matching code from ```strcmp(arr[0], "FREE")``` to ```strcmp(arr[0], "FREE\n")```. Furthermore, a newline character has been included at the end of all cache test input files and "mt19937.c".

Close: #155


jserv:

PR #255, opened 2023-11-04
Fix NULL pointer dereference in remove_next_nth_ir()

visitorckw:
During the execution of Doom in rv32emu, a null pointer dereference issue was observed. After a thorough investigation, it was determined that this problem was introduced by the commit [ac05003](https://github.com/sysprog21/rv32emu/commit/ac050039e10637fe702300cd8790106fdde8051a), which added the 'remove_next_nth_ir()' function. The issue arose when attempting to remove more ir than were actually present in the linked list. To address this, we now check for the existence of the next ir in the loop. If there is no next ir, we terminate the loop early. This change ensures that we do not attempt to remove non-existent ir.

### DIFF
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -729,6 +729,10 @@ static inline void remove_next_nth_ir(const riscv_t *rv,
 {
     for (uint8_t i = 0; i < n; i++) {
         rv_insn_t *next = ir->next;
+        if (!next) {
+            n = i;
+            break;
+        }
         ir->next = ir->next->next;
         mpool_free(rv->block_ir_mp, next);
     }
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -138,7 +138,7 @@ RV_EXCEPTION_LIST
         return false;                                                 \
     }
 
-/* get current time in microsecnds and update csr_time register */
+/* get current time in microseconds and update csr_time register */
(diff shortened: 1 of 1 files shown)


qwe661234:
I suspect a bug when encountering a NULL pointer dereference in the 'remove_next_nth_ir()' function. This is because the existence of the IR is a prerequisite for its removal. Therefore, I need to investigate this error further. Could you please provide me with information to reproduce this issue?

> During the execution of Doom in rv32emu, a null pointer dereference issue was observed. After a thorough investigation, it was determined that this problem was introduced by the commit [ac05003](https://github.com/sysprog21/rv32emu/commit/ac050039e10637fe702300cd8790106fdde8051a), which added the 'remove_next_nth_ir()' function. The issue arose when attempting to remove more ir than were actually present in the linked list. To address this, we now check for the existence of the next ir in the loop. If there is no next ir, we terminate the loop early. This change ensures that we do not attempt to remove non-existent ir.

visitorckw:
Using the code from the current master branch, you can directly execute Doom to observe the phenomenon of a NULL pointer dereference. You can use either ```./build/rv32emu ./build/doom.elf``` or ```make doom```.

visitorckw:
The issue has been resolved by PR #256 ,so I am closing this PR.

PR #256, opened 2023-11-05
Fix the inaccuracy in the number of IR elements to be removed

qwe661234:
In the `libc_substitute`, only one IR needs to be removed, but the original version removes two. Therefore, the error of NULL pointer dereference occurs when running doom.


jserv:
> see #255

Enhance the descriptions and clarify the necessary actions, which may involve either deprecating the current pull request or treating the suggested modifications as a separate and distinct request.

visitorckw (approved):

PR #257, opened 2023-11-05
Fix syscall_read to handle partial reads and errors

### LINKED ISSUE #214: Assertion failed: (total_read == rv_get_reg(rv, rv_reg_a2)), function syscall_read
After commit 56766c0653526c546f00784bbc0cefb800a6f9a2, Doom fails to launch:
```
V_Init: allocate screens.
M_LoadDefaults: Load system defaults.
Assertion failed: (total_read == rv_get_reg(rv, rv_reg_a2)), function syscall_read, file syscall.c, line 290.
/bin/sh: line 1: 12197 Abort trap: 6           ../build/rv32emu doom.elf
make: *** [doom] Error 134
```

Tested on both Ubuntu Linux 20.04 (Intel Xeon E5-2650) and macOS 14.1 (Apple M1).

visitorckw:
In the original syscall_read function, an assertion error could occur when the number of bytes read didn't match the number of bytes requested. This was because the function didn't handle partial reads and errors properly.

The modified syscall_read function now handles partial reads by breaking the loop when a read operation doesn't fill the entire buffer. Additionally, it checks for errors during the read operation and sets the return value to -1 if an error occurs.

Close: #214

### DIFF
--- a/src/syscall.c
+++ b/src/syscall.c
@@ -281,13 +281,19 @@ static void syscall_read(riscv_t *rv)
     while (count > PREALLOC_SIZE) {
         size_t r = fread(tmp, 1, PREALLOC_SIZE, handle);
         memory_write(s->mem, buf + total_read, tmp, r);
-        count -= PREALLOC_SIZE;
-        total_read += PREALLOC_SIZE;
+        count -= r;
+        total_read += r;
+        if (r != PREALLOC_SIZE)
+            break;
     }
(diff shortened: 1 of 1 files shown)


qwe661234 (approved):
I truly observed that the `count` is 1024, but the `total_read` is 444 when running quake, and this patch sloved it.

jserv:

PR #260, opened 2023-11-10
Comprehensive F extension support with SoftFloat integration

### LINKED ISSUE #78: any idea why instruction like `fadd` is not passing some of the test cases such as `fadd_b10-01` ?
is there some corner cases need to be considered?

visitorckw:
This PR aimed at achieving full F extension support within rv32emu. It incorporates the SoftFloat open-source library into the rv32emu project to provide comprehensive support for the F extension. 

The key changes in this PR include:
1. Integrated the SoftFloat library into rv32emu, completing the F extension support within the project.
2. Introduced F extension testing in the CI pipeline, ensuring thorough validation of the implemented functionality.
3. Updated the README to correct the description of F extension support from incomplete to full support.

Close: #78

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "mini-gdbstub"]
 	path = src/mini-gdbstub
 	url = https://github.com/RinHizakura/mini-gdbstub
+[submodule "softfloat"]
+	path = src/softfloat
+	url = https://github.com/visitorckw/berkeley-softfloat-3
--- a/Makefile
+++ b/Makefile
@@ -64,6 +64,14 @@ $(call set-feature, EXT_C)
 ENABLE_EXT_F ?= 1
 $(call set-feature, EXT_F)
 ifeq ($(call has, EXT_F), 1)
+SOFTFLOAT_OUT = $(abspath $(OUT)/softfloat)
+src/softfloat/build/Linux-RISCV-GCC/Makefile:
+	git submodule update --init src/softfloat/
+SOFTFLOAT_LIB := $(SOFTFLOAT_OUT)/softfloat.a
+$(SOFTFLOAT_LIB): src/softfloat/build/Linux-RISCV-GCC/Makefile
+	$(MAKE) -C $(dir $<) BUILD_DIR=$(SOFTFLOAT_OUT)
+$(OUT)/decode.o $(OUT)/riscv.o: $(SOFTFLOAT_LIB)
+LDFLAGS += $(SOFTFLOAT_LIB)
 LDFLAGS += -lm
 endif
 
@@ -211,5 +219,6 @@ distclean: clean
 	$(RM) *.zip
 	$(RM) -r $(OUT)/mini-gdbstub
 	-$(RM) $(OUT)/.config
+	-$(RM) -r $(OUT)/softfloat
 
 -include $(deps)
--- a/src/riscv.h
+++ b/src/riscv.h
@@ -6,6 +6,9 @@
 #pragma once
 #include <stdbool.h>
 #include <stdint.h>
+#if RV32_HAS(EXT_F)
+#include "softfloat/softfloat.h"
+#endif
 
 #ifdef __cplusplus
 extern "C" {
@@ -89,7 +92,9 @@ typedef uint32_t riscv_word_t;
 typedef uint16_t riscv_half_t;
 typedef uint8_t riscv_byte_t;
 typedef uint32_t riscv_exception_t;
-typedef float riscv_float_t;
+#if RV32_HAS(EXT_F)
+typedef float32_t riscv_float_t;
+#endif
 
 /* memory read handlers */
 typedef riscv_word_t (*riscv_mem_ifetch)(riscv_word_t addr);
--- a/src/softfloat.h
+++ b/src/softfloat.h
@@ -8,6 +8,7 @@
 #include <stdint.h>
 
 #include "riscv.h"
+#include "riscv_private.h"
 
 /* clang-format off */
 enum {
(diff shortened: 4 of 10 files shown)


jserv:
Is it possible to use git submodule instead of copying SoftFloat files in this repository?

jserv:
> The key changes in this PR include:
> 1. Integrated the SoftFloat library into rv32emu, completing the F extension support within the project.

I would like to revoke my opinion on the implementation of the RISC-V F extension. Integrating SoftFloat would achieve full compatibility and reduce maintenance efforts. There is no need to retain the original RV32F-specific code. That means the conditional build between the original and SoftFloat-specific code can be removed, and let's focus on SoftFloat.

visitorckw:
> Is it possible to use git submodule instead of copying SoftFloat files in this repository?

Currently, I've copied all files as I added additional files in the SoftFloat folder to comply with the risc-v F extension standard. Perhaps we can explore alternatives, such as copying these files into the SoftFloat directory during compilation.

visitorckw:
> I would like to revoke my opinion on the implementation of the RISC-V F extension. Integrating SoftFloat would achieve full compatibility and reduce maintenance efforts. There is no need to retain the original RV32F-specific code. That means the conditional build between the original and SoftFloat-specific code can be removed, and let's focus on SoftFloat.

Sounds like a good idea, I'll make the change later.

jserv:
> Currently, I've copied all files as I added additional files in the SoftFloat folder to comply with the risc-v F extension standard. Perhaps we can explore alternatives, such as copying these files into the SoftFloat directory during compilation.

To incorporate external source code, please refer to the `Makefile` and examine the sections related to `mini-gdbstub`. These sections contain information on how to utilize `git submodule` and their associated build rules.

Makefile
```diff
@@ -211,5 +223,6 @@ distclean: clean
 	$(RM) *.zip
 	$(RM) -r $(OUT)/mini-gdbstub
 	-$(RM) $(OUT)/.config
+	$(MAKE) -C src/SoftFloat-3e/build/Linux-RISCV-GCC/ clean
```
jserv: It should begin with a `-` character, indicating that it is acceptable to fail.
See https://www.gnu.org/software/make/manual/html_node/Errors.html

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "mini-gdbstub"]
 	path = src/mini-gdbstub
 	url = https://github.com/RinHizakura/mini-gdbstub
+[submodule "SoftFloat-3e"]
+	path = src/SoftFloat-3e
```
jserv: Simply name it `src/softfloat` as QEMU does.
See https://github.com/qemu/qemu/blob/master/fpu/softfloat.c

src/riscv.h
```diff
@@ -6,6 +6,7 @@
 #pragma once
 #include <stdbool.h>
 #include <stdint.h>
+#include "SoftFloat-3e/source/include/softfloat.h"
```
jserv: The header inclusion is a bit lenthy. Can you shorten it?

jserv:
Can you review `src/softfloat.h` and determine whether we should retain it (partially) or not?

visitorckw:
> Can you review `src/softfloat.h` and determine whether we should retain it (partially) or not?

We can remove the is_snan function as the softfloat library already provides the same functionality with isSignalingNaN. However, the rest of the content needs to be retained. Additionally, I'm considering whether renaming src/softfloat.h could prevent confusion.

visitorckw:
I'm uncertain why the current code is failing the Codacy Security Scan test.

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "mini-gdbstub"]
 	path = src/mini-gdbstub
 	url = https://github.com/RinHizakura/mini-gdbstub
+[submodule "softfloat"]
+	path = src/softfloat
+	url = https://github.com/visitorckw/SoftFloat-3e.git
```
jserv: Use shorter URL. i.e., https://github.com/visitorckw/SoftFloat-3e`
Can you utilize https://github.com/ucb-bar/berkeley-softfloat-3 instead?

src/softfloat.h
```diff
@@ -85,9 +85,32 @@ static inline bool is_nan(uint32_t f)
     return (expn == FMASK_EXPN && frac);
 }
 
-static inline bool is_snan(uint32_t f)
-{
-    const uint32_t expn = f & FMASK_EXPN;
-    const uint32_t frac = f & FMASK_FRAC;
-    return (expn == FMASK_EXPN && frac && !(frac & FMASK_QNAN));
-}
+#define set_fflag()                                              \
```
jserv: Can you rewrite it by using a `static inline` function?

.gitmodules
visitorckw: The Makefile for building softfloat.a at https://github.com/ucb-bar/berkeley-softfloat-3 doesn't specify an output path parameter. As a result, using it may require keeping softfloat.a within the src/softfloat instead of the build/softfloat. Additionally, the previously mentioned issue of lengthy header inclusion in riscv.h might be temporarily challenging to resolve.

jserv:
> I'm uncertain why the current code is failing the Codacy Security Scan test.

It seems that something is wrong with Docker Hub.

.gitmodules
jserv: Rather than duplicating files without their historical commits, consider forking `ucb-bar/berkeley-softfloat-3` and incorporating your own modifications and bug fixes. Does this approach seem reasonable?

visitorckw: Agreed. I'll make the adjustment.

jserv:
When rv32emu is built without F extension, i.e., `make ENABLE_EXT_F=0`, I got the following error messages:
```
src/riscv.h:9:10: fatal error: softfloat/softfloat.h: No such file or directory
    9 | #include "softfloat/softfloat.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

jserv (changes requested):
Fix build without F extension support.

visitorckw:
Apologies for the error. I've corrected it. It should be resolved now.

jserv:

PR #265, opened 2023-11-16
Fix a potential issue where the block_find() might return NULL

henrybear327:
As `block_find` might return NULL,  `prev->ir_tail` might be dereferencing a NULL pointer.

### DIFF
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -1004,20 +1004,22 @@ void rv_step(riscv_t *rv, int32_t cycles)
             if (prev->pc_start != last_pc)
                 prev = block_find(&rv->block_map, last_pc);
 
-            rv_insn_t *last_ir = prev->ir_tail;
-            /* chain block */
-            if (!insn_is_unconditional_branch(last_ir->opcode)) {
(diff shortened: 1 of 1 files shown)


henrybear327:
Not sure if an assertion is actually a better idea here, as a NULL pointer would most likely be an issue in the control flow?

qwe661234 (approved):

qwe661234:
This situation arises when the cache is full, yet the current test cases do not produce this particular scenario.

qwe661234:
> This situation arises when the cache is full, yet the current test cases do not produce this particular scenario.

jserv:
> This situation arises when the cache is full, yet the current test cases do not produce this particular scenario.

Shall we extend the test cases accordingly?

qwe661234:
> > This situation arises when the cache is full, yet the current test cases do not produce this particular scenario.
> 
> Shall we extend the test cases accordingly?

Actually, I encountered it in JIT mode and I has made the same modification in JIT branch. However, the block cache of interpreter would be clear if it is full, so it is difficult to produce this.

Issue #268, opened 2023-11-18
Apply branch prediction for branch instructions


jserv:
To speed up the emulation of branch instructions, the following C code serves as a PoC:
```c
#include <stdio.h>
#include <stdbool.h>
#include <stdint.h>

#define HISTORY_SIZE 16

/* Define a structure for the branch history buffer */
typedef struct {
    uint32_t instruction_address;
    bool branch_taken;
} branch_history_entry;

/* Initialize the branch history buffer */
static branch_history_entry history_buffer[HISTORY_SIZE];
static int buffer_index = 0;

/* Function to predict the outcome of a branch instruction */
static bool predict(uint32_t instruction_address)
{
    /* Check if the instruction address is in the history buffer */
    for (int i = 0; i < HISTORY_SIZE; i++) {
        if (history_buffer[i].instruction_address == instruction_address) {
            /* Make a prediction based on historical behavior */
            return history_buffer[i].branch_taken;
        }
    }
    /* Default prediction if no history is available */
    return false;
}

/* Function to update the branch history buffer */
static void update_history(uint32_t instruction_address, bool branch_outcome)
{
    /* Update the branch history buffer with the outcome of the instruction */
    history_buffer[buffer_index].instruction_address = instruction_address;
    history_buffer[buffer_index].branch_taken = branch_outcome;
    buffer_index = (buffer_index + 1) % HISTORY_SIZE;
}

int main()
{
    /* Emulation loop */
    while (!end_of_program) {
        uint32_t current_instruction = fetch_instruction();

        if (is_branch_instruction(current_instruction)) {
            bool prediction = predict(current_instruction.instruction_address);
            if (prediction) {
                jump_to(predicted_jump_target);
            } else {
                continue_execution();
            }
        }

        /* Execute the instruction as usual */

        /* Update branch history based on the actual outcome */
        update_history(current_instruction.instruction_address, actual_outcome);
    }

    return 0;
}
```

`The history_buffer` stores the history of branch instructions, including their addresses and whether they were taken or not. This allows the emulator to learn from past behavior. The `predict` function looks up the history buffer to predict the outcome of the current branch instruction. If the instruction's address is found in the history buffer, it makes a prediction based on historical behavior. After executing a branch instruction and determining the actual outcome, the `update_history` function updates the history buffer with the actual outcome. This information helps refine future predictions.

[...]

PR #269, opened 2023-11-19
Apply branch prediction for indirect jump

qwe661234:
Previously, it was necessary to perform a block cache lookup at the end of an indirect jump emulation; however, the associated overhead of this operation proved to be substantial. To mitigate this overhead, we have introduced a branch history table that captures the historical data of indirect jump targets. Given the limited number of entries in the branch history table, the lookup overhead is significantly reduced.

As shown in the performance analysis provided below, the branch history table has demonstrably enhanced the overall performance.

|  Metric   |   original   |   proposed   |
|-----------|--------------|--------------|
| Dhrystone | 2932.3 DMIPS | 2985.2 DMIPS |
| CoreMark  | 2231 iter/s  | 2236 iter/s  |
| Stream    | 76.04 sec    | 75.299 sec   |
| Nqueens   | 4.069 sec    | 3.933 sec    |

See: #268

### DIFF
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -371,21 +371,21 @@ static bool is_branch_taken = false;
 static uint32_t last_pc = 0;
 
 /* Interpreter-based execution path */
-#define RVOP(inst, code)                                                    \
-    static bool do_##inst(riscv_t *rv, const rv_insn_t *ir, uint64_t cycle, \
-                          uint32_t PC)                                      \
-    {                                                                       \
-        cycle++;                                                            \
-        code;                                                               \
-    nextop:                                                                 \
-        PC += __rv_insn_##inst##_len;                                       \
-        if (unlikely(RVOP_NO_NEXT(ir))) {                                   \
-            rv->csr_cycle = cycle;                                          \
-            rv->PC = PC;                                                    \
-            return true;                                                    \
-        }                                                                   \
-        const rv_insn_t *next = ir->next;                                   \
-        MUST_TAIL return next->impl(rv, next, cycle, PC);                   \
+#define RVOP(inst, code)                                              \
+    static bool do_##inst(riscv_t *rv, rv_insn_t *ir, uint64_t cycle, \
+                          uint32_t PC)                                \
+    {                                                                 \
+        cycle++;                                                      \
+        code;                                                         \
+    nextop:                                                           \
+        PC += __rv_insn_##inst##_len;                                 \
+        if (unlikely(RVOP_NO_NEXT(ir))) {                             \
+            rv->csr_cycle = cycle;                                    \
+            rv->PC = PC;                                              \
+            return true;                                              \
+        }                                                             \
+        const rv_insn_t *next = ir->next;                             \
+        MUST_TAIL return next->impl(rv, next, cycle, PC);             \
     }
 
 #include "rv32_template.c"
@@ -633,8 +633,16 @@ static void block_translate(riscv_t *rv, block_t *block)
         block->n_insn++;
         prev_ir = ir;
         /* stop on branch */
-        if (insn_is_branch(ir->opcode))
+        if (insn_is_branch(ir->opcode)) {
+            if (ir->opcode == rv_insn_jalr
+#if RV32_HAS(EXT_C)
+                || ir->opcode == rv_insn_cjalr || ir->opcode == rv_insn_cjr
+#endif
+            )
+                ir->branch_table =
+                    calloc(1, HISTORY_SIZE * sizeof(branch_history_entry_t));
             break;
+        }
 
         ir = mpool_alloc(rv->block_ir_mp);
     }
--- a/src/rv32_template.c
+++ b/src/rv32_template.c
@@ -41,6 +41,30 @@ RVOP(jal, {
     return true;
 })
 
+/* The branch history table records historical data pertaining to indirect jump
+ * targets. This functionality alleviates the need to invoke block_find() and
+ * incurs overhead only when the indirect jump targets are not previously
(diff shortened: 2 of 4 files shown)


jserv:
Can you check the strategies and implementations at https://github.com/bucaps/marss-riscv/tree/master/src/riscvsim/bpu as well?

src/rv32_template.c
```diff
@@ -4,6 +4,8 @@
  */
 
 /* Internal */
+#include <stdio.h>
```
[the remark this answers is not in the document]
jserv: Is the inclusion `<stdio.h>` necessary?

src/rv32_template.c
```diff
@@ -4,6 +4,8 @@
...
+#include "riscv_private.h"
```
jserv: As the header name suggests, this template is intended for both interpreter and JIT compiler. Whenever feasible, we aim to eliminate the need for including this header. Additionally, each function's body should be concise, making use of macros and helper functions from `src/emulate.c`. This approach will help minimize the proposed changes in the `src/rv32_template.c` file.

RinHizakura:
I'm not sure whether the history table has a significant effect on the performance. It looks like the history table is the cache of the block map, doesn't it?

According to my understanding, `block_find()` should be O(1) in the average case, and the entries in the block cache should only be evicted when doing the `block_map_clear()`. Since `block_map_clear()` should happen rarely, the probability of missing on the block map would be low.

qwe661234:
> I'm not sure whether the history table has a significant effect on the performance. It looks like the history table is the cache of the block map, doesn't it?
> 
> 
> 
> According to my understanding, `block_find()` should be O(1) in the average case, and the entries in the block cache should only be evicted when doing the `block_map_clear()`. Since `block_map_clear()` should happen rarely, the probability of missing on the block map would be low. 

The original design is block_find().

RinHizakura:
> The original design is block_find().

Sure, I know it is the original design.

No disrespect, I am just curious to know what's the meaning of the "overhead" on the original design. I assume that the original design has O(1) complexity on average, and the extra cost of the original design should only happen when "find miss" on the map. Because of these great features, I can't feel how the history table can improve on top of this. On the other hand, although the metric shows the proposed design is better, I suspect this improvement is still in the margin of error.

qwe661234:
> > The original design is block_find().
> 
> 
> 
> Sure, I know it is the original design.
> 
> 
> 
> No disrespect, I am just curious to know what's the meaning of the "overhead" on the original design. I assume that the original design has O(1) complexity on average, and the extra cost of the original design should only happen when "find miss" on the map. On the other hand, although the metric shows the proposed design is better, I suspect this improvement is still in the margin of error.
> 
> 

In the block_find(), it needs hash function, comparison and the overhead of function call. The overhead of branch history table is extra memory, but it only needs several comparisons without hash function and function call.

Actually, the branch table is designed for further improving JIT indirect jump codegen and its benefit on interpreter is not substantial.

RinHizakura:
>In the block_find(), it needs hash function, comparison and the overhead of function call. The overhead of branch history table is extra memory, but it only needs several comparisons without hash function and function call.
>
> Actually, the branch table is designed for further improving JIT indirect jump codegen and its benefits on interpreter is not substantial.

Got it. Possibly because of the simple mechanism now in the history table with the interpreter mode, I can't see its power on the emulator. I also assume that the cost of hash function should be low which can be ignored, but this might not be true.

Thanks for your kind reply!

src/emulate.c
```diff
@@ -633,8 +633,16 @@ static void block_translate(riscv_t *rv, block_t *block)
...
         prev_ir = ir;
         /* stop on branch */
-        if (insn_is_branch(ir->opcode))
+        if (insn_is_branch(ir->opcode)) {
+            if (ir->opcode == rv_insn_jalr
+#if RV32_HAS(EXT_C)
+                || ir->opcode == rv_insn_cjalr || ir->opcode == rv_insn_cjr
+#endif
+            )
+                ir->branch_table =
```
jserv, line 642: When can we release the allocated space?

src/rv32_template.c
```diff
@@ -57,9 +57,21 @@ RVOP(jalr, {
         rv->X[ir->rd] = pc + 4;
     /* check instruction misaligned */
     RV_EXC_MISALIGN_HANDLER(pc, insn, false, 0);
+    /* lookup branch table */
```
jserv: Could you consider eliminating redundant code for branch lookups in `jalr`, `c.jr`, and `cjalr`? Macros might be a feasible solution.

qwe661234: I eliminate redundant code by macro.

RinHizakura (approved):

src/rv32_template.c
```diff
@@ -41,6 +41,24 @@ RVOP(jal, {
     return true;
 })
 
+#define LOOKUP_OR_UPDATE_BRANCH_HISTORY_TABLE()                               \
```
jserv, line 50: Add explanation comments for this macro.

jserv:
> Can you check the strategies and implementations at https://github.com/bucaps/marss-riscv/tree/master/src/riscvsim/bpu as well?

The [two-level adaptive predictor](https://en.wikipedia.org/wiki/Branch_predictor#Two-level_adaptive_predictor)'s strength lies in its ability to swiftly adapt to and predict repetitive patterns effectively. This prediction technique is employed in many contemporary microprocessors, and rv32emu could potentially gain advantages from adopting such an approach. Let's assess its performance using the reference code.

jserv (changes requested):
Fix typo ("proposed") in git commit message. Update the numbers if you have recent measurements.

PR #271, opened 2023-11-22
Improve efficiency of branch history table

RinHizakura:
There are some small mistakes/defects in the original history table design:
* The least recent branch history won't be checked because the latest history will be put at index `ir->branch_table_count`, but we only check the range in [0, `ir->branch_table_count`)
* If `ir->branch_table_count` rings back to index 0, the previous history will be dropped because we only check the range in [0, `ir->branch_table_count`)
* The `PC` for every branch history entry can be put densely since they are the key for comparison, so they could be beneficial for cache and probably provide chance to optimize using SIMD operation

### DIFF
--- a/src/decode.h
+++ b/src/decode.h
@@ -254,9 +254,10 @@ typedef struct {
 
 #define HISTORY_SIZE 16
 typedef struct {
-    uint32_t PC;
-    struct rv_insn *branch_target;
-} branch_history_entry_t;
+    uint8_t idx;
+    uint32_t PC[HISTORY_SIZE];
+    struct rv_insn *target[HISTORY_SIZE];
+} branch_history_table_t;
 
 typedef struct rv_insn {
     union {
@@ -300,8 +301,7 @@ typedef struct rv_insn {
      * specific IR array without the need for additional copying.
      */
     struct rv_insn *branch_taken, *branch_untaken;
-    uint8_t branch_table_count;
-    branch_history_entry_t *branch_table;
+    branch_history_table_t *branch_table;
 } rv_insn_t;
 
 /* decode the RISC-V instruction */
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -639,8 +639,7 @@ static void block_translate(riscv_t *rv, block_t *block)
                 || ir->opcode == rv_insn_cjalr || ir->opcode == rv_insn_cjr
 #endif
             )
-                ir->branch_table =
-                    calloc(1, HISTORY_SIZE * sizeof(branch_history_entry_t));
+                ir->branch_table = calloc(1, sizeof(branch_history_table_t));
             break;
         }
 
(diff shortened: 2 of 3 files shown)


RinHizakura:
It's worth mentioning that the check of the history table is changed to a static range [0, HISTORY_SIZE). I assume that it is impossible to branch to PC 0, so the condition `if (ir->branch_table->PC[i] == PC)` will always be false for a non-initialized entry. If the assumption is not always correct, more checks should be introduced to avoid accessing the non-initialized object.

qwe661234 (approved):
Right, the line 52 should be `HISTORY_SIZE`, instead of `ir->branch_table_count`. The other modification save 1 byte memory (maybe more bytes becasue of memory alignment) for non indirect jump instruction IR.

qwe661234:
> It's worth mentioning that the check of the history table is changed to a static range [0, HISTORY_SIZE). I assume that it is impossible to branch to PC 0, so the condition `if (ir->branch_table->PC[i] == PC)` will always be false for a non-initialized entry. If the assumption is not always correct, more check should be introduced to avoid to access the NULL object.

The error only occurs when jump target is PC 0, but it is not common. Other solution is initializing the PC array to an odd value because the odd jump target is impossible.

src/decode.h
```diff
@@ -254,9 +254,10 @@ typedef struct {
 
 #define HISTORY_SIZE 16
 typedef struct {
-    uint32_t PC;
-    struct rv_insn *branch_target;
-} branch_history_entry_t;
+    uint8_t idx;
+    uint32_t PC[HISTORY_SIZE];
+    struct rv_insn *branch_target[HISTORY_SIZE];
```
jserv: Rename it `target` from `branch_target` since it explicitly appears in a structure dealing with branch history.

jserv (changes requested):
Improve the git commit messages. Make it informative.

jserv:
By the way, I am wondering which value should be chosen for `HISTORY_SIZE`.
@qwe661234, Any considerations or explanations for making this adjustment?

qwe661234:
> By the way, I am wondering which value should be chosen for `HISTORY_SIZE`. @qwe661234, Any considerations or explanations for making this adjustment?

Based on my observations, the majority of branch targets typically number less than 10. However, for certain hard-to-predict branches, the count exceeds 20. I believe the branch history table is tailored to enhance the performance of indirect jump instructions featuring a limited number of branch targets. Therefore, I propose that the value 16 is suited for this purpose.

jserv (changes requested):
Consider the following change for git commit message:
> The previous implementation of the branch history table had some limitations that prevented us from fully leveraging its capabilities. This commit aims to improve access to recent history entries and make it more cache-friendly.

RinHizakura:
> By the way, I am wondering which value should be chosen for HISTORY_SIZE.

Actually I am thinking to optimize this with SIMD operation, and HISTORY_SIZE needs to be 8 consider 256 bits vector. But the idea seems to be a sledgehammer to crack a nut at the first glance.

https://github.com/RinHizakura/rv32emu/commit/ad1c7e5fc988a382fc9564eb29ead3ddf96c8fec

PR #273, opened 2023-11-22
Implement file path sanitation

henrybear327:
As mentioned in issue #137, the elf path is not sanitized before opening, which might be a security risk. This PR addresses this issue by porting the path sanitation logic from Golang to C.

Reference code: https://cs.opensource.google/go/go/+/refs/tags/go1.21.4:src/path/path.go;l=70

Unit test -> https://github.com/sysprog21/rv32emu/pull/274

PR #274, opened 2023-11-22
The unit tests for PR #273

henrybear327:
This PR ported the unit tests for path sanitation from the Golang library.

Reference: https://cs.opensource.google/go/go/+/refs/tags/go1.21.4:src/path/path_test.go

### DIFF
--- a/src/elf.c
+++ b/src/elf.c
@@ -8,6 +8,7 @@
 
 #include "elf.h"
 #include "io.h"
+#include "utils.h"
 
 #if defined(_WIN32)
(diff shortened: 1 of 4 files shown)


henrybear327:
Issues:
- I am not certain how we should be including unit tests in the rv32emu codebase
- The current diff is messy, as the base PR https://github.com/sysprog21/rv32emu/pull/273 can't be used as the base commit for this PR

jserv:
Close in favor of #273

PR #276, opened 2023-11-23
rv_histogram: Make it more verbose

### LINKED ISSUE #272: histogram: Fail to execute with `cc.elf`
when running `cc.elf`, rv_histogram will occur Arithmetic exception.

After slightly debugging with gdb, I discovered that max_insn_freq was 0 and that dividing by 0 resulted in an arithmetic exception.

```
Program received signal SIGFPE, Arithmetic exception.
0x000055555555b460 in gen_hist_bar (hist_bar=0x7fffffffd630 " ", hist_bar_len=388, insn_freq=0, max_insn_freq=0, max_col=129, used_col=35) at tools/rv_histogram.c:121
121         size_t v = insn_freq * (max_col - used_col) * 8 / max_insn_freq;
(gdb) p max_insn_freq 
```

ChinYikMing:
Section headers must be present for static analysis since the tool uses the `SHT_PROGBITS` type and `SHF_EXECINSTR` flag, which are defined in section headers, to determine the executable machine instructions of the target program

Closed #272

### DIFF
--- a/tools/rv_histogram.c
+++ b/tools/rv_histogram.c
@@ -241,11 +241,17 @@ int main(int argc, const char *args[])
 
     elf_t *e = elf_new();
     if (!elf_open(e, elf_prog)) {
-        fprintf(stderr, "elf_open failed\n");
+        (void) fprintf(stderr, "Failed to open %s\n", elf_prog);
         return 1;
     }
 
     struct Elf32_Ehdr *hdr = get_elf_header(e);
+    if (!hdr->e_shnum) {
+        (void) fprintf(stderr, "no section headers are found in %s\n",
+                       elf_prog);
+        return 1;
+    }
+
     uint8_t *elf_first_byte = get_elf_first_byte(e);
     const struct Elf32_Shdr **shdrs =
         (const struct Elf32_Shdr **) &elf_first_byte[hdr->e_shoff];


tools/rv_histogram.c
```diff
@@ -231,21 +231,32 @@ static void insn_hist_incr(const rv_insn_t *ir)
...
     }
 
     /* resolver of histogram accounting */
     hist_record_handler hist_record = show_reg ? reg_hist_incr : insn_hist_incr;
 
     elf_t *e = elf_new();
     if (!elf_open(e, elf_prog)) {
-        fprintf(stderr, "elf_open failed\n");
-        return 1;
+        (void) fprintf(stderr, "elf_open %s failed\n", elf_prog);
```
jserv: Replace the message with something like "Failed to open"

jserv (changes requested):
Explain why ELF section headers are necessary for static analysis in the git commit message.

tools/rv_histogram.c
```diff
@@ -317,5 +328,6 @@ int main(int argc, const char *args[])
 
     elf_delete(e);
 
-    return 0;
+end:
+    return !!ret_val;
```
jserv: Simply return `ret_val`.

tools/rv_histogram.c
```diff
@@ -231,21 +231,32 @@ static void insn_hist_incr(const rv_insn_t *ir)
 
 int main(int argc, const char *args[])
 {
+    int ret_val = 0;
+
     if (!parse_args(argc, args) || !elf_prog) {
         print_usage(args[0]);
-        return 1;
+        ret_val = 1;
+        goto end;
```
jserv: Instead of using `goto`, you can use `return` instead.

jserv: Alternatively, define a function named `fatal`, which dumps the message and invokes [exit](https://man7.org/linux/man-pages/man3/exit.3.html).

ChinYikMing: MISRA checker reported that single function should have only one return statement. So I made this change. Shall I stick with early return?

jserv: You can safely ignore the warnings reported by Cppcheck at the moment. I shall suppress some of them.

jserv:

PR #281, opened 2023-11-26
CI: Publish Docker image to Docker Hub

henrybear327:
Docker images will be built on merge requests.

As discussed, we will employ a multi-stage docker build to reduce build time and image size, but still, making keep the image reproducible when the environment needs to be changed.


jserv:
How much time does it take to build a Docker image from scratch? I am concerned about the considerably longer build time in the CI pipeline, which could lead to potential problems.

henrybear327:
> How much time does it take to build a Docker image from scratch? I am concerned about the considerably longer build time in the CI pipeline, which could lead to potential problems.

This is under testing now. I have no idea how long this will take for the CI on GitHub. Locally, it will be ~20 min IIRC.

jserv:
> This is under testing now. I have no idea how long this will take for the CI on GitHub. Locally, it will be ~20 min IIRC.

20 min is too long. Can we concentrate on x86-64 target instead?

henrybear327:
> > This is under testing now. I have no idea how long this will take for the CI on GitHub. Locally, it will be ~20 min IIRC.
> 
> 20 min is too long. Can we concentrate on x86-64 target instead?

I think I would propose the following:
- make a separate DockerHub image, called `rv32emu-base`
- unless we change the base environment, we will just pull the base image and build on top of it, which I believe will only be several seconds to complete!

Dockerfile-base
```diff
@@ -0,0 +1,51 @@
...
+
+# when using apt install gcc-riscv64-unknown-elf, this will cause "unsupported ISA subset 'z'" during compilation
+# thus, we are building from scratch, following the version here -> https://github.com/sysprog21/rv32emu/blob/master/.ci/riscv-toolchain-install.sh
+# for x86, we can optimize this part and take the nightly build directly, but not for aarch64 
+ENV RISCV=/opt/riscv
+ENV PATH=$RISCV/bin:$PATH
+WORKDIR $RISCV
+RUN apt install -y autoconf automake autotools-dev curl libmpc-dev libmpfr-dev libgmp-dev gawk build-essential bison flex texinfo gperf libtool patchutils bc zlib1g-dev libexpat-dev
+RUN git clone --recursive https://github.com/riscv/riscv-gnu-toolchain
+RUN cd riscv-gnu-toolchain && \
```
jserv: Can we maintain a repository which preserves the prebuilt GNU Toolchain for RISC-V? Therefore, it saves unnecessary build time.

henrybear327: 2 separate images are currently being created, namely `sysprog21/rv32emu-gcc` and `sysprog21/rv32emu-sail`. 

For `gcc`, we are using the compiler version `tags/2023.10.06`, as using `apt install gcc-riscv64-unknown-elf` will cause "unsupported ISA subset 'z'" during the compilation of the emulator on M1. 

For `sail`, we set up the toolchain to use `sail-0.16`, as the latest version `0.17.x` series will cause compilation errors. 

For the reference emulator, we are using the commit `9547a30bf84572c458476591b569a95f5232c1c7` from `sail-riscv`, since this is the commit of the emulator that is closest to the time when https://github.com/sysprog21/rv32emu/commit/01b00b6f175f57ef39ffd1f4fa6a611891e36df3#diff-3b436c5e32c40ecca4095bdacc1fb69c0759096f86e029238ce34bbe73c6e68f is added.

The eventual `rv32emu` on Dockerhub will be built using the multi-stage feature, where both of the aforementioned images will be used (copying the necessary binaries only), and then install extra packages for the emulator to compile and execute with full capabilities correctly, then packaged into an image.

dockerhub/build.sh
```diff
@@ -1,9 +1,13 @@
...
+docker build --progress=plain -t sysprog21/rv32emu-gcc -f Dockerfile-gcc . 2>&1 | tee build-gcc.log
+rm build-gcc.log
+
+# will take around 10 minutes if the internet connection is fast
+docker build --progress=plain -t sysprog21/rv32emu-sail -f Dockerfile-sail . 2>&1 | tee build-sail.log
+rm build-sail.log
 
-# will take around 10 minutes
-docker build --progress=plain -t sysprog21/rv32emu-sail -f Dockerfile-sail . 2>&1 | tee build.log
+# push to dockerhub
```
jserv: Always refer it as Docker Hub.

dockerhub/README.md
```diff
@@ -1,9 +1,13 @@
 # Base images
 
-This folder contains the Dockerfiles of the base images.
+This folder contains the Dockerfiles of the base images, containing the `gcc` and `sail` toolchain.
```
jserv: Use "directory" instead of "folder" in UNIX or similar systems.

dockerhub/README.md
```diff
@@ -1,9 +1,13 @@
...
+
+## Details
+
+For `gcc`, we are using the compiler version `tags/2023.10.06`, as using the binaries from `apt install gcc-riscv64-unknown-elf` will cause `"unsupported ISA subset 'z'"` error during the compilation of the emulator on M1.
 
 For `sail`, we set up the toolchain to use `sail-0.16`, as the latest version `0.17.x` series will cause compilation errors.
 
-For the reference emulator, we are using the commit `9547a30bf84572c458476591b569a95f5232c1c7` from sail-riscv, since this is the commit of the emulator that is closest to the time when the [x86 reference sail emulator](https://github.com/sysprog21/rv32emu/commit/01b00b6f175f57ef39ffd1f4fa6a611891e36df3#diff-3b436c5e32c40ecca4095bdacc1fb69c0759096f86e029238ce34bbe73c6e68f) is added.
+For the reference emulator, we are using the commit `9547a30bf84572c458476591b569a95f5232c1c7` from `sail-riscv`, since this is the commit of the emulator that is closest to the time when the [x86 reference sail emulator](https://github.com/sysprog21/rv32emu/commit/01b00b6f175f57ef39ffd1f4fa6a611891e36df3#diff-3b436c5e32c40ecca4095bdacc1fb69c0759096f86e029238ce34bbe73c6e68f) is added to the `rv32emu` Github repo.
```
jserv: Improve the writing.

jserv: Improve the writing.

dockerhub/README.md
```diff
@@ -0,0 +1,13 @@
+# Base images
```
jserv: Move this file to the `docs` directory.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,33 @@
+name: Publish Docker image to DockerHub
```
jserv: Always call it Docker Hub.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,33 @@
...
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+# https://docs.docker.com/build/ci/github-actions/multi-platform/
+jobs:
+  docker:
+    name: Build Docker Image and Publish (only on push)
+    runs-on: ubuntu-latest
```
jserv: Should Ubuntu Linux 22.04-LTS be specified?

jserv:
Instead of a directory named "dockerhub," should we put the Docker related files in somewhere like `docker` directory?

Dockerfile
```diff
@@ -1,55 +1,29 @@
...
-    git checkout 9547a30bf84572c458476591b569a95f5232c1c7 && \
-    eval $(opam env) && \
-    make && \
-    ARCH=RV32 make 
-
 # copy in the source code
 WORKDIR /home/root/rv32emu
 COPY . .
 
+# Copy the GCC binaries
```
jserv: Should be "GNU Toolchain files"

.github/workflows/docker.yml
henrybear327: I have also changed other GitHub workflows!

henrybear327:
> Instead of a directory named "dockerhub," should we put the Docker related files in somewhere like `docker` directory?

Directory renamed.

The Dockerfile in the root directory isn't moved, since we can't copy source files from `../`, as the rule is the following

```
The <src> path must be inside the context of the build; you cannot COPY ../something /something, because the first step of a docker build is to send the context directory (and subdirectories) to the docker daemon.
```

docker/build.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
jserv: Use `#!/usr/bin/env bash`.
See https://stackoverflow.com/questions/21612980/why-is-usr-bin-env-bash-superior-to-bin-bash

henrybear327: Interesting read! Thanks for the reference!

henrybear327:
The GitHub secrets aren't set properly. @jserv can you fix it so the CI can work properly :) 

Thank you!

jserv:
> The GitHub secrets aren't set properly. @jserv can you fix it so the CI can work properly :)

I do set the following:
* `DOCKERHUB_ACCESS_TOKEN`
* `DOCKERHUB_USERNAME`

.github/workflows/docker.yml
```diff
@@ -0,0 +1,33 @@
...
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+# https://docs.docker.com/build/ci/github-actions/multi-platform/
+jobs:
+  docker:
+    name: Build Docker Image and Publish (only on push)
```
jserv: Use shorter name, such as `build-and-publish`.

docs/base-image.md
```diff
@@ -0,0 +1,13 @@
+# Base images
+
+This directory contains the Dockerfiles of the base images, containing the `gcc` and `sail` toolchain.
```
jserv: Refine the descriptions to give an overview about the motivations and purpose.

jserv (changes requested):
Document the elapsed time for the completion of building, generating, and publishing the Docker image, in git commit messages.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,33 @@
...
+      
+      - name: Check out the repo
+        uses: actions/checkout@v4
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Login to Docker Hub
+        uses: docker/login-action@v3
+        # if: github.event_name == 'push'
```
jserv: Don't leave the dead code.

docker/build.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env
```
jserv: It lacks of `bash` or `sh`.

henrybear327:
> > The GitHub secrets aren't set properly. @jserv can you fix it so the CI can work properly :)
> 
> I do set the following:
> 
> * `DOCKERHUB_ACCESS_TOKEN`
> * `DOCKERHUB_USERNAME`

Based on this [comment](https://github.com/pypa/gh-action-pypi-publish/discussions/49#discussioncomment-162653), I think it's a security feature, that a pull request from another repo shouldn't be able to access the secret, which makes sense. 

I have tested the Docker Hub login and push feature on my branch, so the PR should work when merged!

docs/base-image.md
```diff
@@ -0,0 +1,15 @@
+# Base images
+
+The goal is to support both `x86` and `aarch64` docker images for development and testing.
```
jserv: Should be `x86-64`

jserv:
> The build process will take around 20 and 10 minutes when building a native image on the native arch for gcc and sail, respectively, if  internet connection is fast

The proposed change is currently too lengthy to be accepted. We should consider building the Docker image only when files in the `src` directory are modified. Additionally, to optimize the process, we should implement incremental generation. This means that we should update only the `rv32emu` executable instead of rebuilding the entire gcc and sail components.

henrybear327:
> > The build process will take around 20 and 10 minutes when building a native image on the native arch for gcc and sail, respectively, if  internet connection is fast
> 
> The proposed change is currently too lengthy to be accepted. We should consider building the Docker image only when files in the `src` directory are modified. Additionally, to optimize the process, we should implement incremental generation. This means that we should update only the `rv32emu` executable instead of rebuilding the entire gcc and sail components.

Sorry that the commit message is misleading. Please see the updated version. 

To sum up, the incremental build is taking around 6 minutes to generate a new rv32emu image and push it to Docker Hub.

jserv:
> To sum up, the incremental build is taking around 6 minutes to generate a new rv32emu image and push it to Docker Hub.

A filter should be implemented to monitor modified files. See https://github.com/sysprog21/rv32emu/blob/master/.github/workflows/benchmark.yml

henrybear327:
> > To sum up, the incremental build is taking around 6 minutes to generate a new rv32emu image and push it to Docker Hub.
> 
> A filter should be implemented to monitor modified files. See https://github.com/sysprog21/rv32emu/blob/master/.github/workflows/benchmark.yml

The rule is set up to only produce a new `rv32emu` image when there are any changes in the following directories (as there are files that might impact code compilation, benchmarking, arch tests, etc.
- `src/**`
- `build/**`
- `mk/**`
- `tests/**`
- `tools/**`

However, I think it might be better that all PR be merged to the master branch to produce a new image. In this case, any image that we pull from the Docker Hub will be the latest commit on the master branch.

jserv (changes requested):
Modify `README.md` and mention the instructions to pull built images from Docker Hub.

.github/workflows/codacy.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       security-events: write # for github/codeql-action/upload-sarif to upload SARIF results
       actions: read # only required for a private repository by github/codeql-action/upload-sarif to get the Action run status
     name: Codacy Security Scan
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
jserv: Revoke the changes on {codacy,codeql,compliance}.yml.

jserv:

Issue #238, opened 2023-10-04
jit: Add predecessor-indexed CFG profiling infrastructure


qwe661234:
## Goal

Improve control-flow profiling and block-cache maintenance without committing to a graph-based SSA IR or whole-function lifting.

## Context

The emulator already records learned taken and untaken block edges, tracks indirect-target history, and uses block chaining to bypass the dispatcher. Current work also includes packed interpreter traces and trace-local fusion. The missing capability is efficient reverse-edge tracking: when a cached block is evicted or invalidated, its incoming links must currently be found by scanning translated blocks.

A single `prev` pointer is not sufficient because a CFG node can have multiple predecessors, including indirect control-flow targets.

## Proposed scope

- Introduce a bounded, ownership-safe predecessor index for translated-block entry edges.
- Keep it consistent when direct, conditional, and learned indirect edges are installed, replaced, evicted, or invalidated.
- Use it to unlink incoming edges without a full block-list traversal.
- Expose CFG-edge and hot-path profiling data through the existing profiler/tools.
- Add focused tests for edge installation, multiple predecessors, cache eviction, invalidation, and stale-edge prevention.

## Non-goals

- No whole-function lifting based on ELF metadata.
- No new graph/SSA/use-def IR in this issue.
- No cross-function optimization or replacement of the existing Tier-1 or LLVM Tier-2 pipelines.

## Success criteria

- Correctness is preserved under cache replacement, address-space changes, and indirect branches.
- The implementation removes or materially reduces global scans during block unlinking.
- Profiling identifies useful hot-edge/trace candidates.
- Benchmark results show neutral or positive steady-state performance and acceptable memory overhead.

## Follow-up decision

Consider a trace-local or graph-IR prototype only if this infrastructure and benchmark data demonstrate an optimization opportunity that the existing packed IR, block chaining, fusion, and LLVM Tier-2 compiler cannot address.

jserv:
The [sfuzz](https://github.com/seal9055/sfuzz) project employs a [custom intermediate representation](https://github.com/seal9055/sfuzz/blob/main/docs/code_gen.md). The initial step in the actual code generation process involves lifting the entire function into this intermediate representation. During the initialization phase, when the target is first loaded, the size of the function is determined. This is achieved by parsing the elf metadata and creating a hashmap that maps function start addresses to their respective sizes.

[...]

PR #289, opened 2023-12-12
Introduce a tier-1 JIT compiler based on x86-64 architecture

qwe661234:
When the using frequency of a block exceeds a predetermined threshold, the tier-1 JIT compiler traces the chained block and generate  corresponding low quailty machine code. The resulting target machine code is stored in the code cache for future utilization.

The primary objective of introducing the tier-1 JIT compiler is to enhance the execution speed of RISC-V instructions. This implementation requires two additional components: a tier-1 machine code generator,  and code cache. Furthermore, this tier-1 JIT compiler serves as the foundational target for future improvements.

In addition, we have developed a Python script that effectively traces code templates and automatically generates JIT code templates. This approach eliminates the need for manually writing duplicated code.

As shown in the performance analysis below, the tier-1 JIT compiler's  performance closely parallels that of QEMU in benchmarks with a  constrained dynamic instruction count. However, for benchmarks  featuring a substantial dynamic instruction count or lacking specific hotspots—examples include pi and STRINGSORT—the tier-1 JIT compiler demonstrates noticeably slower execution compared to QEMU.

Hence, a robust tier-2 compiler is essential to generate optimized machine code across diverse execution paths, coupled with a runtime profiler for detecting hotspots.

* Performance

| Metric   | rv32emu(T1C) | qemu  |
|----------|-----------------|-------|
|aes	   |             0.02|  0.031|
|mandelbrot|	        0.029| 0.0115|
|puzzle	   |           0.0115|	0.009|
|pi	       |           0.0413| 0.0177|
|dhrystone |	        0.331|	0.393|
|Nqeueens  |	        0.854|	0.749|
|qsort-O2  |	        2.384|	 2.16|
|miniz-O2  |	         1.33|	 1.01|
|primes-O2 |	         2.93|	1.069|
|sha512-O2 |	        2.057|	0.939|
|stream	   |           12.747|	10.36|
|STRINGSORT|	       89.012| 11.496|

As demonstrated in the memory usage analysis below, the tier-1 JIT compiler utilizes less memory than QEMU across all benchmarks.

* Memory usage

| Metric   | rv32emu(T1C) |   qemu  |
|----------|-----------------|---------|
|aes	   |          186,228|1,343,012|
|mandelbrot|	      152,203|  841,841|
|puzzle	   |          153,423|  890,225|
|pi	       |          152,923|  879,957|
|dhrystone |	      154,466|  856,404|
|Nqeueens  |	      154,880|  858,618|
|qsort-O2  |	      155,091|  933,506|
|miniz-O2  |	      165,627|1,076,682|
|primes-O2 |	      150,540|  928,446|
|sha512-O2 |	      153,553|	978,177|
|stream	   |          165,911|  957,845|
|STRINGSORT|	      167,871|1,104,702|

Related: #238

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,6 +118,21 @@ gdbstub-test: $(BIN)
 	$(Q).ci/gdbstub-test.sh && $(call notice, [OK])
 endif
 
+ENABLE_JIT ?= 0
+$(call set-feature, JIT)
+ifeq ($(call has, JIT), 1)
+OBJS_EXT += jit_x64.o 
+ifneq ($(processor), x86_64)
+$(error JIT mode only supports for x64 target currently.)
+endif
+
+src/rv32_jit_template.c:
+	$(Q)tools/gen-jit-template.py $(CFLAGS) > $@
+
+$(OUT)/jit_x64.o: src/jit_x64.c src/rv32_jit_template.c
+	$(VECHO) "  CC\t$@\n"
+	$(Q)$(CC) -o $@ $(CFLAGS) -c -MMD -MF $@.d $<
+endif
 # For tail-call elimination, we need a specific set of build flags applied.
 # FIXME: On macOS + Apple Silicon, -fno-stack-protector might have a negative impact.
 $(OUT)/emulate.o: CFLAGS += -foptimize-sibling-calls -fomit-frame-pointer -fno-stack-check -fno-stack-protector
@@ -214,7 +229,7 @@ endif
 endif
 
 clean:
-	$(RM) $(BIN) $(OBJS) $(HIST_BIN) $(HIST_OBJS) $(deps) $(CACHE_OUT)
+	$(RM) $(BIN) $(OBJS) $(HIST_BIN) $(HIST_OBJS) $(deps) $(CACHE_OUT) src/rv32_jit_template.c
 distclean: clean
 	-$(RM) $(DOOM_DATA) $(QUAKE_DATA)
 	$(RM) -r $(OUT)/id1
--- a/src/cache.c
+++ b/src/cache.c
@@ -13,10 +13,13 @@
 #include "mpool.h"
 #include "utils.h"
 
-/* THRESHOLD is set to identify hot spots. Once the frequency of use for a block
- * exceeds the THRESHOLD, the JIT compiler flow is triggered.
+/* Currently, THRESHOLD is set to identify hot spots. Once the using frequency
+ * for a block exceeds the THRESHOLD, the tier-1 JIT compiler process is
+ * triggered.
+ * FIXME: Implement effective  profiler to detect hot spots, instead of simply
+ * relying on THRESHOLD.
  */
-#define THRESHOLD 1000
+#define THRESHOLD 4096
 
 static uint32_t cache_size, cache_size_bits;
 static struct mpool *cache_mp;
@@ -530,3 +533,62 @@ void cache_free(cache_t *cache)
     free(cache->map);
     free(cache);
 }
+
+#if !RV32_HAS(ARC)
+uint32_t cache_freq(struct cache *cache, uint32_t key)
+{
+    if (!cache->capacity ||
+        hlist_empty(&cache->map->ht_list_head[cache_hash(key)]))
+        return 0;
+    lfu_entry_t *entry = NULL;
+#ifdef __HAVE_TYPEOF
+    hlist_for_each_entry (entry, &cache->map->ht_list_head[cache_hash(key)],
+                          ht_list)
+#else
+    hlist_for_each_entry (entry, &cache->map->ht_list_head[cache_hash(key)],
+                          ht_list, lfu_entry_t)
+#endif
+    {
+        if (entry->key == key)
+            return entry->frequency;
+    }
+    return 0;
+}
+#endif
+
+#if RV32_HAS(JIT)
+bool cache_hot(struct cache *cache, uint32_t key)
+{
+    if (!cache->capacity ||
+        hlist_empty(&cache->map->ht_list_head[cache_hash(key)]))
+        return false;
+#if RV32_HAS(ARC)
+    arc_entry_t *entry = NULL;
+#ifdef __HAVE_TYPEOF
+    hlist_for_each_entry (entry, &cache->map->ht_list_head[cache_hash(key)],
+                          ht_list)
+#else
+    hlist_for_each_entry (entry, &cache->map->ht_list_head[cache_hash(key)],
+                          ht_list, arc_entry_t)
+#endif
+    {
+        if (entry->key == key && entry->frequency == THRESHOLD)
+            return true;
+    }
+#else
+    lfu_entry_t *entry = NULL;
+#ifdef __HAVE_TYPEOF
+    hlist_for_each_entry (entry, &cache->map->ht_list_head[cache_hash(key)],
+                          ht_list)
+#else
+    hlist_for_each_entry (entry, &cache->map->ht_list_head[cache_hash(key)],
+                          ht_list, lfu_entry_t)
+#endif
+    {
+        if (entry->key == key && entry->frequency == THRESHOLD)
+            return true;
+    }
+#endif
+    return false;
+}
+#endif
--- a/src/decode.h
+++ b/src/decode.h
@@ -179,13 +179,31 @@ enum op_field {
     )
 /* clang-format on */
 
+/* Macro operation fusion */
+
+/* macro operation fusion: convert specific RISC-V instruction patterns
+ * into faster and equivalent code
+ */
+#define FUSE_INSN_LIST \
+    _(fuse1)           \
+    _(fuse2)           \
+    _(fuse3)           \
+    _(fuse4)           \
+    _(fuse5)           \
+    _(fuse6)           \
+    _(fuse7)
+
 /* clang-format off */
 /* IR list */
 enum {
 #define _(inst, can_branch, insn_len, reg_mask) rv_insn_##inst,
     RV_INSN_LIST
 #undef _
-    N_RV_INSNS
+    N_RV_INSNS,
+#define _(inst) rv_insn_##inst,
+    FUSE_INSN_LIST
+#undef _
+    N_TOTAL_INSNS,
 };
 /* clang-format on */
 
--- a/src/decode.h
+++ b/src/decode.h
@@ -29,153 +29,153 @@ enum op_field {
 #define ENC(...) ENC_GEN(ENC, COUNT_VARARGS(__VA_ARGS__))(__VA_ARGS__)
 
 /* RISC-V instruction list in format _(instruction-name, can-branch, insn_len,
- * reg-mask)
+ * translatable, reg-mask)
  */
 /* clang-format off */
-#define RV_INSN_LIST                                \
-    _(nop, 0, 4, ENC(rs1, rd))                      \
-    /* RV32I Base Instruction Set */                \
-    _(lui, 0, 4, ENC(rd))                           \
-    _(auipc, 0, 4, ENC(rd))                         \
-    _(jal, 1, 4, ENC(rd))                           \
-    _(jalr, 1, 4, ENC(rs1, rd))                     \
-    _(beq, 1, 4, ENC(rs1, rs2))                     \
-    _(bne, 1, 4, ENC(rs1, rs2))                     \
-    _(blt, 1, 4, ENC(rs1, rs2))                     \
-    _(bge, 1, 4, ENC(rs1, rs2))                     \
-    _(bltu, 1, 4, ENC(rs1, rs2))                    \
-    _(bgeu, 1, 4, ENC(rs1, rs2))                    \
-    _(lb, 0, 4, ENC(rs1, rd))                       \
-    _(lh, 0, 4, ENC(rs1, rd))                       \
-    _(lw, 0, 4, ENC(rs1, rd))                       \
-    _(lbu, 0, 4, ENC(rs1, rd))                      \
-    _(lhu, 0, 4, ENC(rs1, rd))                      \
-    _(sb, 0, 4, ENC(rs1, rs2))                      \
-    _(sh, 0, 4, ENC(rs1, rs2))                      \
-    _(sw, 0, 4, ENC(rs1, rs2))                      \
-    _(addi, 0, 4, ENC(rs1, rd))                     \
-    _(slti, 0, 4, ENC(rs1, rd))                     \
-    _(sltiu, 0, 4, ENC(rs1, rd))                    \
-    _(xori, 0, 4, ENC(rs1, rd))                     \
-    _(ori, 0, 4, ENC(rs1, rd))                      \
-    _(andi, 0, 4, ENC(rs1, rd))                     \
-    _(slli, 0, 4, ENC(rs1, rd))                     \
-    _(srli, 0, 4, ENC(rs1, rd))                     \
-    _(srai, 0, 4, ENC(rs1, rd))                     \
-    _(add, 0, 4, ENC(rs1, rs2, rd))                 \
-    _(sub, 0, 4, ENC(rs1, rs2, rd))                 \
-    _(sll, 0, 4, ENC(rs1, rs2, rd))                 \
-    _(slt, 0, 4, ENC(rs1, rs2, rd))                 \
-    _(sltu, 0, 4, ENC(rs1, rs2, rd))                \
-    _(xor, 0, 4, ENC(rs1, rs2, rd))                 \
-    _(srl, 0, 4, ENC(rs1, rs2, rd))                 \
-    _(sra, 0, 4, ENC(rs1, rs2, rd))                 \
-    _(or, 0, 4, ENC(rs1, rs2, rd))                  \
-    _(and, 0, 4, ENC(rs1, rs2, rd))                 \
-    _(ecall, 1, 4, ENC(rs1, rd))                    \
-    _(ebreak, 1, 4, ENC(rs1, rd))                   \
-    /* RISC-V Privileged Instruction */             \
-    _(wfi, 0, 4, ENC(rs1, rd))                      \
-    _(uret, 0, 4, ENC(rs1, rd))                     \
-    _(sret, 0, 4, ENC(rs1, rd))                     \
-    _(hret, 0, 4, ENC(rs1, rd))                     \
-    _(mret, 1, 4, ENC(rs1, rd))                     \
-    /* RV32 Zifencei Standard Extension */          \
-    IIF(RV32_HAS(Zifencei))(                        \
-        _(fencei, 1, 4, ENC(rs1, rd))               \
-    )                                               \
-    /* RV32 Zicsr Standard Extension */             \
-    IIF(RV32_HAS(Zicsr))(                           \
-        _(csrrw, 0, 4, ENC(rs1, rd))                \
-        _(csrrs, 0, 4, ENC(rs1, rd))                \
-        _(csrrc, 0, 4, ENC(rs1, rd))                \
-        _(csrrwi, 0, 4, ENC(rs1, rd))               \
-        _(csrrsi, 0, 4, ENC(rs1, rd))               \
-        _(csrrci, 0, 4, ENC(rs1, rd))               \
-    )                                               \
-    /* RV32M Standard Extension */                  \
-    IIF(RV32_HAS(EXT_M))(                           \
-        _(mul, 0, 4, ENC(rs1, rs2, rd))             \
-        _(mulh, 0, 4, ENC(rs1, rs2, rd))            \
-        _(mulhsu, 0, 4, ENC(rs1, rs2, rd))          \
-        _(mulhu, 0, 4, ENC(rs1, rs2, rd))           \
-        _(div, 0, 4, ENC(rs1, rs2, rd))             \
-        _(divu, 0, 4, ENC(rs1, rs2, rd))            \
-        _(rem, 0, 4, ENC(rs1, rs2, rd))             \
-        _(remu, 0, 4, ENC(rs1, rs2, rd))            \
-    )                                               \
-    /* RV32A Standard Extension */                  \
-    IIF(RV32_HAS(EXT_A))(                           \
-        _(lrw, 0, 4, ENC(rs1, rs2, rd))             \
-        _(scw, 0, 4, ENC(rs1, rs2, rd))             \
-        _(amoswapw, 0, 4, ENC(rs1, rs2, rd))        \
-        _(amoaddw, 0, 4, ENC(rs1, rs2, rd))         \
-        _(amoxorw, 0, 4, ENC(rs1, rs2, rd))         \
-        _(amoandw, 0, 4, ENC(rs1, rs2, rd))         \
-        _(amoorw, 0, 4, ENC(rs1, rs2, rd))          \
-        _(amominw, 0, 4, ENC(rs1, rs2, rd))         \
-        _(amomaxw, 0, 4, ENC(rs1, rs2, rd))         \
-        _(amominuw, 0, 4, ENC(rs1, rs2, rd))        \
-        _(amomaxuw, 0, 4, ENC(rs1, rs2, rd))        \
-    )                                               \
-    /* RV32F Standard Extension */                  \
-    IIF(RV32_HAS(EXT_F))(                           \
-        _(flw, 0, 4, ENC(rs1, rd))                  \
-        _(fsw, 0, 4, ENC(rs1, rs2))                 \
-        _(fmadds, 0, 4, ENC(rs1, rs2, rs3, rd))     \
-        _(fmsubs, 0, 4, ENC(rs1, rs2, rs3, rd))     \
-        _(fnmsubs, 0, 4, ENC(rs1, rs2, rs3, rd))    \
-        _(fnmadds, 0, 4, ENC(rs1, rs2, rs3, rd))    \
-        _(fadds, 0, 4, ENC(rs1, rs2, rd))           \
-        _(fsubs, 0, 4, ENC(rs1, rs2, rd))           \
-        _(fmuls, 0, 4, ENC(rs1, rs2, rd))           \
-        _(fdivs, 0, 4, ENC(rs1, rs2, rd))           \
-        _(fsqrts, 0, 4, ENC(rs1, rs2, rd))          \
-        _(fsgnjs, 0, 4, ENC(rs1, rs2, rd))          \
-        _(fsgnjns, 0, 4, ENC(rs1, rs2, rd))         \
-        _(fsgnjxs, 0, 4, ENC(rs1, rs2, rd))         \
-        _(fmins, 0, 4, ENC(rs1, rs2, rd))           \
-        _(fmaxs, 0, 4, ENC(rs1, rs2, rd))           \
-        _(fcvtws, 0, 4, ENC(rs1, rs2, rd))          \
-        _(fcvtwus, 0, 4, ENC(rs1, rs2, rd))         \
-        _(fmvxw, 0, 4, ENC(rs1, rs2, rd))           \
-        _(feqs, 0, 4, ENC(rs1, rs2, rd))            \
-        _(flts, 0, 4, ENC(rs1, rs2, rd))            \
-        _(fles, 0, 4, ENC(rs1, rs2, rd))            \
-        _(fclasss, 0, 4, ENC(rs1, rs2, rd))         \
-        _(fcvtsw, 0, 4, ENC(rs1, rs2, rd))          \
-        _(fcvtswu, 0, 4, ENC(rs1, rs2, rd))         \
-        _(fmvwx, 0, 4, ENC(rs1, rs2, rd))           \
-    )                                               \
-    /* RV32C Standard Extension */                  \
-    IIF(RV32_HAS(EXT_C))(                           \
-        _(caddi4spn, 0, 2, ENC(rd))                 \
-        _(clw, 0, 2, ENC(rs1, rd))                  \
-        _(csw, 0, 2, ENC(rs1, rs2))                 \
-        _(cnop, 0, 2, ENC())                        \
-        _(caddi, 0, 2, ENC(rd))                     \
-        _(cjal, 1, 2, ENC())                        \
-        _(cli, 0, 2, ENC(rd))                       \
-        _(caddi16sp, 0, 2, ENC())                   \
-        _(clui, 0, 2, ENC(rd))                      \
-        _(csrli, 0, 2, ENC(rs1))                    \
-        _(csrai, 0, 2, ENC(rs1))                    \
-        _(candi, 0, 2, ENC(rs1))                    \
-        _(csub, 0, 2, ENC(rs1, rs2, rd))            \
-        _(cxor, 0, 2, ENC(rs1, rs2, rd))            \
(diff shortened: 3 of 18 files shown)


jserv (changes requested):
Eliminate the ubpf specific code. In fact, you already construct a new JIT backend.

src/riscv_private.h
```diff
@@ -116,11 +110,32 @@ struct riscv_internal {
     uint32_t csr_mip;      /* Machine interrupt pending */
     uint32_t csr_mbadaddr;
 
-    bool compressed;       /**< current instruction is compressed or not */
+    bool compressed; /**< current instruction is compressed or not */
+#if !RV32_HAS(JIT)
     block_map_t block_map; /**< basic block map */
+#else
+    struct cache *block_cache;
+#ifdef MIR
```
jserv: Drop MIR specific code.

src/rv32_template.c
```diff
@@ -4,42 +4,73 @@
...
-RVOP(nop, { rv->X[rv_reg_zero] = 0; })
+RVOP(nop, { rv->X[rv_reg_zero] = 0; }, {RVCODEGEN(nop, {})})
 
 /* LUI is used to build 32-bit constants and uses the U-type format. LUI
  * places the U-immediate value in the top 20 bits of the destination
  * register rd, filling in the lowest 12 bits with zeros. The 32-bit
  * result is sign-extended to 64 bits.
  */
-RVOP(lui, { rv->X[ir->rd] = ir->imm; })
+RVOP(lui, { rv->X[ir->rd] = ir->imm; }, {RVCODEGEN(lui, {
```
jserv: Instead of `RVCODEGEN`, use `X64` identifier instead. Later, we may integrate A64 backend as well.

src/rv32_template.c
```diff
@@ -4,42 +4,73 @@
...
+         emit_load_imm(state, RAX, ir->imm);
+         emit_store(state, S32, RAX, platform_parameter_registers[0],
```
jserv: Replace `platform_parameter_registers[0]` with something shorter.

src/rv32_template.c
```diff
@@ -4,42 +4,73 @@
...
+                    offsetof(struct riscv_internal, X) + 4 * ir->rd);
```
jserv: I anticipate the presence of template-based code generation, which necessitates minimal descriptions of target instructions without the need for complete C source code.

src/rv32_template.c
jserv: Can you avoid the repeated `lui` declaration?

src/cache.c
```diff
@@ -3,20 +3,19 @@
...
-
-#include "cache.h"
 #include "mpool.h"
 #include "utils.h"
 
 /* THRESHOLD is set to identify hot spots. Once the frequency of use for a block
  * exceeds the THRESHOLD, the JIT compiler flow is triggered.
  */
-#define THRESHOLD 1000
+#define THRESHOLD 4096
```
jserv, line 22: Add `FIXME` to indicate the need of an effective (dynamic) profiler.

src/rv32_template.c
qwe661234: > Can you avoid the repeated `lui` declaration?

Does you mean rewrite to `RVCODEGEN({})`?

src/rv32_template.c
qwe661234: Does you mean writing IR here, and this IR would be parsed and translatedin to the current C code by python script?

src/rv32_template.c
qwe661234: > Instead of `RVCODEGEN`, use `X64` identifier instead. Later, we may integrate A64 backend as well.

Actually, every lines of machine code are generated from API, so we can switch the target machine in the API.

jserv: I expect the instruction such as "lui" appears once within the scope of `RVOP` instead of twice.

src/rv32_template.c
jserv: Exactly. The template-based JIT code generation is mapped to the use of a custom domain-specific language (DSL), which is interpreted by our Python script. Having minimal descriptions of target instructions without the need for complete C source code makes sense for both maintainability and future machine code lifting. See [Rellume — Lift machine code to LLVM IR](https://github.com/aengelke/rellume),

jserv: Meanwhile, you can include a brief documentation section at the beginning of `rv32_template.c` explaining the use of DSL during JIT code generation.

Makefile
```diff
@@ -118,6 +118,22 @@ gdbstub-test: $(BIN)
 	$(Q).ci/gdbstub-test.sh && $(call notice, [OK])
 endif
 
+ENABLE_JIT ?= 0
+$(call set-feature, JIT)
+ifeq ($(call has, JIT), 1)
+OBJS_EXT += ubpf_jit_x86_64.o 
+uname_p := $(shell uname -p)
+ifneq ($(uname_p), x86_64)
+$(error JIT mode only support for x86_64 target now.)
```
jserv: Improve the grammar.

jserv (changes requested):
After removing the outdated ubpf-jit code, rename `ubpf_jit_x86_64.[ch]` to `jit-x64.[ch]`, which serves as the infrastructure for the tier-1 JIT compiler.

src/jit_x64.c
```diff
@@ -0,0 +1,745 @@
...
+    BPF_REG_2,
+    BPF_REG_3,
+    BPF_REG_4,
+    BPF_REG_5,
+    BPF_REG_6,
+    BPF_REG_7,
+    BPF_REG_8,
+    BPF_REG_9,
+    BPF_REG_10,
+    _BPF_REG_MAX,
```
jserv: Refine these names for identifiers.

src/jit_x64.c
```diff
@@ -0,0 +1,745 @@
...
+#define EBPF_SIZE_W 0x00
+#define EBPF_SIZE_H 0x08
+#define EBPF_SIZE_B 0x10
+#define EBPF_SIZE_DW 0x18
+
+/* Other memory modes are not yet supported */
+#define EBPF_MODE_IMM 0x00
+#define EBPF_MODE_MEM 0x60
+
+#define EBPF_OP_ADD_IMM (EBPF_CLS_ALU | EBPF_SRC_IMM | 0x00)
```
jserv, line 79: Ditto. Use consistent naming.

src/rv32_template.c
```diff
@@ -3,43 +3,75 @@
...
  */
 
+/* Please check files gen-jit-template.py and jit_x64.[ch] to understand custom
+ * domain-specific language of tiered 1 JIT compiler, ecah opcode has
+ * corresonding API. For example, the API of `ld_imm` is `emit_load_imm`.
+ */
+
 /* Internal */
-RVOP(nop, { rv->X[rv_reg_zero] = 0; })
+RVOP(nop, { rv->X[rv_reg_zero] = 0; }, {X64({})})
```
jserv: How about using shorter form? That is,
```c
RVOP(nop, { rv->X[rv_reg_zero] = 0; }, X64 { } )
```

For interpreter path, we can define a dummy `X64` macro which ignores everything.

qwe661234: > How about using shorter form? That is,
> 
> ```c
> RVOP(nop, { rv->X[rv_reg_zero] = 0; }, X64 { } )
> ```
> 
> For interpreter path, we can define a dummy `X64` macro which ignores everything.

It can only be (X64{}), the compiler would print error for X64 {} because the MACRO of RVOP would extend incorrectly.

qwe661234: The compiler would print error for X64 {} because the MACRO of RVOP are extended incorrectly.

jserv: We can still stick to the shorter form `RVOP(nop, { rv->X[rv_reg_zero] = 0; }, X64({ }) )`, right?

Makefile
```diff
@@ -118,6 +118,22 @@ gdbstub-test: $(BIN)
 	$(Q).ci/gdbstub-test.sh && $(call notice, [OK])
 endif
 
+ENABLE_JIT ?= 0
+$(call set-feature, JIT)
+ifeq ($(call has, JIT), 1)
+OBJS_EXT += jit_x64.o 
+uname_p := $(shell uname -p)
+ifneq ($(uname_p), x86_64)
```
jserv: Reuse the variable `processor` defined in `mk/toolchain.mk`.

Makefile
```diff
@@ -118,6 +118,22 @@ gdbstub-test: $(BIN)
 	$(Q).ci/gdbstub-test.sh && $(call notice, [OK])
 endif
 
+ENABLE_JIT ?= 0
+$(call set-feature, JIT)
+ifeq ($(call has, JIT), 1)
+OBJS_EXT += jit_x64.o 
+uname_p := $(shell uname -p)
```
jserv: Drop unnecessary `uname -p`.

src/rv32_template.c
qwe661234: This form is OK.

jserv (changes requested):
Reuse processor detection in build system.

src/common.h
```diff
@@ -6,7 +6,6 @@
 #pragma once
 
 #include "feature.h"
-
```
jserv: Don't change this line.

src/jit_x64.c
jserv: Replace all identifiers which contain BPF with the refined ones.

src/jit_x64.c
jserv: Use consistent naming. No BPF should appear.

src/jit_x64.c
```diff
@@ -0,0 +1,745 @@
...
+#include <unistd.h>
+
+#include "cache.h"
+#include "decode.h"
+#include "io.h"
+#include "jit_x64.h"
+#include "state.h"
+#include "utils.h"
+
+enum bpf_register {
```
jserv: Since this C source is intended for internal use, specifically for JIT code generation, the naming scheme can follow the principle of simplicity. Therefore, we can safely drop the prefix `bpf_` or `BPF_`.

src/jit_x64.h
```diff
@@ -0,0 +1,436 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * Generic x86-64 code generation functions
+ */
+
+#ifndef JIT_X64_H
```
jserv: Use `#pragma once` instead.

src/feature.h
```diff
@@ -52,5 +52,10 @@
 #define RV32_FEATURE_ARC 0
 #endif
 
+/* Just-in-time compiler */
```
jserv: Rephrase as "Experimental just-in-time compiler"

src/jit_x64.h
```diff
@@ -0,0 +1,436 @@
...
+
+#include <assert.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+
+#include "riscv_private.h"
+
+#define RAX 0
```
jserv: Use `enum` instead of defines for these registers.

src/jit_x64.h
```diff
@@ -0,0 +1,436 @@
...
+    uint32_t retpoline_loc;
+    struct offset_map *offset_map;
+    int num_insn;
+    struct jump *jumps;
+    int num_jumps;
+};
+
+struct jit_state *init_state(size_t size);
+
+uint32_t translate_x86_64(riscv_t *rv, block_t *block);
```
jserv: Always refer to the architecture as "x64" instead of "x86-64" since we are unlikely to support 32-bit x86 for the JIT compiler. We have plans and commitments to support two targets: x64 and Arm64.

src/cache.c
```diff
@@ -3,20 +3,21 @@
  * "LICENSE" for information on usage and redistribution of this file.
  */
 
+#include "cache.h"
```
jserv: Don't change the order of local header inclusion.

src/jit_x64.c
```diff
@@ -0,0 +1,745 @@
+// Copyright (c) 2015 Big Switch Networks, Inc
```
jserv: Drop the original licensing declaration since we are reworking the ubpf implementation. We can mention the following:
> This JIT implementation has undergone extensive modifications, heavily relying on the ubpf_jit_x86_64.[ch] from ubpf. The original ubpf_jit_x86_64.[ch] file served as the foundation and source of inspiration for adapting and tailoring it specifically for this JIT implementation. Therefore, credit and sincere thanks are extended to ubpf for their invaluable work.

Check src/map.[ch] for the way in which we provide credit for the previous work.

src/jit_x64.h
```diff
@@ -0,0 +1,436 @@
+// Copyright (c) 2015 Big Switch Networks, Inc
```
jserv: Ditto. Rework the licensing notice and credit information.

src/jit_x64.h
```diff
@@ -0,0 +1,436 @@
...
+{
+    emit_basic_rex(state, size == S64, dst, src);
+    if (size == S8 || size == S16) {
+        /* movzx */
+        emit1(state, 0x0f);
+        emit1(state, size == S8 ? 0xb6 : 0xb7);
+    } else if (size == S32) {
+        /* mov */
+        emit1(state, 0x8b);
+    } else if (size == S64) {
```
jserv: Shall we care about S64 case?

src/jit_x64.c
jserv: Apply consistent naming scheme. That is,
```c
enum vm_reg {
    VM_REG0 = 0,
    VM_REG1,
    ...
    VM_REG10,
    N_VM_REGS,
};
```

src/emulate.c
```diff
@@ -1030,11 +924,33 @@ void rv_step(riscv_t *rv, int32_t cycles)
             }
         }
         last_pc = rv->PC;
-
-        /* execute the block */
+#if RV32_HAS(JIT)
+        /* execute by tiered 1 JIT compiler */
```
jserv: The correct term is "tier 1 JIT compiler" and "tiered compilation."

src/jit_x64.c
```diff
@@ -0,0 +1,745 @@
...
+#define EBPF_MODE_JSGT 0x60
+#define EBPF_MODE_JSGE 0x70
+#define EBPF_MODE_CALL 0x80
+#define EBPF_MODE_EXIT 0x90
+#define EBPF_MODE_JLT 0xa0
+#define EBPF_MODE_JLE 0xb0
+#define EBPF_MODE_JSLT 0xc0
+#define EBPF_MODE_JSLE 0xd0
+
+#define EBPF_OP_JA (EBPF_CLS_JMP | EBPF_MODE_JA)
```
jserv: Drop unused macros. Check if the following macros are required or not.

src/rv32_template.c
```diff
@@ -3,43 +3,85 @@
  * specification version 20191213.
  */
 
+/* Please check files gen-jit-template.py and jit_x64.[ch] to understand custom
```
jserv: Improve the documentation.

jserv (changes requested):
Improve the git commit messages to explicitly highlight the performance comparison between the rv32emu interpreter, QEMU, and the proposed tier-1 JIT compiler. This comparison will focus on execution time and resource utilization, with particular attention to memory consumption.

src/emulate.c
```diff
@@ -292,9 +166,14 @@ static block_t *block_alloc(riscv_t *rv)
...
     block->n_insn = 0;
     block->predict = NULL;
+#if RV32_HAS(JIT)
+    block->hot = false;
+    block->backward = false;
+#endif
     return block;
 }
 
+#if !RV32_HAS(JIT)
```
jserv, line 309: Function `map_hash` is only used by interpreter.

jserv:
This implementation has been successfully tested on x86-based macOS 10.15.7. However, note that, in compliance with Apple's security policy, additional entitlements are necessary to permit the execution of JIT-compiled code. See https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_cs_allow-jit

Quoted from [JDK-8234930](https://bugs.openjdk.org/browse/JDK-8234930) "Use `MAP_JIT` when allocating pages for code cache on macOS":
> This patch is a strong prerequest for mac-aarch64 port. it was made slightly different to upstream patch so on platforms other than macos this patch equals to "nothing".

qwe661234:
> Improve the git commit messages to explicitly highlight the performance comparison between the rv32emu interpreter, QEMU, and the proposed tier-1 JIT compiler. This comparison will focus on execution time and resource utilization, with particular attention to memory consumption.

Currently, we don't have effective memory examination tool to examinate the real memory usage. The memory consumption we can gain is the result without counting `mmap`.

src/riscv.c
```diff
@@ -143,10 +155,15 @@ bool rv_enables_to_output_exit_code(riscv_t *rv)
...
 }
 
+
 void rv_delete(riscv_t *rv)
 {
     assert(rv);
+#if !RV32_HAS(JIT)
     block_map_destroy(rv);
+#else
+    cache_free(rv->block_cache, NULL);
```
jserv: The second parameter is set to `NULL`, but there is no check for callback in the implementation of `cache_free`, resulting in segmentation fault.
Send another pull request to fix prior to the merge of this pull request.

jserv:
> Currently, we don't have effective memory examination tool to examinate the real memory usage. The memory consumption we can gain is the result without counting `mmap`.

Simply show what you have in git commit messages where we tend to track the development progress.

src/jit_x64.c
```diff
@@ -0,0 +1,594 @@
...
+    VM_REG_2,
+    VM_REG_3,
+    VM_REG_4,
+    VM_REG_5,
+    VM_REG_6,
+    VM_REG_7,
+    VM_REG_8,
+    VM_REG_9,
+    VM_REG_10,
+    _VM_REG_MAX,
```
jserv, line 47: Replace `_VM_REG_MAX` with `N_VM_REGS` for consistent naming scheme.

src/jit_x64.c
```diff
@@ -0,0 +1,594 @@
...
+/* This JIT implementation has undergone extensive modifications, heavily
+ * relying on the ubpf_jit_x86_64.[ch] from ubpf. The original
+ * ubpf_jit_x86_64.[ch] file served as the foundation and source of inspiration
+ * for adapting and tailoring it specifically for this JIT implementation.
+ * Therefore, credit and sincere thanks are extended to ubpf for their
+ * invaluable work.
+ * Reference:
+ *   https://github.com/iovisor/ubpf/blob/main/vm/ubpf_jit_x86_64.c
+ */
+
```
jserv, line 16: Add compilation guard. That is,
```c
#if !defined(__x86_64__)
#error "This implementation is dedicated to x86-64."
#endif
```

See https://github.com/sysprog21/kvm-host/blob/master/src/arch/x86/vm.c as an example.

src/jit_x64.c
```diff
@@ -0,0 +1,594 @@
...
+#define X64_OP_DIV_IMM (X64_CLS_ALU | X64_SRC_IMM | 0x30)
+#define X64_OP_DIV_REG (X64_CLS_ALU | X64_SRC_REG | 0x30)
+#define X64_OP_MOD_IMM (X64_CLS_ALU | X64_SRC_IMM | 0x90)
+#define X64_OP_MOD_REG (X64_CLS_ALU | X64_SRC_REG | 0x90)
+
+#define STACK_SIZE 512
+#define MAX_INSTS 65536
+
+#if !defined(_countof)
+#define _countof(array) (sizeof(array) / sizeof(array[0]))
```
jserv: Use macro `ARRAYS_SIZE` defined in `src/common.h` instead.

src/jit_x64.c
```diff
@@ -0,0 +1,594 @@
...
+#endif
+
+#if RV32_HAS(EXT_M)
+static void muldivmod(struct jit_state *state,
```
jserv, line 71: Is this forward declaration required?

src/jit_x64.c
```diff
@@ -0,0 +1,594 @@
...
+ * There are two common x86-64 calling conventions, as discussed at
+ * https://en.wikipedia.org/wiki/X64_calling_conventions#x86-64_calling_conventions
+ *
+ * Please Note: R12 is special and we are *not* using it. As a result, it is
+ * omitted from the list of non-volatile registers for both platforms (even
+ * though it is, in fact, non-volatile).
+ */
+
+#if defined(_WIN32)
+static int nonvolatile_reg[] = {RBP, RBX, RDI, RSI, R13, R14, R15};
```
jserv, line 215: Add `const` qualifier when possible.

src/emulate.c
```diff
@@ -978,6 +987,10 @@ static block_t *block_find_or_translate(riscv_t *rv)
     return next;
 }
 
+#if RV32_HAS(JIT)
+typedef void (*exec_tired1_block_func_t)(riscv_t *rv, uintptr_t);
```
jserv: Replace `exec_tired1_block_func_t ` with `exec_block_func_t ` because we don't have to distinguish from tier-1 and tier-2 compilers at the moment.

jserv: In addition, "tired1" is considered as a typo.

jserv:
When creating tables or performance metrics, you can simply name this Tier-1 JIT compiler as "rv32emu-T1C" or "T1C."

src/emulate.c
```diff
@@ -370,8 +382,14 @@ static bool is_branch_taken = false;
 /* record the program counter of the previous block */
 static uint32_t last_pc = 0;
 
+#if RV32_HAS(JIT)
+/* record whether the block is replaced by cache. If so, clear the EBB
```
jserv: Refine the use of "EBB" since we move to block chaining.

jserv (changes requested):
Update CI pipeline to validate the JIT compiler.

jserv:
Consistently utilize the phrase "tier-1 JIT compiler" or "tier-1 compiler," and you can opt for the shortened form "T1C" for brevity.

src/emulate.c
```diff
@@ -1030,11 +1051,33 @@ void rv_step(riscv_t *rv, int32_t cycles)
...
+        struct jit_state *state = rv->jit_state;
+        if (block->hot) {
+            ((exec_block_func_t) state->buf)(
+                rv, (uintptr_t) (state->buf + block->offset));
+            prev = NULL;
+            continue;
+        } /* check if using frequency of block exceed threshold */
+        else if ((block->backward &&
+                  cache_freq(rv->block_cache, block->pc_start) >= 1024) ||
+                 (cache_hot(rv->block_cache, block->pc_start))) {
```
jserv: Remove the parentheses out of `cache_hot`.

src/jit_x64.c
```diff
@@ -0,0 +1,593 @@
...
+    VM_REG_2,
+    VM_REG_3,
+    VM_REG_4,
+    VM_REG_5,
+    VM_REG_6,
+    VM_REG_7,
+    VM_REG_8,
+    VM_REG_9,
+    VM_REG_10,
+    _VM_REG_MAX,
```
jserv: Replace `_VM_REG_MAX` with `N_VM_REGS` for consistent naming scheme.

src/jit_x64.c
```diff
@@ -0,0 +1,593 @@
...
+#define X64_SRC_REG 0x08
+#define X64_OP_MUL_IMM (X64_CLS_ALU | X64_SRC_IMM | 0x20)
+#define X64_OP_MUL_REG (X64_CLS_ALU | X64_SRC_REG | 0x20)
+#define X64_OP_DIV_IMM (X64_CLS_ALU | X64_SRC_IMM | 0x30)
+#define X64_OP_DIV_REG (X64_CLS_ALU | X64_SRC_REG | 0x30)
+#define X64_OP_MOD_IMM (X64_CLS_ALU | X64_SRC_IMM | 0x90)
+#define X64_OP_MOD_REG (X64_CLS_ALU | X64_SRC_REG | 0x90)
+
+#define STACK_SIZE 512
+#define MAX_INSTS 1024
```
jserv: Replace `MAX_INSTS` with `MAX_INSNS` for consistency.

src/jit_x64.c
```diff
@@ -0,0 +1,593 @@
...
+#define _(inst, can_branch, insn_len, reg_mask) [rv_insn_##inst] = do_##inst,
+    RV_INSN_LIST
+#undef _
+    /* Macro operation fusion instructions */
+#define _(inst) [rv_insn_##inst] = do_##inst,
+    FUSE_INSN_LIST
+#undef _
+};
+/* clang-format on */
+typedef void (*codegen_tired1_block_func_t)(struct jit_state *,
```
jserv: Replace `codegen_tired1_block_func_t` with `codegen_block_func_t`.

src/rv32_template.c
```diff
@@ -3,43 +3,86 @@
  * specification version 20191213.
  */
 
+/* Please review the files gen-jit-template.py and jit_x64.[ch] to comprehend
```
jserv: Rather than specifying "Please review the files gen-jit-template.py and jit_x64.[ch]," it is more beneficial to provide an overview of the domain-specific language (DSL) specification for JIT code generation. Documentation serves the purpose of simplifying maintenance tasks, rather than merely serving as an overly simplified reference.

jserv:
The proposed JIT compiler does not encompass effective code generation for RV32F. You should clarify which instruction sets and extensions are supported.

While launching the Quake via `make quake`, I am aware of the following:
```
Playing registered version.
Console initialized.
Exe: 20:12:56 Sep  1 2023
 8.0 megabyte heap
792k surface cache
Assertion failed: (NULL), function do_fmvwx, file src/rv32_jit_template.c, line 556.
```

Another reproducer:
```
 $ build/rv32emu build/maj2random.elf
maj_random (NROUNDS=2)                 count     mean(x)     mean(y) variance(x) variance(y)  std-dev(x)  std-dev(y)
(             0, (0 - 1K)/8K )    999999,057     0.49915     0.50594     0.08122     0.08092     0.28499     0.28447
Assertion failed: (NULL), function do_fmvwx, file src/rv32_jit_template.c, line 556.
```

jserv:
Tested on Ubuntu 20.04, Python 3.8 generates the following error messages:
```
  File "tools/gen-jit-template.py", line 153
    match items[0]:
          ^
SyntaxError: invalid syntax
make: *** [Makefile:130: src/rv32_jit_template.c] Error 1
$ python3 --version
Python 3.8.10
```

I installed Python 3.9.5 for comparison, and the above message still appear.
According to [Status of Python versions](https://devguide.python.org/versions/), Python 3.8 and 3.9 are still supported by Python Software Foundation.

qwe661234:
> Tested on Ubuntu 20.04, Python 3.8 generates the following error messages:
> 
> ```
>   File "tools/gen-jit-template.py", line 153
>     match items[0]:
>           ^
> SyntaxError: invalid syntax
> make: *** [Makefile:130: src/rv32_jit_template.c] Error 1
> $ python3 --version
> Python 3.8.10
> ```
> 
> I installed Python 3.9.5 for comparison, and the above message still appear. According to [Status of Python versions](https://devguide.python.org/versions/), Python 3.8 and 3.9 are still supported by Python Software Foundation.

As alternative, I rewrite python switch into if-else.

jserv:
The JIT-compiled code can initiate Doom, but it crashes when attempting to start the game.

Tested on Ubuntu Linux 20.04.

src/rv32_template.c
```diff
@@ -3,43 +3,123 @@
  * specification version 20191213.
  */
 
+/* Currently, the tier-1 JIT compiler only supports for RV32-IMC extension,
+ * RV32-A and RV32-F extension are unsupported.
+ */
+
+/* size: size of data
```
jserv: Add general descriptions prior to the following. Make them informative.

jserv:
While T1C is capable of running Quake, it does not behave as expected when compared to the interpreter. I encountered an unexpected crash with Quake. To reproduce this, first run `make quake` and watch the demo scene without any keystrokes. You will then encounter a segmentation fault.
```
You got the shells
You got the Grenade Launcher
You receive 25 health
You get 2 rockets  
You got the rockets
You got the nailgun
You got the nails
You got the nails
You receive 15 health
You receive 15 health
You got the rockets
/bin/sh: line 1: 71739 Segmentation fault: 11  ../build/rv32emu quake.elf
```

Th log generated by the interpreter:
```
You got the shells
You got the Grenade Launcher
You receive 25 health
You get 2 rockets  
You got the rockets
You got the nailgun
You got the nails
You got the nails
You receive 15 health
You receive 15 health
You got the rockets
You get 2 rockets  
You got the nails
You got the nails
You receive 25 health
You got the gold key
Playing demo from demo2.dem.
You got the shells
You got the shells
You got the nails
You got armor
You got the nails
You receive 25 health
You got the Super Nailgun
You got the nails
You get 2 rockets  
```

qwe661234:
> While T1C is capable of running Quake, it does not behave as expected when compared to the interpreter. I encountered an unexpected crash with Quake. To reproduce this, first run `make quake` and watch the demo scene without any keystrokes. You will then encounter a segmentation fault.
> 
> ```
> You got the shells
> You got the Grenade Launcher
> You receive 25 health
> You get 2 rockets  
> You got the rockets
> You got the nailgun
> You got the nails
> You got the nails
> You receive 15 health
> You receive 15 health
> You got the rockets
> /bin/sh: line 1: 71739 Segmentation fault: 11  ../build/rv32emu quake.elf
> ```
> 
> Th log generated by the interpreter:
> 
> ```
> You got the shells
> You got the Grenade Launcher
> You receive 25 health
> You get 2 rockets  
> You got the rockets
> You got the nailgun
> You got the nails
> You got the nails
> You receive 15 health
> You receive 15 health
> You got the rockets
> You get 2 rockets  
> You got the nails
> You got the nails
> You receive 25 health
> You got the gold key
> Playing demo from demo2.dem.
> You got the shells
> You got the shells
> You got the nails
> You got armor
> You got the nails
> You receive 25 health
> You got the Super Nailgun
> You got the nails
> You get 2 rockets  
> ```

Because quake has RV32F insturction, but jit does not support it now.

jserv:
> Because quake has RV32F instructions, but T1C does not support it now.

We do need an effective fallback for RV32F and RV32A when the JIT compiler is unable to generate the corresponding machine code. A trampoline would be required for switching between translated blocks and non-JIT parts.
See https://github.com/jserv/rv32jit/blob/master/src/codegen/jitabi.cpp

src/decode.h
```diff
@@ -32,150 +32,150 @@ enum op_field {
...
-        _(cbnez, 1, 2, ENC(rs1))                    \
-        _(cslli, 0, 2, ENC(rd))                     \
-        _(clwsp, 0, 2, ENC(rd))                     \
-        _(cjr, 1, 2, ENC(rs1, rs2, rd))             \
-        _(cmv, 0, 2, ENC(rs1, rs2, rd))             \
-        _(cebreak, 1, 2, ENC(rs1, rs2, rd))         \
-        _(cjalr, 1, 2, ENC(rs1, rs2, rd))           \
-        _(cadd, 0, 2, ENC(rs1, rs2, rd))            \
-        _(cswsp, 0, 2, ENC(rs2))                    \
+#define RV_INSN_LIST                                   \
```
jserv, line 35: Refine the above comment for new field.

jserv:
Finally, the JIT compiler is incorporated into this emulator repository.

Issue #286, opened 2023-11-30
CI: Generate Docker image with Alpine Linux


jserv:
Currently, the CI pipeline can generate Docker images, but the size of these images is quite large, which is not ideal for showcasing a small yet powerful RISC-V instruction set simulator. To address this, we can consider rebuilding the Docker image using [Alpine Linux](https://www.alpinelinux.org/), a security-focused, lightweight Linux distribution that uses musl libc and busybox. This would not only reduce the image size but also provide a minimal runtime environment.

henrybear327:
This will take a while, as the package management system is `apk` and the package names are usually different from `apt`. 

I will adapt accordingly.

visitorckw:
The substantial size of the current Docker image is primarily attributed to the riscv-gnu-toolchain, surpassing 1GB. While transitioning to Alpine as the base image can reduce the size (from around 100 MB to approximately 5 MB), it's still likely that the overall image size may well exceed 1GB.

henrybear327:
Maybe we don't need to keep around all the binaries compiled from the toolchain. I will take a look over there! Thanks for the pointers.

henrybear327:
Image size evolution (the total size shown is accumulative, with respect to the previously listed items) 

- `Ubuntu:22:04` -> 69.4 MB
- Copying the source code (`tests` folder itself is about 500 MB) ->  831 MB
- Install extra packages (e.g. `libsdl2-dev`) -> 1.49GB

For the GCC toolchain
- Using xpack toolchain -> 3.56GB
- Copying from the self-built GCC image -> 2.12GB

jserv:
> Maybe we don't need to keep around all the binaries compiled from the toolchain. I will take a look over there! Thanks for the pointers.

I don't believe it is necessary to include the prebuilt GNU Toolchain for RISC-V in the Docker image. Users interested in running RISC-V Architecture Tests can manually download the prebuilt toolchain via package managers like [xpack](https://xpack.github.io/). The Docker image should provide the `rv32emu `executable, sample RISC-V ELF files, and auxiliary scripts for downloading the GNU Toolchain and sail as needed.

henrybear327:
> sample RISC-V ELF files

A subset of the elf files from the build directory will be kept

> auxiliary scripts for downloading the GNU Toolchain and sail as needed

Will add this in the PR #290

henrybear327:
[...]

PR #290, opened 2023-12-12
Remove GNU Toolchain for RISC-V from Dockerfile

henrybear327:
An experiment of using `xpack` as the GCC toolchain instead of our self-compiled docker image.

#286

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,34 +1,38 @@
-FROM sysprog21/rv32emu-gcc as base_gcc
 FROM sysprog21/rv32emu-sail as base_sail
 
-FROM ubuntu:22.04 as final
+FROM ubuntu:22.04 as base_gcc
 
-# Install extra packages for the emulator to compile and execute with full capabilities correctly
+# install extra packages for the emulator to compile with full capabilities correctly
 RUN apt-get update && \
     DEBIAN_FRONTEND=noninteractive apt-get install -y \
-    libsdl2-dev libsdl2-mixer-dev python3-pip git && \
+    build-essential curl git && \
     rm -rf /var/lib/apt/lists/*
-
-RUN python3 -m pip install git+https://github.com/riscv/riscof 
-
-# set up the timezone
-ENV TZ=Asia/Taipei
-RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
 
 # copy in the source code
 WORKDIR /home/root/rv32emu
 COPY . .
 
-# Copy the GNU Toolchain files
-ENV RISCV=/opt/riscv
-ENV PATH=$RISCV/bin:$PATH
-COPY --from=base_gcc /opt/riscv/ /opt/riscv/
-
-# replace the emulator (riscv_sim_RV32) with the arch that the container can execute 
-RUN rm /home/root/rv32emu/tests/arch-test-target/sail_cSim/riscv_sim_RV32
-COPY --from=base_sail /home/root/riscv_sim_RV32 /home/root/rv32emu/tests/arch-test-target/sail_cSim/riscv_sim_RV32
-
 # generate execution file for rv32emu and rv_histogram
 RUN make
 RUN make tool
-ENV PATH=/home/root/rv32emu/build:$PATH
+
+FROM ubuntu:22.04 as final
+
+# set up the timezone
+ENV TZ=Asia/Taipei
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+
+# copy in elf files and reference emulator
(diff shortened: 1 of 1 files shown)


henrybear327:
Image size evolution (the total size shown is accumulative, with respect to the previously listed items) 

- `Ubuntu:22:04` -> 69.4 MB
- Copying the source code (`tests` folder itself is about 500 MB) ->  831 MB
- Install extra packages (e.g. `libsdl2-dev`) -> 1.49GB

For the GCC toolchain
- Using xpack toolchain -> 3.56GB
- Copying from the self-built GCC image -> 2.12GB

Dockerfile
```diff
@@ -1,9 +1,23 @@
-FROM sysprog21/rv32emu-gcc as base_gcc
 FROM sysprog21/rv32emu-sail as base_sail
 
+FROM ubuntu:22.04 as base_gcc
+
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y \
+    build-essential curl && \
```
visitorckw: Due to the presence of a git submodule in the source code, it's necessary to install git here in order to successfully execute the subsequent 'RUN make' command.

henrybear327:
Using xpack will actually increase the image size, so maybe we should drop this PR in the end.

jserv:
> Using xpack will actually increase the image size, so maybe we should drop this PR in the end.

The goal here is to utilize scripts for downloading and extracting the toolchain from xpack, similar to those already in use by our CI pipeline. This approach should help in significantly reducing the size of our Docker image.

henrybear327:
> > Using xpack will actually increase the image size, so maybe we should drop this PR in the end.
> 
> The goal here is to utilize scripts for downloading and extracting the toolchain from xpack, similar to those already in use by our CI pipeline. This approach should help in significantly reducing the size of our Docker image.

Got it. So the following changes would be made
- add a script that can automatically install the GCC toolchain 
- remove the existing toolchain from the image
- compile the rv32emu and relevant binaries in the intermediate images and copy it to the final image

jserv:
> Got it. So the following changes would be made
> * add a script that can automatically install the GCC toolchain
> * remove the existing toolchain from the image
> * compile the rv32emu and relevant binaries in the intermediate images and copy it to the final image

Take [run #7295920920](https://github.com/sysprog21/rv32emu/actions/runs/7295920920), for example. It takes over 11 minutes, yet the actual time spent preparing the build environment and launching the build process accounts for only half of this elapsed time.

jserv (changes requested):
Ensure `git` is installed as a prerequisite; otherwise, you will encounter the following error.

```
 > [linux/amd64 base_gcc 5/6] RUN make:
0.108 Makefile:86: No sdl2-config in $PATH. Check SDL2 installation in advance
0.110 /bin/sh: 1: pkg-config: not found
0.180 Check the file build/.config for configured items.
0.224   CC	build/map.o
0.299   CC	build/utils.o
0.342 git submodule update --init src/softfloat/
0.351 make: git: No such file or directory
0.352 make: *** [Makefile:73: src/softfloat/build/Linux-RISCV-GCC/Makefile] Error 127
```

henrybear327:
Added `libsdl2-dev libsdl2-mixer-dev` to the intermediate image, so we can compile a complete `rv32emu` binary. 

Should `libsdl2-dev libsdl2-mixer-dev` also be presented in the final image?

Also, since this dockerfile is not for development anymore (unless I misunderstand the intention of removal of the GCC toolchain), should we just copy the elf files in the build folders, and leave other source files, etc. out?

henrybear327:
> CI / host-arm64

It's now the same compile time as `CI / host-arm64` Github CI job, since it's the QEMU emulating arm64 that's slow. 

Image size is now basically 65 MB + size of directories copied in from rv32emu repo

jserv:
> Should `libsdl2-dev` and `libsdl2-mixer-dev` also be presented in the final image?

No, the Docker image should be limited to only the essential components of rv32emu, excluding SDL display and audio, as these can lead to issues in containerized environments. Therefore, just remove the `libsdl2-dev` and `libsdl2-mixer-dev` packages.

> should we just copy the elf files in the build directory, and leave other source files, etc. out?

Yes, exactly. Users can always view the source code through the GitHub website. Only the built ELF files (for both the host and the RISC-V target) are preserved.

jserv:
The [recent CI record](https://github.com/sysprog21/rv32emu/actions/runs/7377774989/job/20072140463) indicates that building and publishing a Docker image to Docker Hub requires 8 minutes. It looks pretty well since we can lower the elapsed time by installing less packages.

jserv (changes requested):
Drop unnecessary packages such as SDL2.

henrybear327:
Updated as requested/discussed.

jserv:

PR #291, opened 2023-12-14
Fix cache_free fucntion

qwe661234:
After we introduced the memory pool for blocks and IRs, we don't need to free the memory of blocks and IRs by cache.

### DIFF
--- a/src/cache.c
+++ b/src/cache.c
@@ -515,31 +515,16 @@ void *cache_put(cache_t *cache, uint32_t key, void *value)
     return delete_value;
 }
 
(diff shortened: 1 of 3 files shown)


jserv (changes requested):
Fix undefined behavior reported by the optimizer:
```
src/cache.c:521:26: warning: iteration 4 invokes undefined behavior [-Waggressive-loop-optimizations]
  521 |         free(cache->lists[i]);
      |              ~~~~~~~~~~~~^~~
src/cache.c:520:23: note: within this loop
  520 |     for (int i = 0; i < THRESHOLD; i++)
      |                       ^
```

Issue #75, opened 2022-10-20
Run rv32emu inside web browser via WebAssembly translation


jserv:
[iamlouk/riscv64-sim](https://github.com/iamlouk/riscv64-sim) is written in C and capable of being translated into WebAssembly. Check its [web demo](https://louknr.net/projs/riscv64-sim/www/). Some web-based (open source) RISC-V simulations:
* [WebRISC-V](https://github.com/Mariotti94/WebRISC-V): 5-stage Graphical Pipeline 32/64-bit Simulator
* [emulsiV](https://github.com/ESEO-Tech/emulsiV)

The goal of this task is to expand `rv32emu` to the simulator, which will be used to teach computer architecture fundamentals, based on the existing RISC-V codebase.
Incomplete attempt: [rv32emu-wasm](https://github.com/wokwi/rv32emu-wasm)

jserv:
We might need RISC-V instruction decoder for web-based execution. See [rvcodec.js: an online encoder/decoder for RISC-V instructions](https://luplab.cs.ucdavis.edu/2022/06/03/rvcodec-js.html)

jserv:
Recent rv32emu relies on tail-call optimization (TCO), which causes build problems with [Emscripten](https://emscripten.org/). Thus, we have to turn it off. Apply the following changes:
```diff
diff --git a/src/common.h b/src/common.h
index 62b66bd..374bcf3 100644
--- a/src/common.h
+++ b/src/common.h
@@ -43,7 +43,7 @@
  * around this, the compiler attribute 'musttail' is used, which forces TCO
  * even without optimizations enabled.
  */
-#if defined(__has_attribute) && __has_attribute(musttail)
+#if defined(__has_attribute) && __has_attribute(musttail) && !defined(__wasm__)
 #define MUST_TAIL __attribute__((musttail))
 #else
 #define MUST_TAIL
diff --git a/src/io.c b/src/io.c
index 5d7a37d..f3de262 100644
--- a/src/io.c
+++ b/src/io.c
@@ -17,8 +17,8 @@
 #include "io.h"
 
 static uint8_t *data_memory_base;
-/* set memory size to 2^32 bytes */
-#define MEM_SIZE 0x100000000ULL
+/* set memory size to 2^31 bytes */
+#define MEM_SIZE 0x10000000ULL
 
 memory_t *memory_new()
 {
```

Then, build this project via `emcc`:
```shell
[...]

PR #293, opened 2023-12-15
Enable build rv32emu using Emscripten

ChinYikMing:
- Tail-call and link-time optimization can be enabled during the build
- SoftFloat archive object file needs to be built using appropriate $(CC) and $(AR) in order to make linking phase of rv32emu successful

The output of `fcalc.elf`(running in Chrome) is error-free, indicating that linking with the compiled softfloat archive object is OK.
```
Performed 12 tests, 0 failures, 100% success rate.
inferior exit code 0
```

Afterwards, rv32emu's core can be used as the backend of the simulator (see #75)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,14 +19,14 @@ CFLAGS += -D DEFAULT_ARGS_ADDR=0xFFFFF000
 # Enable link-time optimization (LTO)
 ENABLE_LTO ?= 1
 ifeq ($(call has, LTO), 1)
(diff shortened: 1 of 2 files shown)


jserv:

ChinYikMing:
To reflect the changes of this PR, softfloat package should be modified, see [PR](https://github.com/sysprog21/berkeley-softfloat-3/pull/1), @jserv

PR #295, opened 2023-12-16
Fix potential security issue in benchmark script

### LINKED ISSUE #294: Fix potential security issue in benchmark script
In [tests/bench-aggregator.py:7](https://github.com/sysprog21/rv32emu/blob/612c89da7f9a59881be3baaf525c1b9c09ca5d7e/tests/bench-aggregator.py#L7-L7),  found `subprocess` function `$FUNC` with `shell=True`. This is dangerous because this call will spawn the command using a shell process.

Reported by Codacy.

ChinYikMing:
Close #294

### DIFF
--- a/tests/bench-aggregator.py
+++ b/tests/bench-aggregator.py
@@ -3,8 +3,14 @@
 import json
 import subprocess
 
-def run_benchmark(cmd):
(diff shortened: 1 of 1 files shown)


jserv:

PR #303, opened 2023-12-22
Fix syscall error handling

visitorckw:
Several system call implementations lack proper error handling, potentially returning 0 even when an error occurs. This patch addresses the issue by introducing a proper check to ensure correct return values when errors occur.


qwe661234 (approved):

jserv:

Issue #297, opened 2023-12-16
jit: Incomplete `muldivmod`


jserv:
Function `muldivmod` in `src/jit_x64.c ` is incomplete:
```c
    /* Short circuit for imm == 0 */
    if (!reg && imm == 0) {
        assert(NULL);                                                                                                                                         
        if (div || mul) {
            /* For division and multiplication, set result to zero. */
            emit_alu32(state, 0x31, dst, dst);
        } else {
            /* For modulo, set result to dividend. */
            emit_mov(state, dst, dst);
        }
        return;
    }
```

When `imm` equals to zero, there is no chance to resume the execution.

PR #304, opened 2023-12-22
Introduce a tier-1 JIT compiler based on aarch64 architecture

### LINKED ISSUE #296: jit: code generation tool should be aware of comments
When improving comments in `rv32_template.c`, I discovered that `tools/gen-jit-template.py` does not process comments, meaning it blindly converts the code template without skipping them. This could potentially cause unexpected problems.

That is the reason I wrote
```
 * Example:
 *
 *   addi,
 *   { rv->X[ir->rd] = (int32_t) (rv->X[ir->rs1]) + ir->imm; },
 *   x64({
 *       ld, S32, RAX, X, rs1;
 *       alu32_imm, 32, 0x81, 0, RAX, imm;
 *       st, S32, RAX, X, rd;
 *   })
```

without `RVOP` and `X64` (capitalized).

qwe661234:
We follow the template and API of X64 to implement A64 tier-1 JIT compiler.

* Perfromance

| Metric   | rv32emu-T1C | qemu  |
|----------|-------------|-------|
|aes	   |        0.034|  0.045|
|puzzle	   |       0.0115| 0.0169|
|pi        |        0.035|  0.032|
|dhrystone |	    1.914|  2.005|
|Nqeueens  |	     3.87|  2.898|
|qsort-O2  |	    7.819| 11.614|
|miniz-O2  |	    7.604|  3.803|
|primes-O2 |	   10.551|  5.986|
|sha512-O2 |	    6.497|  2.853|
|stream	   |        52.25| 45.776|

As demonstrated in the memory usage analysis below, the tier-1 JIT compiler utilizes less memory than QEMU across all benchmarks.

* Memory usage

| Metric   | rv32emu-T1C |   qemu  |
|----------|-------------|---------|
|aes	   |      183,212|1,265,962|
|puzzle	   |      145,239|  891,357|
|pi        |      144,739|  872,525|
|dhrystone |	  146,282|  853,256|
|Nqeueens  |	  146,696|  854,174|
|qsort-O2  |	  146,907|  856,721|
|miniz-O2  |	  157,475|  999,897|
|primes-O2 |	  142,356|  851,661|
|sha512-O2 |	  145,369|  901,136|
|stream	   |      157,975|  955,809|

Related: #238
Close: #296


Makefile
```diff
@@ -121,15 +121,17 @@ endif
 ENABLE_JIT ?= 0
 $(call set-feature, JIT)
 ifeq ($(call has, JIT), 1)
-OBJS_EXT += jit_x64.o 
+OBJS_EXT += jit.o 
 ifneq ($(processor), x86_64)
-$(error JIT mode only supports for x64 target currently.)
+ifneq ($(processor), aarch64)
```
jserv: Use `$(filter` instead.
See https://www.gnu.org/software/make/manual/html_node/Text-Functions.html

src/jit.c
```diff
@@ -0,0 +1,1528 @@
...
+    R14,
+    R15,
+};
+
+#elif defined(__aarch64__)
+#define JUMP_LOC jump_loc
+/* Special values for target_pc in struct jump */
+#define TARGET_PC_EXIT ~UINT32_C(0)
+#define TARGET_PC_ENTER (~UINT32_C(0) & 0x0101)
+// This is guaranteed to be an illegal A64 instruction.
```
jserv: Always use C-style comments.

src/jit.c
```diff
@@ -0,0 +1,1528 @@
...
+        emit_loadstorepair_immediate(state, LSP_LDPX, callee_reg[i],
+                                     callee_reg[i + 1], SP, (i + 2) * 8);
+    }
+    emit_loadstorepair_immediate(state, LSP_LDPX, R29, R30, SP, 0);
+    emit_addsub_immediate(state, true, AS_ADD, SP, SP, state->stack_size);
+    emit_unconditionalbranch_register(state, BR_RET, R30);
+#endif
+}
+
+#define X64(inst, code)                                                       \
```
jserv: Rename to `GEN`.

src/jit.c
```diff
@@ -0,0 +1,1528 @@
...
+    static void do_##inst(struct jit_state *state UNUSED, riscv_t *rv UNUSED, \
+                          rv_insn_t *ir UNUSED)                               \
+    {                                                                         \
+        code;                                                                 \
+    }
+#include "rv32_jit_template.c"
```
jserv: Ditto. Replace all occurrences of `X64` with `GEN`.

tools/gen-jit-template.py
```diff
@@ -136,6 +136,7 @@ def remove_comment(str):
 f.close()
 
 fields = {"imm", "pc", "rs1", "rs2", "rd", "shamt", "branch_taken", "branch_untaken"}
+temp_regs = {"TEMP0", "TEMP1", "TEMP2"}
```
jserv: Use `TMP0`, `TMP1`, and `TMP2` instead.

src/jit.c
```diff
@@ -0,0 +1,1528 @@
...
+// Caller saved registers (and parameter registers)
+static const int parameter_reg[] = {R0, R1, R2, R3, R4};
+static const int temp_reg[] = {R6, R7, R8};
+// Temp register for immediate generation
+static const int temp_imm_reg = R24;
+// Temp register for division results
+static const int temp_div_reg = R25;
+
+// Register assignments:
+//   BPF        Arm64       Usage
```
jserv: Correct the information for this project.

src/jit.c
```diff
@@ -0,0 +1,1528 @@
...
+}
+
+static inline void offset_map_insert(struct jit_state *state, int32_t target_pc)
+{
+    struct offset_map *map_entry = &state->offset_map[state->num_insn++];
+    map_entry->PC = target_pc;
+    map_entry->offset = state->offset;
+}
+
+
```
jserv: Remove a blank line.

Makefile
jserv: The result of `uname -m` on macOS/Arm64 is `arm64`.

src/jit.c
```diff
@@ -0,0 +1,1528 @@
...
+out:
+    return entry_loc;
+}
+
+struct jit_state *init_state(size_t size)
+{
+    struct jit_state *state = malloc(sizeof(struct jit_state));
+    state->offset = 0;
+    state->size = size;
+    state->buf = mmap(0, size, PROT_READ | PROT_WRITE | PROT_EXEC,
```
jserv, line 1537: `pthread_jit_write_protect_np` is required for Apple Silicon.
See
* https://lists.gnu.org/archive/html/qemu-devel/2021-01/msg02555.html
* https://github.com/vnmakarov/mir/blob/master/mir.c#L3777
* https://github.com/minoki/icfpc2006/blob/master/aarch64/run_jit.c
* https://github.com/LekKit/RVVM/commit/dd94a057daa508f8829cd403f95a62067f0b46a6

jserv (changes requested):
Rebase the latest `master` branch. In particular, commit 45e3c47a253e6c904156562ad567956d0651edd8 for Arm64-enabled CI pipeline.

jserv:
Consider the following changes for Apple Silicon:
```diff
diff --git a/Makefile b/Makefile
index f8fe8e7..3bc5cc3 100644
--- a/Makefile
+++ b/Makefile
@@ -123,7 +123,7 @@ $(call set-feature, JIT)
 ifeq ($(call has, JIT), 1)
 OBJS_EXT += jit.o 
 ifneq ($(processor), x86_64)
-ifneq ($(processor), aarch64)
+ifneq ($(processor), arm64)
 $(error JIT mode only supports for x64 and arm64 target currently.)
 endif
 endif
diff --git a/src/jit.c b/src/jit.c
index 403f874..89d0fdb 100644
--- a/src/jit.c
+++ b/src/jit.c
@@ -36,6 +36,10 @@
 #include "state.h"
 #include "utils.h"
 
+#if defined(__APPLE__) && defined(__aarch64__)
+#include <libkern/OSCacheControl.h>
+#include <pthread.h>
+#endif
 
 #define JIT_CLS_MASK 0x07
 #define JIT_ALU_OP_MASK 0xf0
@@ -286,9 +290,10 @@ static inline void offset_map_insert(struct jit_state *state, int32_t target_pc)
     map_entry->offset = state->offset;
 }
 
-
+#if !defined(__APPLE__)
 #define sys_icache_invalidate(addr, size) \
     __builtin___clear_cache((char *) (addr), (char *) (addr) + (size));
+#endif
 
 static inline void emit_bytes(struct jit_state *state, void *data, uint32_t len)
 {
@@ -1505,13 +1510,17 @@ struct jit_state *init_state(size_t size)
     struct jit_state *state = malloc(sizeof(struct jit_state));
     state->offset = 0;
     state->size = size;
+#if defined(__APPLE__) && defined(__aarch64__)
+    assert(pthread_jit_write_protect_supported_np());
+    pthread_jit_write_protect_np(0);
+#endif
+#if defined(__APPLE__) && defined(__aarch64__)
+    state->buf = mmap(NULL, size, PROT_EXEC | PROT_WRITE | PROT_READ,
+                      MAP_PRIVATE | MAP_ANONYMOUS | MAP_JIT, -1, 0);
+#else
     state->buf = mmap(0, size, PROT_READ | PROT_WRITE | PROT_EXEC,
-                      MAP_PRIVATE | MAP_ANONYMOUS
-#if defined(__APPLE__)
-                          | MAP_JIT
+                      MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
 #endif
-                      ,
-                      -1, 0);
     assert(state->buf != MAP_FAILED);
     prepare_translate(state);
     state->offset_map = calloc(MAX_INSNS, sizeof(struct offset_map));
```

`build/hello.elf` is known to work.

qwe661234:
> Consider the following changes for Apple Silicon:
> 
> ```diff
> diff --git a/Makefile b/Makefile
> index f8fe8e7..3bc5cc3 100644
> --- a/Makefile
> +++ b/Makefile
> @@ -123,7 +123,7 @@ $(call set-feature, JIT)
>  ifeq ($(call has, JIT), 1)
>  OBJS_EXT += jit.o 
>  ifneq ($(processor), x86_64)
> -ifneq ($(processor), aarch64)
> +ifneq ($(processor), arm64)
>  $(error JIT mode only supports for x64 and arm64 target currently.)
>  endif
>  endif
> diff --git a/src/jit.c b/src/jit.c
> index 403f874..89d0fdb 100644
> --- a/src/jit.c
> +++ b/src/jit.c
> @@ -36,6 +36,10 @@
>  #include "state.h"
>  #include "utils.h"
>  
> +#if defined(__APPLE__) && defined(__aarch64__)
> +#include <libkern/OSCacheControl.h>
> +#include <pthread.h>
> +#endif
>  
>  #define JIT_CLS_MASK 0x07
>  #define JIT_ALU_OP_MASK 0xf0
> @@ -286,9 +290,10 @@ static inline void offset_map_insert(struct jit_state *state, int32_t target_pc)
>      map_entry->offset = state->offset;
>  }
>  
> -
> +#if !defined(__APPLE__)
>  #define sys_icache_invalidate(addr, size) \
>      __builtin___clear_cache((char *) (addr), (char *) (addr) + (size));
> +#endif
>  
>  static inline void emit_bytes(struct jit_state *state, void *data, uint32_t len)
>  {
> @@ -1505,13 +1510,17 @@ struct jit_state *init_state(size_t size)
>      struct jit_state *state = malloc(sizeof(struct jit_state));
>      state->offset = 0;
>      state->size = size;
> +#if defined(__APPLE__) && defined(__aarch64__)
> +    assert(pthread_jit_write_protect_supported_np());
> +    pthread_jit_write_protect_np(0);
> +#endif
> +#if defined(__APPLE__) && defined(__aarch64__)
> +    state->buf = mmap(NULL, size, PROT_EXEC | PROT_WRITE | PROT_READ,
> +                      MAP_PRIVATE | MAP_ANONYMOUS | MAP_JIT, -1, 0);
> +#else
>      state->buf = mmap(0, size, PROT_READ | PROT_WRITE | PROT_EXEC,
> -                      MAP_PRIVATE | MAP_ANONYMOUS
> -#if defined(__APPLE__)
> -                          | MAP_JIT
> +                      MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
>  #endif
> -                      ,
> -                      -1, 0);
>      assert(state->buf != MAP_FAILED);
>      prepare_translate(state);
>      state->offset_map = calloc(MAX_INSNS, sizeof(struct offset_map));
> ```
> 
> `build/hello.elf` is known to work.

I had tried this, but this implementation only allowed writing assembly into memory for the first time. That's the reason why only hello.elf pass.

qwe661234:
> Consider the following changes for Apple Silicon:

It pass all tests now, we gain about 4 times performance improvement on macOS + Apple Silicon.

jserv (changes requested):
Update CI pipeline for JIT/Arm64-enabled build.

src/jit.c
```diff
@@ -0,0 +1,1550 @@
...
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#if HAVE_MMAP
+#include <sys/mman.h>
+#include <unistd.h>
+#else
+/* fallback to standard I/O text stream */
```
jserv: Don't do that. There should be no fallback because we support *BSD (macOS included) and Linux.

src/jit.h
```diff
@@ -0,0 +1,43 @@
...
+    uint32_t retpoline_loc;
+    struct offset_map *offset_map;
+    int num_insn;
+    struct jump *jumps;
+    int num_jumps;
+};
+
+struct jit_state *init_state(size_t size);
+void destroy_state(struct jit_state *state);
+uint32_t translate_JIT(riscv_t *rv, block_t *block);
```
jserv: Enforce consistent style. That is, rename to `jit_translate`.

src/jit.h
```diff
@@ -0,0 +1,43 @@
...
+    uint32_t exit_loc;
+    uint32_t retpoline_loc;
+    struct offset_map *offset_map;
+    int num_insn;
+    struct jump *jumps;
+    int num_jumps;
+};
+
+struct jit_state *init_state(size_t size);
+void destroy_state(struct jit_state *state);
```
jserv: All public functions in JIT module should have prefix `jit_` in common. That is, rename to `jit_state_init` and `jit_state_exit`.

src/jit.h
```diff
@@ -0,0 +1,43 @@
...
+    uint32_t offset;
+    uint32_t stack_size;
+    uint32_t size;
+    uint32_t entry_loc;
+    uint32_t exit_loc;
+    uint32_t retpoline_loc;
+    struct offset_map *offset_map;
+    int num_insn;
+    struct jump *jumps;
+    int num_jumps;
```
jserv: Use prefix `n_` to represent the number of certain storages. That is, rename to `n_insns` and `n_jumps` respectively.

src/jit.c
```diff
@@ -0,0 +1,1550 @@
...
+#if defined(__x86_64__)
+    bool mul = (opcode & JIT_ALU_OP_MASK) == (JIT_OP_MUL_IMM & JIT_ALU_OP_MASK);
+    bool div = (opcode & JIT_ALU_OP_MASK) == (JIT_OP_DIV_IMM & JIT_ALU_OP_MASK);
+    bool mod = (opcode & JIT_ALU_OP_MASK) == (JIT_OP_MOD_IMM & JIT_ALU_OP_MASK);
+    bool is64 = (opcode & JIT_CLS_MASK) == JIT_CLS_ALU64;
+    bool reg = (opcode & JIT_SRC_REG) == JIT_SRC_REG;
+
+    /* Short circuit for imm == 0 */
+    if (!reg && imm == 0) {
+        assert(NULL);
```
jserv, line 1036: Are you ready to close #297 ?

tools/gen-jit-template.py
```diff
@@ -110,23 +110,18 @@ def parse_argv(EXT_LIST, SKIP_LIST):
...
-    return re.sub(r'/\*[\s|\S]+?\*/\n', "", str)
-
-
 # parse_argv(EXT_LIST, SKIP_LIST)
 # prepare PROLOGUE
 output = ""
 f = open('src/rv32_template.c', 'r')
 lines = f.read()
+# remove_comment
+lines = re.sub(r'/\*[\s|\S]+?\*/', "", lines)
```
jserv, line 119: Is it strong enough to resolve #296 ?

src/rv32_template.c
```diff
@@ -14,19 +14,19 @@
...
- *       ld, S32, RAX, X, rs1;
- *       alu32_imm, 32, 0x81, 0, RAX, imm;
- *       st, S32, RAX, X, rd;
+ *   GEN({
+ *       ld, S32, TMP0, X, rs1;
+ *       alu32_imm, 32, 0x81, 0, TMP0, imm;
+ *       st, S32, TMP0, X, rd;
  *   })
  *
  * The block defined as 'X64' is mapped to the generic C code used in the
```
jserv: s/X64/GEN/

Makefile
```diff
@@ -121,15 +121,16 @@ endif
 ENABLE_JIT ?= 0
 $(call set-feature, JIT)
 ifeq ($(call has, JIT), 1)
-OBJS_EXT += jit_x64.o 
-ifneq ($(processor), x86_64)
-$(error JIT mode only supports for x64 target currently.)
+OBJS_EXT += jit.o 
+supported_processor = x86_64 arm64 aarch64
```
jserv: Consider the changes:
```diff
diff --git a/Makefile b/Makefile
index 8719dc6..d2d9329 100644
--- a/Makefile
+++ b/Makefile
@@ -122,8 +122,7 @@ ENABLE_JIT ?= 0
 $(call set-feature, JIT)
 ifeq ($(call has, JIT), 1)
 OBJS_EXT += jit.o 
-supported_processor = x86_64 arm64 aarch64
-ifneq ($(processor), $(findstring $(processor), $(supported_processor)))
+ifneq ($(processor),$(filter $(processor),x86_64 aarch64 arm64))
 $(error JIT mode only supports for x64 and arm64 target currently.)
 endif
```

src/jit.c
```diff
@@ -0,0 +1,1550 @@
...
+    map_entry->PC = target_pc;
+    map_entry->offset = state->offset;
+}
+
+#if !defined(__APPLE__)
+#define sys_icache_invalidate(addr, size) \
+    __builtin___clear_cache((char *) (addr), (char *) (addr) + (size));
+#endif
+
+static inline void emit_bytes(struct jit_state *state, void *data, uint32_t len)
```
jserv: Don't inline this function. Let the compiler decide the optimizations when LTO is enabled.

src/jit.c
```diff
@@ -0,0 +1,1550 @@
...
+
+/* REX.W prefix, ModRM byte, and 8-bit immediate */
+static inline void emit_alu64_imm8(struct jit_state *state,
+                                   int op,
+                                   int src UNUSED,
+                                   int dst,
+                                   int8_t imm)
+{
+#if defined(__x86_64__)
+
```
jserv: Remove this blank line.

src/jit.c
```diff
@@ -0,0 +1,1550 @@
...
+                                    int dst,
+                                    int32_t offset,
+                                    int32_t imm)
+{
+#if defined(__x86_64__)
+    if (size == S16)
+        emit1(state, 0x66); /* 16-bit override */
+    emit1(state, size == S8 ? 0xc6 : 0xc7);
+    emit_modrm_and_displacement(state, 0, dst, offset);
+    if (size == S32)
```
jserv: Rewrite with switch-case and `__UNREACHABLE` since `size` is deterministic.

src/jit.c
```diff
@@ -0,0 +1,1550 @@
...
+#elif defined(__aarch64__)
+    uint32_t stack_movement = align_to(8, 16);
+    emit_addsub_immediate(state, true, AS_SUB, SP, SP, stack_movement);
+    emit_loadstore_immediate(state, LS_STRX, R30, SP, 0);
+
+    emit_movewide_immediate(state, true, temp_imm_reg, target);
+    emit_unconditionalbranch_register(state, BR_BLR, temp_imm_reg);
+
+    /* On exit need to move result from r0 to whichever register we've mapped
+     * EBPF r0 to.  */
```
jserv: Refine this comment to get rid of EBPF.

src/jit.c
```diff
@@ -0,0 +1,1550 @@
...
+                                          int rd,
+                                          int rn,
+                                          int rm)
+{
+    emit_instruction(state, sz(sixty_four) | op | (rm << 16) | (rn << 5) | rd);
+}
+
+
+/* [ArmARM-A H.a]: C4.1.67: Data-processing (3 source).  */
+static void emit_dataprocessing_threesource(struct jit_state *state,
```
jserv: Use shorter name. e.g., `emit_dataproc_3source`.

src/jit.c
```diff
@@ -0,0 +1,1550 @@
...
+/* [ArmARM-A H.a]: C4.1.65: Unconditional branch (register).  */
+static void emit_unconditionalbranch_register(struct jit_state *state,
+                                              enum Opcode op,
+                                              int rn)
+{
+    emit_instruction(state, op | (rn << 5));
+}
+
+/* [ArmARM-A H.a]: C4.1.67: Data-processing (2 source).  */
+static void emit_dataprocessing_twosource(struct jit_state *state,
```
jserv: Ditto. Consider `emit_dataproc_2source`.

src/jit.c
```diff
@@ -0,0 +1,1550 @@
...
+                                         int32_t imm7)
+{
+    int32_t imm_div = ((op == LSP_STPX) || (op == LSP_LDPX)) ? 8 : 4;
+    assert(imm7 % imm_div == 0);
+    imm7 /= imm_div;
+    emit_instruction(state, op | (imm7 << 15) | (rt2 << 10) | (rn << 5) | rt);
+}
+
+/* [ArmARM-A H.a]: C4.1.65: Unconditional branch (register).  */
+static void emit_unconditionalbranch_register(struct jit_state *state,
```
jserv: Consider `emit_uncond_branch_reg`.

src/jit.c
```diff
@@ -0,0 +1,1550 @@
...
+
+    /* Tidy up for the case imm = 0 or imm == -1.  */
+    if (op != MW_MOVK) {
+        emit_instruction(state,
+                         sz(sixty_four) | op | (0 << 21) | (0 << 5) | rd);
+    }
+}
+
+/* [ArmARM-A H.a]: C4.1.66: Load/store register (unscaled immediate).  */
+static void emit_loadstore_immediate(struct jit_state *state,
```
jserv: Consider `emit_loadstore_imm`.

src/jit.c
```diff
@@ -0,0 +1,1550 @@
...
+}
+
+static void emit_instruction(struct jit_state *state, uint32_t instr)
+{
+    assert(instr != BAD_OPCODE);
+    emit_bytes(state, &instr, 4);
+}
+
+/* Get the value of the size bit in most instruction encodings (bit 31). */
+static uint32_t sz(bool sixty_four)
```
jserv: Replace `sixty_four` with `is64`.

src/jit.c
```diff
@@ -0,0 +1,1550 @@
...
+#endif
+
+#include "cache.h"
+#include "decode.h"
+#include "io.h"
+#include "jit.h"
+#include "riscv.h"
+#include "utils.h"
+
+
```
jserv: Remove this blank line.

src/jit.c
```diff
@@ -0,0 +1,1550 @@
...
+    /* DP3Opcode */
+    DP3_MADD = 0x1b000000U,  // 0001_1011_0000_0000_0000_0000_0000_0000
+    DP3_MSUB = 0x1b008000U,  // 0001_1011_0000_0000_1000_0000_0000_0000
+    /* MoveWideOpcode */
+    MW_MOVN = 0x12800000U,  // 0001_0010_1000_0000_0000_0000_0000_0000
+    MW_MOVZ = 0x52800000U,  // 0101_0010_1000_0000_0000_0000_0000_0000
+    MW_MOVK = 0x72800000U,  // 0111_0010_1000_0000_0000_0000_0000_0000
+};
+
+enum Condition {
```
jserv: Enforce consistent style. Use lowercase instead.

src/jit.c
```diff
@@ -0,0 +1,1550 @@
...
+#elif defined(__aarch64__)
+/*
+ * There are two common x86-64 calling conventions, as discussed at
+ * https://en.wikipedia.org/wiki/X86_calling_conventions#x86-64_calling_conventions
+ *
+ * Please Note: R12 is special and we are *not* using it. As a result, it is
+ * omitted from the list of non-volatile registers for both platforms (even
+ * though it is, in fact, non-volatile).
+ *
+ * BPF R0-R4 are "volatile"
```
jserv: Refine the comments to get rid of BPF.

src/jit.c
```diff
@@ -0,0 +1,1550 @@
...
+    N_VM_REGS,
+};
+
+enum operand_size {
+    S8,
+    S16,
+    S32,
+};
+
+#define REGISTER_MAP_SIZE 11
```
jserv: Instead of defining `REGISTER_MAP_SIZE` here, we can make use of macro `ARRAYS_SIZE` after defining `register_map`.

src/jit.c
```diff
@@ -0,0 +1,1550 @@
...
+        emit_logical_register(state, false, LOG_AND, dst, dst, src);
+        break;
+    case 0xd3:
+        if (src == 4) /* SLL */
+            emit_dataprocessing_twosource(state, false, DP2_LSLV, dst, dst, R8);
+        else if (src == 5) /* SRL */
+            emit_dataprocessing_twosource(state, false, DP2_LSRV, dst, dst, R8);
+        else if (src == 7) /* SRA */
+            emit_dataprocessing_twosource(state, false, DP2_ASRV, dst, dst, R8);
+        break;
```
jserv, line 623: See if `__UNREACHABLE` fits here.

jserv:
For both x64 and A64 targets, CoreMark complains as following:
```
2K performance run parameters for coremark.
[0]ERROR! list crc 0x13bc - should be 0xe714
[0]ERROR! matrix crc 0x79d7 - should be 0x1fd7
[0]ERROR! state crc 0x4554 - should be 0x8e3a
CoreMark Size    : 666
...
```

What is wrong?

jserv:
I tried running `build/donut.elf` with the JIT-enabled build and noticed that the animation did not perform as expected. Unlike a typical spinning donut, it was not smooth and failed to render consistently over time, compared to the pure interpreter mode.

src/jit.c
```diff
@@ -0,0 +1,1550 @@
...
+static inline void emit_jump_target_address(struct jit_state *state,
+                                            int32_t target_pc)
+{
+    struct jump *jump = &state->jumps[state->num_jumps++];
+    jump->offset_loc = state->offset;
+    jump->target_pc = target_pc;
+    emit4(state, 0);
+}
+#elif defined(__aarch64__)
+static inline uint32_t align_to(uint32_t amount, uint64_t boundary)
```
jserv: The align routines are used widely, and we can move this implementation to `src/common.h`.
See [mimalloc/internal.h](https://github.com/microsoft/mimalloc/blob/4e50d6714d471b72b2285e25a3df6c92db944593/include/mimalloc/internal.h#L288) for `_mi_align_up` and `_mi_align_down`.

In this case, we shall explicitly mention down/up when aligning.

src/jit.c
```diff
@@ -0,0 +1,1550 @@
...
+                                  int rd,
+                                  int rn,
+                                  uint32_t imm12)
+{
+    const uint32_t imm_op_base = 0x11000000;
+    emit_instruction(state, sz(sixty_four) | (op << 29) | imm_op_base |
+                                (0 << 22) | (imm12 << 10) | (rn << 5) | rd);
+}
+
+/* [ArmARM-A H.a]: C4.1.67: Logical (shifted register).  */
```
jserv: You should mention `ArmARM-A H.a` as the following reference:
> https://developer.arm.com/documentation/ddi0487/ha (Arm Architecture Reference Manual for A-profile architecture)

Alternatively, name it "ARM-A" for short.

src/jit.c
```diff
@@ -0,0 +1,1550 @@
...
+                                        int rm)
+{
+    const uint32_t reg_op_base = 0x0b000000;
+    emit_instruction(state, sz(sixty_four) | (op << 29) | reg_op_base |
+                                (rm << 16) | (rn << 5) | rd);
+}
+
+/* [ArmARM-A H.a]: C4.1.64: Move wide (Immediate).  */
+static inline void emit_movewide_immediate(struct jit_state *state,
+                                           bool sixty_four,
```
jserv: Replace all occurrences of `sixty_four` with `is64`.

tools/gen-jit-template.py
qwe661234: Yes.

src/jit.c
qwe661234: Not yet. I will close this issue by another PR.

tools/gen-jit-template.py
jserv: Append `Close #296` at the end of git commit messages and improve the comments of `src/rv32_template.c` as a complete example starting with `RVOP`.

src/common.h
```diff
@@ -144,3 +144,12 @@
     ((type *) ((char *) (ptr) - (offsetof(type, member))))
 #endif
 #endif
+
+static inline unsigned int align_up(unsigned int sz, unsigned long alignment)
+{
+    unsigned int mask = alignment - 1;
+    if ((alignment & mask) == 0)
+        return ((sz + mask) & ~mask);
+    else
```
jserv: Remove `else` statement here.

src/common.h
```diff
@@ -144,3 +144,12 @@
     ((type *) ((char *) (ptr) - (offsetof(type, member))))
 #endif
 #endif
+
+static inline unsigned int align_up(unsigned int sz, unsigned long alignment)
```
jserv: Be aware of derived types. Should be
```c
static inline uintptr_t align_up(uintptr_t sz, size_t alignment)
```

src/emulate.c
```diff
@@ -9,6 +9,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include "common.h"
```
jserv: There is no need to include `common.h` since the build system already specifies it as the first header for inclusion.

src/jit.c
jserv: > Not yet. I will close this issue by another PR.

Add proper `FIXME` or `TODO` comment to reminder.

src/jit.c
```diff
@@ -0,0 +1,1578 @@
...
+    R26,
+    R27,
+    R28,
+    R29,
+    R30,
+    SP,
+    RZ = 31
+};
+
+enum Opcode {
```
jserv: Always use lowercase for the enum name unless you have strong reasons as an exception.

src/jit.c
```diff
@@ -0,0 +1,1578 @@
...
+    COND_EQ,
+    COND_NE,
+    COND_HS,
+    COND_LO,
+    COND_GE = 10,
+    COND_LT = 11,
+};
+#endif
+
+enum VM_REG {
```
jserv: Use `enum vm_reg` instead.

src/jit.c
```diff
@@ -0,0 +1,1578 @@
...
+static const int nonvolatile_reg[] = {RBP, RBX, R13, R14, R15};
+static const int parameter_reg[] = {RDI, RSI, RDX, RCX, R8, R9};
+static const int temp_reg[] = {RAX, RBX, RCX};
+static const int register_map[] = {
+    RAX, RDI, RSI, RDX, R9, R8, RBX, R13, R14, R15, RBP,
+};
+#endif
+#elif defined(__aarch64__)
+/*
+ * There are two common x86-64 calling conventions, as discussed at
```
jserv: This comment has nothing to Aarch64.

src/jit.c
```diff
@@ -0,0 +1,1578 @@
...
+}
+
+/* Get the value of the size bit in most instruction encodings (bit 31). */
+static uint32_t sz(bool is64)
+{
+    return (is64 ? UINT32_C(1) : UINT32_C(0)) << 31;
+}
+
+
+/* For details on ARM instructions, users can refer to
```
jserv: You should refer it to "Arm instructions" instead of "ARM instructions."
See https://www.arm.com/company/policies/trademarks/guidelines-trademarks
> Do not alter the spelling or form of Arm's trademarks by abbreviating them, creating acronyms, translating them, joining them to other words, symbols or numbers (either as one word or with a hyphen - unless otherwise permitted, e.g. Arm® Cortex®-A15), or using improper capitalization. However, permitted capitalization occurs when using an Arm word trademark in headlines, titles or text where all of the surrounding words are shown in uppercase characters. In this situation, you may use the relevant Arm trademark in uppercase characters, provided that such use complies with these guidelines.

src/jit.c
```diff
@@ -0,0 +1,1578 @@
...
+#endif
+}
+
+#define GEN(inst, code)                                                       \
+    static void do_##inst(struct jit_state *state UNUSED, riscv_t *rv UNUSED, \
+                          rv_insn_t *ir UNUSED)                               \
+    {                                                                         \
+        code;                                                                 \
+    }
+#include "rv32_jit_template.c"
```
jserv: Let's rename `rv32_jit_template.c` to `rv32_jit.c`, since the latter is actually a specialized form tailored for JIT code generation, using `rv32_template.c` as its input.

src/jit.c
```diff
@@ -0,0 +1,1578 @@
...
+out:
+    return entry_loc;
+}
+
+struct jit_state *jit_state_init(size_t size)
+{
+    struct jit_state *state = malloc(sizeof(struct jit_state));
+    state->offset = 0;
+    state->size = size;
+#if HAVE_MMAP
```
jserv: Remove the `HAVE_MMAP` check, as our JIT compiler will not support systems without `mmap`. In the future, we may introduce additional checks to identify operating systems supported by the build system.

src/jit.c
```diff
@@ -0,0 +1,1578 @@
...
+    state->buf = mmap(0, size, PROT_READ | PROT_WRITE | PROT_EXEC,
+                      MAP_PRIVATE | MAP_ANONYMOUS
+#if defined(__APPLE__)
+                          | MAP_JIT
+#endif
+                      ,
+                      -1, 0);
+    assert(state->buf != MAP_FAILED);
+#else
```
jserv: Ditto. Remove `assert(NULL)` and `#else`.

src/jit.h
```diff
@@ -0,0 +1,43 @@
...
+#include "riscv_private.h"
+
+struct jump {
+    uint32_t offset_loc;
+    uint32_t target_pc;
+    uint32_t target_offset;
+};
+
+struct offset_map {
+    uint32_t PC;
```
jserv: Use lowercase `pc`.

src/jit.h
```diff
@@ -0,0 +1,43 @@
+/*
+ * rv32emu is freely redistributable under the MIT License. See the file
+ * "LICENSE" for information on usage and redistribution of this file.
+ */
+
+#pragma once
+
+#include <assert.h>
```
jserv: Remove the inclusion of `<assert.h>`, `<stdio.h>`, `<string.h>`, and `<unistd.h>` because they are not really used in this header.

src/common.h
qwe661234: If we want to use `uintptr_t` and `size_t`, we need to include other header file in common.h.

jserv: > If we want to use `uintptr_t` and `size_t`, we need to include other header file in common.h.

Alternatively, implement `align_up()` in `src/utils.h`.

src/jit.c
```diff
@@ -0,0 +1,1566 @@
...
+#error "This implementation is dedicated to x64 and arm64."
+#endif
+
+#include <assert.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#if HAVE_MMAP
```
jserv: No `HAVE_MMAP` check should appear here since we assume BSD/macOS/Linux is used. For Windows support, we may provide a wrapper header `mman.h` instead. Again, in current JIT compiler, no more `mmap` checks.

src/rv32_jit.c
```diff
@@ -0,0 +1,824 @@
+GEN(nop, {})
```
jserv: Don't check in this file, which is generated. Update `.gitignore` accordingly.

qwe661234:
> For both x64 and A64 targets, CoreMark complains as following:
> 
> ```
> 2K performance run parameters for coremark.
> [0]ERROR! list crc 0x13bc - should be 0xe714
> [0]ERROR! matrix crc 0x79d7 - should be 0x1fd7
> [0]ERROR! state crc 0x4554 - should be 0x8e3a
> CoreMark Size    : 666
> ...
> ```
> 
> What is wrong?

This is a known issue, but I am unable to identify where the error is occurring currently. Can we handle this issue by another PR?

jserv:
> > For both x64 and A64 targets, CoreMark complains as following:
> > ```
> > [0]ERROR! list crc 0x13bc - should be 0xe714
> > [0]ERROR! matrix crc 0x79d7 - should be 0x1fd7
> > [0]ERROR! state crc 0x4554 - should be 0x8e3a
> > ```
>
> This is a known issue, but I am unable to identify where the error is occurring currently. Can we handle this issue by another PR?

Sure, start by describing the issue on GitHub, and include any relevant materials you have.

src/jit.c
```diff
@@ -0,0 +1,1564 @@
...
+        emit_loadstore_imm(state, LS_STRH, src, dst, offset);
+        break;
+    case S32:
+        emit_loadstore_imm(state, LS_STRW, src, dst, offset);
+        break;
+    default:
+        __UNREACHABLE;
+        break;
+    }
+
```
jserv: Remove this blank line.

src/jit.c
```diff
@@ -0,0 +1,1564 @@
...
+    const uint32_t index = set_hash(key);
+    for (uint8_t count = 0; set->table[index][count]; count++) {
+        if (set->table[index][count] == key)
+            return true;
+    }
+    return false;
+}
+
+#define UPDATE_PC(pc)                             \
+    emit_load_imm(state, RAX, (pc));              \
```
jserv: There is no need to write `(pc)` for an inline function. Use plain `pc`.

src/jit.c
```diff
@@ -0,0 +1,1564 @@
...
+    emit_store(state, S32, RAX, parameter_reg[0], \
+               offsetof(struct riscv_internal, PC));
```
jserv: It is a bit confusing for mixing `PC` and `pc` in a macro. Can you clarify?

qwe661234: This macro is redundant, we can remove it directly.

src/jit.c
```diff
@@ -0,0 +1,1557 @@
...
+
+#include <assert.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/mman.h>
+#include <unistd.h>
+#if defined(__APPLE__) && defined(__aarch64__)
```
jserv: Consider the following changes to satisfy macOS/x86-64:
```diff
--- a/src/jit.c
+++ b/src/jit.c
@@ -27,10 +27,12 @@
 #include <stdlib.h>
 #include <sys/mman.h>
 #include <unistd.h>
-#if defined(__APPLE__) && defined(__aarch64__)
+#if defined(__APPLE__)
 #include <libkern/OSCacheControl.h>
+#if defined(__aarch64__)
 #include <pthread.h>
 #endif
+#endif
 
 #include "cache.h"
 #include "decode.h"
```

jserv (changes requested):
Fix CI breakage:
```
Check failure on line 59 in .github/workflows/main.yml

GitHub Actions
/ .github/workflows/main.yml
Invalid workflow file
You have an error in your yaml syntax on line 59
```

> [runs #7315225700](https://github.com/sysprog21/rv32emu/actions/runs/7315225700/workflow)

src/jit.c
```diff
@@ -0,0 +1,1559 @@
...
+
+#if !defined(__APPLE__)
+#define sys_icache_invalidate(addr, size) \
+    __builtin___clear_cache((char *) (addr), (char *) (addr) + (size));
+#endif
+
+static void emit_bytes(struct jit_state *state, void *data, uint32_t len)
+{
+    assert(state->offset <= state->size - len);
+    if ((state->offset + len) > state->size) {
```
jserv: Add `unlikely` hint in this `if` statement since this case is almost false.

.github/workflows/main.yml
```diff
@@ -55,6 +55,9 @@ jobs:
           git config --global --add safe.directory /home/runner/work/rv32emu/rv32emu
           make
           make check
+    - name: JIT test
+      run: |
+            make ENABLE_JIT=1 clean check
```
jserv: Replace with `make ENABLE_JIT=1 distclean check`

src/jit.c
```diff
@@ -0,0 +1,1559 @@
...
+static inline void emit_jump_target_address(struct jit_state *state,
+                                            int32_t target_pc)
+{
+    struct jump *jump = &state->jumps[state->n_jumps++];
+    jump->offset_loc = state->offset;
+    jump->target_pc = target_pc;
+    emit4(state, 0);
+}
+#elif defined(__aarch64__)
+static void emit_instruction(struct jit_state *state, uint32_t instr)
```
jserv: `emit_instruction` is quite confusing since we mix both A64 and x64 implementations here. Replace it with something like `emit_a64` instead.

jserv: For consistency, the parameter name should be `insn` rather than `instr`.

jserv (changes requested):
Rebase the latest `master` branch and be aware of the following changes:
```diff
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -58,6 +58,7 @@ jobs:
         run: |
           make
           make check
+          make ENABLE_JIT=1 clean check
 
   coding-style:
     runs-on: ubuntu-22.04
```

Your previous attempt was rejected by [uraimo/run-on-arch-action](https://github.com/uraimo/run-on-arch-action) because Arm64 is emulated via QEMU, and we cannot run commands outside the emulated environment.

jserv (changes requested):
Improve the comments in `src/rv32_template.c` with a complete example below.

```c
 * RVOP(
 *   addi,
 *   { rv->X[ir->rd] = (int32_t) (rv->X[ir->rs1]) + ir->imm; }, 
 *   GEN({
 *       ld, S32, TMP0, X, rs1;
 *       alu32_imm, 32, 0x81, 0, TMP0, imm;
 *       st, S32, TMP0, X, rd;
 *   }))
```

Also, explain the functionality of TMP{0,1,2}.

src/jit.c
```diff
@@ -0,0 +1,1557 @@
...
+    R26,
+    R27,
+    R28,
+    R29,
+    R30,
+    SP,
+    RZ = 31
+};
+
+enum arm_opcode {
```
jserv: Given the frequent use of this enumeration, we can establish its derived type. Therefore, let's use:
```c
typedef enum {
    AS_ADD = 0,
    ... 
} a64opcode_t;
```

Subsequently, we should consistently apply `a64opcode_t` in function parameters and declarations.

src/jit.c
```diff
@@ -0,0 +1,1557 @@
...
+    if (imm >= INT32_MIN && imm <= INT32_MAX)
+        emit_alu64_imm32(state, 0xc7, 0, dst, imm);
+    else {
+        /* movabs $imm, dst */
+        emit_basic_rex(state, 1, 0, dst);
+        emit1(state, 0xb8 | (dst & 7));
+        emit8(state, imm);
+    }
+#elif defined(__aarch64__)
+    if (imm >= INT32_MIN && imm <= INT32_MAX)
```
jserv: Rewrite as `if ((int32_t)imm == imm)`, which is concise and should be as fast or faster than the original, depending on the compiler optimizations.

Let's prove the equivalence of these two statements:

Case for `imm` within `int32_t` Range: If `imm` is within the `int32_t` range (`INT32_MIN` to `INT32_MAX`), then casting `imm` to `int32_t` will not change its value. The comparison `(int32_t)imm == imm` will be true. This is equivalent to saying `imm >= INT32_MIN && imm <= INT32_MAX`.

Case for `imm` outside `int32_t` Range: If `imm` is outside the `int32_t` range, casting it to `int32_t` will result in overflow, changing the value. Thus, `(int32_t)imm == imm` will be false. This is equivalent to `imm` not satisfying `imm >= INT32_MIN && imm <= INT32_MAX`.

The equivalence of `(int32_t)imm == imm` and `imm >= INT32_MIN && imm <= INT32_MAX` lies in how C handles type casting and comparisons. In both cases, the condition checks whether imm lies within the range that can be accurately represented by an `int32_t`. Thus, these statements are functionally equivalent.

src/jit.c
```diff
@@ -0,0 +1,1557 @@
...
+    case 0x98:
+        divmod(state, JIT_OP_MOD_REG, dst, dst, src);
+        break;
+    default:
+        __UNREACHABLE;
+        break;
+    }
+#endif
+}
+#endif
```
jserv: Append `/* RV32_HAS(EXT_M) */` right after `#endif`.

src/jit.c
```diff
@@ -0,0 +1,1557 @@
...
+#include "rv32_jit.c"
+#undef GEN
+
+static void do_fuse1(struct jit_state *state, riscv_t *rv UNUSED, rv_insn_t *ir)
+{
+    opcode_fuse_t *fuse = ir->fuse;
+    for (int i = 0; i < ir->imm2; i++) {
+        emit_load_imm(state, temp_reg[0], fuse[i].imm);
+        emit_store(state, S32, temp_reg[0], parameter_reg[0],
+                   offsetof(struct riscv_internal, X) + 4 * fuse[i].rd);
```
jserv: Why not using `offsetof(riscv_t, X)`?

Since `offsetof(struct riscv_internal, X)` plus a certain offset is frequently used, we can introduce a macro to simplify the specification of memory locations.

jserv (changes requested):
Update top-level `README.md` to address `ENABLE_JIT` option.

src/jit.c
```diff
@@ -0,0 +1,1557 @@
...
+};
+#endif
+#elif defined(__aarch64__)
+/* callee_reg - this must be a multiple of two because of how we save the stack
+ * later on. */
+static const int callee_reg[] = {R19, R20, R21, R22, R23, R24, R25, R26};
+/* parameter_reg (Caller saved registers) */
+static const int parameter_reg[] = {R0, R1, R2, R3, R4};
+static const int temp_reg[] = {R6, R7, R8};
+/* Temp register for immediate generation */
```
jserv: Use `enum` for both `temp_imm_reg` and `temp_div_reg` because `const` is just a qualifier and the variables still occupy space. That is,
```c
enum {
    temp_imm_reg = R24, /* Temp register for immediate generation */
    temp_div_reg = R25, /* Temp register for division results */
};
```

src/jit.c
```diff
@@ -0,0 +1,1557 @@
...
+    assert(insn != BAD_OPCODE);
+    emit_bytes(state, &insn, 4);
+}
+
+/* Get the value of the size bit in most instruction encodings (bit 31). */
+static uint32_t sz(bool is64)
+{
+    return (is64 ? UINT32_C(1) : UINT32_C(0)) << 31;
+}
+
```
jserv, line 397: Remove this blank line.

src/jit.c
```diff
@@ -0,0 +1,1557 @@
...
+}
+#elif defined(__aarch64__)
+static void emit_a64(struct jit_state *state, uint32_t insn)
+{
+    assert(insn != BAD_OPCODE);
+    emit_bytes(state, &insn, 4);
+}
+
+/* Get the value of the size bit in most instruction encodings (bit 31). */
+static uint32_t sz(bool is64)
```
jserv: Add `inline` qualifier.

src/jit.c
```diff
@@ -0,0 +1,1557 @@
...
+    memset(state->offset_map, 0, MAX_INSNS * sizeof(struct offset_map));
+    memset(state->jumps, 0, MAX_INSNS * sizeof(struct jump));
+    state->n_insn = 0;
+    state->n_jumps = 0;
+    uint32_t entry_loc = state->offset;
+    set_t set;
+    set_reset(&set);
+    translate_chained_block(&(*state), rv, block, &set);
+
+    if (state->offset == state->size) {
```
jserv, line 1523: If target buffer is too small, does it make sense to continue translating? Shall we raise abort or exception?

src/jit.c
```diff
@@ -0,0 +1,1557 @@
...
+}
+
+
+/* For details on Arm instructions, users can refer to
+ * https://developer.arm.com/documentation/ddi0487/ha (Arm Architecture
+ * Reference Manual for A-profile architecture).
+ */
+
+/* [ARM-A]: C4.1.64: Add/subtract (immediate).  */
+static void emit_addsub_immediate(struct jit_state *state,
```
jserv: We can abbreviate 'immediate' as "imm" in all occurrences within function names.

src/jit.c
```diff
@@ -0,0 +1,1557 @@
...
+                                        int rn,
+                                        int rm)
+{
+    const uint32_t reg_op_base = 0x0b000000;
+    emit_a64(state,
+             sz(is64) | (op << 29) | reg_op_base | (rm << 16) | (rn << 5) | rd);
+}
+
+/* [ARM-A]: C4.1.64: Move wide (Immediate).  */
+static inline void emit_movewide_immediate(struct jit_state *state,
```
jserv: Ditto. Use shorter `imm`.

src/jit.c
```diff
@@ -0,0 +1,1557 @@
...
+                               int16_t imm9)
+{
+    const uint32_t imm_op_base = 0x38000000U;
+    assert(imm9 >= -256 && imm9 < 256);
+    imm9 &= 0x1ff;
+    emit_a64(state, imm_op_base | op | (imm9 << 12) | (rn << 5) | rt);
+}
+
+/* [ARM-A]: C4.1.66: Load/store register pair (offset).  */
+static void emit_loadstorepair_immediate(struct jit_state *state,
```
jserv: Ditto. Use shorter `imm`.

src/jit.c
```diff
@@ -0,0 +1,1557 @@
...
+                                  enum arm_opcode op,
+                                  int rd,
+                                  int rn,
+                                  int rm,
+                                  int ra)
+{
+    emit_a64(state, sz(is64) | op | (rm << 16) | (ra << 10) | (rn << 5) | rd);
+}
+
+static void update_branch_immediate(struct jit_state *state,
```
jserv: Ditto. Use shorter `imm`.

src/jit.c
qwe661234: In the function `emit_bytes`, we have ` assert(state->offset <= state->size - len);` to detect the machine code size is overflow or not.

src/jit.c
```diff
@@ -0,0 +1,1557 @@
...
+}
+
+static inline void emit_modrm_and_displacement(struct jit_state *state,
+                                               int r,
+                                               int m,
+                                               int32_t d)
+{
+    if (d == 0 && (m & 7) != RBP) {
+        emit_modrm(state, 0x00, r, m);
+    } else if (d >= -128 && d <= 127) {
```
jserv: Rewrite as `if ((int8_t) d == d) {`.

src/rv32_template.c
```diff
@@ -12,24 +12,32 @@
...
+ * execution. The block defined as 'GEN' is mapped to the generic C code used in
+ * the interpreter. The following instructions will be generated by JIT
+ * compiler:
+ * - Load X->rs1 (target field) from the rv data structure to TMP0 (destination
  *   register).
- * - Do ALU operation on 0 (source register) and imm and store the result into
- *   RAX (destination register).
- * - Store RAX (source register) value to the X->rd (target field) of the rv
- * data structure.
+ * - Do ALU operation on TMP0 (source register) and imm and store the result
```
jserv: Refine this statement to ensure that it is technically correct.

src/rv32_template.c
```diff
@@ -12,24 +12,32 @@
...
+ *   into TMP0 (destination register).
+ * - Store TMP0 (source register) value to the X->rd (target field) of the rv
+ *   data structure.
  *
- * The parameter of x64 instruction API
+ * The instruction sequence of addi:
```
jserv: Rephrase as
> The sequence of host instructions generated during dynamic binary translation for the `addi` instruction:

src/utils.h
```diff
@@ -47,3 +48,11 @@ void rv_clock_gettime(struct timespec *tp);
  * https://9p.io/sys/doc/lexnames.html
  */
 char *sanitize_path(const char *input);
+
+static inline uintptr_t align_up(uintptr_t sz, size_t alignment)
+{
+    uintptr_t mask = alignment - 1;
+    if ((alignment & mask) == 0)
```
jserv: Add `likely` hint.

src/jit.c
```diff
@@ -0,0 +1,1557 @@
...
+static inline void emit_jump_target_offset(struct jit_state *state,
+                                           uint32_t jump_loc,
+                                           uint32_t jump_state_offset)
+{
+    struct jump *jump = &state->jumps[state->n_jumps++];
+    jump->offset_loc = jump_loc;
+    jump->target_offset = jump_state_offset;
+}
+
+/* The REX prefix and ModRM byte are emitted.
```
jserv: This comment should only make sense to x64 backend. Move into the function body.

src/jit.c
```diff
@@ -0,0 +1,1557 @@
...
+            emit_dataproc_2source(state, false, DP2_ASRV, dst, dst, R8);
+        break;
+    default:
+        __UNREACHABLE;
+        break;
+    }
+#endif
+}
+
+/* REX prefix, ModRM byte, and 32-bit immediate */
```
jserv: This comment should only make sense to x64 backend. Move into the function body.

src/jit.c
```diff
@@ -0,0 +1,1557 @@
...
+        emit_logical_register(state, false, LOG_EOR, dst, src, R10);
+        break;
+    default:
+        __UNREACHABLE;
+        break;
+    }
+#endif
+}
+
+/* REX prefix, ModRM byte, and 8-bit immediate */
```
jserv: This comment should only make sense to x64 backend. Move into the function body.

src/jit.c
```diff
@@ -0,0 +1,1557 @@
...
+        emit_dataproc_2source(state, false, DP2_ASRV, dst, dst, R10);
+        break;
+    default:
+        __UNREACHABLE;
+        break;
+    }
+#endif
+}
+
+/* The REX.W prefix and ModRM byte are emitted.
```
jserv: This comment should only make sense to x64 backend. Move into the function body.

src/jit.c
```diff
@@ -0,0 +1,1557 @@
...
+    emit_basic_rex(state, 1, src, dst);
+    emit1(state, op);
+    emit_modrm_reg2reg(state, src, dst);
+#elif defined(__aarch64__)
+    if (op == 0x01)
+        emit_addsub_register(state, true, AS_ADD, dst, dst, src);
+#endif
+}
+
+/* REX.W prefix, ModRM byte, and 8-bit immediate */
```
jserv: This comment should only make sense to x64 backend. Move into the function body.

PR #305, opened 2023-12-22
Consolidate public header by integrating src/state.h

### LINKED ISSUE #301: Consolidate public header by integrating src/state.h
To fully(as much as possible) leverage rv32emu's core, I've attempted to export the `state_new` function from "state.h" for usage in WebAssembly, but it doesn't appear to be functioning. It works after splitting to "state.[ch]". I think the reason is that `state_new` is not found in any translation units since they have been inlined by the compiler.

May I know any concerns with this change? I might send a PR for this modification if it doesn't. 

Refer to #75

ChinYikMing:
- Merge "src/state.h" into "riscv.[ch]" to consolidate public header.
- Handle malloc failed in the "state_new" and "memory_new" functions.

Close #301


jserv:

Issue #28, opened 2022-08-02
Integrate `embench-iot` for benchmarking


jserv:
[Embench benchmark suite](https://github.com/embench/embench-iot) is designed to test the performance of deeply embedded systems.
- The measurement of execution performance is designed to use "hot" caches. Thus each benchmark executes its entire code several times, before starting a timing run.
- Execution runs are scaled to take approximately 4 second of CPU time. This is large enough to be accurately measured, yet means all benchmarks, including cache warm up can be run in a few minutes.
- To facilitate execution on machines of different performance, the tests are scaled by the clock speed of the processor.

jserv:
[riscv-benchmark-automation](https://github.com/ekliptik/riscv-benchmark-automation) provides the necessary changes to adopt embench-iot for RISC-V.

jserv:
@fourcolor, can you make a list for known pending issues to evaluate embench-iot?

PR #313, opened 2023-12-30
Fix incorrect mcycle csr address

fourcolor:
According to [RISC-V Specifications](https://riscv.org/technical/specifications/) Volume 2, Privileged Specification, the correct CSR address for the mcycle is 0xB00.

### DIFF
--- a/src/riscv_private.h
+++ b/src/riscv_private.h
@@ -40,7 +40,7 @@ enum {
     CSR_MIP = 0x344,      /* Machine interrupt pending */
 
     /* low words */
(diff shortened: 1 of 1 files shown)


jserv:
`CSR_MCYCLE` (Machine cycle counter low word) is `0xb00` while `CSR_CYCLE` (Cycle counter low word) is `0xc00`. The mnemonic identifier used is correct. Why would you propose such a change?

fourcolor:
Sorry, I realized that I mixed up 'mcycle' and 'cycle.' I made this change because I'm working on issue #28 . I believe this pr can be closed.

PR #316, opened 2024-01-01
Support RV32FC

### LINKED ISSUE #315: Incorrect global/static float array initialization
The following code run through rv32emu  did not function as expected. I have also tested it with spike, confirming that the issue is specific to rv32emu. This error will lead to an error when running #28 [embench-iot minver](https://github.com/embench/embench-iot/blob/master/src/minver/libminver.c)
* global static float array initialization
```c
#include <stdio.h>
static float test[3] = {-3.0, 0.0, 2.0};
int main()
{
    printf("%f\n", test[0]);
    return 0;
}
```
compile
```
riscv32-unknown-elf-gcc test.c -march=rv32imafc -mabi=ilp32 -o test.elf
```
expect output
```
-3.000000
inferior exit code 0
```
actual output
```
0.000000
inferior exit code 0
```
Furthermore, I have observed that the same issue exists in the following code.
* global float array initialization
```c
#include <stdio.h>
float test[3] = {-3.0, 0.0, 2.0};
int main()
{
    printf("%f\n", test[0]);
    return 0;
}
```
* local static float array initialization
```c
#include <stdio.h>
int main()
{
    static float test[3] = {-3.0, 0.0, 2.0};
    printf("%f\n", test[0]);
    return 0;
}
```
I notice that it only occurred with the `global float array initialization` and `static float array initialization`

fourcolor:
This patch is aimed at supporting RV32FC-only instructions.
* The following instructions are implemented: C.FSW, C.FLW, C.FSWSP, C.FLWSP.
* Update README.md for the corresponding extension description.
* Add corresponding RISC-V architectural tests suites

Close #315


jserv (changes requested):
Refine the README.md for the corresponding extension description.

jserv (changes requested):
1. Append `Close #315` at the end of git commit message.
2. Attempt to run the corresponding RISC-V Architectural Tests when possible.

src/decode.c
```diff
@@ -1668,15 +1668,92 @@ static inline bool op_cbnez(rv_insn_t *ir, const uint32_t insn)
...
+     * -------+------+-------------+---+--
+     * C.FSWSP 111    uimm[5:2|7:6] rs2 10
+     */
+    ir->imm = (insn & 0x1e00) >> 7 | (insn & 0x180) >> 1;
+    ir->rs2 = c_decode_rs2(insn);
+    ir->opcode = rv_insn_cfswsp;
+    return true;
+}
+
+/* C.LW: CL-format
```
visitorckw: "C.LW" should be corrected to "C.FLW".

visitorckw:
The current [riscv-arch-test](https://github.com/riscv-non-isa/riscv-arch-test) does not include a test suite for RV32FC. However, we could explore using [riscv-ctg](https://github.com/riscv-software-src/riscv-ctg) to generate a suitable test suite.

visitorckw:
Similar to commit [cfa1c9f](https://github.com/sysprog21/rv32emu/commit/cfa1c9f8a3331eb53239bd284bcf298fab933793),  we need to conduct alignment checks on load/store addresses.

fourcolor:
After a few days of researching [riscv-ctg](https://github.com/riscv-software-src/riscv-ctg), I consider it is challenging to develop a related test suite in a short period. Therefore, I would like to inquire whether adding a TODO comment in the code or opening an issue, along with explaining in the README that certain parts have not undergone arch-test testing, would be a better practice.

jserv:
> After a few days of researching [riscv-ctg](https://github.com/riscv-software-src/riscv-ctg), I consider it is challenging to develop a related test suite in a short period.

Rather than expanding existing compilation suites, it would be more effective to develop a straightforward, specialized test program specifically for validating RV32FC support.

fourcolor:
I ultimately managed to figure out how to use riscv-ctg to generate relevant test suites, and I also added tests for RV32FC in the CI.

src/decode.h
```diff
@@ -176,6 +176,13 @@ enum op_field {
         _(cjalr, 1, 2, 1, ENC(rs1, rs2, rd))           \
         _(cadd, 0, 2, 1, ENC(rs1, rs2, rd))            \
         _(cswsp, 0, 2, 1, ENC(rs2))                    \
+        /* RV32FC Instruction */                       \
+        IIF(RV32_HAS(EXT_F))(                          \
+        _(cflwsp, 0, 2, 1, ENC(rd))                    \
+        _(cfswsp, 0, 2, 1, ENC(rs2))                   \
+        _(cflw, 0, 2, 1, ENC(rs1, rd))                 \
+        _(cfsw, 0, 2, 1, ENC(rs1, rs2))                \
+        )                                              \
```
visitorckw, line 185: Please correct the indentation here.

visitorckw (approved):

jserv:
Thank @fourcolor for contributing!

Issue #287, opened 2023-11-30
CI: Generate Docker image with commit SHA1


henrybear327:
Currently, the generated docker image is tagged with `latest`. 

Docker Hub doesn't have a storage limit, and we would like to offer users the ability to switch between commits, thus we should append the commit SHA1 to the image tag.

henrybear327:
Can only verify when there is a PR with code changes (since docker hub images will only be updated when certain directories have files gotten modified)

henrybear327:
https://hub.docker.com/r/sysprog21/rv32emu/tags

verified

PR #320, opened 2024-01-04
CI: Generate Docker image with commit SHA1 #287

henrybear327:


jserv (changes requested):
Improve the git commit messages.
Append `Close #287` at the end of git commit message body.

.github/workflows/docker.yml
```diff
@@ -39,11 +39,15 @@ jobs:
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_ACCESS_TOKEN }}
+      - name: Get short commit SHA1
+        shell: bash
+        run: |
+          echo "sha_short=$(git rev-parse --short "$GITHUB_SHA")" >> "$GITHUB_ENV"
```
jserv: `GITHUB_SHA` looks weird.

henrybear327: According to the [docs](https://docs.github.com/en/actions/learn-github-actions/variables)
Screenshot 2024-01-05 at 2 46 17 AM

This returns the full commit hash. 

Is there anything weird with it?

jserv: It would be short hash instead of short sha.
https://git-scm.com/docs/git-rev-parse

henrybear327: Sorry, I am not sure if I understand this correctly. 

Isn't the commit SHA, which is the hash that identifies a specific commit, after passing it through `git rev-parse --short`, a short version of the commit SHA (a.k.a the short hash) will be returned?

As the GitHub CI environmental variable doesn't provide a short commit SHA as a variable to be referenced directly, and, the full string is named `GITHUB_SHA` by them, I can only use `GITHUB_SHA` directly for processing and I can't change the name of it.

jserv: > Isn't the commit SHA, which is the hash that identifies a specific commit, after passing it through `git rev-parse --short`, a short version of the commit SHA (a.k.a the short hash) will be returned?

I was concerned about the naming scheme instead of the logic behind this pull request.

henrybear327: Updated

PR #323, opened 2024-01-06
Fix rounding mode selection

visitorckw:
Some F extension instructions include an rm field, indicating the rounding mode to be used. When an instruction has an rm field, the rounding mode should be chosen based on the value of that field. If the rm field is set to 0b111, it signifies dynamic selection of the rounding mode based on the fcsr's frm field.

The current implementation always selected the rounding mode dynamically based on fcsr's frm, regardless of the rm field in the instruction. This patch corrects this error, ensuring that the rounding mode is now chosen according to the rm field when present.

### DIFF
--- a/src/decode.h
+++ b/src/decode.h
@@ -288,6 +288,20 @@ typedef struct rv_insn {
 #if RV32_HAS(EXT_C)
     uint8_t shamt;
 #endif
+
+#if RV32_HAS(EXT_F)
+    /* Floating-point operations use either a static rounding mode encoded in
+     * the instruction, or a dynamic rounding mode held in frm. A value of 111
+     * in the instruction’s rm field selects the dynamic rounding mode held in
+     * frm. If frm is set to an invalid value (101–111), any subsequent attempt
+     * to execute a floating-point operation with a dynamic rounding mode will
+     * cause an illegal instruction trap. Some instructions that have the rm
+     * field are nevertheless unaffected by the rounding mode; they should have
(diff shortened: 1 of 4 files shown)


src/decode.h
```diff
@@ -288,6 +288,11 @@ typedef struct rv_insn {
 #if RV32_HAS(EXT_C)
     uint8_t shamt;
 #endif
+
+#if RV32_HAS(EXT_F)
+    uint8_t rm;
```
jserv, line 302: Add corresponding comment. i.e., rounding mode.

src/softfloat.h
```diff
@@ -115,3 +115,17 @@ static inline void set_rounding_mode(riscv_t *rv)
     if (frm == 0b100)
         softfloat_roundingMode = softfloat_round_near_maxMag;
 }
+
+static inline void set_static_rounding_mode(uint8_t rm)
+{
+    if (rm == 0b000)
```
jserv: Rewrite with switch-case. In addition, consider the use of `__UNREACHABLE` when possible.

src/decode.h
```diff
@@ -288,6 +288,12 @@ typedef struct rv_insn {
 #if RV32_HAS(EXT_C)
     uint8_t shamt;
 #endif
+
+#if RV32_HAS(EXT_F)
+    /* rounding mode */
```
jserv: Outline the various rounding modes in this comment as specified in the RISC-V standard.

jserv:

PR #267, opened 2023-11-17
Integrate LibFuzzer

henrybear327:
By leveraging the LLVM's LibFuzzer, we will be able to test our emulator code more thoroughly.

The seed corpus being added automatically are all the elf files in the build folder.

Major changes:
- Conditional compilation for the main function, as the fuzzer will generate its own main function
- Add Clang to the Dockerfile 
- Introduce loading binary stream as elf file, as the input from fuzzer will be passed into the emulator directly instead of going through a file

Minor changes:
- Fixed ELF verification logic 

Currently, known improvements are:
- Fix the Makefile to handle the compilation flags and compile coexisting .c and .cc files correctly

PR #324, opened 2024-01-06
Fix signed integer overflow in RV32IM

visitorckw:
UBSAN has identified several signed integer overflow issues in the RV32IM implementation, including incorrect type conversions to int32_t for add/sub operations and a lack of prevention for overflow in multiplication.

Address these issues by ensuring proper type handling and preventing integer overflow in the RV32IM implementation.

### DIFF
--- a/src/rv32_constopt.c
+++ b/src/rv32_constopt.c
@@ -408,8 +408,9 @@ CONSTOPT(csrrci, { info->is_constant[ir->rd] = false; })
 CONSTOPT(mul, {
     if (info->is_constant[ir->rs1] && info->is_constant[ir->rs2]) {
         info->is_constant[ir->rd] = true;
-        ir->imm = (int32_t) info->const_val[ir->rs1] *
-                  (int32_t) info->const_val[ir->rs2];
+        const int64_t multiplicand = (int32_t) info->const_val[ir->rs1];
+        const int64_t multiplier = (int32_t) info->const_val[ir->rs2];
+        ir->imm = ((uint64_t) (multiplicand * multiplier)) & ((1ULL << 32) - 1);
         info->const_val[ir->rd] = ir->imm;
         ir->opcode = rv_insn_lui;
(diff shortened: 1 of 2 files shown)


jserv:
I defer to @qwe661234 for conformation. Meanwhile, I am thinking of enabling more checks about undefined behavior in CI pipeline.

qwe661234:
This modification seems reasonable, but we need a test case to verify it.

visitorckw:
I observe this undefined behavior in several prebuilt ELF binaries. Taking scimark2.elf as an example, by adding ```-g -fsanitize=undefined -fno-sanitize=alignment``` to CFLAGS and LDFLAGS in the Makefile, recompiling, and executing, we can observe the following errors:
```
src/rv32_template.c:1052:1: runtime error: signed integer overflow: 41505 * 63875 cannot be represented in type 'int'
src/rv32_template.c:733:1: runtime error: signed integer overflow: -1833556764 + -1620771432 cannot be represented in type 'int'
src/rv32_template.c:746:1: runtime error: signed integer overflow: -2095260944 - 984558896 cannot be represented in type 'int'
```

visitorckw:
Additionally, if the ```-fno-sanitize=alignment``` is removed, UBSAN will complain about a load of misaligned address caused by io.c. 
```
src/io.c:84:12: runtime error: load of misaligned address 0x7f0b7abff92e for type 'const uint32_t', which requires 4 byte alignment
```

visitorckw:
> I defer to @qwe661234 for conformation. Meanwhile, I am thinking of enabling more checks about undefined behavior in CI pipeline.

Perhaps we can consider incorporating UBSAN along with prebuilt ELF binaries for testing in CI. Alternatively, we may reconsider using a fuzzer for detection.

See: #267

jserv:
> Perhaps we can consider incorporating UBSAN along with prebuilt ELF binaries for testing in CI. Alternatively, we may reconsider using a fuzzer for detection.

Cc. @henrybear327

PR #326, opened 2024-01-08
CI: Skip pipelines if no source files are changed

### LINKED ISSUE #284: CI: Skip pipelines if no source files are changed
In the current CI pipeline, several actions are specifically designed for C source files. Consequently, when no changes are detected in the source files, it becomes reasonable to bypass most of the CI pipeline, including generic architecture tests and build checks.

henrybear327:
Based on some research ([1](https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow) [2](https://github.com/orgs/community/discussions/25280) [3](https://github.com/tj-actions/changed-files) [4](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions) [5](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)) regarding job dependencies on GitHub CI: 
- in order to define job dependencies, all jobs that have dependencies need to be defined in the same `yml` file
- there doesn't exist a way to cancel the started job gracefully 

Thus, based on the current knowledge
- a job to test for code-related changes in certain directories will be the pre-req job for jobs that only need to be run if there are code-related changes
- because of the GitHub CI restrictions, jobs that require file-change checks are moved to `mail.yml`. By using dependencies, we can avoid adding `if` checks for all the steps that require file-change checks

Close #284


.github/workflows/main.yml
```diff
@@ -76,3 +112,50 @@ jobs:
             .ci/check-newline.sh
             .ci/check-format.sh
       shell: bash
+  
+  arch-tests:
```
jserv: Should be compliance test.

PR #327, opened 2024-01-10
Misc. bug and typo fixes

henrybear327:
During the meeting, infer was mentioned, so I did a set up and ran infer against our codebase.

### DIFF
--- a/src/elf.c
+++ b/src/elf.c
@@ -300,8 +300,7 @@ bool elf_open(elf_t *e, const char *input)
 #if HAVE_MMAP
     int fd = open(path, O_RDONLY);
     if (fd < 0) {
-        free(path);
-        return false;
+        goto free_path;
     }
 
     /* get file size */
@@ -314,17 +313,14 @@ bool elf_open(elf_t *e, const char *input)
      */
     e->raw_data = mmap(0, st.st_size, PROT_READ, MAP_PRIVATE, fd, 0);
     if (e->raw_data == MAP_FAILED) {
-        release(e);
-        free(path);
-        return false;
+        goto free_fd;
     }
     close(fd);
 
 #else  /* fallback to standard I/O text stream */
     FILE *f = fopen(path, "rb");
     if (!f) {
-        free(path);
-        return false;
+        goto free_path;
     }
 
     /* get file size */
@@ -332,9 +328,7 @@ bool elf_open(elf_t *e, const char *input)
     e->raw_size = ftell(f);
     fseek(f, 0, SEEK_SET);
     if (e->raw_size == 0) {
-        fclose(f);
-        free(path);
-        return false;
+        goto free_fd;
     }
 
     /* allocate memory */
@@ -346,9 +340,7 @@ bool elf_open(elf_t *e, const char *input)
     const size_t r = fread(e->raw_data, 1, e->raw_size, f);
     fclose(f);
     if (r != e->raw_size) {
-        release(e);
-        free(path);
-        return false;
+        goto free_path;
     }
 #endif /* HAVE_MMAP */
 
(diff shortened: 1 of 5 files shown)


henrybear327:
Other errors detected by infer

The second one is `malloc` related (we didn't check if the `malloc` returns NULL or not). 

The first and third one I actually don't have a clue on what it's implying.

```
src/elf.c:194: error: Uninitialized Value
  `node.right_red` is read without initialization during the call to `map_insert()`. 
  192.     /* initialize the symbol table */
  193.     map_clear(e->symbols);
  194.     map_insert(e->symbols, &(int){0}, &(char *){NULL});
           ^
  195. 
  196.     /* get the string table */

src/jit.c:1535: error: Null Dereference
  `state` could be null (null value originating from line 1534) and is dereferenced. 
  1533. {
  1534.     struct jit_state *state = malloc(sizeof(struct jit_state));
  1535.     state->offset = 0;
            ^
  1536.     state->size = size;
  1537.     state->buf = mmap(0, size, PROT_READ | PROT_WRITE | PROT_EXEC,

src/riscv.c:310: error: Uninitialized Value
  `node.right_red` is read without initialization during the call to `map_insert()`. 
  308.     FILE *files[] = {stdin, stdout, stderr};
  309.     for (size_t i = 0; i < ARRAYS_SIZE(files); i++)
  310.         map_insert(s->fd_map, &i, &files[i]);
               ^
  311. 
  312.     return s;
```

jserv:
> Other errors detected by infer

Have you been resolving all the errors raised by infer?

src/elf.c
```diff
@@ -320,6 +320,7 @@ bool elf_open(elf_t *e, const char *input)
     if (e->raw_data == MAP_FAILED) {
         release(e);
         free(path);
+        close(fd);
```
jserv: Follow the method used by `cache_create` for employing `goto` to release resources, which results in more concise code.

src/emulate.c
```diff
@@ -1135,4 +1135,8 @@ void dump_registers(riscv_t *rv, char *out_file_path)
         fprintf(f, "  \"x%d\": %u%s\n", i, rv->X[i], comma);
     }
     fprintf(f, "}\n");
+
+    if (out_file_path[0] != '-') {
```
jserv: Remove `{` and `}`.

henrybear327:
> > Other errors detected by infer
> 
> Have you been resolving all the errors raised by infer?

I do not understand the first error raised by infer. Can you provide some hints if you have any? 

I will fix the second one by adding necessary NULL checks for `malloc` and `calloc`. 

The third one I actually don't see the issue though. Can you also provide some hints if you have any? 

Thank you!

jserv:
> I do not understand the first error raised by infer. Can you provide some hints if you have any?

You should check src/map.[ch] instead. Using static analysis is a bit tricky. Sometimes, you just figure out something considered as false alarm.

jserv (changes requested):
Refine the subject of this issue and git commit messages.

The statement "caught by infer" should only appear in the commit message body rather than the subject.

henrybear327:
The only remaining issue lies in the `map.c` code, which I tried to read the RB-tree source code and couldn't quite spot the issue so far. I suggest that we create another issue and debug the potential issue in the RB-tree in a separate PR.

```
src/elf.c:194: error: Uninitialized Value
`node.right_red` is read without initialization during the call to `map_insert()`.
  192.     /* initialize the symbol table */
  193.     map_clear(e->symbols);
  194.     map_insert(e->symbols, &(int){0}, &(char *){NULL});
              ^
  195.
  196.     /* get the string table */

Found 1 issue
    Issue Type(ISSUED_TYPE_ID): # 
    Uninitialized Value(PULSE_UNINITIALIZED_VALUE): 1
```

src/jit.c
```diff
@@ -1481,6 +1481,7 @@ uint32_t jit_translate(riscv_t *rv, block_t *block)
 struct jit_state *jit_state_init(size_t size)
 {
     struct jit_state *state = malloc(sizeof(struct jit_state));
+    assert(state != NULL);
```
jserv: Rewrite as `assert(state)`.

jserv (changes requested):
Split this pull requests into individual functional ones.

henrybear327:
> Split this pull requests into individual functional ones.

Split as requested

PR #328, opened 2024-01-11
Add dockerfile that contains the development environment

henrybear327:
We currently have a demo image that contains the `rv32emu` binary. 

This PR introduces a new rv32emu development image containing all the toolchains required in the docker directory, providing contributors with an easier way to get on board.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,11 @@ The usage and limitations of Doom and Quake demo are listed in [docs/demo.md](do
 
 ### Docker image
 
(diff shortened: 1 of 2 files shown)


jserv:
I am inclined to conclude this pull request due to the specific utilization context of Docker images, which are designed to streamline deployment processes for particular objectives, rather than encompassing a comprehensive solution. Concerning the development environment, it would be more efficient to offer specialized shell scripts. These scripts would automate tasks such as downloading, extracting, and executing additional modifications, thereby reducing dependency on tailor-made Docker images.

PR #333, opened 2024-01-21
Add profiler

### LINKED ISSUE #189: jit: Implement runtime instruction profiling
The existing JIT compilation process depends on heuristics that utilize fixed thresholds (#159) to determine when to transition from interpretation to JIT compilation when executing RISC-V instructions. This approach lacks flexibility and leads to inconsistent performance patterns. Consequently, there is a clear need for a more pragmatic method that involves gathering profiling data during interpretation. Furthermore, we require a defined strategy for making this transition based on the sampled data rather than relying on predetermined thresholds.

Let's consider the Java virtual machine (JVM), particularly HotSpot, which has a crucial objective: generating efficient machine code while minimizing runtime costs. To accomplish this, HotSpot employs a range of strategies, including tiered compilation, dynamic profiling, speculation, deoptimization, and various compiler optimizations, both architecture-specific and architecture-independent.

Typically, the execution of a method begins in the interpreter, which is the simplest and most cost-effective means available to HotSpot for executing code. During method execution, whether through interpretation or in compiled form, the dynamic profile of the method is collected through instrumentation. This profile is then used by several heuristics to make decisions, such as whether the method should be compiled, recompiled at a different optimization level, and which optimizations should be applied.

When an application starts, the JVM initially interprets all bytecode while gathering profiling information about it. The JIT compiler then leverages this collected profiling information to identify hotspots. Initially, the JIT compiler compiles frequently executed code sections with C1 to rapidly achieve native code performance. Later, as more profiling information becomes available, C2 comes into play. C2 recompiles the code with more aggressive and time-intensive optimizations to further enhance performance.
tiered-compilation

Another advantageous aspect of tiered compilation is the acquisition of more accurate profiling information. Prior to tiered compilation, the JVM collected profiling information only during interpretation. However, with tiered compilation enabled, the JVM also gathers profiling information on the code compiled with C1. As the compiled code delivers better performance, it allows the JVM to accumulate more profiling samples.

Reference:
* [How Tiered Compilation works in OpenJDK](https://devblogs.microsoft.com/java/how-tiered-compilation-works-in-openjdk/)
* [How the Intel® Simics® Simulator Executes Instructions](https://community.intel.com/t5/Blogs/Products-and-Solutions/Software/How-the-Intel-Simics-Simulator-Executes-Instructions/post/1543049)

qwe661234:
Based on our observation, a high percentage of true hotspots involve loops or backward jumps, but the number of IR is unstable within these true hotspots.

Therefore, we believe our profiler can use three indices to detect hotspots:

1. Backward jump
2. Loop
3. Used frequency

Close: #189

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,6 +255,23 @@ _Example Instructions Histogram_
 _Example Registers Histogram_
 ![Registers Hisrogram Example](docs/histogram-reg.png)
 
+## RISC-V Basic Block Usage Statistics
+
+To install [lolviz](https://github.com/parrt/lolviz)
+```shell
+$ pip install lolviz
+```
+For macOS users, installing might be required:
+```shell
+$ brew install graphviz
+```
+First, user need to crate the directory `prof` and build profiling data through executing `rv32emu`
+
+```shell
+$ ./build/rv32emu -p ./build/[test_program].elf
+$ ./tools/rv_profiler [--start-address|--stop-address|--graph-ir] [test_program]
+```
+
 ## Contributing
 
 See [CONTRIBUTING.md](CONTRIBUTING.md) for contribution guidelines.
--- a/src/cache.c
+++ b/src/cache.c
@@ -13,14 +13,6 @@
 #include "mpool.h"
 #include "utils.h"
 
-/* Currently, THRESHOLD is set to identify hot spots. Once the using frequency
- * for a block exceeds the THRESHOLD, the tier-1 JIT compiler process is
- * triggered.
- * FIXME: Implement effective  profiler to detect hot spots, instead of simply
- * relying on THRESHOLD.
- */
-#define THRESHOLD 4096
-
 static uint32_t cache_size, cache_size_bits;
 static struct mpool *cache_mp;
 
--- a/src/cache.c
+++ b/src/cache.c
@@ -166,7 +166,7 @@ free_lists:
     return NULL;
 }
 
-void *cache_get(const cache_t *cache, uint32_t key)
+void *cache_get(const cache_t *cache, uint32_t key, bool update)
 {
     if (!cache->capacity ||
         hlist_empty(&cache->map->ht_list_head[cache_hash(key)]))
@@ -192,7 +192,7 @@ void *cache_get(const cache_t *cache, uint32_t key)
      * code. The generated C code is then compiled into machine code by the
      * target compiler.
      */
-    if (entry->frequency < THRESHOLD) {
+    if (update && entry->frequency < THRESHOLD) {
         list_del_init(&entry->list);
         list_add(&entry->list, cache->lists[entry->frequency++]);
     }
@@ -242,7 +242,7 @@ void cache_free(cache_t *cache)
     free(cache);
 }
 
-uint32_t cache_freq(struct cache *cache, uint32_t key)
+uint32_t cache_freq(const struct cache *cache, uint32_t key)
 {
     if (!cache->capacity ||
         hlist_empty(&cache->map->ht_list_head[cache_hash(key)]))
--- a/src/cache.c
+++ b/src/cache.c
@@ -6,6 +6,7 @@
 #include <assert.h>
 #include <stdbool.h>
 #include <stddef.h>
+#include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 
@@ -273,5 +274,17 @@ bool cache_hot(const struct cache *cache, uint32_t key)
             return true;
     }
     return false;
+}
+void cache_profile(const struct cache *cache,
+                   FILE *output_file,
+                   prof_func_t func)
+{
+    assert(func);
+    for (int i = 0; i < THRESHOLD; i++) {
+        lfu_entry_t *entry, *safe;
+        list_for_each_entry_safe (entry, safe, cache->lists[i], list) {
+            func(entry->value, entry->frequency, output_file);
+        }
+    }
 }
 #endif
--- a/src/cache.h
+++ b/src/cache.h
@@ -8,6 +8,12 @@
 #include <stdbool.h>
 #include <stdint.h>
 
+/* Currently, THRESHOLD is set to identify hot spots. Once the using frequency
+ * for a block exceeds the THRESHOLD, the tier-1 JIT compiler process is
+ * triggered.
+ */
+#define THRESHOLD 4096
+
 struct cache;
 
 /** cache_create - crate a new cache
--- a/src/cache.h
+++ b/src/cache.h
@@ -7,6 +7,7 @@
 
 #include <stdbool.h>
 #include <stdint.h>
+#include <stdio.h>
 
 /* Currently, THRESHOLD is set to identify hot spots. Once the using frequency
  * for a block exceeds the THRESHOLD, the tier-1 JIT compiler process is
@@ -55,6 +56,11 @@ void cache_free(struct cache *cache);
  * @key: the key of the specified entry
  */
 bool cache_hot(const struct cache *cache, uint32_t key);
+
+typedef void (*prof_func_t)(void *, uint32_t, FILE *);
+void cache_profile(const struct cache *cache,
+                   FILE *output_file,
+                   prof_func_t func);
 #endif
 
 uint32_t cache_freq(const struct cache *cache, uint32_t key);
--- a/src/cache.h
+++ b/src/cache.h
@@ -20,9 +20,10 @@ struct cache *cache_create(int size_bits);
  * cache_get - retrieve the specified entry from the cache
  * @cache: a pointer points to target cache
  * @key: the key of the specified entry
+ * @update: update frequency or not
  * @return: the specified entry or NULL
  */
-void *cache_get(const struct cache *cache, uint32_t key);
+void *cache_get(const struct cache *cache, uint32_t key, bool update);
 
 /**
  * cache_put - insert a new entry into the cache
@@ -50,4 +51,4 @@ void cache_free(struct cache *cache);
 bool cache_hot(const struct cache *cache, uint32_t key);
 #endif
 
-uint32_t cache_freq(struct cache *cache, uint32_t key);
+uint32_t cache_freq(const struct cache *cache, uint32_t key);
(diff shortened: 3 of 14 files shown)


jserv (changes requested):
Check https://github.com/sysprog21/rv32emu/issues/189#issuecomment-1852165303 carefully. You should implement some of the following:
* Allow Linux perf to profile JITted code as [QEMU](https://github.com/qemu/qemu/blob/master/docs/devel/tcg.rst).
* Generate profiling logs and use report program to view as [mono](https://www.mono-project.com/docs/debug+profile/profile/profiler/) does.
* Display in tree-based view as [vmprof](https://github.com/vmprof/vmprof-python) does.

It is important to provide interactive ways to browse profiling data with external tools. See [All my favorite tracing tools: eBPF, QEMU, Perfetto, new ones I built and more](https://thume.ca/2023/12/02/tracing-methods/).

qwe661234:
> Check [#189 (comment)](https://github.com/sysprog21/rv32emu/issues/189#issuecomment-1852165303) carefully. You should implement some of the following:
> 
> * Allow Linux perf to profile JITted code as [QEMU](https://github.com/qemu/qemu/blob/master/docs/devel/tcg.rst).
> * Generate profiling logs and use report program to view as [mono](https://www.mono-project.com/docs/debug+profile/profile/profiler/) does.
> * Display in tree-based view as [vmprof](https://github.com/vmprof/vmprof-python) does.
> 
> It is important to provide interactive ways to browse profiling data with external tools. See [All my favorite tracing tools: eBPF, QEMU, Perfetto, new ones I built and more](https://thume.ca/2023/12/02/tracing-methods/).

This runtime profiler serves as a trigger for the tier-1 JIT compiler. Do you mean we should provide a log mode for users to dump the profiling information?

jserv:
> This runtime profiler serves as a trigger for the tier-1 JIT compiler. Do you mean we should provide a log mode for users to dump the profiling information?

If you are going to close #189, you should consolidate the profiling facilities.

src/riscv.c
```diff
@@ -325,3 +325,19 @@ void state_delete(state_t *s)
     memory_delete(s->mem);
     free(s);
 }
+
+#if RV32_HAS(JIT)
+void profile(block_t *block)
+{
+    printf("%#-8x|", block->pc_start);
```
jserv: Instead of displaying messages, configure them to be written to files. You can use the `dump_registers` function as a reference for this process. Furthermore, ensure the IR sequence is logged in files for future reference and analysis. This implies the need to develop a standalone tool specifically for analyzing the profiling data.

src/riscv.h
```diff
@@ -227,6 +227,10 @@ state_t *state_new(void);
 /* delete a state */
 void state_delete(state_t *s);
 
+#if RV32_HAS(JIT)
+void rv_profile(riscv_t *rv);
```
jserv: The function `rv_profile` needs to be designed as independent of the JIT configuration. This means it should be callable in both interpreter-only mode and when using the tier-1 JIT compiler, enabling it to gather runtime profiling data for subsequent analysis. Additionally, we should provide tools capable of handling graphical representations of this data.

src/emulate.c
```diff
@@ -1039,6 +1048,22 @@ static block_t *block_find_or_translate(riscv_t *rv)
...
 #if RV32_HAS(JIT)
+static bool runtime_profiler(riscv_t *rv, block_t *block)
+{
+    /* Based on our observation, a high percentage of true hotspots involve high
+     * using frequency, loops or backward jumps. Therefore, we believe our
+     * profiler can use three indices to detect hotspots */
+    uint32_t freq = cache_freq(rv->block_cache, block->pc_start);
+    /* to profile the block after chaining, the block should be executed first
+     */
+    if (freq >= 2 && (block->backward || block->loop))
```
jserv: Add `likely` and/or `unlikely` hints.

src/riscv_private.h
```diff
@@ -62,12 +62,13 @@ typedef struct block {
...
     rv_insn_t *ir_head, *ir_tail; /**< the first and last ir for this block */
-    bool backward;
 #if RV32_HAS(JIT)
-    bool hot; /**< Determine the block is hotspot or not */
+    bool backward; /**< Determine the block has backward jump or not */
+    bool hot;      /**< Determine the block is hotspot or not */
     uint32_t offset;
     bool
         translatable; /**< Determine the block has RV32AF insturctions or not */
+    bool loop;        /**< Determine the block has loop or not */
```
jserv: Rename to `has_loops`.

qwe661234:
Currently, we only support for providing the information about branch, branch-untaken, and IRs of a basic block. I think we can create another issue for supporting visualizing graph IR.

src/main.c
```diff
@@ -137,6 +142,11 @@ static bool parse_args(int argc, char **args)
         case 'm':
             opt_misaligned = true;
             break;
+        // case 'p':
```
jserv: It does not work.
```
$ build/rv32emu -p build/dhrystone.elf 
RV32I[MA] Emulator which loads an ELF file to execute.
Usage: ./build/rv32emu [options] [filename] [arguments]
Options:
  -t : print executable trace
  -d [filename]: dump registers as JSON to the given file or `-` (STDOUT)
  -q : Suppress outputs other than `dump-registers`
  -a [filename] : dump signature to the given file, required by arch-test test
  -m : enable misaligned memory access
  -p : print profiling data
  -h : show this message
```

jserv (changes requested):
1. Separate the processes of generating and rendering profiling data. Specifically, rv32emu (activated with the `-p` option) will be responsible for generating profiling data, while the rendering will be handled by the newly added tool `rv_profile`.
2. Ensure that block information, including the instructions (referencing ELF files for disassembly as needed), is captured as part of the profiling data. This approach will make the data more informative and valuable for analysis.
3. The proposed changes aim to equip the system with adequate capabilities for depicting runtime profiling information, a crucial aspect for ongoing JIT development.

src/main.c
qwe661234: I have uncommented, the profiling data log works for interpreter mode and JIT mode now. Please check the profiling data log meets the requirement ot not.

* command
```
./build/rv32emu -p prof_log ./build/pi.elf
```

jserv: > `build/rv32emu -p prof_log ./build/pi.elf`

There is no need to specify a distinct file name for the profiling data. You can just add a suffix, like `.prof`, to the existing file name. At the same time, make sure to create a header for the profiling data, enabling it to be easily identified by the profiler.

src/main.c
```diff
@@ -108,6 +112,7 @@ static void print_usage(const char *filename)
             "  -a [filename] : dump signature to the given file, "
             "required by arch-test test\n"
             "  -m : enable misaligned memory access\n"
+            "  -p : print profiling data\n"
```
jserv: Change the description of option `-p` to "generate profiling data."

src/cache.h
```diff
@@ -48,6 +56,11 @@ void cache_free(struct cache *cache);
  * @key: the key of the specified entry
  */
 bool cache_hot(const struct cache *cache, uint32_t key);
+
+typedef void (*profile_func)(void *, uint32_t, FILE *);
```
jserv: Rename to `prof_func_t`.

src/cache.c
```diff
@@ -282,4 +275,15 @@ bool cache_hot(const struct cache *cache, uint32_t key)
     }
     return false;
 }
+void cache_profile(const struct cache *cache,
+                   FILE *output_file,
+                   profile_func func)
+{
```
jserv, line 281: Add `assert(func)`.

jserv:
The rationale behind segregating the generation and rendering of profiling data includes:
1. Enabling graphical representations without complicating the main rv32emu program.
2. Facilitating the use of tools like objdump for disassembling specific blocks, exemplified by commands such as `objdump --start-address=0x50c40 --stop-address=0x50c60 -d`.
3. Allowing for the profiling of RISC-V executables in a separate, offline run to collect profiling data. This data can then be utilized to assign accurate weights to the predecessors and successors in [superoptimizers](https://www.usenix.org/legacy/event/osdi08/tech/full_papers/bansal/bansal.pdf).

qwe661234:
> The rationale behind segregating the generation and rendering of profiling data includes:
> 
> 1. Enabling graphical representations without complicating the main rv32emu program.
> 2. Facilitating the use of tools like objdump for disassembling specific blocks, exemplified by commands such as `objdump --start-address=0x50c40 --stop-address=0x50c60 -d`.
> 3. Allowing for the profiling of RISC-V executables in a separate, offline run to collect profiling data. This data can then be utilized to assign accurate weights to the predecessors and successors in [superoptimizers](https://www.usenix.org/legacy/event/osdi08/tech/full_papers/bansal/bansal.pdf).

1. We can achieve this by writing a python script to parse generated profiling data. 
2. We need another debug mode for this? because we don't have debug mode currently.

jserv:
> 1. We can achieve this by writing a python script to parse generated profiling data.

Yes, you can write a Python-based renderer for profiling data.

> 2. We need another debug mode for this? because we don't have debug mode currently.

No, you can specify the start/stop address for `objdump -d` without any debugging symbols.

qwe661234:
> Facilitating the use of tools like objdump for disassembling specific blocks, exemplified by commands such as objdump --start-address=0x50c40 --stop-address=0x50c60 -d.

Does you means this function supports for providing the instruction sequence of a basic block starting from address `start-address` without other profiling data?

jserv:
> Does you means this function supports for providing the instruction sequence of a basic block starting from address `start-address` without other profiling data?

It depends on the format/layout of profiling data recorded during the execution of RISC-V programs. You can simply record the entry/exit address of each block, so that you can consult binutils.

qwe661234:
Currently, the profiler script support for:
1. printing profiling data from `start_address` to `stop_address`
2. visualizing graph IR of the specficed program counter like example below
image

README.md
```diff
@@ -255,6 +255,23 @@ _Example Instructions Histogram_
...
+```
+For macOS users, installing might be required:
+```shell
+$ brew install graphviz
+```
+First, user need to crate the directory `prof` and build profiling data through executing `rv32emu`
+
+```shell
+$ ./build/rv32emu -p ./build/[test_program].elf
+$ ./tools/profiler.py [--start-address|--stop-address|--graph-ir] [test_program]
```
jserv: Rename to `tools/rv_profiler` without `.py`

PR #337, opened 2024-01-27
Refine file naming scheme for profiling data

### LINKED ISSUE #336: Incorrect basename generated by dynamic profiler
How to reproduce?
1. `git clone https://github.com/sysprog21/rv32emu.git --depth=1`
2. `cd rv32emu`
3. `make`
4. `mkdir prof`
5. `build/rv32emu -p build/hello.elf`
6. `ls prof`

You will see:
```
'llo'$'\206\025\177''.prof' 
```

Reason:
"main.c"
```
if (opt_prof_data) {
        char *prog_name = malloc(strlen(opt_prog_name) - 11);
        strncpy(prog_name, opt_prog_name + 8, strlen(opt_prog_name) - 12);
        prof_out_file = malloc(strlen(opt_prog_name) + 1);
        sprintf(prof_out_file, "./prof/%s.prof", prog_name);
    }
```

I think the above codes want to get the basename of the target program. But, it is only valid  if target program is specified as `./build/xxx.elf`. 

Solution:
Maybe using `man 3 basename` is easier and sufficient.

ChinYikMing:
Close #336

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -3,6 +3,9 @@
  * "LICENSE" for information on usage and redistribution of this file.
  */
 
+#include <assert.h>
+#include <libgen.h>
+#include <limits.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -32,7 +35,7 @@ static char *signature_out_file;
 static bool opt_quiet_outputs = false;
 
 /* target executable */
-static const char *opt_prog_name = "a.out";
+static char *opt_prog_name = "a.out";
 
 /* target argc and argv */
 static int prog_argc;
@@ -166,11 +169,21 @@ static bool parse_args(int argc, char **args)
      */
     prog_args = &args[optind];
     opt_prog_name = prog_args[0];
+
     if (opt_prof_data) {
-        char *prog_name = malloc(strlen(opt_prog_name) - 11);
-        strncpy(prog_name, opt_prog_name + 8, strlen(opt_prog_name) - 12);
-        prof_out_file = malloc(strlen(opt_prog_name) + 1);
-        sprintf(prof_out_file, "./prof/%s.prof", prog_name);
+        char cwd_path[PATH_MAX] = {0};
+        assert(getcwd(cwd_path, PATH_MAX));
+
+        char rel_path[PATH_MAX] = {0};
+        memcpy(rel_path, args[0], strlen(args[0]) - 7 /* strlen("rv32emu")*/);
+
+        char *prog_basename = basename(opt_prog_name);
+        prof_out_file = malloc(strlen(cwd_path) + 1 + strlen(rel_path) +
+                               strlen(prog_basename) + 5 + 1);
+        assert(prof_out_file);
+
+        sprintf(prof_out_file, "%s/%s%s.prof", cwd_path, rel_path,
+                prog_basename);
     }
     return true;
 }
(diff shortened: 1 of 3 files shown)


src/main.c
```diff
@@ -167,10 +169,10 @@ static bool parse_args(int argc, char **args)
...
     opt_prog_name = prog_args[0];
     if (opt_prof_data) {
-        char *prog_name = malloc(strlen(opt_prog_name) - 11);
-        strncpy(prog_name, opt_prog_name + 8, strlen(opt_prog_name) - 12);
-        prof_out_file = malloc(strlen(opt_prog_name) + 1);
-        sprintf(prof_out_file, "./prof/%s.prof", prog_name);
+        char *prog_basename = basename(opt_prog_name);
+        prof_out_file = malloc(strlen(prog_basename) + 12 + 1);
+        assert(prof_out_file);
+        sprintf(prof_out_file, "./prof/%s.prof", prog_basename);
```
jserv: The leading `prof/` directory name is not really making sense. Instead, let's place the profiling data in the same directory where the RV32 ELF file locates.

ChinYikMing: I think you mean `build` directory

ChinYikMing:
Note that the trailing spaces are removed automatically by editor when I motify "README" and "rv_profiler", not my primary intention.

qwe661234 (approved):

src/main.c
qwe661234: I think putting it in the `build` directory is reasonable, we can distinguish the executable files and profiling data by filename extension.

src/main.c
```diff
@@ -167,10 +169,10 @@ static bool parse_args(int argc, char **args)
...
     opt_prog_name = prog_args[0];
     if (opt_prof_data) {
-        char *prog_name = malloc(strlen(opt_prog_name) - 11);
-        strncpy(prog_name, opt_prog_name + 8, strlen(opt_prog_name) - 12);
-        prof_out_file = malloc(strlen(opt_prog_name) + 1);
-        sprintf(prof_out_file, "./prof/%s.prof", prog_name);
+        char *prog_basename = basename(opt_prog_name);
+        prof_out_file = malloc(strlen(prog_basename) + 11 + 1);
+        assert(prof_out_file);
+        sprintf(prof_out_file, "build/%s.prof", prog_basename);
```
jserv: Is the leading `build/` necessary?

ChinYikMing: It depends on working directory of `rv32emu` executable. There are 2 cases to run profiler:

1. 
```
// outside build directory
xxx/build/rv32emu -p xxx/build/xxx.elf
```

2. 
```
// inside build directory
./rv32emu -p xxx.elf
```

Without leading `build/`, the first case will always write to cwd which is parent directory of `xxx`. For second case, it is all right.

I think the new approach could be concatenating the cwd and relative path (e.g., `xxx/build/` in first case and `./` in second case) to `rv32emu` executable. This concatenated path will always point to the `build` directory.

jserv: > I think the new approach could be concatenating the cwd and relative path (e.g., `xxx/build/` in first case and `./` in second case) to `rv32emu` executable. This concatenated path will always point to the `build` directory.

I think the new one is better, and I expect a cleaner and elegant implementation.

jserv:

PR #338, opened 2024-01-28
Fix csrrc instruction behavior when rs1 is 0

visitorckw:
According to the RISC-V specification, no write operation should occur when rs1 is equal to 0. The current implementation incorrectly clears all bits in the specified CSR.

Introduces a fix that ensures the csrrc instruction refrains from performing any write operation when rs1 is 0.


jserv:

PR #340, opened 2024-01-31
Refine the API in the public header

ChinYikMing:
The following should be included in an emulator's simple and clear public API:
1. create/init core
2. run emulation
3. delete/destroy core

Other components, including as memory, file systems, program data, etc., should be abstracted from the user, as a result, setting a configuration value (`vm_attr_t`) is sufficient. The user should manage about memory (`state_t`) and elf stuff before this PR. The user may just construct a core, run it, and shut it down after this PR, so they won't need to worry about them anymore.

For stdio remapping, `rv_remap_stdstream` function is introduced.

The `vm_attr_t` has multiple fields and they are commented clearly in the code.

elf is reopened in `dump_test_signature` because elf is allocated during `rv_create`. It is acceptable to reopen elf since it is only for testing. Print inferior exit code to console inside main instead of `syscall_exit` because the actual usage of exit code depends on applications of using riscv public API.

The `io` interface is not changed in this PR because it could maybe reused with `semu` in some way, still need to be investigated. Also, Logging feature and system emulator integration are not implemented yet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,11 +11,6 @@ CFLAGS = -std=gnu99 -O2 -Wall -Wextra
 CFLAGS += -Wno-unused-label
 CFLAGS += -include src/common.h
 
-# Set the default stack pointer
-CFLAGS += -D DEFAULT_STACK_ADDR=0xFFFFE000
-# Set the default args starting address
-CFLAGS += -D DEFAULT_ARGS_ADDR=0xFFFFF000
-
 # Enable link-time optimization (LTO)
 ENABLE_LTO ?= 1
 ifeq ($(call has, LTO), 1)
@@ -121,7 +116,7 @@ endif
 ENABLE_JIT ?= 0
 $(call set-feature, JIT)
 ifeq ($(call has, JIT), 1)
-OBJS_EXT += jit.o 
+OBJS_EXT += jit.o
 ifneq ($(processor),$(filter $(processor),x86_64 aarch64 arm64))
 $(error JIT mode only supports for x64 and arm64 target currently.)
 endif
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -125,7 +125,7 @@ RV_EXCEPTION_LIST
  */
 #define RV_EXC_MISALIGN_HANDLER(mask_or_pc, type, compress, IO)       \
     IIF(IO)                                                           \
-    (if (!rv->io.allow_misalign && unlikely(addr & (mask_or_pc))),    \
+    (if (!PRIV(rv)->allow_misalign && unlikely(addr & (mask_or_pc))), \
      if (unlikely(insn_is_misaligned(PC))))                           \
     {                                                                 \
         rv->compressed = compress;                                    \
@@ -1182,7 +1182,7 @@ void ecall_handler(riscv_t *rv)
 
 void memset_handler(riscv_t *rv)
 {
-    memory_t *m = ((state_t *) rv->userdata)->mem;
+    memory_t *m = PRIV(rv)->mem;
     memset((char *) m->mem_base + rv->X[rv_reg_a0], rv->X[rv_reg_a1],
            rv->X[rv_reg_a2]);
     rv->PC = rv->X[rv_reg_ra] & ~1U;
@@ -1190,7 +1190,7 @@ void memset_handler(riscv_t *rv)
 
 void memcpy_handler(riscv_t *rv)
 {
-    memory_t *m = ((state_t *) rv->userdata)->mem;
+    memory_t *m = PRIV(rv)->mem;
     memcpy((char *) m->mem_base + rv->X[rv_reg_a0],
            (char *) m->mem_base + rv->X[rv_reg_a1], rv->X[rv_reg_a2]);
     rv->PC = rv->X[rv_reg_ra] & ~1U;
--- a/src/main.c
+++ b/src/main.c
@@ -74,35 +74,10 @@ IO_HANDLER_IMPL(byte, write_b, W)
 #undef R
 #undef W
 
-/* run: printing out an instruction trace */
-static void run_and_trace(riscv_t *rv, elf_t *elf)
-{
-    const uint32_t cycles_per_step = 1;
-
-    for (; !rv_has_halted(rv);) { /* run until the flag is done */
-        /* trace execution */
-        uint32_t pc = rv_get_pc(rv);
-        const char *sym = elf_find_symbol(elf, pc);
-        printf("%08x  %s\n", pc, (sym ? sym : ""));
-
-        /* step instructions */
-        rv_step(rv, cycles_per_step);
-    }
-}
-
-static void run(riscv_t *rv)
-{
-    const uint32_t cycles_per_step = 100;
-    for (; !rv_has_halted(rv);) { /* run until the flag is done */
-        /* step instructions */
-        rv_step(rv, cycles_per_step);
-    }
-}
-
 static void print_usage(const char *filename)
 {
     fprintf(stderr,
-            "RV32I[MA] Emulator which loads an ELF file to execute.\n"
+            "RV32I[MACF] Emulator which loads an ELF file to execute.\n"
             "Usage: %s [options] [filename] [arguments]\n"
             "Options:\n"
             "  -t : print executable trace\n"
@@ -188,8 +163,11 @@ static bool parse_args(int argc, char **args)
     return true;
 }
 
-static void dump_test_signature(elf_t *elf)
+static void dump_test_signature(const char *prog_name)
 {
+    elf_t *elf = elf_new();
+    assert(elf && elf_open(elf, prog_name));
+
     uint32_t start = 0, end = 0;
     const struct Elf32_Sym *sym;
     FILE *f = fopen(signature_out_file, "w");
@@ -212,7 +190,12 @@ static void dump_test_signature(elf_t *elf)
         fprintf(f, "%08x\n", memory_read_w(addr));
 
     fclose(f);
+    elf_delete(elf);
 }
+
+#define MEM_SIZE 0xFFFFFFFFULL  /* 2^32 - 1 */
+#define STACK_SIZE 0x1000       /* 4096 */
+#define ARGS_OFFSET_SIZE 0x1000 /* 4096 */
 
 int main(int argc, char **args)
 {
@@ -221,12 +204,28 @@ int main(int argc, char **args)
         return 1;
     }
 
-    /* open the ELF file from the file system */
-    elf_t *elf = elf_new();
-    if (!elf_open(elf, opt_prog_name)) {
-        fprintf(stderr, "Unable to open ELF file '%s'\n", opt_prog_name);
-        return 1;
-    }
+    int run_flag = 0;
+    run_flag |= opt_trace;
+#if RV32_HAS(GDBSTUB)
+    run_flag |= opt_gdbstub << 1;
+#endif
+    run_flag |= opt_prof_data << 2;
+
+    vm_attr_t attr = {
+        .mem_size = MEM_SIZE,
+        .stack_size = STACK_SIZE,
+        .args_offset_size = ARGS_OFFSET_SIZE,
+        .argc = prog_argc,
+        .argv = prog_args,
+        .log_level = 0,
+        .run_flag = run_flag,
+        .profile_output_file = prof_out_file,
+        .data.user = malloc(sizeof(vm_user_t)),
+        .cycle_per_step = 100,
+        .allow_misalign = opt_misaligned,
+    };
+    assert(attr.data.user);
+    attr.data.user->elf_program = opt_prog_name;
 
     /* install the I/O handlers for the RISC-V runtime */
     const riscv_io_t io = {
@@ -246,42 +245,16 @@ int main(int argc, char **args)
         .on_ebreak = ebreak_handler,
         .on_memcpy = memcpy_handler,
         .on_memset = memset_handler,
-        .allow_misalign = opt_misaligned,
     };
 
-    state_t *state = state_new();
-
-    /* find the start of the heap */
-    const struct Elf32_Sym *end;
-    if ((end = elf_get_symbol(elf, "_end")))
-        state->break_addr = end->st_value;
-
     /* create the RISC-V runtime */
-    riscv_t *rv =
-        rv_create(&io, state, prog_argc, prog_args, !opt_quiet_outputs);
+    riscv_t *rv = rv_create(&io, &attr);
     if (!rv) {
         fprintf(stderr, "Unable to create riscv emulator\n");
         return 1;
     }
 
-    /* load the ELF file into the memory abstraction */
-    if (!elf_load(elf, rv, state->mem)) {
-        fprintf(stderr, "Unable to load ELF file '%s'\n", args[1]);
-        return 1;
-    }
-
-    /* run based on the specified mode */
-    if (opt_trace) {
-        run_and_trace(rv, elf);
-    }
-#if RV32_HAS(GDBSTUB)
-    else if (opt_gdbstub) {
-        rv_debug(rv);
-    }
-#endif
-    else {
-        run(rv);
-    }
+    rv_run(rv);
 
     /* dump registers as JSON */
     if (opt_dump_regs)
@@ -289,14 +262,11 @@ int main(int argc, char **args)
 
     /* dump test result in test mode */
     if (opt_arch_test)
-        dump_test_signature(elf);
+        dump_test_signature(opt_prog_name);
 
-    if (opt_prof_data)
-        rv_profile(rv, prof_out_file);
     /* finalize the RISC-V runtime */
-    elf_delete(elf);
     rv_delete(rv);
-    state_delete(state);
 
+    printf("inferior exit code %d\n", attr.exit_code);
     return 0;
 }
--- a/src/riscv.c
+++ b/src/riscv.c
@@ -8,6 +8,17 @@
 #include <stdlib.h>
 #include <string.h>
 
+#if !defined(_WIN32) && !defined(_WIN64)
+#include <unistd.h>
+#define FILENO(x) fileno(x)
+#else
+#define FILENO(x) _fileno(x)
+#define STDIN_FILENO FILENO(stdin)
+#define STDOUT_FILENO FILENO(stdout)
+#define STDERR_FILENO FILENO(stderr)
+#endif
+
+#include "elf.h"
 #include "mpool.h"
 #include "riscv.h"
 #include "riscv_private.h"
@@ -63,12 +74,6 @@ static void block_map_destroy(riscv_t *rv)
 }
 #endif
 
-riscv_user_t rv_userdata(riscv_t *rv)
-{
-    assert(rv);
-    return rv->userdata;
-}
-
 bool rv_set_pc(riscv_t *rv, riscv_word_t pc)
 {
     assert(rv);
@@ -105,13 +110,55 @@ riscv_word_t rv_get_reg(riscv_t *rv, uint32_t reg)
     return ~0U;
 }
 
-riscv_t *rv_create(const riscv_io_t *io,
-                   riscv_user_t userdata,
-                   int argc,
-                   char **args,
-                   bool output_exit_code)
+/*
+ * Remap standard stream
+ *
+ * @rv: riscv
+ * @fsp: a list of pair of mapping from fd to FILE *
+ * @fsp_size: list size
+ *
+ * Note: fd inside fsp should be 0 or 1 or 2 only
+ *
+ */
+void rv_remap_stdstream(riscv_t *rv, fd_stream_pair_t *fsp, uint32_t fsp_size)
 {
-    assert(io);
+    assert(rv);
+
+    vm_attr_t *attr = PRIV(rv);
+    assert(attr && attr->fd_map);
+
+    for (uint32_t i = 0; i < fsp_size; i++) {
+        int fd = fsp[i].fd;
+        FILE *file = fsp[i].file;
+
+        if (!file)
+            continue;
+        if (fd != STDIN_FILENO && fd != STDOUT_FILENO && fd != STDERR_FILENO)
+            continue;
+
+        /* check if standard stream refered by fd exists or not */
+        map_iter_t it;
+        map_find(attr->fd_map, &it, &fd);
+        if (it.node) /* found, remove first */
+            map_erase(attr->fd_map, &it);
+        map_insert(attr->fd_map, &fd, &file);
+
+        /* store new fd to make the vm_attr_t consistent */
+        int new_fd = FILENO(file);
+        assert(new_fd != -1);
+
+        if (fd == STDIN_FILENO)
+            attr->fd_stdin = new_fd;
+        else if (fd == STDOUT_FILENO)
+            attr->fd_stdout = new_fd;
+        else
+            attr->fd_stderr = new_fd;
+    }
+}
+
+riscv_t *rv_create(const riscv_io_t *io, riscv_user_t rv_attr)
+{
+    assert(io && rv_attr);
 
     riscv_t *rv = calloc(1, sizeof(riscv_t));
     assert(rv);
@@ -119,10 +166,48 @@ riscv_t *rv_create(const riscv_io_t *io,
     /* copy over the IO interface */
     memcpy(&rv->io, io, sizeof(riscv_io_t));
 
-    /* copy over the userdata */
-    rv->userdata = userdata;
+    /* copy over the attr */
+    rv->data = rv_attr;
 
-    rv->output_exit_code = output_exit_code;
+    vm_attr_t *attr = PRIV(rv);
+    attr->mem = memory_new(attr->mem_size);
+    assert(attr->mem);
+
+    /* reset */
+    rv_reset(rv, 0U);
+
+    if (attr->data.user) {
+        elf_t *elf = elf_new();
+        assert(elf && elf_open(elf, (attr->data.user)->elf_program));
+
+        const struct Elf32_Sym *end;
+        if ((end = elf_get_symbol(elf, "_end")))
+            attr->break_addr = end->st_value;
+
+        assert(elf_load(elf, attr->mem));
+
+        /* set the entry pc */
+        const struct Elf32_Ehdr *hdr = get_elf_header(elf);
+        assert(rv_set_pc(rv, hdr->e_entry));
+
+        elf_delete(elf);
+    } else {
+        /* TODO: system emulator */
+    }
+
+    /*
+     * default standard stream
+     *
+     * rv_remap_stdstream can be called to overwrite them
+     */
+    attr->fd_map = map_init(int, FILE *, map_cmp_int);
+    rv_remap_stdstream(rv,
+                       (fd_stream_pair_t[]){
+                           {STDIN_FILENO, stdin},
+                           {STDOUT_FILENO, stdout},
+                           {STDERR_FILENO, stderr},
+                       },
+                       3);
 
     /* create block and IRs memory pool */
     rv->block_mp = mpool_create(sizeof(block_t) << BLOCK_MAP_CAPACITY_BITS,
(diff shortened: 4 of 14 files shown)


src/syscall_sdl.c
```diff
@@ -149,7 +149,7 @@ static submission_queue_t submission_queue = {
 
 static submission_t submission_pop(riscv_t *rv)
 {
-    state_t *s = rv_userdata(rv);
+    vm_attr_t *s = rv_userdata(rv);
```
ChinYikMing: `s` should be renamed to `vm_attr`

src/syscall_sdl.c
```diff
@@ -162,7 +162,7 @@ static submission_t submission_pop(riscv_t *rv)
 
 static void event_push(riscv_t *rv, event_t event)
 {
-    state_t *s = rv_userdata(rv);
+    vm_attr_t *s = rv_userdata(rv);
```
ChinYikMing: `s` should be renamed to `vm_attr`

src/syscall_sdl.c
```diff
@@ -279,7 +279,7 @@ static bool check_sdl(riscv_t *rv, int width, int height)
 
 void syscall_draw_frame(riscv_t *rv)
 {
-    state_t *s = rv_userdata(rv); /* access userdata */
+    vm_attr_t *s = rv_userdata(rv); /* access userdata */
```
ChinYikMing: `s` should be renamed to `vm_attr`

src/syscall_sdl.c
```diff
@@ -677,7 +677,7 @@ static void *music_handler(void *arg)
 
 static void play_sfx(riscv_t *rv)
 {
-    state_t *s = rv_userdata(rv); /* access userdata */
+    vm_attr_t *s = rv_userdata(rv); /* access userdata */
```
ChinYikMing: `s` should be renamed to `vm_attr`

src/syscall_sdl.c
```diff
@@ -705,7 +705,7 @@ static void play_sfx(riscv_t *rv)
 
 static void play_music(riscv_t *rv)
 {
-    state_t *s = rv_userdata(rv); /* access userdata */
+    vm_attr_t *s = rv_userdata(rv); /* access userdata */
```
ChinYikMing: `s` should be renamed to `vm_attr`

src/emulate.c
```diff
@@ -1182,15 +1183,15 @@ void ecall_handler(riscv_t *rv)
 
 void memset_handler(riscv_t *rv)
 {
-    memory_t *m = ((state_t *) rv->userdata)->mem;
+    memory_t *m = ((vm_attr_t *) rv->userdata)->mem;
```
jserv: We can introduce a helper macro similar to how `PRIV` is used in semu. That is,
```c
#define PRIV(x) ((vm_attr_t *) x->userdata)
```

src/main.c
```diff
@@ -221,12 +215,19 @@ int main(int argc, char **args)
         return 1;
     }
 
-    /* open the ELF file from the file system */
-    elf_t *elf = elf_new();
-    if (!elf_open(elf, opt_prog_name)) {
-        fprintf(stderr, "Unable to open ELF file '%s'\n", opt_prog_name);
-        return 1;
-    }
+    vm_attr_t vm_attr = {
```
jserv: Name the instance of `vm_attr_t` simply as `attr`.

src/riscv.c
```diff
@@ -105,11 +106,57 @@ riscv_word_t rv_get_reg(riscv_t *rv, uint32_t reg)
...
-                   bool output_exit_code)
+/*
+ * Register non-common stdio
+ *
+ * @rv: riscv
+ * @stdio: 0 or 1 or 2
+ * @new_fptr: FILE * for new redirection
+ *
+ */
+void rv_register_stdio(riscv_t *rv, int stdio_fd, FILE *new_fptr)
```
jserv: Quote [Wikipedia](https://en.wikipedia.org/wiki/Standard_streams):
> In computer programming, standard streams are preconnected input and output communication channels between a computer program and its environment when it begins execution. The three input/output (I/O) connections are called standard input (stdin), standard output (stdout) and standard error (stderr).

In the proposed changed code, the use of stdio is confusing.

src/syscall_sdl.c
jserv: Or, `attr`.

ChinYikMing:
As discuss before, exit code shall be stored and maybe not to output to console directly. 

Unfortunately, CI/host-x64 and CI/host-arm run `make check` and `make misalign` for testing, and the compare signature includes "inferior exit code 0" during exit but I do not output them in this PR. Shall we modify `make check` and `make misalign` ?

src/syscall.c
```diff
@@ -376,4 +380,12 @@ void syscall_handler(riscv_t *rv)
         fprintf(stderr, "unknown syscall %d\n", (int) syscall);
         break;
     }
+
+    /*
+     * save return code
+     *
+     * the application decides the usage of the return code
+     */
+    vm_attr_t *vm_attr = rv_userdata(rv); /* access userdata */
```
jserv: The line `vm_attr_t *vm_attr = rv_userdata(rv);` might be perplexing due to the similar naming of the object type and the `vm_attr`, while the value comes from the `rv_userdata` call. This suggests a presumed direct link between `rv_userdata` and `vm_attr_t`. However, this connection is not explicitly clarified in the API documentation, leading to potential confusion regarding their relationship.

jserv:
> As discuss before, exit code shall be stored and maybe not to output to console directly.
> Unfortunately, CI/host-x64 and CI/host-arm run `make check` and `make misalign` for testing, and the compare signature includes "inferior exit code 0" during exit but I do not output them in this PR. Shall we modify `make check` and `make misalign` ?

After a RISC-V program ends, it is important to verify the exit code (i.e., `$?`). Consequently, modifications to 'check' and related build targets are necessary to accommodate this.

ChinYikMing:
> After a RISC-V program ends, it is important to verify the exit code (i.e., `$?`). Consequently, modifications to 'check' and related build targets are necessary to accommodate this.

Sure, a build macro can be added to `syscall_exit` and print  the exit code only on test for passing CI.

visitorckw:
Due to the memory size being changed to a runtime-defined value, do we need to check if the addresses of load and store instructions exceed the specified range?

src/riscv.c
ChinYikMing: I think `rv_remap_stream(riscv_t *rv, int fd, FILE *fptr)` is more comprehensive. More comments can be added to classify the parameters.

jserv: How about `rv_remap_stdstream(riscv_t *rv, int fd, FILE *filev)`?
Here `filev` means a vector of `FILE *`.

ChinYikMing: A fd is given in a call, is it necessary required a vector of `FILE *`?

jserv: > A fd is given in a call, is it necessary required a vector of `FILE *`?

No, my intention was to encapsulate the pair of a file descriptor and standard streams. This was meant to form a vector, but I misplaced the parameters.

ChinYikMing: > How about rv_remap_stdstream(riscv_t *rv, int fd, FILE *filev)?

Much more sense than `rv_register_stdio`.

> > A fd is given in a call, is it necessary required a vector of `FILE *`?
> 
> No, my intention was to encapsulate the pair of a file descriptor and standard streams. This was meant to form a vector, but I misplaced the parameters.

Got it. So, the signature becomes `rv_remap_stdstream(riscv_t *rv, int *fdv, FILE *filev)`. We could assume the length of the vector is 3, otherwise undefined behavior. If any `fd` in `fdv` is -1, it fallback to use default stdio.

jserv: > So, the signature becomes `rv_remap_stdstream(riscv_t *rv, int *fdv, FILE *filev)`. We could assume the length of the vector is 3, otherwise undefined behavior. If any `fd` in `fdv` is -1, it fallback to use default stdio.

To mitigate the risk of passing null or malformed pointers, we could introduce a structure, `fd_stream_pair`, to encapsulate file descriptors and `FILE` pointers. This structure can then be reliably passed as a pointer to the `rv_remap_stdstream` function, ensuring safer and more organized handling of these elements.

jserv:
I would like to express a concern regarding the `rv_userdata` function being part of the public API. Originally, it was designed for internal implementation, which suggests there is no necessity to make `rv_userdata` accessible to users interacting with the public API. It might be more appropriate to keep this function internal to maintain clarity and simplicity in the API's usage.

src/syscall.c
ChinYikMing: Added more comments about them. Hopefully it is clearer now.

ChinYikMing:
> Sure, a build macro can be added to `syscall_exit` and print the exit code only on test for passing CI.

The PRINT_EXIT_CODE build macro do not used in default build, thus make CI/host-x64 or CI/host-arm64 failed. Shall I add it to  default build?  Personally, I am prefer default build without PRINT_EXIT_CODE.

Another way to pass CI is that modifying the CI flow, build `rv32emu` with PRINT_EXIT_CODE  in CI tests.

jserv:
> > Sure, a build macro can be added to `syscall_exit` and print the exit code only on test for passing CI.
> 
> The PRINT_EXIT_CODE build macro do not used in default build, thus make CI/host-x64 or CI/host-arm64 failed. Shall I add it to default build? Personally, I am prefer default build without PRINT_EXIT_CODE.
> Another way to pass CI is that modifying the CI flow, build `rv32emu` with PRINT_EXIT_CODE in CI tests.

Drop macro `PRINT_EXIT_CODE` because it is not necessary to have multiple builds. Instead, let the shell or builtin commands to check exit code of RISC-V program emulated.

ChinYikMing:
> Due to the memory size being changed to a runtime-defined value, do we need to check if the addresses of load and store instructions exceed the specified range?

Sound reasonable. Maybe in future PR.

Makefile
```diff
@@ -214,6 +210,7 @@ check: $(BIN)
 	)
 
 EXPECTED_aes_sha1 = 1242a6757c8aef23e50b5264f5941a2f4b4a347e  -
```
ChinYikMing: May I know the original input of this hash value?

ChinYikMing: Maybe I guessed it

ChinYikMing:
> Drop macro `PRINT_EXIT_CODE` because it is not necessary to have multiple builds. Instead, let the shell or builtin commands to check exit code of RISC-V program emulated.

For minimal changes, printing `inferior exit code %d` inside `main` is able to pass the CI tests.

ChinYikMing:
> I would like to express a concern regarding the `rv_userdata` function being part of the public API. Originally, it was designed for internal implementation, which suggests there is no necessity to make `rv_userdata` accessible to users interacting with the public API. It might be more appropriate to keep this function internal to maintain clarity and simplicity in the API's usage.

Actually, we could introduce some helper functions or getter functions to access the specific field of userdata without losing the encapsulation. For example, some helper functions for memory viewing instead of returning whole userdata.

src/main.c
```diff
@@ -212,21 +210,33 @@ static void dump_test_signature(elf_t *elf)
...
+        .mem_size = MEM_SIZE,
+        .stack_size = STACK_SIZE,
+        .args_offset_size = ARGS_OFFSET_SIZE,
+        .argc = prog_argc,
+        .argv = prog_args,
+        .logging_level = 0,
+        .emu_data.vm_user = malloc(sizeof(vm_user_t)),
+        .cycle_per_step = 100,
+        .allow_misalign = opt_misaligned,
+    };
```
RinHizakura, line 226: Since we are going to make these parameters user-defined now, I will suggest having some checks (`check_vm_attr()`) on the `vm_attr_t` before the emulator creation stage, otherwise bad parameters can lead to a crash which is not easy to figure out quickly. For example:  `mem_size` should be larger than `stack_size + args_offset_size`.

Note that `mem_size` should also be larger than the ELF mapping range to load correctly, but this may not be easy to check before the emulator creation stage. As an alternative, maybe applying the range check for `memory_write` at the emulator running stage is a simple idea to solve this. We may need to take care of the possible performance drop for the extra comparison on this solution.

src/main.c
```diff
@@ -246,33 +256,18 @@ int main(int argc, char **args)
...
-    /* load the ELF file into the memory abstraction */
-    if (!elf_load(elf, rv, state->mem)) {
-        fprintf(stderr, "Unable to load ELF file '%s'\n", args[1]);
-        return 1;
-    }
-
     /* run based on the specified mode */
     if (opt_trace) {
-        run_and_trace(rv, elf);
+        run_and_trace(rv, opt_prog_name);
```
RinHizakura: Why don't we just realize the elf name by `PRIV(rv)`? It seems that we don't need to pass `opt_prog_name` again here.

src/riscv.c
```diff
@@ -105,11 +106,65 @@ riscv_word_t rv_get_reg(riscv_t *rv, uint32_t reg)
...
+ * @fsp: a list of pair of mapping from fd to FILE *
+ * @fsp_size: list size
+ *
+ * Note: fd inside fsp should be 0 or 1 or 2 only
+ *
+ */
+void rv_remap_stdstream(riscv_t *rv, fd_stream_pair_t *fsp, uint32_t fsp_size)
+{
+    if (!rv)
+        return;
```
RinHizakura: To make the implementation consistent, replace with `assert(rv)`.

src/riscv.c
```diff
@@ -105,11 +106,65 @@ riscv_word_t rv_get_reg(riscv_t *rv, uint32_t reg)
...
+
+    vm_attr_t *attr = PRIV(rv);
+    if (!attr)
+        return;
```
RinHizakura: Similarly, replace with `assert(attr)`.

src/riscv.c
```diff
@@ -105,11 +106,65 @@ riscv_word_t rv_get_reg(riscv_t *rv, uint32_t reg)
...
+            break;
+        default:
+            break;
+        }
+    }
+}
+
+riscv_t *rv_create(const riscv_io_t *io, riscv_user_t userdata)
 {
     assert(io);
```
RinHizakura: Let's `assert(attr)` here since we'll access it without checking in the future.

src/riscv.c
```diff
@@ -105,11 +106,65 @@ riscv_word_t rv_get_reg(riscv_t *rv, uint32_t reg)
...
+    vm_attr_t *attr = PRIV(rv);
+    if (!attr)
+        return;
+    if (!attr->fd_map)
+        return;
+
+    for (uint32_t i = 0; i < fsp_size; i++) {
+        int fd = fsp[i].fd;
+        FILE *file = fsp[i].file;
+        if (fd != 0 && fd != 1 && fd != 2)
```
RinHizakura: Maybe it's not a great idea to use 0, 1, and 2 here directly. @jserv Do you think we should define macros for each and expand them to the correct value considering the operating system(e.g.  `STDIN_FILENO` for POSIX and _fileno(stdin) on WIN)?

jserv: > Do you think we should define macros for each and expand them to the correct value considering the operating system(e.g. `STDIN_FILENO` for POSIX and _fileno(stdin) on WIN)?

Agree. We shall proceed in a portable way.

src/main.c
ChinYikMing: Nice catch!

src/riscv.c
ChinYikMing: See [commit](https://github.com/sysprog21/rv32emu/pull/340/commits/f953ac1e051558623ae3d1bfd4cee70077e47c53#diff-f99b8d1f48ef2257a9bbbe6b487c621f74ef233c00858df5a37b06e89441e836)

src/main.c
ChinYikMing: As `PRIV(rv)` requires internal structure, `rv_userdata(rv)` as alternative.

src/riscv.c
```diff
@@ -105,13 +114,63 @@ riscv_word_t rv_get_reg(riscv_t *rv, uint32_t reg)
...
+        int fd = fsp[i].fd;
+        FILE *file = fsp[i].file;
+
+        if (!file)
+            continue;
+        if (fd != STDIN_FILENO && fd != STDOUT_FILENO && fd != STDERR_FILENO)
+            continue;
+
+        /*
+         * check if stdio_fd exists or not
```
jserv: Refine the use of `stdio_fd`.

src/riscv.h
```diff
@@ -97,6 +98,8 @@ enum {
 
 #define BLOCK_MAP_CAPACITY_BITS 10
 
+#define PRIV(x) ((vm_attr_t *) x->userdata)
```
jserv: The macro `PRIV` should not appear in public header, `riscv.h`.

src/riscv.h
```diff
@@ -143,22 +146,19 @@ typedef struct {
     riscv_on_ebreak on_ebreak;
     riscv_on_memset on_memset;
     riscv_on_memcpy on_memcpy;
-    /* enable misaligned memory access */
-    bool allow_misalign;
 } riscv_io_t;
 
+/* run emulation */
+void run(riscv_t *rv);
```
jserv: To maintain a consistent API naming scheme and to prevent potential symbol conflicts, it is advisable to choose a name other than `run` for the function.

src/riscv.h
```diff
@@ -168,7 +168,19 @@ void rv_debug(riscv_t *rv);
 /* step the RISC-V emulator */
 void rv_step(riscv_t *rv, int32_t cycles);
 
-/* get RISC-V user data bound to an emulator */
+/*
+ * @rv: RICV-V
```
jserv: In this context, `rv` represents the instance of the RISC-V emulator.

src/riscv.h
```diff
@@ -207,25 +227,82 @@ void rv_halt(riscv_t *rv);
 /* return the halt state */
 bool rv_has_halted(riscv_t *rv);
 
-/* return the flag of outputting exit code */
-bool rv_enables_to_output_exit_code(riscv_t *rv);
+typedef struct {
+    char *elf_program;
+} vm_user_t;
+
+typedef struct {
```
jserv: Let's hold off on including the system emulation components until we are fully prepared to merge them.

src/riscv.h
```diff
@@ -207,25 +227,82 @@ void rv_halt(riscv_t *rv);
...
+    char *kernel_img;
+    char *dtb;
+    char *rootfs_img;
+} vm_sys_t;
+
+typedef union {
+    vm_user_t *vm_user;
+    vm_sys_t *vm_sys;
```
jserv: Ditto. Get rid of `vm_sys` and place proper TODO comments.

src/riscv.h
```diff
@@ -207,25 +227,82 @@ void rv_halt(riscv_t *rv);
...
+
+    /* allow misaligned memory access */
+    bool allow_misalign;
+
+    /*
+     * default stdin, stdout, stderr, set by rv_create
+     *
+     * use rv_remap_stdstream to overwrite them
+     */
+    int stdin_fd;
```
jserv: Unify the naming in the form of `fd_stdxxx`. That is,
```c
int fd_stdin, fd_stdout, fd_stderr;
```

src/riscv.h
```diff
@@ -180,6 +192,14 @@ riscv_word_t rv_get_pc(riscv_t *rv);
 /* set a register of the RISC-V emulator */
 void rv_set_reg(riscv_t *rv, uint32_t reg, riscv_word_t in);
 
+typedef struct {
+    int fd;
+    FILE *file;
+} fd_stream_pair_t;
+
+/* remap stdio stream */
```
jserv: Refine the comment.

ChinYikMing:
> Actually, we could introduce some helper functions or getter functions to access the specific field of userdata without losing the encapsulation. For example, some helper functions for memory viewing instead of returning whole userdata.

Shall we drop it for now? Later on, we can think about what fields are gonna export to applications.

Note that the `run_and_trace` function requires program name to run for now, see [here](https://github.com/sysprog21/rv32emu/pull/340/files#diff-e0cf5b28d9b6b600f0af2bc78e8fd30ec675fd731a5da86f0c4283ffc0e40176). If we drop `rv_userdata`, the `run_and_trace` can only get program name by passing `opt_prog_name` or `attr` to it.

src/riscv.h
ChinYikMing: Stored it in `riscv_private.h`

src/riscv.h
ChinYikMing: How about `rv_start_emu`?

jserv: > How about `rv_start_emu`?

Opt for concise and meaningful names. e.g., `rv_mainloop` and `rv_run`.

jserv:
> Note that the run_and_trace function requires program name to run for now, see [here](https://github.com/sysprog21/rv32emu/pull/340/files#diff-e0cf5b28d9b6b600f0af2bc78e8fd30ec675fd731a5da86f0c4283ffc0e40176). If we drop rv_userdata, the run_and_trace can only get program name by passing opt_prog_name or attr to it.

We could streamline the `run` and `run_and_trace` functions by treating `run_and_trace` as a specialized variant of `rv_run`, activated through specific callback functions or flags. This strategy would enable us to eliminate the need for exposing `rv_userdata` directly.

src/riscv.h
jserv: Check the manner of GLib. The [GMainLoop struct](https://docs.gtk.org/glib/struct.MainLoop.html) is an opaque data type representing the main event loop of a GLib or GTK application.

src/main.c
ChinYikMing: Nice perspective. I think validation of `vm_attr_t` could be complex. For example, shall we check the invalid file during validation or propagate to `elf_open` to verify the invalidation. Similar situation for memory, if the user-provided memory size is not enough to run emulation, shall we simply abort before `rv_create` or propagate to segmentation fault. We may need more discussion about this. Maybe in future PR?

ChinYikMing:
> > Note that the run_and_trace function requires program name to run for now, see [here](https://github.com/sysprog21/rv32emu/pull/340/files#diff-e0cf5b28d9b6b600f0af2bc78e8fd30ec675fd731a5da86f0c4283ffc0e40176). If we drop rv_userdata, the run_and_trace can only get program name by passing opt_prog_name or attr to it.
> 
> We could streamline the `run` and `run_and_trace` functions by treating `run_and_trace` as a specialized variant of `rv_run`, activated through specific callback functions or flags. This strategy would enable us to eliminate the need for exposing `rv_userdata` directly.

Since we have introduced `vm_attr_t`, we can simply add one more field (e.g., `trace`). `rv_run` can check that easily.

ChinYikMing:
> Since we have introduced `vm_attr_t`, we can simply add one more field (e.g., `trace`). `rv_run` can check that easily.

Another way: we can introduce `rv_set_main_loop(rv, void callback(void *arg), void *arg)`. The application can pass their own callback to register main loop. In this way, we can simply call their callback and decouple for many scenario. Inspired by [emscripten](https://emscripten.org/docs/api_reference/emscripten.h.html?highlight=main_loop#c.emscripten_set_main_loop). If callback is NULL, we can fallback to normal `rv_run`. If callback is NULL and `trace` field in `vm_attr_t` is set, then fallback to `rv_run_and_trace`. Specifically, `rv_run` and `rv_run_and_trace` are private function in "riscv.c".

jserv:
> Since we have introduced `vm_attr_t`, we can simply add one more field (e.g., `trace`). `rv_run` can check that easily.

Exactly. It is time to drop `rv_userdata` by extensive use of `vm_attr_t`. We may refine the term "userdata" as well to avoid unintended confusion.

ChinYikMing:
> > > > Note that the run_and_trace function requires program name to run for now, see [here](https://github.com/sysprog21/rv32emu/pull/340/files#diff-e0cf5b28d9b6b600f0af2bc78e8fd30ec675fd731a5da86f0c4283ffc0e40176). If we drop rv_userdata, the run_and_trace can only get program name by passing opt_prog_name or attr to it.
> > > 
> > > 
> > > We could streamline the `run` and `run_and_trace` functions by treating `run_and_trace` as a specialized variant of `rv_run`, activated through specific callback functions or flags. This strategy would enable us to eliminate the need for exposing `rv_userdata` directly.
> > 
> > 
> > Since we have introduced `vm_attr_t`, we can simply add one more field (e.g., `trace`). `rv_run` can check that easily.
> 
> Another way: we can introduce `rv_set_main_loop(void callback(void *arg), void *arg)`. The application can pass their own callback to register main loop. In this way, we can simply call their callback and decouple for many scenario. Inspired by [emscripten](https://emscripten.org/docs/api_reference/emscripten.h.html?highlight=main_loop#c.emscripten_set_main_loop). If callback is NULL, we can fallback to normal `rv_run`.

How about this way?

jserv:
> Another way: we can introduce `rv_set_main_loop(void callback(void *arg), void *arg)`. The application can pass their own callback to register main loop. In this way, we can simply call their callback and decouple for many scenario. Inspired by [emscripten](https://emscripten.org/docs/api_reference/emscripten.h.html?highlight=main_loop#c.emscripten_set_main_loop). If callback is NULL, we can fallback to normal `rv_run`. If callback is NULL and `trace` field in `vm_attr_t` is set, then fallback to `rv_run_and_trace`. Specifically, `rv_run` and `rv_run_and_trace` are private function in "riscv.c".

Including the callback function as part of the emulator's context is possible, but it’s not necessary to have a separate function like `rv_set_main_loop` for assigning callbacks. Specifying the callback function directly using C99 designated initializers is more efficient, especially since it is unlikely we'll need to change the callback during the execution of the RISC-V emulation main loop.

src/riscv.h
ChinYikMing: Since we drop the `rv_userdata`, no more comment about them.

src/riscv.h
```diff
@@ -207,27 +210,97 @@ void rv_halt(riscv_t *rv);
...
+    /* FIXME: rv32emu cannot access envp yet */
+
+    /* emulation program exit code */
+    int exit_code;
+
+    /* emulation program error code */
+    int error;
+
+    /* TODO: for logging feature */
+    int logging_level;
```
jserv: Rename to `log_level` for short.

src/riscv_private.h
```diff
@@ -100,7 +102,7 @@ struct riscv_internal {
     riscv_word_t PC;
 
     /* user provided data */
-    riscv_user_t userdata;
+    riscv_user_t attr;
```
jserv: Would it be better to name this field as `priv` or `data`?

src/riscv.h
```diff
@@ -207,27 +210,97 @@ void rv_halt(riscv_t *rv);
...
-/* return the flag of outputting exit code */
-bool rv_enables_to_output_exit_code(riscv_t *rv);
+typedef struct {
+    char *elf_program;
+} vm_user_t;
+
+typedef union {
+    vm_user_t *vm_user;
+    /* TODO: system emulator stuff */
+} vm_emu_data_t;
```
jserv: It is quite confusing to have `vm` and `emu` here.

src/main.c
RinHizakura: Sure. We can leave these at other place.

src/riscv.c
```diff
@@ -105,24 +108,112 @@ riscv_word_t rv_get_reg(riscv_t *rv, uint32_t reg)
...
+        map_insert(attr->fd_map, &fd, &file);
+
+        /*
+         * store new fd to make the vm_attr_t consistent
+         */
+#if !defined(_WIN32) && !defined(_WIN64)
+        int new_fd = fileno(file);
+#else
+        int new_fd = _fileno(file);
+#endif
```
RinHizakura: How about adding another macro like `FILENO()` here for both, so we can make the codes here clean? Then you can define platform-specific `FILENO()`  at the place you defined things like `STDIN_FILENO`.

src/main.c
RinHizakura: Sure. We can leave these at other places.

src/riscv.c
ChinYikMing: Got it.

src/riscv_private.h
ChinYikMing: I think `data` is decent.

src/riscv.h
ChinYikMing: How about `vm_data_t`? The instance of it can simply called `data` in `vm_attr_t` struct. Hopefully, it will not confuse with the `data` in internal implementation. I think it is not because the latter is private.

src/riscv.c
ChinYikMing: Check this [commit](https://github.com/sysprog21/rv32emu/pull/340/commits/ea51b2d1300c3d2379c4274af1faa74423fece89#diff-f99b8d1f48ef2257a9bbbe6b487c621f74ef233c00858df5a37b06e89441e836) out.

src/main.c
ChinYikMing: > Sure. We can leave these at other places.

I have mentioned it in the commit message.

src/riscv.h
jserv: > How about `vm_data_t`? 

It looks good to me.

RinHizakura (approved):
I'm good with the changes now! Let's see if other guys have more comments.

Thank you!

jserv (changes requested):
In your git commit messages, it is more appropriate to mention "this commit" rather than "this PR," as "PR" refers to a pull request, not a commit.

visitorckw (approved):

visitorckw:
The changes also look good to me. Thanks!

PR #341, opened 2024-02-04
Introducing local register allocation for the tier-1 JIT compiler

qwe661234:
Local register allocation effectively reuses the host register value with in a basic block scope, thereby reducing the number of load and store instructions.

Take continuous addi instructions as an example:
```
addi t0, t0, 1
addi t0, t0, 1
addi t0, t0, 1
```
* The generated machine code without register allocation
```
load t0, t0_addr
add t0, 1
sw t0, t0_addr
load t0, t0_addr
add t0, 1
sw t0, t0_addr
load t0, t0_addr
add t0, 1
sw t0, t0_addr
```
* The generated machine code with register allocation
```
load t0, t0_addr
add t0, 1
add t0, 1
add t0, 1
sw t0, t0_addr
```
As shown in the above example, register allocation reuses the host register and reduces the number of load and store instructions.

* x86-64(i7-11700)

| Metric   | w/o RA  |  w/ RA  | SpeedUp |
|----------|----------|----------|---------|
| dhrystone| 0.342 s  | 0.328 s  |  +4.27% |
| miniz    | 1.243 s  | 1.185 s  |  +4.89% |
| primes   | 1.716 s  | 1.689 s  |  +1.60% |
| sha512   | 2.063 s  | 1.880 s  |  +9.73% |
| stream   | 11.619 s  |11.419 s  |  +1.75% |

* Aarch64 (eMag)

| Metric   | w/o RA  |  w/ RA  | SpeedUp |
|----------|----------|----------|---------|
| dhrystone| 1.935 s  | 1.301 s  | +48.73% |
| miniz    |  7.706 s  | 4.362 s  | +76.66% |
| primes   | 10.513 s  | 9.633 s  |  +9.14% |
| sha512   | 6.508 s  | 6.119 s  |  +6.36% |
| stream   | 45.174 s  | 38.037 s  | +18.76% |

As demonstrated in the performance analysis, the register allocation improves the overall performance for the T1C generated machine code. Without RA, the generated machine need to store back the register value in the end of intruction. With RA, we only need to store back theregister value in the end of basic block or when host registers are fully occupied. The performance enhancement is particularly pronounced on Aarch64 due to its increased availability of registers, providing a more extensive mapping capability for VM registers.


jserv:
Prior to our review of this pull request, please provide a detailed explanation of the register allocation (RA) algorithm you've implemented, including the rationale behind your choice. It would also be beneficial to discuss how this approach compares with traditional algorithms like graph coloring and linear scan, highlighting the advantages and considerations of each.

Referece:
* [linear scan register allocation](https://github.com/dotnet/runtime/blob/main/docs/design/coreclr/jit/lsra-heuristic-tuning.md)
* [Linear Scan Register Allocation: Design and Implementation Notes](https://github.com/dotnet/runtime/blob/main/docs/design/coreclr/jit/lsra-detail.md)
* [Trace-based Register Allocation in a JIT Compiler](https://ssw.jku.at/General/Staff/Eisl/papers/2016_PPPJ_TraceRA-preprint.pdf)

tools/gen-jit-template.py
```diff
@@ -131,7 +131,7 @@ def parse_argv(EXT_LIST, SKIP_LIST):
 f.close()
 
 fields = {"imm", "pc", "rs1", "rs2", "rd", "shamt", "branch_taken", "branch_untaken"}
-temp_regs = {"TMP0", "TMP1", "TMP2"}
+temp_regs = {"VM_REG0", "VM_REG1", "VM_REG2"}
```
jserv: These mnemonics are lengthy. Shorten them with `VR0`, `VR1`, and `VR2`.

qwe661234:
I implemented register allocation using an available host register table, a counter, and a VM register table.

* Available host register table: Records the available host registers for register allocation.
* Counter: Points to the available host register.
* VM register table: Records the mapping between host registers and VM registers.

Take the instruction sequence below as an example:
```
addi t0, a0, 1
addi t0, t0, 1

counter = 0
avalible_host_register = [RAX, RBX, RCX]
```

In the first `addi` instruction, we check the VM register table and find that `reg_a0` has not been mapped to a host register. Therefore, we map `reg_a0` to `RAX` and load the value from the RV data into it. Next, we need to map the destination VM register `t0`. We check the VM register table and find that `reg_t0` has not been mapped to a host register, so we map `reg_t0` to `RBX`, then move the value in `RAX` (`reg_a0`) to `RBX` (`reg_t0`) and add `RBX` (`reg_t0`) with 1.

In the second `addi` instruction, we check the VM register table and find that `reg_t0` has been mapped to the host register `RBX`, so there is no need to load `reg_t0`. Next, we need to map the destination VM register `t0`, and `reg_t0` has been mapped to the host register `RBX`. Therefore, we directly add `RBX` (`reg_t0`) with 1.

tools/gen-jit-template.py
```diff
@@ -131,7 +131,7 @@ def parse_argv(EXT_LIST, SKIP_LIST):
 f.close()
 
 fields = {"imm", "pc", "rs1", "rs2", "rd", "shamt", "branch_taken", "branch_untaken"}
-temp_regs = {"TMP0", "TMP1", "TMP2"}
+temp_regs = {"VR0", "VR1", "VR2"}
```
jserv: Rename `temp_regs` to `virt_regs` to emphasize the use of virtual registers.

tools/gen-jit-template.py
```diff
@@ -146,7 +146,9 @@ def parse_argv(EXT_LIST, SKIP_LIST):
                 if items[i] in fields:
                     items[i] = "ir->" + items[i]
                 if items[i] in temp_regs:
-                    items[i] = "temp_reg[" + items[i][-1] + "]"
+                    items[i] = "vm_reg[" + items[i][-1] + "]"
+                if items[i] == "TMP":
```
jserv, line 150: Replace `TMP` with `VIRT`. or `VR`.

tools/gen-jit-template.py
```diff
@@ -183,6 +185,14 @@ def parse_argv(EXT_LIST, SKIP_LIST):
                 else:
                     asm = "emit_load_sext(state, {}, {}, {}, {});".format(
                     items[1], items[2],  items[3],  items[4])
+            elif items[0] == "opt_ld":
```
jserv: Avoid using `opt` because someone might consider it as "optimization."

src/rv32_template.c
```diff
@@ -69,10 +73,18 @@
  * |                                | using the memory address calculated as |
  * |                                | the sum of the src and the specified   |
  * |                                | offset.                                |
+ * | opt_ld, dst, field             | Map VM register to host register, and  |
```
jserv: Apply the following refinement:
* Change from `opt_ld` to `ld`
* Change from `opt_ld_2` to `ld2`
* Change from `opt_ld_2_sext` `ld2s`

src/rv32_template.c
```diff
@@ -93,7 +105,11 @@
  * | mod, op, src, dst, imm;        | Do mod operation on src and dst and    |
  * |                                | store the result into dst.             |
  * | cond, src;                     | set condition if (src)                 |
- * | end;                           |  set the end of condition if (src)     |
+ * | end;                           | set the end of condition if (src)      |
+ * | st_back;                       | In the end of a basic block, we need   |
```
jserv: How about renaming from `st_back` to `break`?

src/rv32_template.c
qwe661234: However, the `ld` already exists; it serves another purpose. Maybe we can rename it to `ra_ld`, `ra_ld2` and `ra_ld2s`

jserv: Avoid using underscore (`_`) for mnemonics. Simply use something like `rald` or `vld`.

jserv (changes requested):
Rebase the latest `master` branch for API refinement.

src/jit.c
```diff
@@ -219,26 +205,24 @@ enum operand_size {
...
+static int temp_reg = RCX;
 #else
-#define RCX_ALT R9
 static const int nonvolatile_reg[] = {RBP, RBX, R13, R14, R15};
 static const int parameter_reg[] = {RDI, RSI, RDX, RCX, R8, R9};
-static const int temp_reg[] = {RAX, RBX, RCX};
-static const int register_map[] = {
-    RAX, RDI, RSI, RDX, R9, R8, RBX, R13, R14, R15, RBP,
-};
+static const int register_map[] = {RAX, RBX, RDX, R8,  R9,
```
jserv: Add a comma (`,`) right after `R15`, so that clang-format can indent better.

src/jit.c
```diff
@@ -249,21 +233,11 @@ static const int temp_reg[] = {R6, R7, R8};
  *   r25         Temp - used for modulous calculations
  */
 
-static const int register_map[] = {
-    R5,                      /* result */
-    R0,  R1,  R2,  R3,  R4,  /* parameters */
-    R19, R20, R21, R22, R23, /* callee-saved */
-};
+static const int register_map[] = {R5,  R6,  R7,  R9,  R11, R12, R13,
```
jserv: Ditto. Tweak for clang-format.

jserv:
> As shown in the above example, register allocation reuses the host register and reduces the number of load and store instructions.
> Metric	w/o RA	w/ RA	Speedup

Provide benchmarks for both x86-64 and Aarch64.

src/jit.c
```diff
@@ -1252,6 +1232,147 @@ static void prepare_translate(struct jit_state *state)
...
+    }
+}
+
+FORCE_INLINE void store_back_target(struct jit_state *state, int target_reg)
+{
+    for (int i = 0; i < 32; i++) {
+        if (reg_table[i] == target_reg) {
+            reg_table[i] = -1;
+            emit_store(state, S32, target_reg, parameter_reg[0],
+                       offsetof(riscv_t, X) + 4 * i);
```
vacantron, line 1268: Can we do an early return here? The target should be unique in the `reg_table`.

qwe661234: Yes, we should do early return here.

qwe661234:
> > As shown in the above example, register allocation reuses the host register and reduces the number of load and store instructions.
> > Metric	w/o RA	w/ RA	Speedup
> 
> Provide benchmarks for both x86-64 and Aarch64.

Added

jserv:
> > Provide benchmarks for both x86-64 and Aarch64.
>
> Added

Refine git commit message with correct column description -- `w/o` vs. `w/`.
Can you explain the measurements of RA for X64 and A64 code generation?

src/jit.c
```diff
@@ -1252,6 +1234,148 @@ static void prepare_translate(struct jit_state *state)
...
+        }
+    }
+}
+
+static int map_reg(struct jit_state *state, int reg_number)
+{
+    int target_reg = -1;
+    if (reg_table[reg_number] != -1)
+        return reg_table[reg_number];
+    else {
```
jserv: Remove this `else` statement since it is not needed.

qwe661234:
> > > Provide benchmarks for both x86-64 and Aarch64.
> > 
> > 
> > Added
> 
> Refine git commit message with correct column description -- `w/o` vs. `w/`. Can you explain the measurements of RA for X64 and A64 code generation?

Sure, I explained in the git message.

src/rv32_template.c
```diff
@@ -26,27 +26,31 @@
...
+ *       cond, regneq;
+ *       mov, VR0, VR1;
+ *       end;
+ *       alu32imm, 32, 0x81, 0, VR1, imm;
  *   }))
  *
- * TMP0, TMP1, TMP2 are host registers for storing calculated value during
- * execution. The block defined as 'GEN' is mapped to the generic C code used in
- * the interpreter. The following instructions will be generated by JIT
+ * VR0, VR1, VR2 are host registers for storing calculated value
```
jserv, line 37: You should explain the use of `TMP` as well.

src/jit.c
```diff
@@ -1252,6 +1234,146 @@ static void prepare_translate(struct jit_state *state)
 #endif
 }
 
+static int n_reg = ARRAYS_SIZE(register_map);
```
jserv: Add comments for these variables.

src/jit.c
```diff
@@ -1252,6 +1234,146 @@ static void prepare_translate(struct jit_state *state)
...
+            vm_reg[1] = register_map[count];
+        }
+        store_back_target(state, vm_reg[1]);
+        reg_table[reg_number2] = vm_reg[1];
+        emit_load(state, S32, parameter_reg[0], reg_table[reg_number2],
+                  offsetof(riscv_t, X) + 4 * reg_number2);
+    }
+}
+
+static void ra_load_2reg_sext(struct jit_state *state,
```
jserv: Simply name it `ra_load2_sext`.

jserv:
In the realm of compiler optimizations, register allocation strategies are broadly categorized into Local, Global, and Interprocedural register allocation. The proposed change specifically pertains to local register allocation. This approach focuses on tracking register contents within a basic block -- a linear sequence of instructions -- allowing for the efficient reuse of variables and constants directly from registers. This detail should be clearly communicated in the context of both pull request and git commit messages to ensure clarity regarding the scope and nature of the register allocation.

vacantron (approved):

PR #342, opened 2024-02-05
Fix a bug cache_create()

henrybear327:
The array index is wrongly written to `i` instead of `j`, which is the wrong logic.


jserv:

PR #348, opened 2024-02-07
Fix illegal instruction handling for SLLI, SRLI, and SRAI

visitorckw:
According to the RISC-V specification, specifically for RV32I, instructions like SLLI, SRLI, and SRAI should generate an illegal instruction exception if the imm[5] bit is not zero. However, the current implementation does not handle this error condition.

Addresse the issue by adding proper handling for illegal instruction exceptions in cases where imm[5] is non-zero for SLLI, SRLI, and SRAI instructions.


jserv:

PR #350, opened 2024-02-11
Enable shallow git submodule

howjmay:
Shallow update git submodule to save space and time.


jserv:
Why did you close this pull request?

howjmay:
I haven't properly checked whether shallow update may cause error when we change the submodule commit

PR #351, opened 2024-02-18
Drop memory_read_str function

ChinYikMing:
The memory_read_str function tends to terminate the string with '\0'
character but the string is only terminated with '\0' character if src is
less than max characters long according to "man 3 strncpy". As a result,
there is a chance to make the dst is not terminated with '\0' character,
e.g., the length of filename equals 256 characters long. It can simply
solved by passing "sizeof(name_str) - 1" to memory_read_str function
instead of "sizeof(name_str)" when calling memory_read_str function.
But, to deal with the filename >= 256 characters long, malloc the desired
length of filename during runtime is a better way.

### DIFF
--- a/src/syscall.c
+++ b/src/syscall.c
@@ -7,6 +7,7 @@
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <string.h>
 #include <sys/time.h>
 
 #include "riscv.h"
@@ -84,7 +85,7 @@ static void syscall_write(riscv_t *rv)
 {
     vm_attr_t *attr = PRIV(rv);
 
-    /* _write(fde, buffer, count) */
+    /* _write(fd, buffer, count) */
     riscv_word_t fd = rv_get_reg(rv, rv_reg_a0);
     riscv_word_t buffer = rv_get_reg(rv, rv_reg_a1);
     riscv_word_t count = rv_get_reg(rv, rv_reg_a2);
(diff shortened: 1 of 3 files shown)


visitorckw:
If the string length could be >= 256 causing dst is not terminated with '\0' character, would this modification result in opening an incorrect file when the length is >= 256?

ChinYikMing:
> If the string length could be >= 256 causing dst is not terminated with '\0' character, would this modification result in opening an incorrect file when the length is >= 256?

In this commit, `memory_read` reads max 255 long characters and the dst (capacity = 256) is confirmed to be terminated with '\0' character. If the length of filename >= 255, then filename will be truncated to 255 and `syscall_open` will failed.

visitorckw:
Should we consider failing the open operation for filenames with a length >= 256 characters? The idea of truncating the filename to 255 characters seems somewhat odd to me because it appears that converting an illegal filename to a legal one and attempting to open the file could potentially create a security issue.

qwe661234 (approved):

ChinYikMing:
> Should we consider failing the open operation for filenames with a length >= 256 characters? The idea of truncating the filename to 255 characters seems somewhat odd to me because it appears that converting an illegal filename to a legal one and attempting to open the file could potentially create a security issue.

Alternatively, `malloc` the required length of filename in runtime is another option.

jserv:
> Alternatively, `malloc` the required length of filename in runtime is another option.

It seems more suitable for that purpose.

src/syscall.c
```diff
@@ -327,13 +328,11 @@ static void syscall_open(riscv_t *rv)
...
     /* read name from runtime memory */
-    char name_str[256] = {'\0'};
-    uint32_t read = memory_read_str(attr->mem, (uint8_t *) name_str, name,
-                                    sizeof(name_str));
-    if (read > sizeof(name_str)) {
-        rv_set_reg(rv, rv_reg_a0, -1);
-        return;
-    }
+    const size_t name_len = strlen(attr->mem->mem_base + name);
+    char *name_str = malloc(name_len + 1);
```
jserv, line 332: Would the allocated memory be returned back later?

ChinYikMing: > Would the allocated memory be returned back later?

I should free it after `fopen`.

jserv:

PR #353, opened 2024-02-20
Fix pushing images automatically to DockerHub

henrybear327:
Addresses the problem where docker images aren't pushed to DockerHub automatically by GithubCI


henrybear327:
According to the [pricing](https://www.docker.com/pricing/) page, there isn't a limit on how many images we can push, but there is a rate-limiting mechanism on the image pull within a certain timeframe.

PR #355, opened 2024-02-23
Improve the performance of indirect jump for T1C

qwe661234:
Considering the T1C ends with the indirect jump, our interpreter mode now records the target of the indirect jump, identifying the most  frequent jump target for T1C. The T1 generated code for the indirect jump compares the selected jump target and executes the jump if the comparison results in equality. With this enhancement, T1C and T2C which are integrated later can proceed with an indirect jump if the target PC matches the selected jump target.

Based on the performance analysis, the benchmarks with significant number of indirect jump effectively improve the performance.

| Metric   | N_in_jmp  | Original | Propused | Speedup |
|----------|-----------|----------|----------|---------|
|miniz	   |    2098313|   1.266 s|   1.225 s|   +3.35%|
|sha512    |   10500727|   1.905 s|   1.861 s|   +2.36%|
|dhrystone |   20000618|   0.325 s|   0.253 s|  +28.46%|
|nqueens   |   44658722|   1.051 s|    0.79 s|  +33.04%|
|qsort	   |  275000250|   1.978 s|   1.517 s|  +30.39%|

### DIFF
--- a/src/jit.c
+++ b/src/jit.c
@@ -1336,6 +1336,26 @@ static void ra_load2_sext(struct jit_state *state,
     }
 }
 
+void parse_branch_history_table(struct jit_state *state, rv_insn_t *ir)
+{
+    int max_idx = 0;
+    branch_history_table_t *bt = ir->branch_table;
+    for (int i = 0; i < HISTORY_SIZE; i++) {
+        if (!bt->times[i])
+            break;
+        if (bt->times[max_idx] < bt->times[i])
+            max_idx = i;
+    }
+    if (bt->PC[max_idx]) {
+        emit_load_imm(state, register_map[0], bt->PC[max_idx]);
+        emit_cmp32(state, temp_reg, register_map[0]);
+        uint32_t jump_loc = state->offset;
+        emit_jcc_offset(state, 0x85);
+        emit_jmp(state, bt->PC[max_idx]);
+        emit_jump_target_offset(state, JUMP_LOC, state->offset);
+    }
+}
+
 #define GEN(inst, code)                                                       \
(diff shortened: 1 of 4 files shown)


jserv (changes requested):
Fix typo and improve the descriptions about the beneficial scenario in git commit messages.

src/rv32_template.c
```diff
@@ -250,8 +275,9 @@ RVOP(
         mov, VR1, TMP;
         alu32imm, 32, 0x81, 0, TMP, imm;
         alu32imm, 32, 0x81, 4, TMP, ~1U;
-        st, S32, TMP, PC;
         break;
+        bht;
```
jserv: `bht` might not be an intuitive name.

src/jit.c
```diff
@@ -1336,6 +1336,25 @@ static void ra_load2_sext(struct jit_state *state,
     }
 }
 
+void parse_branch_history_table(struct jit_state *state, rv_insn_t *ir)
+{
+    int max_idx = 0;
+    for (int i = 0; i < HISTORY_SIZE; i++) {
```
jserv, line 1343: Declare a variable which accesses `ir->branch_table`, so that you can write shorter code.

src/rv32_template.c
```diff
@@ -107,6 +107,11 @@
  * |                                | store the result into dst.             |
  * | cond, src;                     | set condition if (src)                 |
  * | end;                           | set the end of condition if (src)      |
+ * | parsebt;                       | parse the branch table of indirect     |
```
jserv: Avoid using "parse", which might confuse. The intention was related to branch prediction. How about something like `predict`?

qwe661234: Ok, I think predict is better.

Issue #310, opened 2023-12-25
Refactoring RISC-V emulation APIs for easier adoption and porting


ChinYikMing:
First trial of refactoring, the [wasm branch](https://github.com/ChinYikMing/rv32emu/tree/wasm)'s latest commit is the result. 

Since `state_t` is a user-provided data, so all runtime defined value(often change) shall be stored there. For instance, the emulated target program's argc and argv, and the emulator's parameter. The following have been adjusted to reflect the changes:
1. `state_t *state_new(void)` -----> `state_t *state_new(uint32_t mem_size, int argc, char **argv, bool allow_misalign, bool quiet_output)`
`mem_size` is used for `memory_new` because different runtimes may have memory requirements (for example, the page size in WebAssembly is 64KiB), the default `MEM_SIZE`(2^32 - 1) is not appropriate for that. The rest of parameters are runtime defined value.

2. `riscv_t *rv_create(const riscv_io_t *io, riscv_user_t userdata, int argc, char **args, bool output_exit_code)` -----> `riscv_t *rv_create(riscv_user_t userdata)`
Much cleaner function signature.

3. `void rv_reset(riscv_t *rv, riscv_word_t pc, int argc, char **args)` -----> `void rv_reset(riscv_t *rv, riscv_word_t pc)`
We can use `rv->userdata` to get the required `argc` and `argv`.

4. Since memory I/O handlers are rarely changed, it makes less sense to define them during runtime (makes porting difficult). Instead, I believe it is preferable to link them during build time. If really want to change the implementations, make a new C file and link it during build time might be a better choice.
To do this, some changes are made:
    - define memory I/O handlers in `rv_create` and link during build time
    - to make memory write interfaces match to compatible with the function pointers, `MEM_WRITE_IMPL` macro has to be changed:
    - [ ] `src/io.c`
```c
// old
#define MEM_WRITE_IMPL(size, type)                                 \
    void memory_write_##size(uint32_t addr, const uint8_t *src)    \
    {                                                              \
        *(type *) (data_memory_base + addr) = *(const type *) src; \
    }

// new
#define MEM_WRITE_IMPL(size, type)                                 \
    void memory_write_##size(uint32_t addr, const type src)    \
    {                                                              \
        *(type *) (data_memory_base + addr) = src; \
    }
```
- the calling of `memory_write_w` in "src/syscall.c" shall be changed accordingly:
- [ ] `src/syscall.c`
```c
// old
[...]

PR #357, opened 2024-02-25
Bind I/O handlers during emulator initialization

ChinYikMing:
It is not required to give an application the opportunity to bind I/O handlers because I/O handlers are rarely altered during the creation of an emulator.

With this commit, the application can now build a emulator much more easier by only taking the emulator's attribute (vm_attr_t) into consideration.

In order to facilitate further integration with the RISC-V system emulator (semu), I have included a TODO inside the I/O interface.

Related: #310


jserv:

Issue #349, opened 2024-02-07
Potential uninitialized value read without initialization 


henrybear327:
Running [infer](https://github.com/facebook/infer) on the codebase, we would get the following report

```c
src/elf.c:194: error: Uninitialized Value
`node.right_red` is read without initialization during the call to `map_insert()`.
  192.     /* initialize the symbol table */
  193.     map_clear(e->symbols);
  194.     map_insert(e->symbols, &(int){0}, &(char *){NULL});
              ^
  195.
  196.     /* get the string table */

Found 1 issue
    Issue Type(ISSUED_TYPE_ID): # 
    Uninitialized Value(PULSE_UNINITIALIZED_VALUE): 1
```

henrybear327:
Root cause is discussed in PR #362

jserv:
Close by commit af24e0ca06c17bda8014880e6d5d218dbc762656

PR #362, opened 2024-02-26
Fix reading uninitialized variable right_red in map.c (#349)

henrybear327:
The struct member right_red is used in several functions, such as  rb_node_set_right(). It has been reported by infer and LLVM static analyzer that right_red isn't initialized before being used.

By tracing the node initialization function calls (starting from `map_create_node`), it can be seen that `rb_node_set_right` is the function where `right_red` is attempted to be initialized, but we are indeed performing `&1` on an uninitialized value. 

In this commit, a change to using `calloc` guarantees the struct members will be zeroed out during allocation, which in terms serves as initialization.

### DIFF
--- a/src/map.c
+++ b/src/map.c
@@ -548,7 +548,7 @@ static map_node_t *map_create_node(void *key,
                                    size_t ksize,
                                    size_t vsize)
 {
-    map_node_t *node = malloc(sizeof(map_node_t));
+    map_node_t *node = calloc(1, sizeof(map_node_t));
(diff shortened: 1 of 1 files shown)


henrybear327:
> In this commit, a change to using calloc guarantees the struct members will be zeroed out during allocation, which in terms serves as initialization.

@jserv I am not very certain if this is a good idea to fix this problem. Maybe explicitly setting the initial values of the member variables in the node structure is a better idea?

PR #367, opened 2024-02-28
Fix RV32A error

visitorckw:
Fix a potential issue where A extension instructions could inadvertently write to the x0 register, ensuring that such instructions do not modify the x0 register. Additionally, address a requirement to handle RV32A instructions that may perform register writes during constant optimization. Given the possibility of these instructions modifying registers, it's crucial to incorporate appropriate handling for such cases within constant optimization routines.

### DIFF
--- a/src/rv32_constopt.c
+++ b/src/rv32_constopt.c
@@ -554,37 +554,67 @@ CONSTOPT(remu, {
 #if RV32_HAS(EXT_A)
 
 /* LR.W: Load Reserved */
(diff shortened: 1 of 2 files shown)


visitorckw:
The current compliance tests and prebuilt ELF files do not utilize RV32A instructions. Therefore, I would like to discuss whether we should consider introducing relevant tests.

visitorckw:
C.c. @fourcolor

jserv:

PR #368, opened 2024-02-28
Fix compile warning in JIT without RV32F/C

visitorckw:
Fix warnings in JIT compiler caused by unused functions 'do_cfsw', 'do_cflw', 'do_cfswsp', and 'do_cflwsp' when JIT compiler is enabled without F or C extensions.


jserv:

PR #369, opened 2024-03-01
Introduce Valgrind to the CI pipeline

henrybear327:
The static analyzer alone can't catch all the runtime issues. In this commit, the dynamic analysis tool Valgrind is added to the CI pipeline.

ENABLE_SDL is set to 0, in order to reduce noise caused by the external
libraries.

Reference:
- https://valgrind.org/docs/manual/quick-start.html
- https://linux.die.net/man/1/gcc

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,14 @@ CFLAGS = -std=gnu99 -O2 -Wall -Wextra
 CFLAGS += -Wno-unused-label
 CFLAGS += -include src/common.h
 
+ENABLE_VALGRIND ?= 0
+ifeq ("$(ENABLE_UBSAN)", "1")
+# according to gcc's man page: "If you use multiple -O options, with or without level numbers, the last such option is the one that is effective."
+# In order to use Valgrind, we need to compile with -g
+CFLAGS += -g
+LDFLAGS += -g
+endif
+
 # Enable link-time optimization (LTO)
 ENABLE_LTO ?= 1
 ifeq ($(call has, LTO), 1)
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -85,7 +85,7 @@ jobs:
     - name: install-dependencies
       run: |
             sudo apt-get update -q -y
-            sudo apt-get install -q -y libsdl2-dev libsdl2-mixer-dev
+            sudo apt-get install -q -y libsdl2-dev libsdl2-mixer-dev valgrind
             .ci/riscv-toolchain-install.sh
             wget https://apt.llvm.org/llvm.sh
             sudo chmod +x ./llvm.sh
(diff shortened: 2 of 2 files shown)


henrybear327:
To merge this pull request, all the memory-related leaks need to be addressed first. 

Currently, based on my initial testing, when disabling SDL and JIT, no memory leaks are found. When enabling JIT, there will be plenty of leaks.

Makefile
```diff
@@ -11,6 +11,14 @@ CFLAGS = -std=gnu99 -O2 -Wall -Wextra
 CFLAGS += -Wno-unused-label
 CFLAGS += -include src/common.h
 
+ENABLE_VALGRIND ?= 0
+ifeq ("$(ENABLE_UBSAN)", "1")
```
visitorckw, line 15: It looks like a typo to me. Did you mean 'ENABLE_VALGRIND' instead of 'ENABLE_UBSAN'?

Makefile
```diff
@@ -11,6 +11,14 @@ CFLAGS = -std=gnu99 -O2 -Wall -Wextra
...
+# according to gcc's man page: "If you use multiple -O options, with or without level numbers, the last such option is the one that is effective."
+# In order to use Valgrind, we need to compile with -g
+CFLAGS += -g
+LDFLAGS += -g
```
visitorckw, line 19: The '-g' flag is not specifically designed for valgrind. Have you considered using a different name such as ENABLE_DEBUG instead of ENABLE_VALGRIND?

jserv:
Memory leaks of JIT compiler reported by Valgrind:
- definitely lost: 144 bytes in 4 blocks
- indirectly lost: 132 bytes in 9 blocks
- still reachable: 64 bytes in 3 blocks

henrybear327:
> Memory leaks of JIT compiler reported by Valgrind:
> 
> * definitely lost: 144 bytes in 4 blocks
> * indirectly lost: 132 bytes in 9 blocks
> * still reachable: 64 bytes in 3 blocks

Fix in progress!

jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: 53dabb26346a98be32f79c3c34f9d36802f01a6a | Previous: b80dcb4ea83b26e0de49fe947c30ecc3e73d3939 | Ratio |
|-|-|-|-|
| `Dhrystone` | `3.44` Average DMIPS over 10 runs | `12.37` Average DMIPS over 10 runs | `3.60` |
| `Coremark` | `0.003` Average iterations/sec over 10 runs | `0.005` Average iterations/sec over 10 runs | `1.67` |

</details>

Makefile
henrybear327: Using `ENABLE_UBSAN` will enable `-g`, so dropping `ENABLE_VALGRIND` for now!

henrybear327:
> Memory leaks of JIT compiler reported by Valgrind:
> 
> * definitely lost: 144 bytes in 4 blocks
> * indirectly lost: 132 bytes in 9 blocks
> * still reachable: 64 bytes in 3 blocks

The initial investigation is that the memory leak is detected only when JIT is enabled.

Currently, there are 2 identified leaks so far
- `ir->fuse`: it allocates memory outside the memory pool. When JIT is enabled, the clean-up logic is missing.
- releasing allocated memory after the memory arena expansion

jserv:
> The initial investigation is that the memory leak is detected only when JIT is enabled.
> Currently, there are 2 identified leaks so far
> * `ir->fuse`: it allocates memory outside the memory pool. When JIT is enabled, the clean-up logic is missing.
> * releasing allocated memory after the memory arena expansion

@qwe661234, can you comment this?

henrybear327:
> > The initial investigation is that the memory leak is detected only when JIT is enabled.
> > Currently, there are 2 identified leaks so far
> > 
> > * `ir->fuse`: it allocates memory outside the memory pool. When JIT is enabled, the clean-up logic is missing.
> > * releasing allocated memory after the memory arena expansion
> 
> @qwe661234, can you comment this?

Reference: https://github.com/sysprog21/rv32emu/blob/ef151e2bb4666716f86d8b1ffac349a164b7cf85/src/riscv.c#L61

qwe661234:
> > > The initial investigation is that the memory leak is detected only when JIT is enabled.
> > > Currently, there are 2 identified leaks so far
> > > 
> > > * `ir->fuse`: it allocates memory outside the memory pool. When JIT is enabled, the clean-up logic is missing.
> > > * releasing allocated memory after the memory arena expansion
> > 
> > 
> > @qwe661234, can you comment this?
> 
> Reference:
> 
> https://github.com/sysprog21/rv32emu/blob/ef151e2bb4666716f86d8b1ffac349a164b7cf85/src/riscv.c#L61

This function is only using in non-JIT mode, the basic block information is stored in block cache in JIT mode. Therefore, I need to add a free for `ir->fuse` when clear block cache.

jserv (changes requested):

jserv:
The build system is being reworked (#473), and I am about to close this pull request. Let's introduce Valgrind to CI pipeline later.

PR #371, opened 2024-03-04
End the indirect jump when potential hotspot is detected

qwe661234:
Currently, the indirect jump does not switch to T1C when potential hotspot is detected, it would make execution stuck in interpreter mode. This modification make execution end when potential hotspot is detected.

### DIFF
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -399,12 +399,14 @@ static bool has_loops = false;
     nextop:                                                           \
         PC += __rv_insn_##inst##_len;                                 \
         if (unlikely(RVOP_NO_NEXT(ir))) {                             \
-            rv->csr_cycle = cycle;                                    \
-            rv->PC = PC;                                              \
-            return true;                                              \
+            goto end_op;                                              \
         }                                                             \
         const rv_insn_t *next = ir->next;                             \
         MUST_TAIL return next->impl(rv, next, cycle, PC);             \
+    end_op:                                                           \
+        rv->csr_cycle = cycle;                                        \
+        rv->PC = PC;                                                  \
+        return true;                                                  \
     }
 
 #include "rv32_template.c"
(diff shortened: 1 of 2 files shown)


jserv:
The following code snip appears several times in `src/emulate.c` and `src/rv32_template.c`, and can you unify by jumping the label in `RVOP`?
```c
        rv->csr_cycle = cycle;
        rv->PC = PC;
        return true;
```

qwe661234:
> The following code snip appears several times in `src/emulate.c` and `src/rv32_template.c`, and can you unify by jumping the label in `RVOP`?
> 
> ```c
>         rv->csr_cycle = cycle;
>         rv->PC = PC;
>         return true;
> ```

However, since this label is within the scope of a function, I can replace them with a MACRO.

jserv:
> However, since this label is within the scope of a function, I can replace them with a MACRO.

That sounds good. The label will be named `end_op`. The term "op" (operation) is preferred over "insn" (instruction) to clearly differentiate between RISC-V instructions and the IR operations derived from them. Our interpreter specifically manipulates these IR operations, rather than the direct RISC-V instructions.

src/rv32_template.c
```diff
@@ -150,6 +150,11 @@ RVOP(
         ldimm, VR0, pc, imm;
     }))
 
+#define END_OP             \
```
jserv: This macro should appear in `src/emulate.c`, renamed to `RVOP_END`.

src/emulate.c
```diff
@@ -389,6 +389,13 @@ static set_t pc_set;
 static bool has_loops = false;
 #endif
 
+
+#define RVOP_END           \
```
jserv: Modify macro `RVOP` to facilitate `RVOP_END`.

jserv (changes requested):
Rephrase "This modification make execution end when potential hotspot is detected" to
> This revision ensures the execution halts upon identifying a potential hotspot. It aims to make the execution process more efficient by stopping when a significant area of code usage is detected, allowing for targeted optimization.

PR #372, opened 2024-03-04
Refine cleanup logic in main.c

ChinYikMing:


jserv (changes requested):
Make both git commit message and the subject informative.

src/main.c
```diff
@@ -224,6 +224,10 @@ int main(int argc, char **args)
     rv_delete(rv);
 
     printf("inferior exit code %d\n", attr.exit_code);
+
+end:
     free(attr.data.user);
-    return 0;
+    if (prof_out_file)
```
jserv: Use `free(prof_out_file)` without `if` statement.

ChinYikMing: > Use `free(prof_out_file)` without `if` statement.

Sure. As man page of `free` says: "If ptr is a NULL pointer, no operation is performed.".

PR #376, opened 2024-03-09
Add assertion to prevent buffer overflow

vacantron:
The index `n_insn` or `n_jumps` might exceed the boundary of the allocated space. Check the index first before writing any value into the space.


jserv:

PR #377, opened 2024-03-10
Tweak build rules for WebAssembly

### LINKED ISSUE #375: Build using emcc failed
After this [commit](https://github.com/sysprog21/rv32emu/commit/f484a4b92cd7b26779f44613b973fe1be83511aa), `emcc` fails to build the source. On my machine, the command in [mk/toolchain.mk](https://github.com/sysprog21/rv32emu/blob/master/mk/toolchain.mk): `emcc --version | head -n 1` will output `emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.51 (c0c2ca1314672a25699846b4663701bcb6f69cca)`. The output contains the keywords `clang` and `emcc`, thus both `CC_IS_CLANG` and `CC_IS_EMCC` variables will be set. Consequently, the `CFLAGS` from `clang` and `emcc` will be mixed. This causes the build to fail.

Note that supposed the build machine has installed both `emcc` and `clang`

ChinYikMing:
The "emcc --version" command displays information about the version of emcc, which contains the "gcc" or "clang" string. As a result, the build rule should only match one compiler and stop checking anymore.  If not, the build could be failed by a mix of different compiler's CFLAGS. For example, -flto from clang and -flto=thin from emcc.

After refactoring the riscv.[ch] public APIs in commit 820cd9b, it is now possible to adjust the memory size to accommodate varying runtime requirements. The memory size must align to a 4KB page size for the WASM runtime and I set 1GB as default memory size for WASM runtime.

The LTO is not supported to build emscripten-port SDL library, so disable it when both ENABLE_LTO=1 and ENABLE_SDL=1 are set.

Related to: #75
Close #375

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ CONFIG_FILE := $(OUT)/.config
 CFLAGS = -std=gnu99 -O2 -Wall -Wextra
 CFLAGS += -Wno-unused-label
 CFLAGS += -include src/common.h
+CFLAGS_emcc ?=
 
 # Enable link-time optimization (LTO)
 ENABLE_LTO ?= 1
@@ -21,7 +22,14 @@ endif
 endif
 $(call set-feature, LTO)
 ifeq ($(call has, LTO), 1)
-ifeq ("$(CC_IS_GCC)$(CC_IS_EMCC)", "1")
+ifeq ("$(CC_IS_EMCC)", "1")
+ifeq ($(call has, SDL), 1)
+$(warning LTO is not supported to build emscripten-port SDL using emcc.)
+else
+CFLAGS += -flto
+endif
+endif
+ifeq ("$(CC_IS_GCC)", "1")
 CFLAGS += -flto
 endif
 ifeq ("$(CC_IS_CLANG)", "1")
@@ -136,6 +144,13 @@ ifeq ("$(CC_IS_EMCC)", "1")
 CFLAGS += -mtail-call
 endif
 
+# Build emscripten-port SDL
+ifeq ("$(CC_IS_EMCC)", "1")
+ifeq ($(call has, SDL), 1)
+CFLAGS_emcc += -sUSE_SDL=2 -sSDL2_MIXER_FORMATS=wav,mid -sUSE_SDL_MIXER=2
+endif
+endif
+
 ENABLE_UBSAN ?= 0
 ifeq ("$(ENABLE_UBSAN)", "1")
 CFLAGS += -fsanitize=undefined -fno-sanitize=alignment -fno-sanitize-recover=all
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -167,13 +190,25 @@ OBJS := \
 OBJS := $(addprefix $(OUT)/, $(OBJS))
 deps := $(OBJS:%.o=%.o.d)
 
+EXPORTED_FUNCS := _main
+EMCC_CFLAGS :=
+EMCC_MACRO :=
+ifeq ("$(CC_IS_EMCC)", "1")
+EMCC_MACRO += -DMEM_SIZE=0x40000000    # 1GB
```
jserv: The name `EMCC_MACRO` is misleading. Drop it and move to `CFLAGS_emcc`.

jserv: Instead of "1GB," write "1 GiB" in [binary prefix](https://en.wikipedia.org/wiki/Binary_prefix) form.

Makefile
```diff
@@ -167,13 +190,25 @@ OBJS := \
...
+EMCC_CFLAGS += -sINITIAL_MEMORY=2GB \
```
jserv: Rename to `CFLAGS_emcc` to emphasize the use of emcc-specific C flags.

Makefile
```diff
@@ -167,13 +190,25 @@ OBJS := \
...
+	       -sALLOW_MEMORY_GROWTH \
+	       --embed-file build \
+	       -s"EXPORTED_FUNCTIONS=$(EXPORTED_FUNCS)" \
+	       -sMALLOC=mimalloc
```
jserv: Since you are using mimalloc, you have to check the versions of emcc which bundle the corresponding packages.

Makefile
```diff
@@ -148,6 +164,13 @@ $(OUT)/emulate.o: CFLAGS += -foptimize-sibling-calls -fomit-frame-pointer -fno-s
 # to the first target after .DEFAULT_GOAL is not set.
 .DEFAULT_GOAL :=
 
+EMCC_OBJS += $(BIN).js \
```
jserv: It is confusing to name them object files.

ChinYikMing: > It is confusing to name them object files.

What about `EMCC_OUT`, these files are generated by `emcc`.

jserv: Use `WEB_FILES` instead.

Makefile
```diff
@@ -136,6 +143,15 @@ ifeq ("$(CC_IS_EMCC)", "1")
 CFLAGS += -mtail-call
 endif
 
+# Build emscripten-port SDL
+ifeq ("$(CC_IS_EMCC)", "1")
+ifeq ($(call has, SDL), 1)
+EMCC_PORT_FLAGS := -sUSE_SDL=2 -sSDL2_MIXER_FORMATS=wav,mid -sUSE_SDL_MIXER=2
```
jserv: Simply name it `CFLAGS_emcc_port`.

Makefile
```diff
@@ -167,13 +190,47 @@ OBJS := \
...
+CFLAGS_emcc :=
+ifeq ("$(CC_IS_EMCC)", "1")
+CFLAGS_emcc += -sINITIAL_MEMORY=2GB \
+	       -sALLOW_MEMORY_GROWTH \
+	       -s"EXPORTED_FUNCTIONS=$(EXPORTED_FUNCS)" \
+	       --embed-file build \
+	       -DMEM_SIZE=0x40000000 \
+	       -w
+
+# see commit 165c1a3 of emscripten
```
jserv: Move the emcc specific rules to `mk/toolchain.mk`.

Makefile
ChinYikMing: > Simply name it `CFLAGS_emcc_port`.

No more need  `CFLAGS_emcc_port` since we can simply concatenate flags to `CFLAGS_emcc`.

jserv:

PR #378, opened 2024-03-11
Avoid generating machine code repeatly for same basic block

qwe661234:
Originally, every execution path had its own machine code. Therefore, even if some basic blocks had been traced in other execution paths, we still generated a new copy of machine code. To use the code cache space more efficiently, we modified the recorded table to be global, allowing every execution path to share the machine code. This modification ensures that every basic block has only one copy of machine code, thus saving code cache space.

For example, the code cache stored 1,926,471 (1.9 MB) bytes of machine code for AES previously. After the modification, the code cache stored 182,730 (0.18 MB) bytes of machine code.

### DIFF
--- a/src/jit.c
+++ b/src/jit.c
@@ -56,7 +56,8 @@
 #define JIT_OP_MOD_REG (JIT_CLS_ALU | JIT_SRC_REG | 0x90)
 
 #define STACK_SIZE 512
-#define MAX_INSNS 1024
+#define MAX_JUMPS 1024
+#define MAX_BLOCKS 8192
 #define IN_JUMP_THRESHOLD 256
 #if defined(__x86_64__)
 #define JUMP_LOC jump_loc + 2
@@ -243,8 +244,8 @@ static inline void emit_load_imm(struct jit_state *state, int dst, int64_t imm);
 
 static inline void offset_map_insert(struct jit_state *state, int32_t target_pc)
 {
-    struct offset_map *map_entry = &state->offset_map[state->n_insn++];
-    assert(state->n_insn < MAX_INSNS);
+    struct offset_map *map_entry = &state->offset_map[state->n_blocks++];
+    assert(state->n_blocks < MAX_BLOCKS);
     map_entry->pc = target_pc;
     map_entry->offset = state->offset;
 }
@@ -354,7 +355,7 @@ static inline void emit_jump_target_address(struct jit_state *state,
                                             int32_t target_pc)
 {
     struct jump *jump = &state->jumps[state->n_jumps++];
-    assert(state->n_jumps < MAX_INSNS);
+    assert(state->n_jumps < MAX_JUMPS);
     jump->offset_loc = state->offset;
     jump->target_pc = target_pc;
     emit4(state, 0);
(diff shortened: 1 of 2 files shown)


src/jit.c
```diff
@@ -1564,26 +1567,31 @@ static void translate_chained_block(struct jit_state *state,
...
-    memset(state->offset_map, 0, MAX_INSNS * sizeof(struct offset_map));
-    memset(state->jumps, 0, MAX_INSNS * sizeof(struct jump));
-    state->n_insn = 0;
+    if (set_has(&set, block->pc_start)) {
+        for (int i = 0; i < state->n_insn; i++) {
+            if (block->pc_start == state->offset_map[i].pc) {
+                return state->offset_map[i].offset;
+            }
+        }
+        assert(NULL);
```
jserv: Can it be replaced with `__UNREACHABLE`?

src/jit.c
```diff
@@ -56,7 +56,8 @@
 #define JIT_OP_MOD_REG (JIT_CLS_ALU | JIT_SRC_REG | 0x90)
 
 #define STACK_SIZE 512
-#define MAX_INSNS 1024
+#define MAX_JUMPS 1024
```
jserv, line 59: Can you show empirical adjustments for both `MAX_JUMPS` and `MAX_INSNS`. I wonder how they can be calculated.

src/jit.c
```diff
@@ -1531,29 +1532,30 @@ static void resolve_jumps(struct jit_state *state)
     }
 }
 
+set_t set;
```
jserv: Can it be declared as a local variable?

src/jit.c
```diff
@@ -1349,7 +1350,7 @@ void parse_branch_history_table(struct jit_state *state, rv_insn_t *ir)
         if (bt->times[max_idx] < bt->times[i])
             max_idx = i;
     }
-    if (bt->PC[max_idx]) {
+    if (bt->PC[max_idx] && bt->times[max_idx] >= 256) {
```
jserv: Why 256 here?

jserv (changes requested):
Drop the trailing dot in the subject of git commit messages.
See https://cbea.ms/git-commit/

src/jit.c
qwe661234: As mentioned in the first commit of this PR, I set 256 as the threshold because we recorded 16 indirect jumps, and
the threshold for launching T1C is 4096. 256 == 4096 / 16. I think the indirect jump with the highest frequency should exceed this threshold.

src/jit.c
qwe661234: The `MAX_JUMPS` means the number of host branch instructions within a chained block collection need to be resolved.

qwe661234: I modified `MAX_INSNS` into `MAX_BLOCKS` because it refers to the total number of basic blocks within our code cache. Considering that our block cache size is 4096 and some blocks may be replaced, I set `MAX_BLOCKS` to 8192.

src/jit.c
jserv: Avoid hardcoded value. Use certain identifier instead.

src/jit.c
```diff
@@ -1564,26 +1564,31 @@ static void translate_chained_block(struct jit_state *state,
             if (bt->times[max_idx] < bt->times[i])
                 max_idx = i;
         }
-        if (bt->PC[max_idx] && !set_has(set, bt->PC[max_idx])) {
+        if (bt->PC[max_idx] && bt->times[max_idx] >= 256 &&
```
jserv: Ditto. Avoid hardcoded value.

PR #379, opened 2024-03-11
Introduce Infer static analyzer to the CI pipeline

henrybear327:
The [Infer Docker Image](https://hub.docker.com/repository/docker/sysprog21/infer/general) in use is hosted on Docker Hub, which was [built and pushed to sysprog21 by us](https://henrybear327.github.io/notes/posts/container/infer/) since [Infer's latest upstream release](https://github.com/facebook/infer/releases/tag/v1.1.0) is very outdated at the time of writing.

The docker image size is around 470 MB, after numerous attempts to reduce its size from 3GB.

This PR also fixes issues reported by Infer. (The value written to `&target_reg` is never used.)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -127,12 +127,23 @@ jobs:
     - name: set up scan-build  
       run: |
             sudo apt-get update -q -y
-            sudo apt-get install -q -y clang clang-tools libsdl2-dev libsdl2-mixer-dev
+            sudo apt-get install -q -y clang clang-tools
       shell: bash
     - name: run scan-build without JIT
       run: make clean && make distclean && scan-build -v -o ~/scan-build --status-bugs --use-cc=clang --force-analyze-debug-code --show-description -analyzer-config stable-report-filename=true -enable-checker valist,nullability make ENABLE_EXT_F=0 ENABLE_SDL=0 ENABLE_JIT=0
     - name: run scan-build with JIT
       run: make clean && make distclean && scan-build -v -o ~/scan-build --status-bugs --use-cc=clang --force-analyze-debug-code --show-description -analyzer-config stable-report-filename=true -enable-checker valist,nullability make ENABLE_EXT_F=0 ENABLE_SDL=0 ENABLE_JIT=1
+  
+  static-analysis-infer:
+    needs: [detect-code-related-file-changes]
+    if: needs.detect-code-related-file-changes.outputs.has_code_related_changes == 'true'
+    runs-on: sysprog21/infer:latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: run infer without JIT
+      run: infer --fail-on-issue -- make ENABLE_SDL=0 ENABLE_LTO=0 ENABLE_EXT_F=0 ENABLE_JIT=0
+    - name: run infer with JIT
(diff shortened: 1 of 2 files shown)


jserv (changes requested):
Skip Arm64 based host during static analysis.

henrybear327:
> Skip Arm64 based host during static analysis.

I am unaware that the Arm64-based host will be run during the static analysis phase, as I didn't specify `arch: aarch64`. Or maybe I am mistaken?

jserv:
> The docker image size is around 470 MB, after numerous attempts to reduce its size from 3GB.

Can you rebuild Infer static analysis tool with `-Os` set? And then, strip the ELF files accordingly.

henrybear327:
> > The docker image size is around 470 MB, after numerous attempts to reduce its size from 3GB.
> 
> Can you rebuild Infer static analysis tool with `-Os` set? And then, strip the ELF files accordingly.

I can give this a try, but I have no guarantee on this.

jserv:
If Clang scan-build already makes sense to CI, do we need Infer? Can you show the considerations?

jserv:
The [build-infer.sh](https://github.com/facebook/infer/blob/main/build-infer.sh) invokes `autogen.sh`, which generates GNU Automake style scripts. At the moment, only C and Python are required for static analysis.

Reference: [The Growth of the OCaml Distribution](https://ocamlpro.com/blog/2023_01_02_ocaml_distribution/)

henrybear327:
> The [build-infer.sh](https://github.com/facebook/infer/blob/main/build-infer.sh) invokes `autogen.sh`, which generates GNU Automake style scripts. At the moment, only C and Python are required for static analysis.
> 
> Reference: [The Growth of the OCaml Distribution](https://ocamlpro.com/blog/2023_01_02_ocaml_distribution/)

The previous Docker Image contained only C and Python analysis capability, but now with the release build, Java is also included. 

Based on prior experience, Java can be excluded by passing in parameters, but some changes to `create_binary_release.sh` will be required as I am now using [create_binary_release.sh](https://github.com/facebook/infer/blob/main/scripts/create_binary_release.sh) to build a standalone binary instead of invoking `build-infer.sh` directly like before.

.github/workflows/main.yml
```diff
@@ -127,12 +127,26 @@ jobs:
...
             sudo apt-get update -q -y
-            sudo apt-get install -q -y clang clang-tools libsdl2-dev libsdl2-mixer-dev
+            sudo apt-get install -q -y clang clang-tools
       shell: bash
     - name: run scan-build without JIT
       run: make clean && make distclean && scan-build -v -o ~/scan-build --status-bugs --use-cc=clang --force-analyze-debug-code --show-description -analyzer-config stable-report-filename=true -enable-checker valist,nullability make ENABLE_EXT_F=0 ENABLE_SDL=0 ENABLE_JIT=0
     - name: run scan-build with JIT
       run: make clean && make distclean && scan-build -v -o ~/scan-build --status-bugs --use-cc=clang --force-analyze-debug-code --show-description -analyzer-config stable-report-filename=true -enable-checker valist,nullability make ENABLE_EXT_F=0 ENABLE_SDL=0 ENABLE_JIT=1
+  
+  static-analysis-infer:
```
jserv, line 137: Don't create a new entry. Use the existing one instead.

henrybear327:
The infer image pipeline broke recently (node11 now has strict internet bandwidth restrictions), and the effort to maintain an image is a bit cumbersome. Thus, closing this PR for now.

PR #380, opened 2024-03-12
Refine rv_step function prototype

ChinYikMing:
After refinement of riscv.[ch] public APIs in commit 820cd9b, cycle_per_step is specified in vm_attr_t, the argument cycles in the function prototype of rv_step is no longer required.


jserv:

PR #383, opened 2024-03-13
Tweak build rules to link correct SDL library

### LINKED ISSUE #381: Failed to build using emcc when ENABLE_SDL=1 is set outside macOS
Error:
```
In file included from src/syscall_sdl.c:16:
In file included from /usr/include/SDL2/SDL.h:32:
In file included from /usr/include/SDL2/SDL_main.h:25:
In file included from /usr/include/SDL2/SDL_stdinc.h:31:
/usr/include/SDL2/SDL_config.h:4:10: fatal error: 'SDL2/_real_SDL_config.h' file not found
    4 | #include <SDL2/_real_SDL_config.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

The linker links system SDL library instead of Emscripten-port SDL library in commit 446b95b without properly checking which compiler is used.

ChinYikMing:
Close #381

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,6 +84,7 @@ $(call set-feature, Zifencei)
 
 # Experimental SDL oriented system calls
 ENABLE_SDL ?= 1
+ifneq ("$(CC_IS_EMCC)", "1") # note that emcc generates port SDL headers/library, so it does not requires system SDL headers/library
 ifeq ($(call has, SDL), 1)
 ifeq (, $(shell which sdl2-config))
 $(warning No sdl2-config in $$PATH. Check SDL2 installation in advance)
@@ -100,6 +101,7 @@ OBJS_EXT += syscall_sdl.o
 $(OUT)/syscall_sdl.o: CFLAGS += $(shell sdl2-config --cflags)
 LDFLAGS += $(shell sdl2-config --libs) -pthread
 LDFLAGS += $(shell pkg-config --libs SDL2_mixer)
+endif
 endif
 
 ENABLE_GDBSTUB ?= 0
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -84,6 +84,7 @@ $(call set-feature, Zifencei)
 
 # Experimental SDL oriented system calls
 ENABLE_SDL ?= 1
+ifneq ("$(CC_IS_EMCC)", "1")
```
jserv: Why does the detection of SDL headers/library shipped with Emscripten rely on sdl2-config?

ChinYikMing: > Why does the detection of SDL headers/library shipped with Emscripten rely on sdl2-config?

No, `ifneq` here means we need to rely on `sdl2-config` when using compiler other than `emcc`.

jserv: Got it. Add some comments in `Makefile` to address the relationship between system SDL headers and the emscripten counterpart.

ChinYikMing:
One more thing, I see there are new preferred syntax to link emscripten-port library, see [building project](https://emscripten.org/docs/compiling/Building-Projects.html?highlight=sdl2) and search keyword: `--use-port`. There is a note states that:
> Since emscripten 3.1.54, `--use-port` is the preferred syntax to use a port in your project. The legacy syntax (for example `-sUSE_SDL2, -sUSE_SDL_IMAGE=2`) remains available.

Shall we stick with the new syntax? I prefer old syntax since it can compatible with older version of `emcc`

jserv:
> Shall we stick with the new syntax? I prefer old syntax since it can compatible with older version of `emcc`

Up to you. Keep in mind with the compatibility.

jserv:

Issue #330, opened 2024-01-20
jit: Implement register allocation for T1C


jserv:
Register allocation is the process of allocating local automatic variables and expression results to a finite set of processor registers. This process involves the compiler tracking register occupancy for each slot using abstract values, allowing it often to skip over code for local accesses and primarily update the abstract state. This allocation can be executed over a basic block (local register allocation), across an entire function or procedure (global register allocation), or even beyond function boundaries in the call-graph (interprocedural register allocation).

When register allocation is applied to an entire function or procedure, it may necessitate the insertion of save/restore instructions around each call site, as per the calling convention. For the tier-1 compiler (T1C), implementing register allocation could significantly enhance code generation, reducing the time taken before the tier-2 JIT compiler (T2C) takes over. This optimized allocation strategy ensures more efficient use of processor registers, which is crucial for faster and more effective code execution.

jserv:
#341 implements local register allocation, and it is still feasible to introduce more advanced approach for register allocation.

jserv:
> #341 implements local register allocation, and it is still feasible to introduce more advanced approach for register allocation.

The basic register allocation feature has already been incorporated into the current tier-1 JIT compiler. However, compared to QEMU, the code generated by this compiler, even with register allocation, is not as efficient. There is an interest in reevaluating the register allocation process to identify opportunities for further optimizations.  @vacantron would look for the chance to enhance register allocation efficiency without modifying the existing linear IR structure.

PR #384, opened 2024-03-13
Eliminate redundant register swapping in JIT

vacantron:
In the current register allocation of JIT, we save the host registers before using to keep the consistency of the content between the vm and the host register, even it had not been changed by any other operation after loaded from the stack. This generates lots of redundant `store` related instructions on the host and reduces the performance of vm.

In this patch, we introduce the `dirty` flag to record the status of the host register. Once the content of the host register is changed by an operation, the flag is set to 1. The register swapping will only be invoked when the register is dirty in next allocation.

The performance improvement after this patch is shown below:

* x86-64

| Metric    | Original | Improved | SpeedUp |
| --------- | -------- | -------- | ------- |
| dhrystone |  0.623 s |  0.463 s |  +25.6% |
| miniz     |  3.437 s |  2.837 s |  +17.5% |
| primes    |  4.435 s |  3.205 s |  +27.7% |
| sha512    |  3.212 s |  3.024 s |   +5.9% |
| aes       |  1.708 s |  1.545 s |   +9.5% |
| qsort     |  5.236 s |  5.219 s |   +0.3% |

* aarch64

| Metric    | Original | Improved | SpeedUp |
| --------- | -------- | -------- | ------- |
| dhrystone |  1.003 s |  0.897 s |  +10.6% |
| miniz     |  4.312 s |  4.107 s |   +4.8% |
| primes    |  9.605 s |  8.879 s |   +7.6% |
| sha512    |  5.902 s |  5.796 s |   +1.8% |
| aes       |  2.932 s |  2.891 s |   +1.4% |
| qsort     |  6.123 s |  5.988 s |   +2.2% |

### DIFF
--- a/src/jit.c
+++ b/src/jit.c
@@ -207,15 +207,17 @@ enum operand_size {
 #if defined(_WIN32)
 static const int nonvolatile_reg[] = {RBP, RBX, RDI, RSI, R13, R14, R15};
 static const int parameter_reg[] = {RCX, RDX, R8, R9};
-static const int register_map[] = {
-    RAX, R10, RDX, R8, R9, R14, R15, RDI, RSI, RBX, RBP,
+static struct host_reg register_map[] = {
+    {RAX, 0}, {R10, 0}, {RDX, 0}, {R8, 0},  {R9, 0},  {R14, 0},
+    {R15, 0}, {RDI, 0}, {RSI, 0}, {RBX, 0}, {RBP, 0},
 };
 static int temp_reg = RCX;
 #else
 static const int nonvolatile_reg[] = {RBP, RBX, R13, R14, R15};
 static const int parameter_reg[] = {RDI, RSI, RDX, RCX, R8, R9};
-static const int register_map[] = {
-    RAX, RBX, RDX, R8, R9, R10, R11, R13, R14, R15,
+static struct host_reg register_map[] = {
+    {RAX, 0}, {RBX, 0}, {RDX, 0}, {R8, 0},  {R9, 0},
+    {R10, 0}, {R11, 0}, {R13, 0}, {R14, 0}, {R15, 0},
 };
 static int temp_reg = RCX;
 #endif
@@ -236,11 +238,26 @@ static int temp_reg = R8;
  *   r24       Temp - used for generating 32-bit immediates
  *   r25       Temp - used for modulous calculations
  */
-static const int register_map[] = {
-    R5, R6, R7, R9, R11, R12, R13, R14, R15, R16, R17, R18, R26,
+static struct host_reg register_map[] = {
+    {R5, 0},  {R6, 0},  {R7, 0},  {R9, 0},  {R11, 0}, {R12, 0}, {R13, 0},
+    {R14, 0}, {R15, 0}, {R16, 0}, {R17, 0}, {R18, 0}, {R26, 0},
 };
-static inline void emit_load_imm(struct jit_state *state, int dst, int64_t imm);
 #endif
+
+static const int n_host_regs =
+    ARRAYS_SIZE(register_map); /* the number of avavliable host register */
+
+static inline void set_dirty(int reg_idx, bool is_dirty)
+{
+    for (int i = 0; i < n_host_regs; i++) {
+        /* ignore nonvolatile and parameter registers */
+        if (register_map[i].reg_idx != reg_idx)
+            continue;
+
+        register_map[i].dirty = is_dirty;
+        return;
+    }
+}
 
 static inline void offset_map_insert(struct jit_state *state, int32_t target_pc)
 {
@@ -361,6 +378,8 @@ static inline void emit_jump_target_address(struct jit_state *state,
     emit4(state, 0);
 }
 #elif defined(__aarch64__)
+static inline void emit_load_imm(struct jit_state *state, int dst, int64_t imm);
+
 static void emit_a64(struct jit_state *state, uint32_t insn)
 {
     assert(insn != BAD_OPCODE);
@@ -389,6 +408,7 @@ static void emit_addsub_imm(struct jit_state *state,
     const uint32_t imm_op_base = 0x11000000;
     emit_a64(state, sz(is64) | (op << 29) | imm_op_base | (0 << 22) |
                         (imm12 << 10) | (rn << 5) | rd);
+    set_dirty(rd, true);
 }
 
 /* [ARM-A]: C4.1.67: Logical (shifted register).  */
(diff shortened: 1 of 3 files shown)


src/jit.c
```diff
@@ -926,6 +973,9 @@ static inline void emit_store(struct jit_state *state,
         break;
     }
 #endif
+
+    if (offset != 0)
```
ChinYikMing: Small advice: simply `if(offset)` would be cleaner

src/jit.c
```diff
@@ -236,12 +238,27 @@ static int temp_reg = R8;
...
-static const int register_map[] = {
-    R5, R6, R7, R9, R11, R12, R13, R14, R15, R16, R17, R18, R26,
+static struct host_reg register_map[] = {
+    {R5, 0},  {R6, 0},  {R7, 0},  {R9, 0},  {R11, 0}, {R12, 0}, {R13, 0},
+    {R14, 0}, {R15, 0}, {R16, 0}, {R17, 0}, {R18, 0}, {R26, 0},
 };
-static inline void emit_load_imm(struct jit_state *state, int dst, int64_t imm);
 #endif
 
+static const int n_reg =
```
jserv: Use `n_regs` for consistency. However, it is a bit confusing for the use of host and virtual registers. Can you clarify?

src/jit.c
```diff
@@ -837,6 +873,8 @@ static inline void emit_load(struct jit_state *state,
         break;
     }
 #endif
+
+    set_dirty(dst, offset ? false : true);
```
jserv: Simply rewrite as `set_dirty(dst, !offset);

src/jit.h
```diff
@@ -36,6 +36,11 @@ struct jit_state {
     int n_jumps;
 };
 
+struct host_reg {
+    int reg_name; /* enum x64_reg/a64_reg */
```
jserv: How about using `uint8_t` for `reg_name`?

jserv: This variable is indeed an index rather than a name for registers.

jserv: By shifting the value left by 1, we can incorporate both the register index value and its 'dirty' status within a short integer. Consequently, checking the 'dirty' status becomes a straightforward bitwise AND operation with `0x1`.

src/jit.c
```diff
@@ -942,6 +992,9 @@ static inline void emit_jmp(struct jit_state *state, uint32_t target_pc)
 #endif
 }
 
+static inline bool save_reg(struct jit_state *, int);
```
qwe661234: I found that we do not need the return value, so it can be `void`.

src/jit.c
```diff
@@ -942,6 +992,9 @@ static inline void emit_jmp(struct jit_state *state, uint32_t target_pc)
...
+static inline bool unmap_vm_reg(int);
```
qwe661234: Ditto, we do not need the return value.

src/jit.c
```diff
@@ -1203,141 +1277,186 @@ static void prepare_translate(struct jit_state *state)
...
+    target_reg2 = map_vm_reg_reserved(state, vm_reg_idx2, target_reg1);
+
+    assert(target_reg1 != target_reg2);
+    return (((uint32_t) target_reg2) << 8) | ((uint32_t) target_reg1);
+}
+
+static void ra_load2(struct jit_state *state, int vm_reg_idx1, int vm_reg_idx2)
+{
+    int origin1 = reg_table[vm_reg_idx1], origin2 = reg_table[vm_reg_idx2];
+    uint32_t regs = map_vm_reg_2(state, vm_reg_idx1, vm_reg_idx2);
```
qwe661234: I think `map_vm_reg_2` is less efficient because it needs some logic calculaitons to retrive register numbers, maybe this can be modified it to 
```
if (vm_reg_idx1 == vm_reg_idx2) {
   vm_reg[0] = vm_reg[1] = map_vm_reg(state, vm_reg_idx1);
} else {
    vm_reg[0] = map_vm_reg(state, vm_reg_idx1);
    vm_reg[1] = map_vm_reg_reserved(state, vm_reg_idx2, target_reg1);
}

```
and remove `map_vm_reg_2`.

src/jit.h
vacantron: > This variable is indeed an index rather than a name for registers.

In fact, it is the index to the host's register file, but not the index of the `register_map` array. i.e. the third register (index: 2, zero-based) of x64 is `RCX`, but the one in `register_map` is `RDX`. We want to retrieve the `dirty` flag through the register name `RDX` instead of its index to `register_map`.

vacantron: > By shifting the value left by 1, we can incorporate both the register index value and its 'dirty' status within a short integer. Consequently, checking the 'dirty' status becomes a straightforward bitwise AND operation with `0x1`.

Can we define a new `struct` here in C? Use `int8_t` might be too confusing.

jserv: > Can we define a new `struct` here in C? Use `int8_t` might be too confusing.

Yes, use it with bit field.

qwe661234 (approved):

src/jit.c
```diff
@@ -207,15 +207,17 @@ enum operand_size {
 #if defined(_WIN32)
 static const int nonvolatile_reg[] = {RBP, RBX, RDI, RSI, R13, R14, R15};
 static const int parameter_reg[] = {RCX, RDX, R8, R9};
-static const int register_map[] = {
-    RAX, R10, RDX, R8, R9, R14, R15, RDI, RSI, RBX, RBP,
+static struct host_reg register_map[] = {
+    {RAX, 0}, {R10, 0}, {RDX, 0}, {R8, 0},  {R9, 0},  {R14, 0},
+    {R15, 0}, {RDI, 0}, {RSI, 0}, {RBX, 0}, {RBP, 0},
```
vacantron, line 212: Should we do anything to check the casting from `enum x64_reg` (implementation-defined width, usually unsigned integer) to `uint8_t` ?

jserv: No, by means of bit fields, compilers can help you on type checking.

src/jit.h
```diff
@@ -36,6 +36,11 @@ struct jit_state {
     int n_jumps;
 };
 
+struct host_reg {
+    uint8_t reg_name : 5; /* enum x64_reg/a64_reg */
```
jserv: The term `reg_name` is misleading.

jserv (changes requested):
Append "Close #330" at the end of git commit message because the work is considered as completed after RA is refined.

jserv:

PR #385, opened 2024-03-16
Enable run video games using WebAssembly

ChinYikMing:
To play sound effects and musics, we would use software synthesizer, here I choose Timidity. When doom and quake target are specified, the Timidity related data are downloaded and extracted. As previous dicussion in the issue forum, the web workers are not reaped after return, thus adding pthread_join to reap them. As I tested, the sound effects and musics are only playable using the emcc version 3.1.51 which will fetch specific version of SDL2_mixer library, thus warning the user if the emcc version is not 3.1.51.

We want to minimize the wasm size, thus only embed necessary files when building the wasm. For example, if doom target is specified, only doom related files are embeded to wasm. This could shorten the fetching time of wasm from the server.

Since we are running the wasm program inside web browser, we have to yield to the web browser in some interval, so emscripten_set_main_loop_arg provided by emscripten comes to help. To utilize this function, we have to update the rv_step function signature.

web-resources directory will be used to store some required static web files.

Note that this commit is not done yet, just want for quick code review before further anything wrong that not in consensus.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -193,17 +193,31 @@ OBJS := \
 OBJS := $(addprefix $(OUT)/, $(OBJS))
 deps := $(OBJS:%.o=%.o.d)
 
-EXPORTED_FUNCS := _main
+include mk/external.mk
+
+deps_emcc :=
+ASSETS := assets
+WEB_JS_RESOURCES := $(ASSETS)/js
+EXPORTED_FUNCS := _main,_indirect_rv_halt
 ifeq ("$(CC_IS_EMCC)", "1")
 CFLAGS_emcc += -sINITIAL_MEMORY=2GB \
 	       -sALLOW_MEMORY_GROWTH \
 	       -s"EXPORTED_FUNCTIONS=$(EXPORTED_FUNCS)" \
-	       --embed-file build \
+	       -sSTACK_SIZE=4MB \
+	       -sPTHREAD_POOL_SIZE=navigator.hardwareConcurrency \
+	       --embed-file build@/ \
+	       --embed-file build/timidity@/etc/timidity \
 	       -DMEM_SIZE=0x40000000 \
+	       -DCYCLE_PER_STEP=2000000 \
+	       --pre-js $(WEB_JS_RESOURCES)/pre.js \
+	       -O3 \
 	       -w
+
+# used to download all dependencies of elf executable and bundle into single wasm
+deps_emcc += $(DOOM_DATA) $(QUAKE_DATA) $(TIMIDITY_DATA)
 endif
 
-$(OUT)/%.o: src/%.c
+$(OUT)/%.o: src/%.c $(deps_emcc)
 	$(VECHO) "  CC\t$@\n"
 	$(Q)$(CC) -o $@ $(CFLAGS) $(CFLAGS_emcc) -c -MMD -MF $@.d $<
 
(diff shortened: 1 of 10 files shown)


jserv (changes requested):
Rename the `web-resources` directory to `assets` to allow for the inclusion of a wider range of relevant resource files.

web-resources/js/pre.js
```diff
@@ -0,0 +1,5 @@
+Module['noInitialRun'] = false;
+
+Module['arguments'] = []
+// FIXME
+//Module['arguments'].push("quake.elf");
```
jserv: Don't leave the dead code.

Makefile
```diff
@@ -263,18 +271,50 @@ include mk/external.mk
 
 # Non-trivial demonstration programs
 ifeq ($(call has, SDL), 1)
-doom: $(BIN) $(DOOM_DATA)
-	(cd $(OUT); ../$(BIN) doom.elf)
+WEB_RESOURCES := web-resources
+WEB_JS_RESOURCES := $(WEB_RESOURCES)/js
+doom_action := (cd $(OUT); ../$(BIN) doom.elf)
+ifeq ("$(CC_IS_EMCC)", "1")
```
jserv, line 279: The file `mk/external.mk` was meant to be related to retrieve something from external sources. Don't specify emscripten rules here.

ChinYikMing: > The file `mk/external.mk` was meant to be related to retrieve something from external sources. Don't specify emscripten rules here.

I want to distinguish between the steps that need to be taken when utilizing emcc. Furthermore, when emcc is not being utilized, we do not want the emcc CFLAGS to be used. Also, notice that only just emcc needs the TIMIDITY_DATA.

jserv: Minimize the necessary changes when possible.

ChinYikMing: I would like to embed all build executable and configuration files such as doomrc into wasm. Although the size getting larger (fetching time might be longer but only once because it will be cached by the browser for further fetch), but we can select and run them in a single web page from a menu bar. 

Consequently, we do not need to check the makefile build target and embed corresponding files respectively.

jserv: > I would like to embed all build executable and configuration files such as doomrc into wasm.

It sounds like a good idea.

ChinYikMing:
> Rename the `web-resources` directory to `assets` to allow for the inclusion of a wider range of relevant resource files.

Done.

Makefile
ChinYikMing: > I would like to embed all build executable and configuration files such as doomrc into wasm. Although the size getting larger (fetching time might be longer but only once because it will be cached by the browser for further fetch), but we can select and run them in a single web page from a menu bar.
> 
> Consequently, we do not need to check the makefile build target and embed corresponding files respectively.

The final wasm is 61.40 MB,  this is larger than GitHub's recommended maximum file size of 50.00 MB. I am going to test it out if this is fine to host on GitHub Pages.

ChinYikMing:
See the result in [here](https://chinyikming.github.io/rv32emu/).

Makefile
ChinYikMing: > > I would like to embed all build executable and configuration files such as doomrc into wasm. Although the size getting larger (fetching time might be longer but only once because it will be cached by the browser for further fetch), but we can select and run them in a single web page from a menu bar.
> > Consequently, we do not need to check the makefile build target and embed corresponding files respectively.
> 
> The final wasm is 61.40 MB, this is larger than GitHub's recommended maximum file size of 50.00 MB. I am going to test it out if this is fine to host on GitHub Pages.

Github Pages are fine even exceeds GitHub's recommended maximum file size of 50.00 MB. See [here](https://chinyikming.github.io/rv32emu/).

jserv:

PR #391, opened 2024-03-29
Fix hardcoded sha1sum command

ChinYikMing:


ChinYikMing:
Note that this commit has been reset. The new and similar PR is #393.

PR #395, opened 2024-03-29
Fix missing dirty flag in `cmv` instruction

vacantron:
In the `cmv` instruction, we ignore the data movement of the same destination register, and the `set_dirty()` function is skipped. We need to explicitly set the flag in the `src/rv_template.c` to make sure the context of the register file is correct.

After the changes, the `do_cmv()` becomes:
```c
vm_reg[0] = ra_load(state, ir->rs2);
vm_reg[1] = map_vm_reg(state, ir->rd);

if (vm_reg[0] != vm_reg[1]) {
    emit_mov(state, vm_reg[0], vm_reg[1]);
} else {
    set_dirty(vm_reg[1], true);
}
```

### DIFF
--- a/src/rv32_template.c
+++ b/src/rv32_template.c
@@ -2337,6 +2337,8 @@ RVOP(
         map, VR1, rd;
         cond, regneq;
         mov, VR0, VR1;
+        else;
+        pollute, VR1;
         end;
     }))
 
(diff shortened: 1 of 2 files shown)


jserv:
@qwe661234, can you confirm recursive programs such as factorial work with the proposed change?

qwe661234 (approved):

Issue #400, opened 2024-04-05
CI: WebAssembly action failed when PR merged


ChinYikMing:
See [action](https://github.com/sysprog21/rv32emu/actions/runs/8566505944)

ChinYikMing:
To access the secrets variable (or token) in actions when merging PR, it is necessary to use `pull_request_target` instead of `pull_request`.

PR #401, opened 2024-04-05
CI: Specify token when checking out demo

ChinYikMing:
Similar [issue](https://github.com/ad-m/github-push-action/issues/170) is discussed. I am not very sure we can solve it by this method but let's give it a try.

See issue: #400


jserv (changes requested):
Improve the git commit message, addressing more references.

ChinYikMing:
> Improve the git commit message, addressing more references.

Done.

ChinYikMing:
This PR does not work for issue #400. #400 is solved by PR #403.

PR #413, opened 2024-04-08
Support hosting WebAssembly demo page locally

ChinYikMing:
To enhance the user experience of using WebAssembly version of rv32emu, we shall provide a Makefile target (serve-wasm) to serve the WebAssembly demo page in an easy way.

Since rv32emu leverages TCO (tail call optimization) when being compiled, so we have to tell if the user's browser supports TCO. Currently, only Chrome with MAJOR 112 and Firefox with MAJOR 121 supports TCO execution runtime, thus we focus on this two browsers. Also, use color text to remind the users if they need to update their Chrome or Firefox.

See also: #75


mk/color.mk
```diff
@@ -0,0 +1,3 @@
+YELLOW=\033[0;33m
```
jserv: It is not worthy to create a new file containing only 3 lines. Move these to `mk/common.mk`.

jserv (changes requested):
Move WebAssembly related rules to `mk/wasm.mk`.

ChinYikMing:
Run `make CC=emcc start-web` to test it.

Note that you should install emcc then source your emcc firstly. In macOS or Linux, use `source ~/emsdk/emsdk_env.sh`.

ChinYikMing:
> Move WebAssembly related rules to `mk/wasm.mk`.

Done.

ChinYikMing:
README will be updated on future PR.

mk/wasm.mk
```diff
@@ -0,0 +1,104 @@
...
+define cp-web-file
+    $(Q)cp $(1) $(DEMO_DIR)
+    $(info)
+endef
+
+# WEB_FILES could be cleaned and recompiled, thus do not mix these two files into WEB_FILES
+STATIC_WEB_FILES := $(WEB_HTML_RESOURCES)/index.html \
+		    $(WEB_JS_RESOURCES)/coi-serviceworker.min.js
+
+serve-wasm: $(BIN) check-demo-dir-exist
```
jserv: Rename the target to `start-web`.

jserv (changes requested):
Update `README.md` to reflect the proposed changes.

ChinYikMing:
> Update `README.md` to reflect the proposed changes.

Done.

README.md
```diff
@@ -281,6 +281,33 @@ To analyze the profiling data, use the `rv_profiler` tool with the desired optio
...
+At this point, you can build and start a web server service to serve WebAssembly by running:
+```shell
+$ make CC=emcc start-web
+```
+You would see the server's IP:PORT in your terminal. Copy and paste it to the browsers and
+you just access the index page of `rv32emu`.
+
+### Index page
+You would see a dropdown menu which you can use to select the ELF executable. Select one and
+click the Run button to run it.
```
jserv, line 309: Mention the demo page as the alternative solution.

README.md
```diff
@@ -281,6 +281,33 @@ To analyze the profiling data, use the `rv_profiler` tool with the desired optio
 $ tools/rv_profiler [--start-address|--stop-address|--graph-ir] [test_program]
 ```
 
+## WebAssembly Translation
+### Build and run
+`rv32emu` relies on Emscripten to be compiled to WebAssembly. Thus, the target system should have
```
jserv: Put the hyperlink to Emscripten in the first sentence.

jserv:

PR #415, opened 2024-04-18
Add benchmarks from rv8-bench and nbench

qwe661234:
These benchmarks are compiled from `rv8-bench` and `nbench` using the compiler `riscv32-unknown-elf-gcc`, which incorporates RV32IMC extensions, and the optimization level is set to O2.


jserv (changes requested):
Update `README.md` for the corresponding information.

PR #417, opened 2024-04-22
Add code cache flush mechanism

### LINKED ISSUE #365: jit: Flush the code cache when it is full
If the code cache fills up during the generation of T1C machine code, an assertion error will occur. Therefore, it is necessary to implement a mechanism to flush the code cache.

See: #347

qwe661234:
This modification introduces a code cache flush mechanism to handle the situation when the code cache becomes full. We flush the code cache by setting the offset to its initial value and clearing all the recording data for the code cache. Subsequently, we restart the T1C translation for the target chained-block.

Close: #365

### DIFF
--- a/src/cache.c
+++ b/src/cache.c
@@ -287,4 +287,16 @@ void cache_profile(const struct cache *cache,
         }
     }
 }
+
+void clear_cache_hot(const struct cache *cache, clear_func_t func)
+{
+    assert(cache);
+    assert(func);
+    for (int i = 0; i < THRESHOLD; i++) {
+        lfu_entry_t *entry, *safe;
+        list_for_each_entry_safe (entry, safe, cache->lists[i], list) {
+            func(entry->value);
+        }
+    }
+}
 #endif
--- a/src/jit.c
+++ b/src/jit.c
@@ -272,11 +272,11 @@ static inline void offset_map_insert(struct jit_state *state, int32_t target_pc)
     __builtin___clear_cache((char *) (addr), (char *) (addr) + (size));
 #endif
 
+static bool should_flush = false;
 static void emit_bytes(struct jit_state *state, void *data, uint32_t len)
 {
-    assert(state->offset <= state->size - len);
     if (unlikely((state->offset + len) > state->size)) {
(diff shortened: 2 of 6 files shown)


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: 143a903047472cf85cdf94b40355bef18efacfc7 | Previous: 0d258e6d7d8259d86366811ac4d4cc3293c405d3 | Ratio |
|-|-|-|-|
| `Dhrystone` | `1750.62` Average DMIPS over 10 runs | `1744.6` Average DMIPS over 10 runs | `1.00` |
| `Coremark` | `1502.731` Average iterations/sec over 10 runs | `1496.516` Average iterations/sec over 10 runs | `1.00` |

</details>

src/jit.c
```diff
@@ -272,11 +272,11 @@ static inline void offset_map_insert(struct jit_state *state, int32_t target_pc)
     __builtin___clear_cache((char *) (addr), (char *) (addr) + (size));
 #endif
 
+static bool flush = false;
```
jserv: Rename to `should_flush` or something alike.

src/cache.c
```diff
@@ -287,4 +287,14 @@ void cache_profile(const struct cache *cache,
         }
     }
 }
+
+void clear_cache_hot(const struct cache *cache, clear_func_t func)
+{
```
jserv, line 292: Add assertion on `cache` and `func`.

src/jit.c
```diff
@@ -1684,28 +1703,29 @@ static void translate_chained_block(struct jit_state *state,
...
             }
         }
         __UNREACHABLE;
     }
+restart:
     memset(state->jumps, 0, 1024 * sizeof(struct jump));
     state->n_jumps = 0;
-    uint32_t entry_loc = state->offset;
+    block->offset = state->offset;
     translate_chained_block(&(*state), rv, block);
```
jserv: Can you avoid using `&(*state)`?

jserv (changes requested):
Mention the preliminary explanation about the behavior of cache flushing.

PR #418, opened 2024-04-25
jit: Implement register allocation

vacantron:
Implement the register allocator with linear-scan register allocation. The `reg_table` is removed and integrated to the `host_reg` for convenience of accessing the vm register by the `reg_idx`.

The performance comparsion is shown below:

* x86-64

| Metric    | Original       | w/ reg-alloc   | SpeedUp |
| --------- | --------------- | -------------- | ------- |
| dhrystone |  14060 DMIPS   |  15225 DMIPS   | +8.29%  |
| coremark  |   3874 iters/s |   3828 iters/s | -1.19%  |
| aes       | 0.0297 s       | 0.0323 s       | -8.75%  |
| nqueens   |   2.16 s       |   2.05 s       | +5.09%  |
| stream    |   29.6 s       |   27.6 s       | +6.76%  |


src/reg-alloc.c
```diff
@@ -0,0 +1,242 @@
+static int liveness[N_RV_REGS];
```
jserv: Add copyright notice at the beginning of this file.

src/reg-alloc.c
```diff
@@ -0,0 +1,242 @@
...
+
+/* The priority queue of vm registers. The one which has farthest liveness is
+ * first.
+ */
+static uint8_t candidate_queue[N_RV_REGS];
+
+static inline void reset_liveness()
+{
+    for (int i = 0; i < N_RV_REGS; i++) {
```
jserv: Rewrite with `memset`, which is optimized by C runtime. That is, `memset(liveness, 0xFF, sizeof(liveness))`.

src/reg-alloc.c
```diff
@@ -0,0 +1,242 @@
...
+}
+
+static inline void init_candidate_queue()
+{
+    for (int i = 0; i < N_RV_REGS; i++) {
+        candidate_queue[i] = i;
+    }
+}
+
+int compare_liveness(const void *l, const void *r)
```
jserv: Declare it as a static function.

jserv (changes requested):
Enforce consistent naming scheme, in the form `item_action`. E.g., prefer `liveness_calc` instead of `calc_liveness`.

vacantron:
> Enforce consistent naming scheme, in the form `item_action`. E.g., prefer `liveness_calc` instead of `calc_liveness`.

Is it better to use `liveness_reset`, `reg_pick` instead of `reset_liveness`, `pick_reg` also?

jserv:
> Is it better to use `liveness_reset`, `reg_pick` instead of `reset_liveness`, `pick_reg` also?

Yes, to avoid unintended symbol conflicts, let's stick to the requested naming scheme across compilation units.

src/jit.c
```diff
@@ -1278,92 +1277,84 @@ static void prepare_translate(struct jit_state *state)
...
-        reg_table[i] = -1;
-        return;
-    }
+    /* check dirty before unmap */
+    assert(idx > -1 && idx < n_host_regs);
+    register_map[idx].vm_reg_idx = -1;
 }
 
-static inline void set_reg_table(int vm_reg_idx, int reg_idx)
+#include "reg-alloc.c"
```
jserv: Since the file `reg-alloc.c` is used only in the tier 1 JIT compiler and is relatively shorter, it would make sense to move the implementation directly into file `jit.c`.

src/reg-alloc.c
```diff
@@ -0,0 +1,242 @@
...
+            liveness[ir->rs1] = idx;
+            break;
+        case rv_insn_fuse3:
+            for (int i = 0; i < ir->imm2; i++) {
+                liveness[ir->fuse[i].rs1] = idx;
+                liveness[ir->fuse[i].rs2] = idx;
+            }
+            break;
+        case rv_insn_fuse4:
+        case rv_insn_fuse5:
```
qwe661234: We don't need to set liveness for `fuse5` or `fuse6`, so we should write a `break`.

src/jit.c
```diff
@@ -1310,6 +1315,242 @@ static void store_back(struct jit_state *state)
     }
 }
 
+
```
jserv: Remove this blank line.

src/jit.h
```diff
@@ -38,8 +38,10 @@ struct jit_state {
 };
 
 struct host_reg {
-    uint8_t reg_idx : 5; /* index to the host's register file */
+    uint8_t reg_idx : 5;   /* index to the host's register file */
+    int8_t vm_reg_idx : 6; /* index to the vm register */
     bool dirty : 1;
+    bool alive : 1;
```
jserv: Add comment to address its usage in register allocation.

jserv:

PR #423, opened 2024-05-04
Fix compilation error caused by missing softfloat dependency

visitorckw:
When adding the -j flag to make for faster compilation speeds, it led to the following error:

src/riscv.h:20:10: fatal error: softfloat/softfloat.h: No such file or directory
   20 | #include "softfloat/softfloat.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.

This was due to incorrect dependency settings, where compilation of the source code started before completing the git submodule update, leading to the error. Add appropriate dependency handling to ensure that compilation of the softfloat library is completed before compiling the source code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -175,6 +175,10 @@ OBJS := \
 OBJS := $(addprefix $(OUT)/, $(OBJS))
 deps := $(OBJS:%.o=%.o.d)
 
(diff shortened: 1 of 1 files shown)


jserv:

PR #424, opened 2024-05-04
CI: Parallelize compilation for faster testing 

visitorckw:
Utilize -j$(nproc) flag to accelerate building time in CI. Additionally, prevent duplicate execution of architectural tests to ensure each test suite runs only once. These optimizations in CI reduced testing times on my GitHub Action:

| Test Suite       | Before | After |
|------------------|---------------------|--------------------|
| host-x86 tests   | 2m3s                | 1m26s              |
| host-arm64 tests | 11m0s               | 8m34s              |
| compliance-test  | 10m4s               | 6m53s              |

Note: Depends on #423 to pass the tests.

### DIFF
--- a/.ci/riscv-tests.sh
+++ b/.ci/riscv-tests.sh
@@ -10,11 +10,4 @@ set -x
 export PATH=`pwd`/toolchain/riscv/bin:$PATH
 
 make clean
-make arch-test RISCV_DEVICE=I || exit 1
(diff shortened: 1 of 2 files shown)


jserv (changes requested):
Rebase and refine the git commit message. Use "parallelize" instead of "optimize."

jserv:

PR #425, opened 2024-05-04
Implement tier-2 JIT compiler to accelerate ISS

qwe661234:
The T2C, designed for flexibility, efficiently transforms custom IR within a chained block into LLVM IR. After the translation, the built LLVM IR is offloaded to the LLVM backend,  where it undergoes optimization through several selected LLVM passes. Subsequently, the optimized LLVM IR is passed to the LLVM execution engine, which compiles the optimized LLVM IR and returns a function pointer to the generated machine code. Because our design for T2C opts for a function pointer of the generated machine code instead of storing it in the code cache, we need not account for the binary format discrepancies among different compilers. This process allows for the possibility of substituting LLVM with alternative optimization frameworks if required.

Figure below shows the performance analysis of QEMU and rv32emu tiered-JIT mode.
image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -44,6 +44,9 @@ jobs:
             sudo apt-get update -q -y
             sudo apt-get install -q -y libsdl2-dev libsdl2-mixer-dev
             .ci/riscv-toolchain-install.sh
+            wget https://apt.llvm.org/llvm.sh
+            sudo chmod +x ./llvm.sh
+            sudo ./llvm.sh 17
       shell: bash
     - name: default build
       run: make
@@ -65,14 +68,14 @@ jobs:
             make distclean ENABLE_GDBSTUB=1 gdbstub-test
     - name: JIT test
       run: |
-            make clean && make ENABLE_JIT=1 check -j$(nproc)
-            make clean && make ENABLE_EXT_A=0 ENABLE_JIT=1 check -j$(nproc)
-            make clean && make ENABLE_EXT_F=0 ENABLE_JIT=1 check -j$(nproc)
-            make clean && make ENABLE_EXT_C=0 ENABLE_JIT=1 check -j$(nproc)
+            make ENABLE_JIT=1 clean && make ENABLE_JIT=1 check -j$(nproc)
+            make ENABLE_JIT=1 clean && make ENABLE_EXT_A=0 ENABLE_JIT=1 check -j$(nproc)
+            make ENABLE_JIT=1 clean && make ENABLE_EXT_F=0 ENABLE_JIT=1 check -j$(nproc)
+            make ENABLE_JIT=1 clean && make ENABLE_EXT_C=0 ENABLE_JIT=1 check -j$(nproc)
     - name: undefined behavior test
       run: |
             make clean && make ENABLE_UBSAN=1 check -j$(nproc)
-            make clean && make ENABLE_JIT=1 ENABLE_UBSAN=1 check -j$(nproc)
+            make ENABLE_JIT=1 clean clean && make ENABLE_JIT=1 ENABLE_UBSAN=1 check -j$(nproc)
 
   host-arm64:
     needs: [detect-code-related-file-changes]
@@ -91,18 +94,21 @@ jobs:
         # No 'sudo' is available
         install: |
           apt-get update -q -y
-          apt-get install -q -y git build-essential libsdl2-dev libsdl2-mixer-dev
+          apt-get install -q -y git build-essential libsdl2-dev libsdl2-mixer-dev lsb-release wget software-properties-common gnupg
           git config --global --add safe.directory ${{ github.workspace }}
           git config --global --add safe.directory ${{ github.workspace }}/src/softfloat
           git config --global --add safe.directory ${{ github.workspace }}/src/mini-gdbstub
+          wget https://apt.llvm.org/llvm.sh
+          chmod +x ./llvm.sh
+          ./llvm.sh 17
         # Append custom commands here
         run: |
           make -j$(nproc)
           make check -j$(nproc)
-          make clean && make ENABLE_JIT=1 check -j$(nproc)
-          make clean && make ENABLE_EXT_A=0 ENABLE_JIT=1 check -j$(nproc)
-          make clean && make ENABLE_EXT_F=0 ENABLE_JIT=1 check -j$(nproc)
-          make clean && make ENABLE_EXT_C=0 ENABLE_JIT=1 check -j$(nproc)
+          make ENABLE_JIT=1 clean && make ENABLE_JIT=1 check -j$(nproc)
+          make ENABLE_JIT=1 clean && make ENABLE_EXT_A=0 ENABLE_JIT=1 check -j$(nproc)
+          make ENABLE_JIT=1 clean && make ENABLE_EXT_F=0 ENABLE_JIT=1 check -j$(nproc)
+          make ENABLE_JIT=1 clean && make ENABLE_EXT_C=0 ENABLE_JIT=1 check -j$(nproc)
 
   coding-style:
     needs: [detect-code-related-file-changes]
@@ -132,7 +138,8 @@ jobs:
     - name: run scan-build without JIT
       run: make distclean && scan-build -v -o ~/scan-build --status-bugs --use-cc=clang --force-analyze-debug-code --show-description -analyzer-config stable-report-filename=true -enable-checker valist,nullability make ENABLE_EXT_F=0 ENABLE_SDL=0 ENABLE_JIT=0
     - name: run scan-build with JIT
-      run: make distclean && scan-build -v -o ~/scan-build --status-bugs --use-cc=clang --force-analyze-debug-code --show-description -analyzer-config stable-report-filename=true -enable-checker valist,nullability make ENABLE_EXT_F=0 ENABLE_SDL=0 ENABLE_JIT=1
+      run: |
+          make ENABLE_JIT=1 distclean && scan-build -v -o ~/scan-build --status-bugs --use-cc=clang --force-analyze-debug-code --show-description -analyzer-config stable-report-filename=true -enable-checker valist,nullability make ENABLE_EXT_F=0 ENABLE_SDL=0 ENABLE_JIT=1
 
   compliance-test:
     needs: [detect-code-related-file-changes]
--- a/Makefile
+++ b/Makefile
@@ -125,7 +125,32 @@ endif
 ENABLE_JIT ?= 0
 $(call set-feature, JIT)
 ifeq ($(call has, JIT), 1)
-OBJS_EXT += jit.o
+OBJS_EXT += jit.o 
+# tier-2 JIT compiler powered LLVM
+LLVM_CONFIG = llvm-config-17
+LLVM_CONFIG := $(shell which $(LLVM_CONFIG))
+ifndef LLVM_CONFIG
+# Try Homebrew on macOS
+LLVM_CONFIG = /opt/homebrew/opt/llvm@17/bin/llvm-config
+LLVM_CONFIG := $(shell which $(LLVM_CONFIG))
+ifdef LLVM_CONFIG
+LDFLAGS += -L/opt/homebrew/opt/llvm@17/lib
+endif
+endif
+ifneq ("$(LLVM_CONFIG)", "")
+ifneq ("$(findstring -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS, "$(shell $(LLVM_CONFIG) --cflags)")", "")
+ENABLE_T2C := 1
+$(call set-feature, T2C)
+OBJS_EXT += t2c.o
+CFLAGS += -g $(shell $(LLVM_CONFIG) --cflags)
+LDFLAGS += $(shell $(LLVM_CONFIG) --libs)
+else
+ENABLE_T2C := 0
+$(call set-feature, T2C)
+$(warning No llvm-config-17 installed. Check llvm-config-17 installation in advance)
+endif
+endif
+
 ifneq ($(processor),$(filter $(processor),x86_64 aarch64 arm64))
 $(error JIT mode only supports for x64 and arm64 target currently.)
 endif
@@ -134,6 +159,10 @@ src/rv32_jit.c:
 	$(Q)tools/gen-jit-template.py $(CFLAGS) > $@
 
 $(OUT)/jit.o: src/jit.c src/rv32_jit.c
+	$(VECHO) "  CC\t$@\n"
+	$(Q)$(CC) -o $@ $(CFLAGS) -c -MMD -MF $@.d $<
+
+$(OUT)/t2c.o: src/t2c.c src/t2c_template.c
 	$(VECHO) "  CC\t$@\n"
 	$(Q)$(CC) -o $@ $(CFLAGS) -c -MMD -MF $@.d $<
 endif
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Features:
 * Implementation of commonly used newlib system calls
 * Experimental SDL-based display/event/audio system calls for running video games
 * Support for remote GDB debugging
-* Experimental JIT compiler for performance boost while maintaining a small footprint
+* Tiered JIT compilation for performance boost while maintaining a small footprint
 
 ## Build and Verify
 
@@ -39,6 +39,12 @@ To ensure proper operation, the target system should have the [SDL2 library](htt
 and [SDL2_Mixer library](https://wiki.libsdl.org/SDL2_mixer) installed.
 * macOS: `brew install sdl2 sdl2_mixer`
 * Ubuntu Linux / Debian: `sudo apt install libsdl2-dev libsdl2-mixer-dev`
+
+### JIT compiler
+The tier-2 JIT compiler in `rv32emu` leverages LLVM for powerful optimization. Therefore, the target system must have [`LLVM`](https://llvm.org/) installed, with version 17 recommended. If `LLVM` is not installed, only the tier-1 JIT compiler will be used for performance enhancement.
+
+* macOS: `brew install llvm@17`
+* Ubuntu Linux / Debian: `sudo apt-get install llvm-17`
 
 Build the emulator.
 ```shell
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -304,7 +304,9 @@ static block_t *block_alloc(riscv_t *rv)
 #if RV32_HAS(JIT)
     block->translatable = true;
     block->hot = false;
+    block->hot2 = false;
     block->has_loops = false;
+    block->n_invoke = 0;
     INIT_LIST_HEAD(&block->list);
 #endif
     return block;
@@ -911,8 +913,6 @@ static bool runtime_profiler(riscv_t *rv, block_t *block)
         return true;
     return false;
 }
-
-typedef void (*exec_block_func_t)(riscv_t *rv, uintptr_t);
 #endif
 
 void rv_step(void *arg)
@@ -985,15 +985,31 @@ void rv_step(void *arg)
         }
         last_pc = rv->PC;
 #if RV32_HAS(JIT)
-        /* execute by tier-1 JIT compiler */
+#if RV32_HAS(T2C)
+        /* executed through the tier-2 JIT compiler */
+        if (block->hot2) {
+            ((exec_t2c_func_t) block->func)(rv);
+            prev = NULL;
+            continue;
+        } /* check if the execution path is strong hotspot */
+        if (block->n_invoke >= THRESHOLD) {
+            t2c_compile(block,
+                        (uint64_t) ((memory_t *) PRIV(rv)->mem)->mem_base);
+            ((exec_t2c_func_t) block->func)(rv);
+            prev = NULL;
+            continue;
+        }
+#endif
+        /* executed through the tier-1 JIT compiler */
         struct jit_state *state = rv->jit_state;
         if (block->hot) {
+            block->n_invoke++;
             ((exec_block_func_t) state->buf)(
                 rv, (uintptr_t) (state->buf + block->offset));
             prev = NULL;
             continue;
-        } /* check if using frequency of block exceed threshold */
-        else if (block->translatable && runtime_profiler(rv, block)) {
+        } /* check if the execution path is potential hotspot */
+        if (block->translatable && runtime_profiler(rv, block)) {
             jit_translate(rv, block);
             ((exec_block_func_t) state->buf)(
                 rv, (uintptr_t) (state->buf + block->offset));
--- a/src/t2c.c
+++ b/src/t2c.c
@@ -1,0 +1,302 @@
+/*
+ * rv32emu is freely redistributable under the MIT License. See the file
+ * "LICENSE" for information on usage and redistribution of this file.
+ */
+
+#include <llvm-c/Analysis.h>
+#include <llvm-c/BitWriter.h>
+#include <llvm-c/Core.h>
+#include <llvm-c/ExecutionEngine.h>
+#include <llvm-c/Target.h>
+#include <llvm-c/Transforms/PassBuilder.h>
+#include <stdlib.h>
+
+#include "jit.h"
+#include "riscv_private.h"
+
+#define MAX_BLOCKS 8152
+
+struct LLVM_block_map_entry {
+    uint32_t pc;
+    LLVMBasicBlockRef block;
+};
+
+struct LLVM_block_map {
+    uint32_t count;
+    struct LLVM_block_map_entry map[MAX_BLOCKS];
+};
+
+FORCE_INLINE void t2c_block_map_insert(struct LLVM_block_map *map,
+                                       LLVMBasicBlockRef *entry,
+                                       uint32_t pc)
+{
+    struct LLVM_block_map_entry map_entry = {
+        .block = *entry,
+        .pc = pc,
+    };
+    map->map[map->count++] = map_entry;
+    return;
+}
+
+FORCE_INLINE LLVMBasicBlockRef t2c_block_map_search(struct LLVM_block_map *map,
+                                                    uint32_t pc)
+{
+    for (uint32_t i = 0; i < map->count; i++) {
+        if (map->map[i].pc == pc) {
+            return map->map[i].block;
+        }
+    }
+    return NULL;
+}
+
+#define T2C_OP(inst, code)                                                \
+    static void t2c_##inst(                                               \
+        LLVMBuilderRef *builder UNUSED, LLVMTypeRef *param_types UNUSED,  \
+        LLVMValueRef start UNUSED, LLVMBasicBlockRef *entry UNUSED,       \
+        LLVMBuilderRef *taken_builder UNUSED,                             \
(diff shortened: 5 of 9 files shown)


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: 5fb1609a6ee34dd89769887d10bd1ae6077a65ce | Previous: 6c13a845d4ef1f6cadd5d1592e8b029e1f5ec089 | Ratio |
|-|-|-|-|
| `Dhrystone` | `3.44` Average DMIPS over 10 runs | `5.87` Average DMIPS over 10 runs | `1.71` |

</details>

jserv (changes requested):

.github/workflows/main.yml
```diff
@@ -65,14 +66,16 @@ jobs:
             make distclean ENABLE_GDBSTUB=1 gdbstub-test
     - name: JIT test
       run: |
-            make clean && make ENABLE_JIT=1 check -j$(nproc)
-            make clean && make ENABLE_EXT_A=0 ENABLE_JIT=1 check -j$(nproc)
-            make clean && make ENABLE_EXT_F=0 ENABLE_JIT=1 check -j$(nproc)
-            make clean && make ENABLE_EXT_C=0 ENABLE_JIT=1 check -j$(nproc)
+            export VER=-14
```
jserv: Why `VER` here?

Makefile
```diff
@@ -125,14 +125,20 @@ endif
 ENABLE_JIT ?= 0
 $(call set-feature, JIT)
 ifeq ($(call has, JIT), 1)
-OBJS_EXT += jit.o
+OBJS_EXT += jit.o t2jit.o
```
jserv: All tired-2 JIT compiler (T2C) related files should have the prefix `t2c`. That is, `t2c.h`, `t2c.c`, and `t2c_template.c`.

.github/workflows/main.yml
qwe661234: Because the cloud machine can not find the `llvm-config` when running GitHub CI, but `llvm-config-14` can be found.

jserv: Can you support [recent LLVM releases](https://releases.llvm.org/)? e.g., LLVM 18.x

Makefile
```diff
@@ -125,14 +125,20 @@ endif
...
+CFLAGS += -g `llvm-config${VER} --cflags`
```
jserv: T2C is only built when the supported versions of LLVM packages are available. That is, T2C is ignored if the LLVM check cannot be satisfied.

src/t2_rv32_template.c
```diff
@@ -0,0 +1,2143 @@
+
```
jserv: Add comments at the beginning of this file.

src/t2_rv32_template.c
```diff
@@ -0,0 +1,2143 @@
...
+RVT2OP(nop, { return; })
+
+RVT2OP(lui, {
+    LLVMValueRef rd_offset[1] = {LLVMConstInt(
```
jserv: Could you provide some helper macros to simplify the invocation of LLVM APIs? This could reduce the effort required when we need to upgrade LLVM versions.

src/t2jit.c
```diff
@@ -0,0 +1,222 @@
+#include <llvm-c/Analysis.h>
```
jserv: Add copyright notice.

src/emulate.c
```diff
@@ -1143,9 +1146,23 @@ void rv_step(void *arg)
         }
         last_pc = rv->PC;
 #if RV32_HAS(JIT)
+        /* execute by tier-2 JIT compiler */
+        if (block->hot2) {
+            ((funcPtr_t) block->func)(rv);
+            prev = NULL;
+            continue;
+        } /* check if invoking times of t1 generated code exceed threshold */
+        else if (block->n_invoke >= THRESHOLD) {
```
jserv: Replace `else if` with `if` statement.

jserv:
LLVM-17 is the preferable version for T2C's LLVM integration. Check [Linux_Build.yml](https://github.com/wheremyfoodat/Panda3DS/blob/master/.github/workflows/Linux_Build.yml) for the usage. Alternatively, [xPack built LLVM](https://xpack.github.io/dev-tools/clang/) packages are available for several host environments.

Makefile
```diff
@@ -125,7 +125,16 @@ endif
 ENABLE_JIT ?= 0
 $(call set-feature, JIT)
 ifeq ($(call has, JIT), 1)
-OBJS_EXT += jit.o
+OBJS_EXT += jit.o 
+ifneq ("$(shell which llvm-config-17)", "")
+ENABLE_T2C := 1
+OBJS_EXT += t2c.o
+CFLAGS += -g `llvm-config-17 --cflags`
```
jserv: Replace with `$(shell llvm-config-17 --cflags)`.

src/t2c.c
```diff
@@ -0,0 +1,227 @@
...
+        LLVMBuilderRef *builder UNUSED, LLVMTypeRef *param_types UNUSED,  \
+        LLVMValueRef start UNUSED, LLVMBasicBlockRef *entry UNUSED,       \
+        LLVMBuilderRef *taken_builder UNUSED,                             \
+        LLVMBuilderRef *untaken_builder UNUSED, uint64_t mem_base UNUSED, \
+        rv_insn_t *ir UNUSED)                                             \
+    {                                                                     \
+        code;                                                             \
+    }
+
+#include "t2c_rv32_template.c"
```
jserv: Rename to `t2c_template.c`

src/t2c.h
```diff
@@ -0,0 +1,10 @@
+/*
+ * rv32emu is freely redistributable under the MIT License. See the file
+ * "LICENSE" for information on usage and redistribution of this file.
+ */
+
+#include "riscv_private.h"
+
+typedef intptr_t (*funcPtr_t)(riscv_t *);
```
jserv: Enforce the consistent style as mentioned in `CONTRIBUTING.md`.
Of course, all LLVM related APIs are exceptional.

.github/workflows/main.yml
```diff
@@ -43,6 +43,9 @@ jobs:
       run: |
             sudo apt-get update -q -y
             sudo apt-get install -q -y libsdl2-dev libsdl2-mixer-dev
+            wget https://apt.llvm.org/llvm.sh
+            sudo chmod +x ./llvm.sh
+            sudo ./llvm.sh 17
```
jserv, line 49: Inside the guest environment initiated by [run-on-arch-action](https://github.com/uraimo/run-on-arch-action), the command `sudo` is not allowed. Instead, superuser (aka "root") is the only user there.

src/t2c_rv32_template.c
```diff
@@ -0,0 +1,1451 @@
+/* This file maps each custom IR to the corresponding LLVM IRs and builds LLVM
+ * IR through LLVM-C API. The built LLVM IR is offloaded to the LLVM backend,
+ * where it undergoes optimization through several selected LLVM passes.
+ * Subsequently, the optimized LLVM IR is passed to the LLVM execution engine,
+ * which compiles the optimized LLVM IR and returns a function pointer to the
+ * generated machine code.
+ */
+
+FORCE_INLINE LLVMValueRef build_rs1_addr(LLVMValueRef start,
```
jserv: Does it make more sense to move these helper functions out of `t2c_template.c`?

Makefile
```diff
@@ -125,7 +125,16 @@ endif
 ENABLE_JIT ?= 0
 $(call set-feature, JIT)
 ifeq ($(call has, JIT), 1)
-OBJS_EXT += jit.o
+OBJS_EXT += jit.o 
+ifeq ("$(shell llvm-config-17 --cflags)", "-I/usr/lib/llvm-17/include  -D_GNU_SOURCE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS")
```
jserv: This check might not be invalid for different Linux distributions and macOS (via homebrew).

jserv:
Consider the following quick and dirty fix for macOS enablement:
```diff
--- a/Makefile
+++ b/Makefile
@@ -125,16 +125,28 @@ endif
 ENABLE_JIT ?= 0
 $(call set-feature, JIT)
 ifeq ($(call has, JIT), 1)
-OBJS_EXT += jit.o 
-ifneq ("$(findstring -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS, "$(shell llvm-config-17 --cflags)")", "")
+OBJS_EXT += jit.o
+LLVM_CONFIG = llvm-config-17
+LLVM_CONFIG := $(shell which $(LLVM_CONFIG))
+ifndef LLVM_CONFIG
+    # Try Homebrew on macOS
+    LLVM_CONFIG = /opt/homebrew/opt/llvm@17/bin/llvm-config
+    LLVM_CONFIG := $(shell which $(LLVM_CONFIG))
+    ifdef LLVM_CONFIG
+    LDFLAGS += -L/opt/homebrew/opt/llvm@17/lib
+    endif
+endif
+ifneq ("$(LLVM_CONFIG)", "")
+ifneq ("$(findstring -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS, "$(shell $(LLVM_CONFIG) --cflags)")", "")
 ENABLE_T2C := 1
 OBJS_EXT += t2c.o
-CFLAGS += -g `llvm-config-17 --cflags`
-LDFLAGS += `llvm-config-17 --libs core executionengine mcjit interpreter analysis native --system-libs`
+CFLAGS += -g $(shell $(LLVM_CONFIG) --cflags)
+LDFLAGS += $(shell $(LLVM_CONFIG) --libs core executionengine mcjit interpreter analysis native --system-libs)
 else
 ENABLE_T2C := 0
 $(warning No llvm-config-17 installed. Check llvm-config-17 installation in advance)
 endif
+endif
 ifneq ($(processor),$(filter $(processor),x86_64 aarch64 arm64))
 $(error JIT mode only supports for x64 and arm64 target currently.)
 endif
```

jserv (changes requested):
Update the top-level documentation for the dependent packages for tier-2 JIT compiler.
* Debian/Ubuntu: `apt-get install llvm-17`
* macOS: `brew install llvm@17`

Makefile
```diff
@@ -125,7 +125,16 @@ endif
 ENABLE_JIT ?= 0
 $(call set-feature, JIT)
 ifeq ($(call has, JIT), 1)
-OBJS_EXT += jit.o
+OBJS_EXT += jit.o 
+ifneq ("$(findstring -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS, "$(shell llvm-config-17 --cflags)")", "")
+ENABLE_T2C := 1
```
jserv, line 142: Add some comments here. e.g., tier-2 JIT compiler powered by LLVM.

src/emulate.c
```diff
@@ -32,6 +32,9 @@ extern struct target_ops gdbstub_ops;
 #if RV32_HAS(JIT)
 #include "cache.h"
 #include "jit.h"
+#if RV32_HAS(T2C)
+#include "t2c.h"
```
jserv: The header `t2c.h` should be included in `jit.h`.

src/emulate.c
```diff
@@ -1137,15 +1143,31 @@ void rv_step(void *arg)
         }
         last_pc = rv->PC;
 #if RV32_HAS(JIT)
+#if RV32_HAS(T2C)
+        /* execute by tier-2 JIT compiler */
```
jserv: Improve the grammar.

src/t2c.h
```diff
@@ -0,0 +1,8 @@
+/*
```
jserv: The header `t2c.h` might not be necessary. Merge it into `jit.h`.

Makefile
```diff
@@ -125,7 +125,32 @@ endif
...
-OBJS_EXT += jit.o
+OBJS_EXT += jit.o 
+# tier-2 JIT compiler powered LLVM
+LLVM_CONFIG = llvm-config-17
+LLVM_CONFIG := $(shell which $(LLVM_CONFIG))
+ifndef LLVM_CONFIG
+  # Try Homebrew on macOS
+  LLVM_CONFIG = /opt/homebrew/opt/llvm@17/bin/llvm-config
+  LLVM_CONFIG := $(shell which $(LLVM_CONFIG))
+ifdef LLVM_CONFIG
```
jserv, line 136: Indent properly.

README.md
```diff
@@ -30,7 +30,7 @@ Features:
 * Implementation of commonly used newlib system calls
 * Experimental SDL-based display/event/audio system calls for running video games
 * Support for remote GDB debugging
-* Experimental JIT compiler for performance boost while maintaining a small footprint
+* Two-tier JIT compiler for performance boost while maintaining a small footprint
```
jserv: Replace with "Tiered JIT compilation"

README.md
```diff
@@ -40,6 +40,12 @@ and [SDL2_Mixer library](https://wiki.libsdl.org/SDL2_mixer) installed.
 * macOS: `brew install sdl2 sdl2_mixer`
 * Ubuntu Linux / Debian: `sudo apt install libsdl2-dev libsdl2-mixer-dev`
 
+### JIT compiler
+The tier-2 JIT compiler in `rv32emu` leverages LLVM for powerful optimization. Therefore, the target system must have [`LLVM`](https://github.com/llvm/llvm-project) installed, with version 17 recommended. If `LLVM` is not installed, only the tier-1 JIT compiler will be used for performance enhancement.
```
jserv: Specify LLVM website: https://llvm.org/

README.md
```diff
@@ -81,7 +87,7 @@ The image containing all the necessary tools for development and testing can be
 * `ENABLE_Zifencei`: Instruction-Fetch Fence
 * `ENABLE_GDBSTUB` : GDB remote debugging support
 * `ENABLE_SDL` : Experimental Display and Event System Calls
-* `ENABLE_JIT` : Experimental JIT compiler
+* `ENABLE_JIT` : Two-tier JIT compiler
```
jserv: Don't change this.

src/t2c.c
```diff
@@ -0,0 +1,314 @@
...
+struct LLVM_block_map {
+    uint32_t count;
+    struct LLVM_block_map_entry map[MAX_BLOCKS];
+};
+
+FORCE_INLINE void LLVM_block_map_insert(struct LLVM_block_map *map,
+                                        LLVMBasicBlockRef *entry,
+                                        uint32_t pc)
+{
+    struct LLVM_block_map_entry map_entry;
```
jserv: Use C99 designated initializers when possible.

src/t2c.c
```diff
@@ -0,0 +1,314 @@
...
+
+FORCE_INLINE LLVMValueRef build_rs1_addr(LLVMValueRef start,
+                                         LLVMBuilderRef *builder,
+                                         rv_insn_t *ir)
+{
+    LLVMValueRef rs1_offset[1] = {LLVMConstInt(
+        LLVMInt32Type(), offsetof(riscv_t, X) / sizeof(int) + ir->rs1, true)};
+    return LLVMBuildInBoundsGEP2(*builder, LLVMInt32Type(),
+                                 LLVMGetParam(start, 0), rs1_offset, 1,
+                                 "addr_rs1");
```
jserv: C language standard (i.e. C99 and C11) defines a predefined identifier as follows in clause 6.4.2.2:
> "The identifier `__func__` shall be implicitly declared by the translator as if, immediately following the opening brace of each function definition, the declaration
```c
static const char __func__[] = "function-name";
```
> appeared, where function-name is the name of the lexically-enclosing function."

It means that the C compiler implicitly adds `__func__` in every function so that it can be used in that function to get the function name.

jserv:
I am a bit worried about the lengthy file src/t2c_template.c, which could lead to potential maintenance issues. Can you suggest a more advanced way to reduce the code size for T2C?

qwe661234:
> I am a bit worried about the lengthy file src/t2c_template.c, which could lead to potential maintenance issues. Can you suggest a more advanced way to reduce the code size for T2C?

I reduce the code size through inline function and MARCO, this modification also provides more information about the gnerated machine code and make it more readable.

jserv (changes requested):
Rebase the latest `master` branch since libc substitution is dropped.

jserv (changes requested):
For consistent naming scheme, enforce the following rules:
- functions: specify prefix `t2c_`. e.g., `t2c_compile`
- macros: specify prefix `T2C`. e.g., `T2C_OP` and `T2C_LLVM_STORE_IMM32`

src/t2c.c
```diff
@@ -0,0 +1,302 @@
...
+        LLVMBuilderRef *builder UNUSED, LLVMTypeRef *param_types UNUSED,  \
+        LLVMValueRef start UNUSED, LLVMBasicBlockRef *entry UNUSED,       \
+        LLVMBuilderRef *taken_builder UNUSED,                             \
+        LLVMBuilderRef *untaken_builder UNUSED, uint64_t mem_base UNUSED, \
+        rv_insn_t *ir UNUSED)                                             \
+    {                                                                     \
+        code;                                                             \
+    }
+
+#define T2C_BUILD_LLVM_ADDR(reg, rv_member, ir_member)                       \
```
jserv: Should be `T2C_LLVM_BUILD_ADDR`. It can even be simpler as `T2C_LLVM_GEN`.

src/t2c.c
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+T2C_BUILD_LLVM_ADDR(rs1, X, ir->rs1);
+T2C_BUILD_LLVM_ADDR(rs2, X, ir->rs2);
+T2C_BUILD_LLVM_ADDR(rd, X, ir->rd);
+T2C_BUILD_LLVM_ADDR(ra, X, rv_reg_ra);
+T2C_BUILD_LLVM_ADDR(sp, X, rv_reg_sp);
+T2C_BUILD_LLVM_ADDR(PC, PC, 0);
+
+#define T2C_BUILD_LLVM_STORE_IMM32(builder, val, addr) \
```
jserv: The prefix `T2C_` should be applied to the implementation details, and the macros related to LLVM should have prefix `T2C_LLVM_`.

jserv:
I defer to @vacantron for reviewing.

src/t2c.c
```diff
@@ -0,0 +1,302 @@
...
+
+#define T2C_LLVM_GEN_ADDR(reg, rv_member, ir_member)                         \
+    FORCE_INLINE LLVMValueRef t2c_gen_##reg##_addr(                          \
+        LLVMValueRef start, LLVMBuilderRef *builder, UNUSED rv_insn_t *ir)   \
+    {                                                                        \
+        LLVMValueRef offset[1] = {LLVMConstInt(                              \
+            LLVMInt32Type(),                                                 \
+            offsetof(riscv_t, rv_member) / sizeof(int) + ir_member, true)};  \
+        return LLVMBuildInBoundsGEP2(*builder, LLVMInt32Type(),              \
+                                     LLVMGetParam(start, 0), offset, 1, ""); \
```
vacantron: Is this statement better?

```c
LLVMValueRef offset = LLVMConstInt(LLVMInt32Type(), offsetof(riscv_t, rv_member) / sizeof(int) + ir_member, true);
return LLVMBuildInBoundsGEP2(*builder, LLVMInt32Type(), LLVMGetParam(start, 0), &offset, 1, "");
```

src/t2c.c
```diff
@@ -0,0 +1,302 @@
...
+        *builder, LLVMPointerType(LLVMVoidType(), 0), LLVMGetParam(start, 0),
+        func_offset, 1, "addr_io_func");
+    LLVMValueRef io_func = LLVMBuildLoad2(
+        *builder,
+        LLVMPointerType(LLVMFunctionType(LLVMVoidType(), param_types, 1, 0), 0),
+        addr_io_func, "io_func");
+    LLVMValueRef io_param[1] = {LLVMGetParam(start, 0)};
+    LLVMBuildCall2(*builder,
+                   LLVMFunctionType(LLVMVoidType(), param_types, 1, 0), io_func,
+                   io_param, 1, "");
```
vacantron: Ditto.

src/t2c_template.c
```diff
@@ -0,0 +1,847 @@
...
+
+T2C_OP(fuse1, {
+    opcode_fuse_t *fuse = ir->fuse;
+    for (int i = 0; i < ir->imm2; i++) {
+        LLVMValueRef rd_offset[1] = {LLVMConstInt(
+            LLVMInt32Type(), offsetof(riscv_t, X) / sizeof(int) + fuse[i].rd,
+            true)};
+        LLVMValueRef addr_rd = LLVMBuildInBoundsGEP2(*builder, LLVMInt32Type(),
+                                                     LLVMGetParam(start, 0),
+                                                     rd_offset, 1, "addr_rd");
```
vacantron: Ditto.

vacantron (approved):

src/t2c_template.c
```diff
@@ -0,0 +1,847 @@
...
+T2C_OP(sw, {
+    LLVMValueRef mem_loc = t2c_gen_mem_loc(start, builder, ir, mem_base);
+    T2C_LLVM_GEN_LOAD_VMREG(rs2, 32, t2c_gen_rs2_addr(start, builder, ir));
+    LLVMBuildStore(*builder, val_rs2, mem_loc);
+})
+
+T2C_OP(addi, {
+    T2C_LLVM_GEN_LOAD_VMREG(rs1, 32, t2c_gen_rs1_addr(start, builder, ir));
+    LLVMValueRef res = T2C_LLVM_GEN_ALU32_IMM(Add, val_rs1, ir->imm);
+    LLVMBuildStore(*builder, res, t2c_gen_rd_addr(start, builder, ir));
```
vacantron, line 144: Can we avoid the storing of result after computing here or would it be optimized out in LLVM Pass?

src/t2c.c
```diff
@@ -0,0 +1,302 @@
...
+        if (map->map[i].pc == pc) {
+            return map->map[i].block;
+        }
+    }
+    return NULL;
+}
+
+#define T2C_OP(inst, code)                                                \
+    static void t2c_##inst(                                               \
+        LLVMBuilderRef *builder UNUSED, LLVMTypeRef *param_types UNUSED,  \
```
vacantron, line 54: The `LLVMBuilderRef *builder` is for receiving the change of control flow in most `slt`-like RISC-V insturctions.
Can we use `LLVMBuildSelect` alternatively to eliminate the pinter to builder and reduce the potential control flow change?

src/t2c_template.c
qwe661234: Yes, it will be optimized through LLVM Pass. To be specific, it will be optimized during register allocation stage of LLVM backend.

src/t2c.c
qwe661234: I have tried `LLVMBuildSelect` before the current implementation, but it would cause error. For example, you can try to replace it with `LLVMBuildSelect(*builder, cmp, T2C_LLVM_GEN_STORE_IMM32(*builder, 1, addr_rd), T2C_LLVM_GEN_STORE_IMM32(*builder, 0, addr_rd),"");`, and make the simulation exeute only through T2C path.

qwe661234: The error output of running `fcalc.elf` is
```
./build/rv32emu ./build/fcalc.elf 
Test 'Integer' failed; Expected 0.00000000000000000000000000000000000000000011/0, got 0.000000000000000000000000000000000000000000141937561/0
11Integer11.32Real number.32Sub-0-Real-42Negation1.2 + 5Addition3 - 1.4Subtraction10 * 5Multiplication50 / 10Division3 - 2 * 5PrecedencepiPi3 * (1 + 2)Parentheses(1 + 2Unmatched (Performed 0 tests, 0 failures, 0% success rate.
INFinfNANnan0123456789abcdef0123456789ABCDEF(null)0/POSIXCASCIIISO-8859-1cjksinglecjknarrowcjkwideUTF-8UTF8JISEUCJPEUCJPSJISISO8859ISO-8859-CPKOI8CP20866CP21866GEORGIANPSCP101PT154CP102TIS620CP874LC_ALLLANG.LC_COLLATELC_CTYPELC_MONETARYLC_NUMERICLC_TIMELC_MESSAGESInfinityNaNBalloc succeeded/Users/ilg/actions-runners/xpack-dev-tools/_work/riscv-none-elf-gcc-xpack/riscv-none-elf-gcc-xpack/build/darwin-arm64/sources/newlib-4.3.0.20230120/newlib/libc/stdlib/dtoa.c/Users/ilg/actions-runners/xpack-dev-tools/_work/riscv-none-elf-gcc-xpack/riscv-none-elf-gcc-xpack/build/darwin-arm64/sources/newlib-4.3.0.20230120/newlib/libc/stdlib/mprec.c, function: assertion "(null)" failed: file "(null)", line 011(null)
$��,$�������$�������4�������$���$���$���4���$���������������4���
��,
����,
��,
��,
���
   ��,
��,
����$��,
�������,
������  ���     ���     ���     ���     ���     ���     ���     ���     ��,
��,
��,
��,
��,
��,
��,
��,
��,\��h
��,
��,
��,
��P��,
��,
�����,
��,
��,
��`��,
��4��,
��,
��P��,
��,
��,
��,
��,
��,
��,
��,
��,
����������,
��<��,
��h�����(�����,
��`��8
      �����,
��,
```
Therefore, I believe the current implementation is feasible, as the use of `LLVMBuildSelect` results in incorrect behavior.

vacantron: I tried this and it works for me. It seems that `LLVMBuilderSelect` needs immediate value to be its operands.

```c
LLVMValueRef r1 = LLVMConstInt(LLVMInt32Type(), 1, true);
LLVMValueRef r2 = LLVMConstInt(LLVMInt32Type(), 0, true);
LLVMValueRef _ = LLVMBuildSelect(*builder, cmp, r1, r2, "");
LLVMBuildStore(*builder, _, addr_rd);
```

qwe661234: Right, it works for me as well. I misunderstood the use of `LLVMBuildSelect`. I have now modified the implementation accordingly.

vacantron (approved):

.github/workflows/main.yml
```diff
@@ -65,14 +68,14 @@ jobs:
...
-            make clean && make ENABLE_EXT_C=0 ENABLE_JIT=1 check -j$(nproc)
+            make ENABLE_JIT=1 clean && make ENABLE_JIT=1 check -j$(nproc)
+            make ENABLE_JIT=1 clean && make ENABLE_EXT_A=0 ENABLE_JIT=1 check -j$(nproc)
+            make ENABLE_JIT=1 clean && make ENABLE_EXT_F=0 ENABLE_JIT=1 check -j$(nproc)
+            make ENABLE_JIT=1 clean && make ENABLE_EXT_C=0 ENABLE_JIT=1 check -j$(nproc)
     - name: undefined behavior test
       run: |
             make clean && make ENABLE_UBSAN=1 check -j$(nproc)
-            make clean && make ENABLE_JIT=1 ENABLE_UBSAN=1 check -j$(nproc)
+            make ENABLE_JIT=1 clean clean && make ENABLE_JIT=1 ENABLE_UBSAN=1 check -j$(nproc)
```
henrybear327, line 78: Out of curiosity, is there a reason to run the `clean` target twice? @qwe661234

qwe661234: Sorry, it is my mistake.

PR #432, opened 2024-05-09
Raise illegal instruction exception for Zicsr

ChinYikMing:
According to the RISC-V privileged architecture 20211203 (Section 2.1), write a read-only register should raise illegal instruction exception. Also, unprivileged ISA 20191213 states that a CSR instruction is considered a write when the rs1 or uimm is not zero (check Table 9.1). Since the exceptions are raised during the decoding stage, the execution of csr_csrrw, csr_csrrs, and csr_csrrc function can skip checking the writability of csr anymore.

### DIFF
--- a/src/decode.c
+++ b/src/decode.c
@@ -788,6 +788,11 @@ static inline bool op_jal(rv_insn_t *ir, const uint32_t insn)
     return true;
 }
 
+FORCE_INLINE bool csr_is_writable(const uint32_t csr)
+{
+    return csr < 0xc00;
+}
+
 /* SYSTEM: I-type
  *  31       20 19   15 14    12 11   7 6      0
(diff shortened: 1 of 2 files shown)


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: 644124a1ddc38ff46291bd993bc52b6d96f5b49e | Previous: a3d2418b95ccd3ee7824690f8b49a7fd01952a08 | Ratio |
|-|-|-|-|
| `Dhrystone` | `1754.22` Average DMIPS over 10 runs | `1608` Average DMIPS over 10 runs | `0.92` |
| `Coremark` | `1503.058` Average iterations/sec over 10 runs | `1476.195` Average iterations/sec over 10 runs | `0.98` |

</details>

jserv:

PR #436, opened 2024-05-11
Optimize syscall_write() to avoid redundant checks and FILE* retrievals

visitorckw:
There is redundant checking of map_at_end within a loop, along with repetitive calls to map_iter_value() to retrieve the file pointer. This results in unnecessary overhead.

Address the issue by optimizing the code to perform the map_at_end check and retrieve the file pointer only once at the beginning.


ChinYikMing (approved):

ChinYikMing:
LGTM.

PR #437, opened 2024-05-11
Consolidate the block_translate fucntion

qwe661234:
Replace the `mpool_alloc` and `memset` with `mpool_calloc`.

PR #438, opened 2024-05-12
Preliminary support for MMU emulation

ChinYikMing:
The purpose of this commit is to boot 32-bit RISC-V Linux in the future. The virtual memory scheme to support is Sv32. There are one change to original code base to adapt the MMU: the prototype of `riscv_io_t` interface needs to be changed. Particularly, add a RISC-V instance(`riscv_t`) as the first parameter. MMU related callbacks require to access the satp CSR to perform a page table walk during virtual memory translation but satp CSR is stored in RISC-V instance(`riscv_t`), thus it should have a way to access the satp CSR. The trivial solution is adding RISC-V
instance(`riscv_t`) to the prototype of `riscv_io_t` interface.

After this change, we can reuse riscv_io_t for system emulation afterward.

The rest of changes are implementing the Sv32 virtual memory scheme. For every memory access, it has to walk through the page table to get the corresponding PTE. Depends on the retrieval of PTE, there are several page faults to be handled if necessary, so there are three exceptions handlers have been introduced which are insn_pgfault, load_pgfault, and store_pgfault and they are used in MMU_CHECK_FAULT. In this commit, the access fault are not handled well since they are related to PMA and PMP and they might not the must to boot 32-bit RISC-V Linux (tested on semu). More PTE, S-mode, M-mode CSR helper macro are introduced as well.

Related: #310

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,10 @@ CFLAGS += $(CFLAGS_NO_CET)
 
 OBJS_EXT :=
 
+ifeq ($(call has, SYSTEM), 1)
+OBJS_EXT += system.o
+endif
+
 # Integer Multiplication and Division instructions
 ENABLE_EXT_M ?= 1
 $(call set-feature, EXT_M)
--- a/Makefile
+++ b/Makefile
@@ -303,6 +303,16 @@ misalign-in-blk-emu: $(BIN)
 	             exit 1; \
 	             fi;
 
+EXPECTED_mmu = STORE PAGE FAULT TEST PASSED!
+mmu-test: $(BIN)
+	$(Q)$(PRINTF) "Running vm.elf ... "; \
+	    if [ "$(shell $(BIN) tests/system/mmu/vm.elf | tail -n 2)" = "$(strip $(EXPECTED_mmu)) inferior exit code 0" ]; then \
+	    $(call notice, [OK]); \
+	    else \
+	    $(PRINTF) "Failed.\n"; \
+	    exit 1; \
+	    fi;
+
 # Non-trivial demonstration programs
 ifeq ($(call has, SDL), 1)
 doom_action := (cd $(OUT); ../$(BIN) riscv32/doom)
--- a/src/decode.c
+++ b/src/decode.c
@@ -907,9 +907,8 @@ static inline bool op_system(rv_insn_t *ir, const uint32_t insn)
     default: /* illegal instruction */
         return false;
     }
-    if (!csr_is_writable(ir->imm) && ir->rs1 != rv_reg_zero)
-        return false;
-    return true;
+
+    return csr_is_writable(ir->imm) || (ir->rs1 == rv_reg_zero);
 }
 
 /* MISC-MEM: I-type
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -41,130 +41,15 @@ extern struct target_ops gdbstub_ops;
 #define IF_rs2(i, r) (i->rs2 == rv_reg_##r)
 #define IF_imm(i, v) (i->imm == v)
 
-/* RISC-V exception code list */
-/* clang-format off */
-#define RV_TRAP_LIST                                               \
-    IIF(RV32_HAS(EXT_C))(,                                         \
-        _(insn_misaligned, 0) /* Instruction address misaligned */ \
-    )                                                              \
-    _(illegal_insn, 2)     /* Illegal instruction */               \
-    _(breakpoint, 3)       /* Breakpoint */                        \
-    _(load_misaligned, 4)  /* Load address misaligned */           \
-    _(store_misaligned, 6) /* Store/AMO address misaligned */      \
-    IIF(RV32_HAS(SYSTEM))(,                                        \
-        _(ecall_M, 11)     /* Environment call from M-mode */      \
-    )
-/* clang-format on */
-
-enum {
-#define _(type, code) rv_trap_code_##type = code,
-    RV_TRAP_LIST
-#undef _
-};
-
 static void rv_trap_default_handler(riscv_t *rv)
 {
     rv->csr_mepc += rv->compressed ? 2 : 4;
     rv->PC = rv->csr_mepc; /* mret */
 }
 
-/*
- * Trap might occurs during block emulation. For instance, page fault.
- * In order to handle trap, we have to escape from block and execute
- * registered trap handler. This trap_handler function helps to execute
- * the registered trap handler, PC by PC. Once the trap is handled,
- * resume the previous execution flow where cause the trap.
- *
- * Since the system emulation has not yet included in rv32emu, the page
- * fault is not practical in current test suite. Instead, we try to
- * emulate the misaligned handling in the test suite.
- */
 #if RV32_HAS(SYSTEM)
-static void trap_handler(riscv_t *rv);
-#endif
-
-/* When a trap occurs in M-mode/S-mode, m/stval is either initialized to zero or
- * populated with exception-specific details to assist software in managing
- * the trap. Otherwise, the implementation never modifies m/stval, although
- * software can explicitly write to it. The hardware platform will define
- * which exceptions are required to informatively set mtval and which may
- * consistently set it to zero.
- *
- * When a hardware breakpoint is triggered or an exception like address
- * misalignment, access fault, or page fault occurs during an instruction
- * fetch, load, or store operation, m/stval is updated with the virtual address
- * that caused the fault. In the case of an illegal instruction trap, m/stval
- * might be updated with the first XLEN or ILEN bits of the offending
- * instruction. For all other traps, m/stval is simply set to zero. However,
- * it is worth noting that a future standard could redefine how m/stval is
- * handled for different types of traps.
- *
- * For simplicity and clarity, abstracting stval and mtval into a single
- * identifier called tval, as both are handled by TRAP_HANDLER_IMPL.
- */
-#define TRAP_HANDLER_IMPL(type, code)                                         \
-    static void rv_trap_##type(riscv_t *rv, uint32_t tval)                    \
-    {                                                                         \
-        /* m/stvec (Machine/Supervisor Trap-Vector Base Address Register)     \
-         * m/stvec[MXLEN-1:2]: vector base address                            \
-         * m/stvec[1:0] : vector mode                                         \
-         * m/sepc  (Machine/Supervisor Exception Program Counter)             \
-         * m/stval (Machine/Supervisor Trap Value Register)                   \
-         * m/scause (Machine/Supervisor Cause Register): store exception code \
-         * m/sstatus (Machine/Supervisor Status Register): keep track of and  \
-         * controls the hart’s current operating state                      \
-         */                                                                   \
-        uint32_t base;                                                        \
-        uint32_t mode;                                                        \
-        /* user or supervisor */                                              \
-        if (RV_PRIV_IS_U_OR_S_MODE()) {                                       \
-            const uint32_t sstatus_sie =                                      \
-                (rv->csr_sstatus & SSTATUS_SIE) >> SSTATUS_SIE_SHIFT;         \
-            rv->csr_sstatus |= (sstatus_sie << SSTATUS_SPIE_SHIFT);           \
-            rv->csr_sstatus &= ~(SSTATUS_SIE);                                \
-            rv->csr_sstatus |= (rv->priv_mode << SSTATUS_SPP_SHIFT);          \
-            rv->priv_mode = RV_PRIV_S_MODE;                                   \
-            base = rv->csr_stvec & ~0x3;                                      \
-            mode = rv->csr_stvec & 0x3;                                       \
-            rv->csr_sepc = rv->PC;                                            \
-            rv->csr_stval = tval;                                             \
-            rv->csr_scause = code;                                            \
-        } else { /* machine */                                                \
-            const uint32_t mstatus_mie =                                      \
-                (rv->csr_mstatus & MSTATUS_MIE) >> MSTATUS_MIE_SHIFT;         \
-            rv->csr_mstatus |= (mstatus_mie << MSTATUS_MPIE_SHIFT);           \
-            rv->csr_mstatus &= ~(MSTATUS_MIE);                                \
-            rv->csr_mstatus |= (rv->priv_mode << MSTATUS_MPP_SHIFT);          \
-            rv->priv_mode = RV_PRIV_M_MODE;                                   \
-            base = rv->csr_mtvec & ~0x3;                                      \
-            mode = rv->csr_mtvec & 0x3;                                       \
-            rv->csr_mepc = rv->PC;                                            \
-            rv->csr_mtval = tval;                                             \
-            rv->csr_mcause = code;                                            \
-            if (!rv->csr_mtvec) { /* in case CSR is not configured */         \
-                rv_trap_default_handler(rv);                                  \
-                return;                                                       \
-            }                                                                 \
-        }                                                                     \
-        switch (mode) {                                                       \
-        /* DIRECT: All traps set PC to base */                                \
-        case 0:                                                               \
-            rv->PC = base;                                                    \
-            break;                                                            \
-        /* VECTORED: Asynchronous traps set PC to base + 4 * code */          \
-        case 1:                                                               \
-            /* MSB of code is used to indicate whether the trap is interrupt  \
-             * or exception, so it is not considered as the 'real' code */    \
-            rv->PC = base + 4 * (code & MASK(31));                            \
-            break;                                                            \
-        }                                                                     \
-        IIF(RV32_HAS(SYSTEM))(if (rv->is_trapped) trap_handler(rv);, )        \
-    }
-
-/* RISC-V exception handlers */
-#define _(type, code) TRAP_HANDLER_IMPL(type, code)
-RV_TRAP_LIST
-#undef _
+static void __trap_handler(riscv_t *rv);
+#endif /* RV32_HAS(SYSTEM) */
 
 /* wrap load/store and insn misaligned handler
  * @mask_or_pc: mask for load/store and pc for insn misaligned handler.
@@ -180,8 +65,8 @@ RV_TRAP_LIST
         rv->compressed = compress;                                    \
         rv->csr_cycle = cycle;                                        \
         rv->PC = PC;                                                  \
-        IIF(RV32_HAS(SYSTEM))(rv->is_trapped = true, );               \
-        rv_trap_##type##_misaligned(rv, IIF(IO)(addr, mask_or_pc));   \
+        SET_CAUSE_AND_TVAL_THEN_TRAP(rv, type##_MISALIGNED,           \
+                                     IIF(IO)(addr, mask_or_pc));      \
         return false;                                                 \
     }
 
@@ -531,8 +416,8 @@ static bool do_fuse3(riscv_t *rv, rv_insn_t *ir, uint64_t cycle, uint32_t PC)
      */
     for (int i = 0; i < ir->imm2; i++) {
         uint32_t addr = rv->X[fuse[i].rs1] + fuse[i].imm;
-        RV_EXC_MISALIGN_HANDLER(3, store, false, 1);
-        rv->io.mem_write_w(addr, rv->X[fuse[i].rs2]);
+        RV_EXC_MISALIGN_HANDLER(3, STORE, false, 1);
+        rv->io.mem_write_w(rv, addr, rv->X[fuse[i].rs2]);
     }
     PC += ir->imm2 * 4;
     if (unlikely(RVOP_NO_NEXT(ir))) {
@@ -555,8 +440,8 @@ static bool do_fuse4(riscv_t *rv, rv_insn_t *ir, uint64_t cycle, uint32_t PC)
      */
     for (int i = 0; i < ir->imm2; i++) {
         uint32_t addr = rv->X[fuse[i].rs1] + fuse[i].imm;
-        RV_EXC_MISALIGN_HANDLER(3, load, false, 1);
-        rv->X[fuse[i].rd] = rv->io.mem_read_w(addr);
+        RV_EXC_MISALIGN_HANDLER(3, LOAD, false, 1);
+        rv->X[fuse[i].rd] = rv->io.mem_read_w(rv, addr);
     }
     PC += ir->imm2 * 4;
     if (unlikely(RVOP_NO_NEXT(ir))) {
@@ -666,12 +551,12 @@ static void block_translate(riscv_t *rv, block_t *block)
             prev_ir->next = ir;
 
         /* fetch the next instruction */
-        const uint32_t insn = rv->io.mem_ifetch(block->pc_end);
+        const uint32_t insn = rv->io.mem_ifetch(rv, block->pc_end);
 
         /* decode the instruction */
         if (!rv_decode(ir, insn)) {
             rv->compressed = is_compressed(insn);
-            rv_trap_illegal_insn(rv, insn);
+            SET_CAUSE_AND_TVAL_THEN_TRAP(rv, ILLEGAL_INSN, insn);
             break;
         }
         ir->impl = dispatch_table[ir->opcode];
@@ -1122,15 +1007,14 @@ void rv_step(void *arg)
 }
 
 #if RV32_HAS(SYSTEM)
-static void trap_handler(riscv_t *rv)
+static void __trap_handler(riscv_t *rv)
 {
     rv_insn_t *ir = mpool_alloc(rv->block_ir_mp);
     assert(ir);
 
-    /* set to false by sret/mret implementation */
-    uint32_t insn;
+    /* set to false by sret implementation */
     while (rv->is_trapped && !rv_has_halted(rv)) {
-        insn = rv->io.mem_ifetch(rv->PC);
+        uint32_t insn = rv->io.mem_ifetch(rv, rv->PC);
         assert(insn);
 
         rv_decode(ir, insn);
@@ -1139,12 +1023,94 @@ static void trap_handler(riscv_t *rv)
         ir->impl(rv, ir, rv->csr_cycle, rv->PC);
     }
 }
-#endif
+#endif /* RV32_HAS(SYSTEM) */
+
+/* When a trap occurs in M-mode/S-mode, m/stval is either initialized to zero or
+ * populated with exception-specific details to assist software in managing
+ * the trap. Otherwise, the implementation never modifies m/stval, although
+ * software can explicitly write to it. The hardware platform will define
+ * which exceptions are required to informatively set mtval and which may
+ * consistently set it to zero.
+ *
+ * When a hardware breakpoint is triggered or an exception like address
+ * misalignment, access fault, or page fault occurs during an instruction
+ * fetch, load, or store operation, m/stval is updated with the virtual address
+ * that caused the fault. In the case of an illegal instruction trap, m/stval
+ * might be updated with the first XLEN or ILEN bits of the offending
+ * instruction. For all other traps, m/stval is simply set to zero. However,
+ * it is worth noting that a future standard could redefine how m/stval is
+ * handled for different types of traps.
+ *
+ */
+static void _trap_handler(riscv_t *rv)
+{
+    /* m/stvec (Machine/Supervisor Trap-Vector Base Address Register)
+     * m/stvec[MXLEN-1:2]: vector base address
+     * m/stvec[1:0] : vector mode
+     * m/sepc  (Machine/Supervisor Exception Program Counter)
+     * m/stval (Machine/Supervisor Trap Value Register)
+     * m/scause (Machine/Supervisor Cause Register): store exception code
+     * m/sstatus (Machine/Supervisor Status Register): keep track of and
+     * controls the hart’s current operating state
+     *
+     * m/stval and m/scause are set in SET_CAUSE_AND_TVAL_THEN_TRAP
+     */
+    uint32_t base;
+    uint32_t mode;
+    uint32_t cause;
+    /* user or supervisor */
+    if (RV_PRIV_IS_U_OR_S_MODE()) {
+        const uint32_t sstatus_sie =
+            (rv->csr_sstatus & SSTATUS_SIE) >> SSTATUS_SIE_SHIFT;
+        rv->csr_sstatus |= (sstatus_sie << SSTATUS_SPIE_SHIFT);
+        rv->csr_sstatus &= ~(SSTATUS_SIE);
+        rv->csr_sstatus |= (rv->priv_mode << SSTATUS_SPP_SHIFT);
+        rv->priv_mode = RV_PRIV_S_MODE;
+        base = rv->csr_stvec & ~0x3;
+        mode = rv->csr_stvec & 0x3;
+        cause = rv->csr_scause;
+        rv->csr_sepc = rv->PC;
+    } else { /* machine */
+        const uint32_t mstatus_mie =
+            (rv->csr_mstatus & MSTATUS_MIE) >> MSTATUS_MIE_SHIFT;
+        rv->csr_mstatus |= (mstatus_mie << MSTATUS_MPIE_SHIFT);
+        rv->csr_mstatus &= ~(MSTATUS_MIE);
+        rv->csr_mstatus |= (rv->priv_mode << MSTATUS_MPP_SHIFT);
+        rv->priv_mode = RV_PRIV_M_MODE;
+        base = rv->csr_mtvec & ~0x3;
+        mode = rv->csr_mtvec & 0x3;
+        cause = rv->csr_mcause;
+        rv->csr_mepc = rv->PC;
+        if (!rv->csr_mtvec) { /* in case CSR is not configured */
+            rv_trap_default_handler(rv);
+            return;
+        }
+    }
+    switch (mode) {
+    /* DIRECT: All traps set PC to base */
+    case 0:
+        rv->PC = base;
+        break;
+    /* VECTORED: Asynchronous traps set PC to base + 4 * code */
+    case 1:
+        /* MSB of code is used to indicate whether the trap is interrupt
+         * or exception, so it is not considered as the 'real' code */
+        rv->PC = base + 4 * (cause & MASK(31));
+        break;
+    }
+    IIF(RV32_HAS(SYSTEM))(if (rv->is_trapped) __trap_handler(rv);, )
+}
+
+void trap_handler(riscv_t *rv)
+{
+    assert(rv);
+    _trap_handler(rv);
+}
 
 void ebreak_handler(riscv_t *rv)
 {
     assert(rv);
-    rv_trap_breakpoint(rv, rv->PC);
+    SET_CAUSE_AND_TVAL_THEN_TRAP(rv, BREAKPOINT, rv->PC);
 }
 
 void ecall_handler(riscv_t *rv)
@@ -1154,7 +1120,7 @@ void ecall_handler(riscv_t *rv)
     syscall_handler(rv);
     rv->PC += 4;
 #else
-    rv_trap_ecall_M(rv, 0);
+    SET_CAUSE_AND_TVAL_THEN_TRAP(rv, ECALL_M, 0);
     syscall_handler(rv);
 #endif
 }
--- a/src/main.c
+++ b/src/main.c
@@ -217,12 +217,21 @@ int main(int argc, char **args)
         .log_level = 0,
         .run_flag = run_flag,
         .profile_output_file = prof_out_file,
+#if RV32_HAS(SYSTEM)
+        .data.system = malloc(sizeof(vm_system_t)),
+#else
         .data.user = malloc(sizeof(vm_user_t)),
+#endif
         .cycle_per_step = CYCLE_PER_STEP,
         .allow_misalign = opt_misaligned,
     };
+#if RV32_HAS(SYSTEM)
+    assert(attr.data.system);
+    attr.data.system->elf_program = opt_prog_name;
+#else
     assert(attr.data.user);
     attr.data.user->elf_program = opt_prog_name;
+#endif
 
     /* create the RISC-V runtime */
     rv = rv_create(&attr);
--- a/src/riscv.c
+++ b/src/riscv.c
@@ -165,8 +165,10 @@ void rv_remap_stdstream(riscv_t *rv, fd_stream_pair_t *fsp, uint32_t fsp_size)
 #define MEMIO(op) on_mem_##op
 #define IO_HANDLER_IMPL(type, op, RW)                                     \
     static IIF(RW)(                                                       \
-        /* W */ void MEMIO(op)(riscv_word_t addr, riscv_##type##_t data), \
-        /* R */ riscv_##type##_t MEMIO(op)(riscv_word_t addr))            \
+        /* W */ void MEMIO(op)(UNUSED riscv_t * rv, riscv_word_t addr,    \
+                               riscv_##type##_t data),                    \
+        /* R */ riscv_##type##_t MEMIO(op)(UNUSED riscv_t * rv,           \
+                                           riscv_word_t addr))            \
     {                                                                     \
         IIF(RW)                                                           \
         (memory_##op(addr, (uint8_t *) &data), return memory_##op(addr)); \
@@ -258,11 +260,40 @@ riscv_t *rv_create(riscv_user_t rv_attr)
             .on_ebreak = ebreak_handler,
             .on_memcpy = memcpy_handler,
             .on_memset = memset_handler,
+            .on_trap = trap_handler,
         };
         memcpy(&rv->io, &io, sizeof(riscv_io_t));
-    } else {
-        /* TODO: system emulator */
     }
+#if RV32_HAS(SYSTEM)
+    else {
+        /*
+         * TODO: system emulator
+         * e.g., kernel image, dtb, rootfs
+         *
+         * The test suite is compiled into a single ELF file, so load it as
+         * an ELF executable, just like a userspace ELF.
+         *
+         */
+        elf_t *elf = elf_new();
+        assert(elf && elf_open(elf, (attr->data.system)->elf_program));
+
+        const struct Elf32_Sym *end;
+        if ((end = elf_get_symbol(elf, "_end")))
+            attr->break_addr = end->st_value;
+
+        assert(elf_load(elf, attr->mem));
+
+        /* set the entry pc */
+        const struct Elf32_Ehdr *hdr = get_elf_header(elf);
+        assert(rv_set_pc(rv, hdr->e_entry));
+
+        elf_delete(elf);
+
+        /* this variable has external linkage to mmu_io defined in system.c */
+        extern riscv_io_t mmu_io;
+        memcpy(&rv->io, &mmu_io, sizeof(riscv_io_t));
+    }
+#endif /* SYSTEM */
 
     /* default standard stream.
      * rv_remap_stdstream can be called to overwrite them
@@ -301,20 +332,6 @@ riscv_t *rv_create(riscv_user_t rv_attr)
     /* activate the background compilation thread. */
     pthread_create(&t2c_thread, NULL, t2c_runloop, rv);
 #endif
-#endif
-
-#if RV32_HAS(SYSTEM)
-    /*
-     * System simulation defaults to S-mode as
-     * it does not rely on M-mode software like OpenSBI.
-     */
-    rv->priv_mode = RV_PRIV_S_MODE;
-
-    /* not being trapped */
-    rv->is_trapped = false;
-#else
-    /* ISA simulation defaults to M-mode */
-    rv->priv_mode = RV_PRIV_M_MODE;
 #endif
 
     return rv;
@@ -356,7 +373,13 @@ void rv_run(riscv_t *rv)
     assert(rv);
 
     vm_attr_t *attr = PRIV(rv);
-    assert(attr && attr->data.user && attr->data.user->elf_program);
+    assert(attr &&
+#if RV32_HAS(SYSTEM)
+           attr->data.system && attr->data.system->elf_program
+#else
+           attr->data.user && attr->data.user->elf_program
+#endif
+    );
 
     if (attr->run_flag & RV_RUN_TRACE)
         rv_run_and_trace(rv);
@@ -516,6 +539,21 @@ void rv_reset(riscv_t *rv, riscv_word_t pc)
     /* reset sp pointing to argc */
     rv->X[rv_reg_sp] = stack_top;
 
+    /* reset privilege mode */
+#if RV32_HAS(SYSTEM)
+    /*
+     * System simulation defaults to S-mode as
+     * it does not rely on M-mode software like OpenSBI.
+     */
+    rv->priv_mode = RV_PRIV_S_MODE;
+
+    /* not being trapped */
+    rv->is_trapped = false;
+#else
+    /* ISA simulation defaults to M-mode */
+    rv->priv_mode = RV_PRIV_M_MODE;
+#endif
+
     /* reset the csrs */
     rv->csr_mtvec = 0;
     rv->csr_cycle = 0;
@@ -540,7 +578,6 @@ void rv_reset(riscv_t *rv, riscv_word_t pc)
 #if RV32_HAS(EXT_M)
     rv->csr_misa |= MISA_M;
 #endif
-
 
     rv->halt = false;
 }
--- a/src/riscv.h
+++ b/src/riscv.h
@@ -91,10 +91,129 @@ enum {
 #define MISA_A (1 << ('A' - 'A'))
 #define MISA_F (1 << ('F' - 'A'))
 #define MISA_C (1 << ('C' - 'A'))
+
+/* The mstatus register keeps track of and controls the hart’s current operating
+ * state */
+#define MSTATUS_SIE_SHIFT 1
+#define MSTATUS_MIE_SHIFT 3
+#define MSTATUS_SPIE_SHIFT 5
+#define MSTATUS_UBE_SHIFT 6
 #define MSTATUS_MPIE_SHIFT 7
+#define MSTATUS_SPP_SHIFT 8
 #define MSTATUS_MPP_SHIFT 11
+#define MSTATUS_MPRV_SHIFT 17
+#define MSTATUS_SUM_SHIFT 18
+#define MSTATUS_MXR_SHIFT 18
+#define MSTATUS_TVM_SHIFT 20
+#define MSTATUS_TW_SHIFT 21
+#define MSTATUS_TSR_SHIFT 22
+#define MSTATUS_SIE (1 << MSTATUS_SIE_SHIFT)
+#define MSTATUS_MIE (1 << MSTATUS_MIE_SHIFT)
+#define MSTATUS_SPIE (1 << MSTATUS_SPIE_SHIFT)
+#define MSTATUS_UBE (1 << MSTATUS_UBE_SHIFT)
 #define MSTATUS_MPIE (1 << MSTATUS_MPIE_SHIFT)
+#define MSTATUS_SPP (1 << MSTATUS_SPP_SHIFT)
 #define MSTATUS_MPP (3 << MSTATUS_MPP_SHIFT)
+#define MSTATUS_MPRV (1 << MSTATUS_MPRV_SHIFT)
+#define MSTATUS_SUM (1 << MSTATUS_SUM_SHIFT)
+#define MSTATUS_MXR (1 << MSTATUS_MXR_SHIFT)
+#define MSTATUS_TVM (1 << MSTATUS_TVM_SHIFT)
+#define MSTATUS_TW (1 << MSTATUS_TW_SHIFT)
+#define MSTATUS_TSR (1 << MSTATUS_TSR_SHIFT)
+
+/* A restricted view of mstatus */
+#define SSTATUS_SIE_SHIFT 1
+#define SSTATUS_SPIE_SHIFT 5
+#define SSTATUS_UBE_SHIFT 6
+#define SSTATUS_SPP_SHIFT 8
+#define SSTATUS_SUM_SHIFT 18
+#define SSTATUS_MXR_SHIFT 19
+#define SSTATUS_SIE (1 << SSTATUS_SIE_SHIFT)
+#define SSTATUS_SPIE (1 << SSTATUS_SPIE_SHIFT)
+#define SSTATUS_UBE (1 << SSTATUS_UBE_SHIFT)
+#define SSTATUS_SPP (1 << SSTATUS_SPP_SHIFT)
+#define SSTATUS_SUM (1 << SSTATUS_SUM_SHIFT)
+#define SSTATUS_MXR (1 << SSTATUS_MXR_SHIFT)
+
+#define SIP_SSIP_SHIFT 1
+#define SIP_STIP_SHIFT 5
+#define SIP_SEIP_SHIFT 9
+#define SIP_SSIP (1 << SIP_SSIP_SHIFT)
+#define SIP_STIP (1 << SIP_STIP_SHIFT)
+#define SIP_SEIP (1 << SIP_SEIP_SHIFT)
+
+#define RV_PG_SHIFT 12
+#define RV_PG_SIZE (1 << RV_PG_SHIFT)
+
+#define RV_PRIV_U_MODE 0
+#define RV_PRIV_S_MODE 1
+#define RV_PRIV_M_MODE 3
+
+#define PTE_V (1U)
+#define PTE_R (1U << 1)
+#define PTE_W (1U << 2)
+#define PTE_X (1U << 3)
+#define PTE_U (1U << 4)
+#define PTE_G (1U << 5)
+#define PTE_A (1U << 6)
+#define PTE_D (1U << 7)
+
+/* PTE XWRV bit in order */
+enum SV32_PTE_PERM {
+    NEXT_PG_TBL = 0b0001,
+    RO_PAGE = 0b0011,
+    RW_PAGE = 0b0111,
+    EO_PAGE = 0b1001,
+    RX_PAGE = 0b1011,
+    RWX_PAGE = 0b1111,
+    RESRV_PAGE1 = 0b0101,
+    RESRV_PAGE2 = 0b1101,
+};
+
+/*
+ * SBI functions must return a pair of values:
+ *
+ * struct sbiret {
+ *     long error;
+ *     long value;
+ * };
+ *
+ * The error and value field will be set to register a0 and a1 respectively
+ * after the SBI function return. The error field indicate whether the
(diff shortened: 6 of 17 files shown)


ChinYikMing:
This PR is not fully ready to be merged since testing is not yet  fully designed. PR earlier to get some feedbacks for further design.

jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: 63176056e9b85c1d6c96517b91da250bcc216e74 | Previous: 79302e243631a093d8c647e5e7da621053ab8cb6 | Ratio |
|-|-|-|-|
| `Dhrystone` | `1422` Average DMIPS over 10 runs | `1610` Average DMIPS over 10 runs | `1.13` |
| `Coremark` | `1431.449` Average iterations/sec over 10 runs | `1414.587` Average iterations/sec over 10 runs | `0.99` |

</details>

ChinYikMing:
The initial design mentioned in [here](https://github.com/sysprog21/rv32emu/issues/310#issuecomment-1953683594) does not fully consider the CSR such as satp CSR needs to be accessed during MMU translation. During implementation, the interface shall be changed to adapt MMU translation.

src/common.h
```diff
@@ -25,6 +25,8 @@
 
 #define ARRAYS_SIZE(arr) (sizeof(arr) / sizeof(arr[0]))
 
+#define MASK(n) (~(1 << n))
```
visitorckw: Perhaps replacing "1" with "1U" to avoid unexpected signed integer overflow would be better?

src/emulate.c
```diff
@@ -604,16 +619,18 @@ static void block_translate(riscv_t *rv, block_t *block)
     block->pc_start = block->pc_end = rv->PC;
 
     rv_insn_t *prev_ir = NULL;
-    rv_insn_t *ir = mpool_calloc(rv->block_ir_mp);
+    rv_insn_t *ir = mpool_alloc(rv->block_ir_mp);
```
visitorckw: Why should mpool_calloc() be replaced with mpool_alloc() and then execute memset()?

src/emulate.c
```diff
@@ -604,16 +619,18 @@ static void block_translate(riscv_t *rv, block_t *block)
...
     block->ir_head = ir;
 
     /* translate the basic block */
     while (true) {
+        memset(ir, 0, sizeof(rv_insn_t));
```
visitorckw: Ditto.

src/emulate.c
```diff
@@ -644,7 +661,7 @@ static void block_translate(riscv_t *rv, block_t *block)
             break;
         }
 
-        ir = mpool_calloc(rv->block_ir_mp);
+        ir = mpool_alloc(rv->block_ir_mp);
```
visitorckw: Ditto.

src/emulate.c
```diff
@@ -51,7 +52,10 @@ extern struct target_ops gdbstub_ops;
     _(breakpoint, 3)       /* Breakpoint */                        \
     _(load_misaligned, 4)  /* Load address misaligned */           \
     _(store_misaligned, 6) /* Store/AMO address misaligned */      \
-    _(ecall_M, 11)         /* Environment call from M-mode */
+    _(ecall_M, 11)         /* Environment call from M-mode */      \
+    _(insn_pgfault, 12)    /* Instruction page fault */            \
```
jserv: Rename to `pagefault_insn`, `pagefault_load`, and `pagefault_store`.

src/emulate.c
```diff
@@ -1178,6 +1195,230 @@ void rv_step(void *arg)
...
+#define MMU_FAULT_CHECK(op, rv, pte, addr, access_bits) \
+    mmu_##op##_fault_check(rv, pte, addr, access_bits)
+#define MMU_FAULT_CHECK_IMPL(op, pgfault)                                      \
+    static bool mmu_##op##_fault_check(riscv_t *rv, uint32_t *pte,             \
+                                       uint32_t addr, uint32_t access_bits)    \
+    {                                                                          \
+        if (!pte && rv->csr_satp) { /* not found */                            \
+            rv_except_##pgfault(rv, addr);                                     \
+            return false;                                                      \
+        } else if (pte &&                                                      \
```
jserv: Replace `else if` with `if` statement.

src/emulate.c
```diff
@@ -1178,6 +1195,230 @@ void rv_step(void *arg)
...
+                   (!(*pte & PTE_V) || (!(*pte & PTE_R) && (*pte & PTE_W)))) { \
+            rv_except_##pgfault(rv, addr);                                     \
+            return false;                                                      \
+        } else if (pte && (!(*pte & PTE_X) && (access_bits & PTE_X))) {        \
```
jserv: Ditto. No `else` is required for the sake of previous return.

src/emulate.c
ChinYikMing: It seems like a rebase problem for #437.

src/emulate.c
ChinYikMing: It seems like a rebase problem for #437.

src/emulate.c
ChinYikMing: It seems like a rebase problem for #437.

jserv:
> This PR is not fully ready to be merged since testing is not yet fully designed. PR earlier to get some feedbacks for further design.

How can we test the MMU specific operations?

ChinYikMing:
> > This PR is not fully ready to be merged since testing is not yet fully designed. PR earlier to get some feedbacks for further design.
> 
> How can we test the MMU specific operations?

The testing idea can be break down to following steps:
1. Creating a simple userspace application and kernel supervisor
2. Starts executing with the simple kernel supervisor. Read/write CSR register to install exception vector table to specific address for traps and root page table for MMU translation.
3. After all CSR stuffs are done, switch to user mode and execute userspace application. At this point, I would like to design some scenario to testing all three types of page fault (instruction, load, store page fault). For every userspace memory access, dump the page table could be beneficial for verification or debugging.

If I am at the wrong path, please correct me. 

It take times to design this testing. So, I would try to support other peripherals emulation at the same time such as PLIC.

jserv:
> 1. Creating a simple userspace application and kernel supervisor
> 2. Starts executing with the simple kernel supervisor. Read/write CSR register to install exception vector table to specific address for traps and root page table for MMU translation.
> 3. After all CSR stuffs are done, switch to user mode and execute userspace application. At this point, I would like to design some scenario to testing all three types of page fault (instruction, load, store page fault). For every userspace memory access, dump the page table could be beneficial for verification or debugging.

The above sound great. I expect the lean and reasonably straightforward approach as following:
* [RISC-V Privilege Levels and System Startup](https://www.openeuler.org/en/blog/zhangxinhao/risc-v/2020-11-28-RISC-V.html)
* [RISC-V Bytes](https://danielmangum.com/categories/risc-v-bytes/)

qwe661234 (approved):

src/riscv.h
```diff
@@ -91,10 +91,61 @@ enum {
 #define MISA_A (1 << ('A' - 'A'))
 #define MISA_F (1 << ('F' - 'A'))
 #define MISA_C (1 << ('C' - 'A'))
+
+#define MSTATUS_SIE_SHIFT 1
```
jserv, line 97: Add comments at least for denoting MSTATUS.

ChinYikMing:
During block emulation, I think  the instructions are executed sequentially until block ends. As such, when a page fault exception is generated during block emulation, the RISC-V core has to jump to the corresponding exception handler. The potential problem is that even  thought the PC could be updated in a exception handler, but the block to emulate is not updated, and this cause the page fault cannot be handled properly. 

I have tested calling `rv_step` after updated the PC to exception handler to resolve the potential problem. It works in gdb but not outside the gdb. Any recommend way for this potential problem?

jserv:
> During block emulation, I think the instructions are executed sequentially until block ends. As such, when a page fault exception is generated during block emulation, the RISC-V core has to jump to the corresponding exception handler. The potential problem is that even thought the PC could be updated in a exception handler, but the block to emulate is not updated, and this cause the page fault cannot be handled properly.
> 
> I have tested calling `rv_step` after updated the PC to exception handler to resolve the potential problem. It works in gdb but not outside the gdb. Any recommend way for this potential problem?

Can you provide a [minimal reproducible example](https://en.wikipedia.org/wiki/Minimal_reproducible_example) so that @qwe661234 can verify if the current block chaining is functioning as expected?

ChinYikMing:
Steps to reproduce the VM test:
1. `make ENABLE_SYSTEM=1`
2. Go to the tests/system directory, run `make`
3. `build/rv32emu tests/system/vm.elf`

Some output would look like this:
```
delegated to supervisor
fault addr: 0x4
new PC: 0x800000b0
next insn addr: 0x4, next insn: 0xfe010113
delegated to supervisor
fault addr: 0x8
new PC: 0x800000b0
next insn addr: 0x8, next insn: 0x112e23
delegated to supervisor
fault addr: 0xc
new PC: 0x800000b0
next insn addr: 0xc, next insn: 0x812c23
delegated to supervisor
fault addr: 0x10
new PC: 0x800000b0
next insn addr: 0x10, next insn: 0x2010413
delegated to supervisor
fault addr: 0x14
new PC: 0x800000b0
next insn addr: 0x14, next insn: 0x6400793
delegated to supervisor
fault addr: 0x18
new PC: 0x800000b0
next insn addr: 0x18, next insn: 0xfef42623
delegated to supervisor
fault addr: 0x1c
new PC: 0x800000b0
next insn addr: 0x1c, next insn: 0xc800793
delegated to supervisor
fault addr: 0x20
new PC: 0x800000b0
next insn addr: 0x20, next insn: 0xfef42423
delegated to supervisor
fault addr: 0x24
new PC: 0x800000b0
next insn addr: 0x24, next insn: 0xfec42703
delegated to supervisor
fault addr: 0x28
new PC: 0x800000b0
next insn addr: 0x28, next insn: 0xfe842783
delegated to supervisor
fault addr: 0x2c
new PC: 0x800000b0
next insn addr: 0x2c, next insn: 0xf707b3
delegated to supervisor
fault addr: 0x30
new PC: 0x800000b0
next insn addr: 0x30, next insn: 0xfef42223
delegated to supervisor
fault addr: 0x34
new PC: 0x800000b0
next insn addr: 0x34, next insn: 0x100513
delegated to supervisor
fault addr: 0x38
new PC: 0x800000b0
next insn addr: 0x38, next insn: 0x80000097
delegated to supervisor
fault addr: 0x3c
new PC: 0x800000b0
next insn addr: 0x3c, next insn: 0x64080e7
next insn addr: 0x8000009c, next insn: 0x5d00893
next insn addr: 0x800000a0, next insn: 0x73
a0: 1
exit syscall called
inferior exit code 1
```

Notice that the user space code starts at address "0x4" and the address "0x800000b0" is the supervisor exception handler entry. 

When instruction fetch fault occurs at address "0x4", the PC is updated to "0x800000b0" but the next instruction is still from address "0x4" and I think it should be from address "0x800000b0". The relevant information is as follows:
```
delegated to supervisor
fault addr: 0x4
new PC: 0x800000b0
next insn addr: 0x4, next insn: 0xfe010113
```

The consequent instruction address ( "0x8", "0xc", ... ) face the same problem.

jserv:
> Steps to reproduce the VM test:
> 1. `make ENABLE_SYSTEM=1`
> 2. Go to the tests/system directory, run `make`
> 3. `build/rv32emu tests/system/vm.elf`

At first glance, it appears that the MMU was not set in `tests/system/vm.c`, and exceptions are delegated to S-mode. Could you show the expected flow for exception handling?

qwe661234:
> ext insn addr: 0x4

Could you provide your `printf` format? I want to map the fields in `riscv_t`.

ChinYikMing:
> > Steps to reproduce the VM test:
> > 
> > 1. `make ENABLE_SYSTEM=1`
> > 2. Go to the tests/system directory, run `make`
> > 3. `build/rv32emu tests/system/vm.elf`
> 
> At first glance, it appears that the MMU was not set in `tests/system/vm.c`, and exceptions are delegated to S-mode. Could you show the expected flow for exception handling?

Please ignore the "vm.c" file. The MMU setup is done in "vm_setup.c". 
```
/* Enable paging */
uintptr_t satp_val =((pte_t) &l1pt >> PG_SHIFT) | SV32_MODE;
write_csr(satp, satp_val);
```

The expected flow for exception handling is that:
1. Search the correspond PTE in page table via `mmu_walk`
2. If the PTE is not found, a correspond page fault exception is generated. In this case, it should be instruction fetch page fault.
3. RISC-V core tends to check if the exception is delegated to S-mode or not. If yes, then set the PC to stvec else set the PC to mtvec. Base mode or Vectored mode depends on the implementation. In this case, the exception is delegated to S-mode so PC is set to stvec. The sepc CSR saves the next instruction to be executed which used by sret instruction to resume the execution seamlessly when page fault is handled. The page fault handler will map 4KiB data during the handling.
4.  The instruction address "0x8", "0xc", ... should not cause instruction page fault since 4KiB data are mapped.

ChinYikMing:
> > ext insn addr: 0x4
> 
> Could you provide your `printf` format? I want to map the fields in `riscv_t`.

Sure. Here it is: `"next insn addr: 0x%x, next insn: 0x%x\n"`.

qwe661234:
> > > ext insn addr: 0x4
> > 
> > 
> > Could you provide your `printf` format? I want to map the fields in `riscv_t`.
> 
> Sure. Here it is: `"next insn addr: 0x%x, next insn: 0x%x\n"`.

Is the variable `rv->PC`?

ChinYikMing:
> > > > ext insn addr: 0x4
> > > 
> > > 
> > > Could you provide your `printf` format? I want to map the fields in `riscv_t`.
> > 
> > 
> > Sure. Here it is: `"next insn addr: 0x%x, next insn: 0x%x\n"`.
> 
> Is the variable `rv->PC`?

`printf("next insn addr: 0x%x, next insn: 0x%x\n", block->pc_end, insn);`. Please see 691 lines in src/emulate.c

qwe661234:
> When instruction fetch fault occurs at address "0x4", the PC is updated to "0x800000b0" but the next instruction is still from address "0x4" and I think it should be from address "0x800000b0". The relevant information is as follows:

I want to ensure that your expected exeuction is 0x4 -> 0x800000b0 -> .... -> 0x8 -> 0x800000b0 -> ..., or others? Because ifetch address is basd on `block->pc_end`, if you don't modify this value, it will fetch next instruction like "0x8", "0xc ..".

ChinYikMing:
> > When instruction fetch fault occurs at address "0x4", the PC is updated to "0x800000b0" but the next instruction is still from address "0x4" and I think it should be from address "0x800000b0". The relevant information is as follows:
> 
> I want to ensure that your expected exeuction is 0x4 -> 0x800000b0 -> .... -> 0x8 -> 0x800000b0 -> ..., or others? Because ifetch address is basd on `block->pc_end`, if you don't modify this value, it will fetch next instruction like "0x8", "0xc ..".

My expected execution flow is 0x4 -> 0x800000b0 -> 0x800000b4 -> 0x800000b8 -> ... -> until reach "sret" and return to 0x4. Shall I modify the value to create the execution flow? In other words, modify the PC in the exception handler is not enough to create the execution flow?

qwe661234:
> Shall I modify the value to create the execution flow? In other words, modify the PC in the exception handler is not enough to create the execution flow?

Yes, you need to modify the block->pc_end, and the next instruction to be fetched will become 0x800000b0. You can see the `block_translate`, it only reads start PC in `block->pc_start = block->pc_end = rv->PC;`. Therefore, the modified `rv->PC` in exception handler does not influence the next instruction in `block_translate`.

ChinYikMing:
> > Shall I modify the value to create the execution flow? In other words, modify the PC in the exception handler is not enough to create the execution flow?
> 
> Yes, you need to modify the block->pc_end, and the next instruction to be fetched will become 0x800000b0. You can see the `block_translate`, it only reads start PC in `block->pc_start = block->pc_end = rv->PC;`. Therefore, the modified `rv->PC` in exception handler does not influence the next instruction in `block_translate`.

Thanks for clarify! But I found that If I try to find_or_translate a block after updating the PC in the exception handler, the corresponding block will be generated and might no need to modify the `block->pc_end`. 

In line 1477:
```diff
if (!pte && rv->csr_satp) { /* not found, then map it in handler */    \
            rv_inter_except_##pgfault(rv, addr);                               \
            printf("fault addr: 0x%x\n", addr); \
            printf("new PC: 0x%x\n", rv->PC); \
+          block_t *block = block_find_or_translate(rv); \
+          assert(block); \
            return true;                                                       \
        }                                                                      \
```

This is the new block generated logs:
```
delegated to supervisor
fault addr: 0x4
new PC: 0x800000b0
next insn addr: 0x800000b0, next insn: 0x14051573
next insn addr: 0x800000b4, next insn: 0x152023
next insn addr: 0x800000b8, next insn: 0x252223
next insn addr: 0x800000bc, next insn: 0x352423
next insn addr: 0x800000c0, next insn: 0x452623
next insn addr: 0x800000c4, next insn: 0x552823
next insn addr: 0x800000c8, next insn: 0x652a23
next insn addr: 0x800000cc, next insn: 0x752c23
next insn addr: 0x800000d0, next insn: 0x852e23
next insn addr: 0x800000d4, next insn: 0x2952023
next insn addr: 0x800000d8, next insn: 0x2b52423
next insn addr: 0x800000dc, next insn: 0x2c52623
next insn addr: 0x800000e0, next insn: 0x2d52823
next insn addr: 0x800000e4, next insn: 0x2e52a23
next insn addr: 0x800000e8, next insn: 0x2f52c23
next insn addr: 0x800000ec, next insn: 0x3052e23
next insn addr: 0x800000f0, next insn: 0x5152023
next insn addr: 0x800000f4, next insn: 0x5252223
next insn addr: 0x800000f8, next insn: 0x5352423
next insn addr: 0x800000fc, next insn: 0x5452623
next insn addr: 0x80000100, next insn: 0x5552823
next insn addr: 0x80000104, next insn: 0x5652a23
next insn addr: 0x80000108, next insn: 0x5752c23
next insn addr: 0x8000010c, next insn: 0x5852e23
next insn addr: 0x80000110, next insn: 0x7952023
next insn addr: 0x80000114, next insn: 0x7a52223
next insn addr: 0x80000118, next insn: 0x7b52423
next insn addr: 0x8000011c, next insn: 0x7c52623
next insn addr: 0x80000120, next insn: 0x7d52823
next insn addr: 0x80000124, next insn: 0x7452a23
next insn addr: 0x80000128, next insn: 0x7f52c23
next insn addr: 0x8000012c, next insn: 0x140512f3
next insn addr: 0x80000130, next insn: 0x2552223
next insn addr: 0x80000134, next insn: 0x100022f3
next insn addr: 0x80000138, next insn: 0x6552e23
next insn addr: 0x8000013c, next insn: 0x141022f3
next insn addr: 0x80000140, next insn: 0x8552023
next insn addr: 0x80000144, next insn: 0x143022f3
next insn addr: 0x80000148, next insn: 0x8552223
next insn addr: 0x8000014c, next insn: 0x142022f3
next insn addr: 0x80000150, next insn: 0x8552423
next insn addr: 0x80000154, next insn: 0x3990006f
next insn addr: 0x4, next insn: 0xfe010113
```

Although the new block is generated, the current emulation block is still the block that starts with 0x4 -> 0x8 -> 0xc -> ... . In other words, the `rv_step` still emulate the old block. In order to execute the newly generated block, shall we make some changes?

qwe661234:
Actually, my idea is modified `block->pc_end` to `0x800000b0`, but we cannot pass block to the exception handler. This issue is difficult because the exception is occured during translation. However, I want to ask why this page fault exception is handled after the execution of 0x04. The page fault occurs when the instruction is fetched. Typically, the operating system handles the page fault first, retrieves the necessary page, and then fetches the instruction. Therefore, my expected execution flow is `0x800000b0 -> 0x800000b4 -> 0x800000b8 -> ... -> 0x4 -> 0x8 -> ...`

ChinYikMing:
> Actually, my idea is modified `block->pc_end` to `0x800000b0`, but we cannot pass block to the exception handler. This issue is difficult because the exception is occured during translation. However, I want to ask why this page fault exception is handled after the execution of 0x04. The page fault occurs when the instruction is fetched. Typically, the operating system handles the page fault first, retrieves the necessary page, and then fetches the instruction. Therefore, my expected execution flow is `0x800000b0 -> 0x800000b4 -> 0x800000b8 -> ... -> 0x4 -> 0x8 -> ...`

If the execution flow like this `0x800000b0 -> 0x800000b4 -> 0x800000b8 -> ... -> 0x4 -> 0x8 -> ...`, there will no instruction fetch page fault generated since the instruction are loaded to RAM. In order to emulation the fetich page fault, I think the flow should be `0x4 -> 0x800000b0 -> 0x800000b4 -> 0x800000b8 -> ... -> page fault handled -> 0x4 -> 0x8 -> ...`. Could you comment this? @jserv

jserv:
> If the execution flow like this `0x800000b0 -> 0x800000b4 -> 0x800000b8 -> ... -> 0x4 -> 0x8 -> ...`, there will no instruction fetch page fault generated since the instruction are loaded to RAM. In order to emulation the fetich page fault, I think the flow should be `0x4 -> 0x800000b0 -> 0x800000b4 -> 0x800000b8 -> ... -> page fault handled -> 0x4 -> 0x8 -> ...

Inside not only the interpreter mode but also the JIT compiler, we have to escape from the chained blocks if an exception is raised. Check QEMU's [Translator Internals](https://www.qemu.org/docs/master/devel/tcg.html) for details.

ChinYikMing:
> > If the execution flow like this `0x800000b0 -> 0x800000b4 -> 0x800000b8 -> ... -> 0x4 -> 0x8 -> ...`, there will no instruction fetch page fault generated since the instruction are loaded to RAM. In order to emulation the fetich page fault, I think the flow should be `0x4 -> 0x800000b0 -> 0x800000b4 -> 0x800000b8 -> ... -> page fault handled -> 0x4 -> 0x8 -> ...
> 
> Inside not only the interpreter mode but also the JIT compiler, we have to escape from the chained blocks if an exception is raised. Check QEMU's [Translator Internals](https://www.qemu.org/docs/master/devel/tcg.html) for details.

Maybe we could apply some mechanism to suspend the current emulating block (might need to save some states) and jump to execute the exception handler block. After that, resume the previous emulating block perhaps addressing this problem. I am wondering whether cause any side effects of this approach.

ChinYikMing:
With the [Translator Internals](https://www.qemu.org/docs/master/devel/tcg.html), I also see some source codes of QEMU,  I realized that some helper functions are required to support handling the exception which are [`raise_mmu_exception`](https://github.com/qemu/qemu/blob/d16cab541ab9217977e2a39abf3d79f914146741/target/riscv/cpu_helper.c#L1171) and [`cpu_loop_exit_restore`](https://github.com/qemu/qemu/blob/d16cab541ab9217977e2a39abf3d79f914146741/include/exec/cpu-common.h#L215). By the way, the latter is defined in cpu-common.h, so I think it should be a generic abstract function for multiple kinds of CPU.

In this [section](https://github.com/qemu/qemu/blob/d16cab541ab9217977e2a39abf3d79f914146741/target/riscv/cpu_helper.c#L1427), we can see that if the MMU generates fault, the former is called then latter is called. We could possibly needs to provide such similar mechanism to support exception handling.

jserv:
> In this [section](https://github.com/qemu/qemu/blob/d16cab541ab9217977e2a39abf3d79f914146741/target/riscv/cpu_helper.c#L1427), we can see that if the MMU generates fault, the former is called then latter is called. We could possibly need to provide such similar mechanism to support exception handling.

Agree. Before merging this MMU work, we should refine the existing exception handling for hardware-aware behavior.

ChinYikMing:
> Steps to reproduce the VM test:
> 
> 1. `make ENABLE_SYSTEM=1`
> 2. Go to the tests/system directory, run `make`
> 3. `build/rv32emu tests/system/vm.elf`

Rerun the steps above would see some outputs like below:
```
...
exit syscall called
inferior exit code 0
```
which means that the userspace application's instruction page fault (caused by address 0x4) has been **handled and mapped 4 KiB for the corresponding PTE**. Therefore, the subsequent instructions would not cause any instruction page fault before exceeding 4 KiB. 

Note: you may objdump the user space application main section via: 
`riscv32-unknown-elf-objdump -d -j .text.main tests/system/vm.elf`

src/common.h
```diff
@@ -25,6 +25,18 @@
 
 #define ARRAYS_SIZE(arr) (sizeof(arr) / sizeof(arr[0]))
 
+#define MASK(n) (~((~0U << (n))))
+
+/* Floor log base 2 */
```
jserv: Reimplement `ilog2` with static inline function, allowing type checking. In addition, there should be a generic C implementation.

src/decode.c
```diff
@@ -813,6 +813,8 @@ static inline bool op_system(rv_insn_t *ir, const uint32_t insn)
     /* decode I-type */
     decode_itype(ir, insn);
 
+    //printf("system decode\n");
```
jserv: Remove the dead code.

src/decode.c
```diff
@@ -868,13 +872,18 @@ static inline bool op_system(rv_insn_t *ir, const uint32_t insn)
      * CSRRSI csr       uimm 110    rd 1110011
      * CSRRCI csr       uimm 111    rd 1110011
      */
+     //0b11000000001111        001    00000 1110011
```
jserv: Use C-style comments when possible.

src/emulate.c
```diff
@@ -19,11 +20,14 @@
 #include "softfloat.h"
 #endif /* RV32_HAS(EXT_F) */
 
+static void intr_excp_handle(riscv_t *rv);
```
jserv: The name looks weird. Can you unify as exception handler?

src/emulate.c
```diff
@@ -374,7 +457,7 @@ enum {
 #if RV32_HAS(GDBSTUB)
 #define RVOP_NO_NEXT(ir) (!ir->next | rv->debug_mode)
 #else
-#define RVOP_NO_NEXT(ir) (!ir->next)
+#define RVOP_NO_NEXT(ir) (!ir->next | rv->is_interrupted)
```
jserv: Shall you update the case for `RV32_HAS(GDBSTUB)`?

src/emulate.c
```diff
@@ -551,6 +634,66 @@ static const void *dispatch_table[] = {
...
+	printf("**********************************\n");
+        while (rv->is_interrupted) {
+	    insn = rv->io.mem_ifetch(rv, rv->PC);
+	    printf("PC: 0x%x, insn: 0x%x\n", rv->PC, insn);
+
+            rv_decode(ir, insn);
+            ir->impl = dispatch_table[ir->opcode];
+	    ir->impl(rv, ir, rv->csr_cycle, rv->PC);
+
+            //rv->compressed = is_compressed(insn);
```
jserv: Remove the dead code.

jserv (changes requested):

tests/system/main.c
```diff
@@ -0,0 +1,10 @@
+#define SECTION_TEXT_MAIN __attribute__((section(".text.main")))
+#define SECTION_DATA_MAIN __attribute__((section(".data.main")))
+#define SECTION_BSS_MAIN __attribute__((section(".bss.main")))
+
+int SECTION_TEXT_MAIN main(){
```
jserv: Indent with the consistent style.

Makefile
```diff
@@ -11,6 +11,12 @@ CFLAGS = -std=gnu99 -O2 -Wall -Wextra
 CFLAGS += -Wno-unused-label
 CFLAGS += -include src/common.h
 
+ENABLE_SYSTEM ?= 0
+ifeq ($(call has, SYSTEM), 1)
+#$(info SYSTEM here)
+CFLAGS += -DSYSTEM=1
```
jserv: Use `$(call set-feature, SYSTEM)` instead.

src/emulate.c
```diff
@@ -1178,16 +1363,360 @@ void rv_step(void *arg)
 #endif
 }
 
+void plic_update_interrupts(riscv_t *rv)
```
jserv: Move PLIC specific code out of file `emulate.c`.

src/main.c
```diff
@@ -207,6 +207,7 @@ int main(int argc, char **args)
 #endif
     run_flag |= opt_prof_data << 2;
 
+#ifndef SYSTEM
```
jserv: Use `RV32_HAS` macro instead.

tests/system/vm.c
```diff
@@ -0,0 +1,65 @@
...
+int main(){
+	uint8_t *page_addr = aligned_alloc(PG_SIZE, NR_PG * PG_SIZE);
+	uint32_t *root_page_table = aligned_alloc(PTE_SIZE, PTE_PER_TABLE);
+	/* Disable MMU */
+	const uint32_t satp_val = ((uint32_t)root_page_table >> PG_SHIFT) | SATP_BARE_MODE;
+	write_csr(satp, satp_val);
+	const uint32_t satp_read = read_csr(satp);
+	assert(satp_read == 0);
+
+	//for(uint32_t i = 0; i < NR_PG * PG_SIZE; i += PG_SIZE){
```
jserv: Don't leave the dead code.

tests/system/vm_setup.c
```diff
@@ -0,0 +1,414 @@
...
+#define PTES_PER_PT (1U << 10)
+#define PTE_PPN_SHIFT 10
+pte_t pt[NPT][PTES_PER_PT] SECTION_BSS_VMSETUP __attribute__((aligned(PG_SIZE)));
+
+#define MASK(n) (~((~0U << (n))))
+#define pa2uva(x) ((pte_t) (x) - (pte_t) (_start))
+#define pa2kva(x) ((pte_t) (x) - (pte_t) (_start) - MEGA_PG)
+#define uva2kva(x) ((pte_t) (x) - MEGA_PG)
+
+uint32_t SECTION_TEXT_VMSETUP syscall(uint32_t syscall_nr, uint32_t arg)
```
jserv: `SECTION_TEXT_VMSETUP` should appear prior to `uint32_t`.

src/syscall.c
```diff
@@ -23,23 +23,26 @@
...
+    _(lseek,                62)            \
+    _(read,                 63)            \
+    _(write,                64)            \
+    _(fstat,                80)            \
+    _(exit,                 93)            \
+    _(gettimeofday,         169)           \
+    _(brk,                  214)           \
+    _(clock_gettime,        403)           \
+    _(open,                 1024)          \
+    _(sbi_base,             0x10)          \
```
jserv: The SBI calls should be built conditionally. That is, specify `IIF(RV32_HAS(SYSTEM))`.

src/syscall.c
```diff
@@ -134,6 +137,9 @@ static void syscall_exit(riscv_t *rv)
      * the application decides the usage of exit code
      */
     rv_halt(rv);
+    printf("a0: %d\n", rv_get_reg(rv, rv_reg_a0));
```
jserv: Don't dump such information.

src/syscall.c
```diff
@@ -134,6 +137,9 @@ static void syscall_exit(riscv_t *rv)
...
+
+    printf("exit syscall called\n");
```
jserv: Ditto.

tests/system/Makefile
```diff
@@ -0,0 +1,27 @@
+PREFIX ?= riscv32-unknown-elf-
+ARCH = -march=rv32imaf
```
jserv: Don't specify "f" extension. Minimize when possible.

src/riscv.h
```diff
@@ -228,12 +383,20 @@ typedef struct {
     char *elf_program;
 } vm_user_t;
 
+typedef struct {
+    char *elf_program;
```
jserv, line 520: It is weird that bot `vm_user_t` and `vm_system_t` have only a member `elf_program`. If so, why another structure?

src/emulate.c
```diff
@@ -76,49 +89,84 @@ static void rv_exception_default_handler(riscv_t *rv)
...
+        /* VECTORED: Asynchronous interrupts set PC to base + 4 * code */     \
+        case 1:                                                               \
+            /* MSB of code is used to indicate whether the trap is interrupt  \
+             * or exception, so it is not considered as the 'real' code */    \
+            rv->PC = base + 4 * (code & MASK(31));                            \
+            break;                                                            \
+        }                                                                     \
+	/* block escaping for interrupt or exception handling */ \
+        if(rv->is_interrupted){ \
+	    assert(setjmp(env) == 0); \
```
jserv: If `NDEBUG` is set, `setjmp` will not be called.

src/emulate.c
jserv: For minimal changes, the mixture of interrupt and exception in a single structure is really confusing. RISC-V does have its exception definitions as well as the counterpart for interrupts.

src/io.c
```diff
@@ -15,6 +15,11 @@
 
 #include "io.h"
 
+plic_t *plic_new()
```
jserv: Move PLIC specific code to `plic.c`.

src/decode.c
```diff
@@ -891,8 +900,11 @@ static inline bool op_system(rv_insn_t *ir, const uint32_t insn)
     default: /* illegal instruction */
         return false;
     }
-    if (!csr_is_writable(ir->imm) && ir->rs1 != rv_reg_zero)
+    if (!csr_is_writable(ir->imm) && ir->rs1 != rv_reg_zero){
+	//printf("writable check failed\n");
         return false;
+    }
+    //printf("return true\n");
```
jserv: Remove the dead code.
Simply rewrite as following:
```c
    return csr_is_writable(ir->imm) || (ir->rs1 != rv_reg_zero);
```

src/emulate.c
jserv: By default, M-mode is the first receiver of any interrupts or exceptions (i.e., traps) in the system. Instead of weird "intr_excp_handle," we can call it "trap_handler." The latter is used in semu.

src/riscv.h
ChinYikMing: Since I am loading a single ELF program for MMU testing and I should put a FIXME there to clarify. Eventually, we have to put kernel image, device tree blob, etc in the structure.

src/emulate.c
ChinYikMing: I'll make it into an if conditional statement.

src/emulate.c
ChinYikMing: > For minimal changes, the mixture of interrupt and exception in a single structure is really confusing. RISC-V does have its exception definitions as well as the counterpart for interrupts.

Yes, we can determine a trap is an exception or an interrupt via the MSB of the trap code.

tests/system/Makefile
ChinYikMing: I would refine it as `-march=rv32izicsr`

src/emulate.c
ChinYikMing: > Shall you update the case for `RV32_HAS(GDBSTUB)`?

In that case, we have to introduce a new variable to indicate the state whether in block emulation or exception/interrupt.

ChinYikMing: How about calling `is_trapped`?

ChinYikMing:
> > Steps to reproduce the VM test:
> > 
> > 1. `make ENABLE_SYSTEM=1`
> > 2. Go to the tests/system directory, run `make`
> > 3. `build/rv32emu tests/system/vm.elf`
> 
> Rerun the steps above would see some outputs like below:
> 
> ```
> ...
> exit syscall called
> inferior exit code 0
> ```
> 
> which means that the userspace application's instruction page fault (caused by address 0x4) has been **handled and mapped 4 KiB for the corresponding PTE**. Therefore, the subsequent instructions would not cause any instruction page fault before exceeding 4 KiB.
> 
> Note: you may objdump the user space application main section via: `riscv32-unknown-elf-objdump -d -j .text.main tests/system/vm.elf`

Rerun the steps and will see the output of MMU test suite.
```
INSTRUCTION FETCH PAGE FAULT TEST PASSED!
LOAD PAGE FAULT TEST PASSED!
STORE PAGE FAULT TEST PASSED!
inferior exit code 0
```

src/emulate.c
```diff
@@ -76,49 +93,92 @@ static void rv_exception_default_handler(riscv_t *rv)
...
+            rv->csr_mepc = rv->PC;                                            \
+            rv->csr_mtval = mtval;                                            \
+            rv->csr_mcause = code;                                            \
+            rv->csr_mstatus |= MSTATUS_MPP; /* set privilege mode */          \
+            if (!rv->csr_mtvec) { /* in case CSR is not configured */         \
+                rv_interrupt_exception_default_handler(rv);                   \
+                return;                                                       \
+            }                                                                 \
+        }                                                                     \
+        /* FIXME: ecall is not emulated in kernel so temporarily add this */  \
```
jserv: Mor descriptions here. You should explicitly mention Linux kernel.

src/emulate.c
```diff
@@ -76,49 +93,92 @@ static void rv_exception_default_handler(riscv_t *rv)
...
+            rv->PC = base + 4 * (code & MASK(31));                            \
+            break;                                                            \
+        }                                                                     \
+        /* block escaping for interrupt or exception handling */              \
+        if (rv->is_interrupted) {                                             \
+            if (setjmp(env) == 0) {                                           \
+                trap_handler(rv, code & ~MASK(31));                           \
+            } else {                                                          \
+                fprintf(stderr, "setjmp failed");                             \
+                exit(1);                                                      \
```
jserv: Avoid invoking `exit`, otherwise it is a bit difficult to track the internal state of this virtual machine.

src/emulate.c
```diff
@@ -43,24 +51,33 @@ extern struct target_ops gdbstub_ops;
...
-#define RV_EXCEPTION_LIST                                          \
-    IIF(RV32_HAS(EXT_C))(,                                         \
-        _(insn_misaligned, 0) /* Instruction address misaligned */ \
-    )                                                              \
-    _(illegal_insn, 2)     /* Illegal instruction */               \
-    _(breakpoint, 3)       /* Breakpoint */                        \
-    _(load_misaligned, 4)  /* Load address misaligned */           \
-    _(store_misaligned, 6) /* Store/AMO address misaligned */      \
-    _(ecall_M, 11)         /* Environment call from M-mode */
+#define RV_INTER_EXCEPTION_LIST                                                        \
```
jserv: It is TRAP.

src/emulate.c
```diff
@@ -76,49 +93,92 @@ static void rv_exception_default_handler(riscv_t *rv)
...
-        case 0: /* DIRECT: All exceptions set PC to base */                  \
-            rv->PC = base;                                                   \
-            break;                                                           \
-        /* VECTORED: Asynchronous interrupts set PC to base + 4 * code */    \
-        case 1:                                                              \
-            rv->PC = base + 4 * code;                                        \
-            break;                                                           \
-        }                                                                    \
+static jmp_buf env;
+#define INTER_EXCEPTION_HANDLER_IMPL(type, code)                              \
```
jserv: It is TRAP.

src/emulate.c
jserv: > How about calling `is_trapped`?

It sounds good.

tests/system/main.c
```diff
@@ -0,0 +1,98 @@
+#define SECTION_TEXT_MAIN __attribute__((section(".text.main")))
+#define SECTION_DATA_MAIN __attribute__((section(".data.main")))
+#define SECTION_BSS_MAIN __attribute__((section(".bss.main")))
+
+#define _printf(ptr, length)                    \
```
jserv: Don't call it `_printf` since it has nothing to format string.
Instead, call it `printstr`.

src/decode.c
```diff
@@ -903,9 +903,8 @@ static inline bool op_system(rv_insn_t *ir, const uint32_t insn)
     default: /* illegal instruction */
         return false;
     }
-    if (!csr_is_writable(ir->imm) && ir->rs1 != rv_reg_zero)
-        return false;
-    return true;
+
+    return csr_is_writable(ir->imm) || (ir->rs1 != rv_reg_zero);
```
jserv: This should be `csr_is_writable(ir->imm) || (ir->rs1 == rv_reg_zero)`

jserv:
RISC-V Architecture Test complains:
> ERROR | rv32emu/tests/riscv-arch-test/riscv-test-suite/rv32i_m/privilege/src/ecall.S : -                                        : Failed

src/emulate.c
```diff
@@ -41,31 +43,49 @@ extern struct target_ops gdbstub_ops;
...
 }
 
+typedef enum {
+    EXCEPTION,
+    INTERRUPT,
+} trap_t;
+
+#if RV32_HAS(SYSTEM)
+static void trap_handler(riscv_t *rv, trap_t type UNUSED);
+#else /* should not be called in non-SYSTEM mode as default trap handler is \
```
jserv: Move the comment to the next line of `#else`.

src/emulate.c
```diff
@@ -253,6 +353,7 @@ static uint32_t csr_csrrc(riscv_t *rv, uint32_t csr, uint32_t val)
     if (!c)
         return 0;
 
+
```
jserv: Don't do this.

src/emulate.c
```diff
@@ -584,21 +685,25 @@ static void block_translate(riscv_t *rv, block_t *block)
...
     while (true) {
+        memset(ir, 0, sizeof(rv_insn_t));
+
         if (prev_ir)
             prev_ir->next = ir;
 
         /* fetch the next instruction */
-        const uint32_t insn = rv->io.mem_ifetch(block->pc_end);
+        const uint32_t insn = rv->io.mem_ifetch(rv, block->pc_end);
+        // printf("next insn addr: 0x%x, next insn: 0x%x, PC: 0x%x\n",
```
jserv: Remove the dead code.

src/main.c
```diff
@@ -207,6 +207,23 @@ int main(int argc, char **args)
 #endif
     run_flag |= opt_prof_data << 2;
 
+#if RV32_HAS(SYSTEM)
```
jserv: Move `#if RV32_HAS(SYSTEM)` before `assert`. Minimize the necessary changes.

src/emulate.c
```diff
@@ -41,31 +43,50 @@ extern struct target_ops gdbstub_ops;
...
+static void rv_trap_default_handler(riscv_t *rv)
 {
     rv->csr_mepc += rv->compressed ? 2 : 4;
     rv->PC = rv->csr_mepc; /* mret */
 }
 
+typedef enum {
+    EXCEPTION,
+    INTERRUPT,
+} trap_t;
```
jserv: Does `trap_t` make sense?

ChinYikMing: Actually, it is not used currently. It was introduced for emulating the priority of interrupt when it is supported in the future for example.

ChinYikMing: Shall we drop or preserve it for now?

jserv: I would expect the functional and minimal changes.

ChinYikMing: I'll eliminate it then.

jserv (changes requested):
Update CI pipeline to include system emulation tests.

src/emulate.c
```diff
@@ -76,49 +92,86 @@ static void rv_exception_default_handler(riscv_t *rv)
  * When a hardware breakpoint is triggered or an exception like address
  * misalignment, access fault, or page fault occurs during an instruction
  * fetch, load, or store operation, mtval is updated with the virtual address
- * that caused the fault. In the case of an illegal instruction trap, mtval
- * might be updated with the first XLEN or ILEN bits of the offending
- * instruction. For all other traps, mtval is simply set to zero. However,
- * it is worth noting that a future standard could redefine how mtval is
- * handled for different types of traps.
+ * that caused the fault. In the case of an illegal instruction trap, mtval *
```
jserv: Don't change the existing comments unless you have strong reasons.

src/rv32_template.c
```diff
@@ -158,8 +159,10 @@ RVOP(
     jal,
     {
         const uint32_t pc = PC;
+
```
jserv: Don't do that. Minimize the necessary changes.

src/rv32_template.c
```diff
@@ -257,11 +264,14 @@ RVOP(
     jalr,
     {
         const uint32_t pc = PC;
+
```
jserv: Ditto. Minimize the necessary changes.

jserv:
Quote from the comment of [esp32-running-linux](https://github.com/nodestark/esp32-running-linux/blob/main/riscv-esp32s3r8/main/riscv-esp32s3r8.c)
> The TLB is a high-level cache of the page table, which stores the most recently used translations and makes them quickly and efficiently accessible.
> Instead of accessing the page table in main memory, the processor first checks the TLB to see if the translation of the virtual address is already stored there.
> If it is, the translation is used directly, without the need to access the page table in main memory.
> This reduces memory access latency and increases system performance.

 [esp32-running-linux](https://github.com/nodestark/esp32-running-linux) provides a minimalist implementation of RV32 + MMU capable of running Linux kernel, and its MMU/TLB is worth checking.

src/emulate.c
ChinYikMing: Shall we change `mtval` parameter identifier to `tval` since we handle both supervisor tval and machine tval in the same function?

jserv: > Shall we change `mtval` parameter identifier to `tval` since we handle both supervisor tval and machine tval in the same function?

I think so. Make it informative.

src/emulate.c
```diff
@@ -4,6 +4,7 @@
  */
 
 #include <assert.h>
+#include <setjmp.h>
```
jserv: Is `setjmp` necessary?

ChinYikMing: > Is `setjmp` necessary?

It should be removed.

tests/system/Makefile
```diff
@@ -0,0 +1,27 @@
+PREFIX ?= riscv32-unknown-elf-
+ARCH = -march=rv32izicsr
```
vacantron: Should it be `rv32i_zicsr` instead?

ChinYikMing: > Should it be `rv32i_zicsr` instead?

Yes, given that we're using the xPack toolchain in the CI environment, the `rv32i_zicsr` flag will be the correct flag once the test suite is integrated.

ChinYikMing:
> Update CI pipeline to include system emulation tests.

Done.

src/emulate.c
```diff
@@ -1122,24 +1132,8 @@ void rv_step(void *arg)
...
-        assert(insn);
-
-        rv_decode(ir, insn);
-        ir->impl = dispatch_table[ir->opcode];
-        rv->compressed = is_compressed(insn);
-        ir->impl(rv, ir, rv->csr_cycle, rv->PC);
-    }
-}
-#endif
+#include "system.c"
```
jserv: Why did you include this file instead of making it as a new compilation unit?

src/rv32_template.c
```diff
@@ -155,8 +155,10 @@ RVOP(
     jal,
     {
         const uint32_t pc = PC;
+
```
jserv: Don't do this. Minimize the necessary changes.
Indent with clang-format-18.

tests/system/pgfault/vm_setup.c
```diff
@@ -0,0 +1,420 @@
...
+ * 1. https://github.com/sifive/example-vm-test
+ * 2. https://github.com/yutongshen/RISC-V-Simulator
+ */
+
+#include <stdarg.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#if !defined(__GNUC__) || !defined(__riscv) || 32 != __riscv_xlen
+#error "32-bit RISC-V GNU toolchain is required"
```
jserv: Should be "GNU Toolchain for 32-bit RISC-V"

jserv:

src/emulate.c
ChinYikMing: > Why did you include this file instead of making it as a new compilation unit?

Previously, the implementation are compiled in the same unit of `src/emulate.c`. So, simply separate into new file. 

However, a new compilation unit for `system.[ch]` looks good.

src/rv32_template.c
ChinYikMing: > Indent with clang-format-18.

clang-format-18 above will also work?

jserv: I encountered incompatibility with clang-format-20.

src/emulate.c
ChinYikMing: When attempting to separate into system.[ch], it appears to be non-trivial because `MMU_FAULT_CHECK` depends on the `rv_trap_xx_pgfault` handler, which is a static function within `src/emulate.c`. Also, `trap_handler` depends on `dispatch_table` declared as static within `src/emulate.c`.

The original `#include system.c` is trivial. Or modify `TRAP_HANDLER_IMPL` and `dispatch_table` to non-static?

jserv: > The original `#include system.c` is trivial. Or modify `TRAP_HANDLER_IMPL` and `dispatch_table` to non-static?

It sounds good. However, the file `src/emulate.c` will eventually need to be split into several files to facilitate modularization.

src/emulate.c
```diff
@@ -290,7 +291,16 @@ static uint32_t csr_csrrw(riscv_t *rv, uint32_t csr, uint32_t val)
         out &= FFLAG_MASK;
 #endif
 
-    *c = val;
+    if (c == &rv->csr_satp) {
```
jserv: Add comments to address SATP.

src/emulate.c
ChinYikMing: > > The original `#include system.c` is trivial. Or modify `TRAP_HANDLER_IMPL` and `dispatch_table` to non-static?
> 
> It sounds good. However, the file `src/emulate.c` will eventually need to be split into several files to facilitate modularization.

I would temporarily modify them to non-static for this PR.

src/rv32_template.c
ChinYikMing: > I encountered incompatibility with clang-format-20.

My system is clang-format-19, hope it is fine.

src/emulate.c
ChinYikMing: > Add comments to address SATP.

I think this check is a bit redundant (I will remove them) because the ppn of root page table (`MASK(22)`)  can be simply obtained during the `mmu_walk`. In addition, the ASID bit could be used to implement TLB. So, store the whole 32-bit value is harmless.

src/system.c
```diff
@@ -0,0 +1,309 @@
+/*
+ * rv32emu is freely redistributable under the MIT License. See the file
+ * "LICENSE" for information on usage and redistribution of this file.
+ */
+
+#include "system.h"
+#include <assert.h>
```
jserv: Move `<assert.h>` prior to the local headers. That is.
```c
#include <assert.h>

#include "mpool.h"
#include "system.h"
```

src/system.c
```diff
@@ -0,0 +1,309 @@
...
+
+void trap_handler(riscv_t *rv)
+{
+    rv_insn_t *ir = mpool_alloc(rv->block_ir_mp);
+    assert(ir);
+
+    /* set to false by sret implementation */
+    uint32_t insn;
+    while (rv->is_trapped && !rv_has_halted(rv)) {
+        insn = rv->io.mem_ifetch(rv, rv->PC);
```
jserv: Declare `insn` here.

jserv:
Is it better to rename from `tests/system/pgfault` to `tests/system/mmu` since more tests can be integrated?

src/riscv.c
```diff
@@ -356,7 +372,11 @@ void rv_run(riscv_t *rv)
     assert(rv);
 
     vm_attr_t *attr = PRIV(rv);
+#if RV32_HAS(SYSTEM)
+    assert(attr && attr->data.system && attr->data.system->elf_program);
+#else
     assert(attr && attr->data.user && attr->data.user->elf_program);
+#endif
```
jserv: Rewrite as following:
```c
    assert(attr &&
#if RV32_HAS(SYSTEM)
        attr->data.system && attr->data.system->elf_program
#else
        attr->data.user && attr->data.user->elf_program
#endif
    );
```

src/rv32_template.c
```diff
@@ -155,8 +155,10 @@ RVOP(
...
         /* Jump */
         PC += ir->imm;
+
```
jserv: Don't change this.

src/system.c
```diff
@@ -0,0 +1,309 @@
...
+            ppn = (*pte >> (RV_PG_SHIFT - 2));
+            page_table = PAGE_TABLE(ppn);
+            if (!page_table)
+                return NULL;
+            break;
+        case 0b0011:
+        case 0b0111:
+        case 0b1001:
+        case 0b1011:
+        case 0b1111:
```
jserv: Introduce `enum` for these conditions.

ChinYikMing:
> Is it better to rename from `tests/system/pgfault` to `tests/system/mmu` since more tests can be integrated?

Agree.

src/emulate.c
```diff
@@ -103,7 +95,7 @@ static void trap_handler(riscv_t *rv);
  * identifier called tval, as both are handled by TRAP_HANDLER_IMPL.
  */
 #define TRAP_HANDLER_IMPL(type, code)                                         \
-    static void rv_trap_##type(riscv_t *rv, uint32_t tval)                    \
+    void rv_trap_##type(riscv_t *rv, uint32_t tval)                           \
```
jserv: It is a pity that the helper functions can not be declared as `static`. Can you find a way for that?

src/emulate.c
```diff
@@ -589,7 +581,7 @@ static bool do_fuse5(riscv_t *rv,
 }
 
 /* clang-format off */
-static const void *dispatch_table[] = {
+const void *dispatch_table[] = {
```
jserv: Ditto. The dispatch table should be hidden.

ChinYikMing: > Ditto. The dispatch table should be hidden.

How about following the approach used for the `ebreak_handler`? We could register the `trap_handler` with the I/O interface during RISC-V core initialization. This way, the dispatch table can remain hidden.

src/emulate.c
ChinYikMing: > It is a pity that the helper functions can not be declared as `static`. Can you find a way for that?

We can register the `trap_handler` with the I/O interface during the initialization of the RISC-V core, keeping the dispatch table concealed. When a page fault occurs, the `trap_handler` is invoked, which in turn calls `_trap_handler`, and then `_trap_handler` calls `__trap_handler`. This `__trap_handler` may correspond to a helper function defined in emulate.c. The `__trap_handler` would represent the original implementation of trap_handler, while `_trap_handler` may need to inspect the `scause` CSR to determine which `TRAP_HANDLER_IMPL` to invoke. This approach ensures that the specific `TRAP_HANDLER_IMPL` remains hidden.

ChinYikMing: In detail,  `trap_handler` callback (need to set corresponding scause CSR, in IO interface) 
---(call)-----> `_trap_handler`(check scause CSR) ---(dispatch)-----> `TRAP_HANDLER_IMPL` (depends on scause CSR, static function in emulate.c) 
---(call)-----> `__trap_handler`(actual trap handler, static function in emulate.c)

src/emulate.c
jserv: > How about following the approach used for the `ebreak_handler`? We could register the `trap_handler` with the I/O interface during RISC-V core initialization. This way, the dispatch table can remain hidden.

It sounds good. Let's enforce the consistent schema.

ChinYikMing:
Now, the trap handling flow is as following:

`trap_handler` callback (requires setting the appropriate `scause` CSR) ---(invokes)---> `_trap_handler `(verifies scause CSR) ---(dispatches to)---> `TRAP_HANDLER_IMPL` (based on `scause` CSR, implemented as a static function in emulate.c) ---(calls)---> `__trap_handler` (actual trap handler, also a static function in emulate.c)

vacantron:
Since `m/scause`, `m/stval` has been set in `SET_CAUSE_AND_TVAL_THEN_TRAP()`, setting `tval` in https://github.com/sysprog21/rv32emu/blob/164da62b64fa050569b22910ea207525884802ba/src/emulate.c#L106 is redundant.

src/emulate.c
```diff
@@ -99,11 +89,9 @@ static void trap_handler(riscv_t *rv);
  * it is worth noting that a future standard could redefine how m/stval is
  * handled for different types of traps.
  *
- * For simplicity and clarity, abstracting stval and mtval into a single
- * identifier called tval, as both are handled by TRAP_HANDLER_IMPL.
  */
-#define TRAP_HANDLER_IMPL(type, code)                                         \
-    static void rv_trap_##type(riscv_t *rv, uint32_t tval)                    \
+#define TRAP_HANDLER_IMPL(type)                                               \
+    static void rv_trap_##type(riscv_t *rv)                                   \
```
vacantron: We do not need different handler for different trap type anymore. The information for trapping is stored in `m/scause`, `m/stval`.

ChinYikMing: > We do not need different handler for different trap type anymore. The information for trapping is stored in `m/scause`, `m/stval`.

Could you elaborate more on this?

vacantron: > Could you elaborate more on this?

I meant we could simplify some handlers in `_trap_handler()` (e.g. illegal_insn, breakpoint, misalligned). They have the same routine of reading `m/stval` and `m/scause` to identify the type of traps.

ChinYikMing: > > Could you elaborate more on this?
> 
> I meant we could simplify some handlers in `_trap_handler()` (e.g. illegal_insn, breakpoint, misalligned). They have the same routine of reading `m/stval` and `m/scause` to identify the type of traps.

It make sense.

vacantron (approved):

jserv:
@vacantron, after merging this pull request, let's prepare the prebuilt MMU test executable file.

jserv:
Any reason to expose `mmu_ifetch`, `mmu_read_w`, `mmu_read_s`, etc.? I marked them as static functions, and the SYSTEM configuration still builds.

ChinYikMing:
> Any reason to expose `mmu_ifetch`, `mmu_read_w`, `mmu_read_s`, etc.?

They should be `static` in system.c and only accessible through `riscv_io_t` interface. In this way, the declaration of them shall be removed from the system.h header file.

>  I marked them as static functions, and the SYSTEM configuration still builds.

Do you mark them in system.c?

jserv:
> > Any reason to expose `mmu_ifetch`, `mmu_read_w`, `mmu_read_s`, etc.?
> They should be `static` in system.c and only accessible through `riscv_io_t` interface. In this way, the declaration of them shall be removed from the system.h header file.

Go ahead for avoid disclosing the internal functions.

ChinYikMing:
> > > Any reason to expose `mmu_ifetch`, `mmu_read_w`, `mmu_read_s`, etc.?
> > > They should be `static` in system.c and only accessible through `riscv_io_t` interface. In this way, the declaration of them shall be removed from the system.h header file.
> 
> Go ahead for avoid disclosing the internal functions.

'system.h' is removed due to too less LoC, it is not worth to add a new header file. Thus, I moved the original content to 'riscv.h' instead.

jserv:

ChinYikMing:
Thanks all for reviewing. Ready to bring up Linux kernel on next PR!

PR #441, opened 2024-05-17
Optimize cmp function

huaxinliao:
The number of instructions has decreased, and the number of memory accesses has been reduced.

### DIFF
--- a/src/breakpoint.c
+++ b/src/breakpoint.c
@@ -11,8 +11,8 @@
 
 static inline int cmp(const void *arg0, const void *arg1)
 {
-    riscv_word_t *a = (riscv_word_t *) arg0, *b = (riscv_word_t *) arg1;
-    return (*a < *b) ? _CMP_LESS : (*a > *b) ? _CMP_GREATER : _CMP_EQUAL;
+    riscv_word_t a = *(riscv_word_t *) arg0, b = *(riscv_word_t *) arg1;
+    return (a < b) ? _CMP_LESS : (a > b) ? _CMP_GREATER : _CMP_EQUAL;
 }
 
 breakpoint_map_t breakpoint_map_new()


huaxinliao:
I analyzed the cmp function by converting it into x86 instructions, as https://hackmd.io/ucXfO5ixSTePjNX_KFR_Vg?view#Optimize-cmp-function-441.

visitorckw:
I would be surprised if the compiler does not optimize this.

huaxinliao:
> I would be surprised if the compiler does not optimize this.

You can conduct experiments to verify it.

visitorckw:
Out of curiosity, I used ```objdump -D``` on my Ubuntu x86-64 machine to examine the object code generated by gcc with ```ENABLE_GDBSTUB=1``` and ```ENABLE_LTO=0```. From what I can see, there is no difference in the generated code?

visitorckw:
I believe the difference you observed is due to the O2 optimization not being enabled.

huaxinliao:
How to make the O2 flag be enabled, @visitorckw.

visitorckw:
> How to make the O2 flag be enabled, @visitorckw.

You can check the CFLAGS used for compiling rv32emu in the Makefile. If you are compiling a single C file with gcc, you could add the -O2 option.

huaxinliao:
> > How to make the O2 flag be enabled, @visitorckw.
> 
> You can check the CFLAGS used for compiling rv32emu in the Makefile. If you are compiling a single C file with gcc, you could add the -O2 option.

With the addition of the -O2 flag, the generated code remains consistent. Thank you, @visitorckw .

jserv:
I tend to close this pull request since modern compiler optimizer can always generate good code. There is no need to apply such proposed change.

PR #360, opened 2024-02-26
Fix the microsecond-to-nanosecond conversion in rv_clock_gettime()

alanjian85:
A microsecond is 1000 times bigger than a nanosecond (10<sup>-6</sup> vs 10<sup>-9</sup> seconds). This commit corrects the conversion of microseconds to nanoseconds in `rv_clock_gettime()` by multiplying `tv_usec` by 1000 instead of dividing it.

PR #445, opened 2024-05-21
Modify get_time_info() to return seconds and nanoseconds

visitorckw:
Modify get_time_info() to return seconds and nanoseconds to get higher time resolution. Additionally, remove the outdated TODO comments as it was solved in newlib v4.4.0 [1].

Link: https://sourceware.org/git/?p=newlib-cygwin.git;a=commit;h=5f15d7c5817b07a6b18cbab17342c95cb7b42be4 [1]

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -59,11 +59,6 @@ void rv_gettimeofday(struct timeval *tv)
     tv->tv_usec = tv_usec;
 }
 
-/* TODO: Clarify newlib's handling of time units.
- * It appears that newlib is using millisecond resolution for time manipulation,
- * while clock_gettime expects nanoseconds in the timespec struct.
- * Further investigation are needed.
- */
 void rv_clock_gettime(struct timespec *tp)
 {
     int32_t tv_sec, tv_usec;
--- a/src/utils.c
+++ b/src/utils.c
@@ -26,13 +26,13 @@
 
 #define MAX_PATH_LEN 1024
 
-static void get_time_info(int32_t *tv_sec, int32_t *tv_usec)
+static void get_time_info(int32_t *tv_sec, int32_t *tv_nsec)
 {
 #if defined(HAVE_POSIX_TIMER)
     struct timespec t;
     clock_gettime(CLOCKID, &t);
     *tv_sec = t.tv_sec;
-    *tv_usec = t.tv_nsec / 1000;
+    *tv_nsec = t.tv_nsec;
 #elif defined(HAVE_MACH_TIMER)
     static mach_timebase_info_data_t info;
     /* If it is the first time running, obtain the timebase. Using denom == 0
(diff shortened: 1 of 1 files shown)


jserv:

vacantron:
@visitorckw, I found the results of some tests like `dhrystone` and `coremark` being weird after applying the changes of timespec. Should we also update the binary to reflect the change of newlib?

visitorckw:
> @visitorckw, I found the results of some tests like `drystone` and `coremark` being weird after applying the changes of timespec. Should we also update the binary to reflect the change of newlib?

Absolutely. When I discovered the errors in newlib libgloss, @jserv  asked me to correct the time-related system calls in rv32emu and also update all the binaries. However, I remember that when I applied my newlib patch and compiled the riscv gnu toolchain to generate the test binaries for rv32emu, I encountered an ```undefined reference to `nanosleep'``` error. Since I couldn't figure out the cause of this error at that time, I haven't made this update.

visitorckw:
> @visitorckw, I found the results of some tests like `drystone` and `coremark` being weird after applying the changes of timespec. Should we also update the binary to reflect the change of newlib?

Additionally, I believe the weird results you observed are due to the change in time units introduced by PR #360 , not this PR. This PR merely optimized the precision without changing the time units.

vacantron:
> Additionally, I believe the weird results you observed are due to the change in time units introduced by PR #360 , not this PR. This PR merely optimized the precision without changing the time units.

Yes, I have found that I replied the wrong one. I sorted the PRs with created time and not realized it. 🥲

visitorckw:
> Yes, I have found that I replied the wrong one. I sorted the PRs with created time and not realized it. 🥲

That's fine. Replying to either one is fine.
I just tried using the updated riscv gnu toolchain to compile nyancat.c and still encountered the following error:
```
nyancat.c: In function 'main':
nyancat.c:940:9: warning: implicit declaration of function 'nanosleep' [-Wimplicit-function-declaration]
  940 |         nanosleep(&ts, &ts);
      |         ^~~~~~~~~
in function `main':
nyancat.c:(.text+0x34c): undefined reference to `nanosleep'
collect2: error: ld returned 1 exit status
```
Do you have any idea about this error?

visitorckw:
> That's fine. Replying to either one is fine. I just tried using the updated riscv gnu toolchain to compile nyancat.c and still encountered the following error:
> 
> ```
> nyancat.c: In function 'main':
> nyancat.c:940:9: warning: implicit declaration of function 'nanosleep' [-Wimplicit-function-declaration]
>   940 |         nanosleep(&ts, &ts);
>       |         ^~~~~~~~~
> in function `main':
> nyancat.c:(.text+0x34c): undefined reference to `nanosleep'
> collect2: error: ld returned 1 exit status
> ```
> 
> Do you have any idea about this error?

I encountered this error back in November or December of last year. This issue has indeed been pending for quite a while without resolution. If we still can't identify the cause of this error, perhaps I could write a script to perform a bisection on the riscv gnu toolchain to try and find the cause.

vacantron:
> Do you have any idea about this error?

I recompile `dhrystone`  with the [new GNU toolchain](https://github.com/riscv-collab/riscv-gnu-toolchain/releases/tag/2024.04.12) and solve this problem. Can you confirm this?

> ..., perhaps I could write a script to perform a bisection on the riscv gnu toolchain to try and find the cause.

Maybe just update the pre-built ELF testbench?

vacantron:
> I encountered this error back in November or December of last year. This issue has indeed been pending for quite a while without resolution. If we still can't identify the cause of this error, perhaps I could write a script to perform a bisection on the riscv gnu toolchain to try and find the cause.

I add `-lsemihost` to the compilation option and the problem is solved. It seems they turned some `libgloss`'s static linked libraries into dynamic ones.

PR #446, opened 2024-05-25
Replace sprintf with safer snprintf

huaxinliao:
The sprintf function has been replaced with the safer snprintf function to prevent potential buffer overflow vulnerabilities and enhance security.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -128,12 +128,13 @@ static bool parse_args(int argc, char **args)
         memcpy(rel_path, args[0], strlen(args[0]) - 7 /* strlen("rv32emu") */);
 
         char *prog_basename = basename(opt_prog_name);
-        prof_out_file = malloc(strlen(cwd_path) + 1 + strlen(rel_path) +
-                               strlen(prog_basename) + 5 + 1);
+        size_t prof_out_file_size = strlen(cwd_path) + 1 + strlen(rel_path) +
+                                    strlen(prog_basename) + 5 + 1;
+        prof_out_file = malloc(prof_out_file_size);
         assert(prof_out_file);
 
-        sprintf(prof_out_file, "%s/%s%s.prof", cwd_path, rel_path,
-                prog_basename);
+        snprintf(prof_out_file, prof_out_file_size, "%s/%s%s.prof", cwd_path,
+                 rel_path, prog_basename);
     }
     return true;
 }


p96114175:
This is my development record: https://hackmd.io/ucXfO5ixSTePjNX_KFR_Vg?view#tip1

jserv:
> This is my development record: https://hackmd.io/ucXfO5ixSTePjNX_KFR_Vg?view#tip1

You don't have to put your note here. Instead, you should explain the rationale and make explicit considerations on security concerns.

huaxinliao:
> You don't have to put your note here. Instead, you should explain the rationale and make explicit considerations on security concerns.

The sprintf function lacks boundary checking, making it prone to buffer overflow, while snprintf function ensures that no more characters are written than the size of the buffer, thus avoiding buffer overflow.

src/main.c
```diff
@@ -132,8 +132,8 @@ static bool parse_args(int argc, char **args)
                                strlen(prog_basename) + 5 + 1);
         assert(prof_out_file);
 
-        sprintf(prof_out_file, "%s/%s%s.prof", cwd_path, rel_path,
-                prog_basename);
+        snprintf(prof_out_file, sizeof(prof_out_file), "%s/%s%s.prof", cwd_path,
```
visitorckw: I believe that sizeof(prof_out_file) is incorrect. IIUC, it will always return the size of the pointer (most likely 4 or 8).

visitorckw:
Upon quickly reviewing the code, it appears that the memory allocated for the 'prof_out_file' is precisely calculated to be sufficient, thereby eliminating the risk of buffer overflow. Additional length checks might introduce unnecessary overhead. If an overflow is indeed possible here, could you provide an example?

src/main.c
huaxinliao: Corrected

huaxinliao:
> Upon quickly reviewing the code, it appears that the memory allocated for the 'prof_out_file' is precisely calculated to be sufficient, thereby eliminating the risk of buffer overflow. Additional length checks might introduce unnecessary overhead. If an overflow is indeed possible here, could you provide an example?

I understand that the currently allocated memory is sufficient, but we cannot guarantee that future changes to the code will not cause a buffer overflow. With my enhancements, we can guarantee the security of this section of the code.

The rv32emu paper also mentions: "Considering the demand for secure and robust sandboxed executions, an efficient and embeddable RISC-V simulator that meets these requirements is crucial." Therefore, I believe that rv32emu should meet the essential security requirements while maintaining high performance. We should not solely pursue performance at the expense of security in smaller details.

visitorckw:
> I understand that the currently allocated memory is sufficient, but we cannot guarantee that future changes to the code will not cause a buffer overflow. With my enhancements, we can guarantee the security of this section of the code.
> 
> The rv32emu paper also mentions: "Considering the demand for secure and robust sandboxed executions, an efficient and embeddable RISC-V simulator that meets these requirements is crucial." Therefore, I believe that rv32emu should meet the essential security requirements while maintaining high performance. We should not solely pursue performance at the expense of security in smaller details.

NAK.

You are addressing a security issue that clearly does not exist **and** introducing unnecessary overhead. If future changes to the code result in a buffer overflow, it means those new changes are incorrect and should be avoided. It does not mean we should add unnecessary checks to the code everywhere. This is akin to why the C language standard still has a lot of well-defined undefined behaviors to allow for various optimizations by the compiler.

huaxinliao:
> You are addressing a security issue that clearly does not exist **and** introducing unnecessary overhead. If future changes to the code result in a buffer overflow, it means those new changes are incorrect and should be avoided. It does not mean we should add unnecessary checks to the code everywhere. This is akin to why the C language standard still has a lot of well-defined undefined behaviors to allow for various optimizations by the compiler.

I suggest you take a look at https://security.web.cern.ch/recommendations/en/codetools/c.shtml. It clearly points out that sprintf is a security vulnerability.

visitorckw:
> I suggest you take a look at https://security.web.cern.ch/recommendations/en/codetools/c.shtml. It clearly points out that sprintf is a security vulnerability. Unless you think CERN Computer Security is lying.

So I looked at the link you provided. It points out that sprintf causes problems when the string length exceeds the buffer length, ***which is obviously not the case here***. Again, you only fixed an issue that may or may not occur in the future, ***and*** it slowed down the performance. To me, this is as unreasonable as adding ```if (index < array_size)``` everywhere when accessing an array.

jserv:
It is worth paying attention if you raise a security concern. However, it is not worthwhile to patch code for a non-existent security threat if you only read the CERN Computer Security guidance without thoughtful security advice.

huaxinliao:
> It is worth paying attention if you raise a security concern. However, it is not worthwhile to patch code for a non-existent security threat if you only read the CERN Computer Security guidance without thoughtful security advice.

Thank you for @jserv and @visitorckw's guidance. I will keep this in mind.

jserv:
Close as no security issue has been resolved.

Issue #448, opened 2024-05-27
Crash when the memory on host is insufficient


tony2037:
I was testing on [multipass](https://multipass.run/) (a lightweight VM manager for Linux, Windows and macOS).
The VM was running on a MacBook Pro (13-inch, 2017, Two Thunderbolt 3 ports), the configuration was as follows:
```
<Host>
Processors: 2.3 GHz Dual-Core Intel Core i5
Memory: 8 GB 2133 MHz LPDDR3
<VM>
Distribution: multipass
local cpu: 4
local memory: 4.0GiB
$ free -h
               total        used        free      shared  buff/cache   available
Mem:           3.8Gi       212Mi       2.1Gi       0.0Ki       1.5Gi       3.3Gi
Swap:             0B          0B          0B
```

I was testing `hello.elf`, but this problem happened to all the other binaries.
The steps to reproduce are as follows:
```
$> ./build/rv32emu ./build/hello.elf
rv32emu: src/riscv.c:199: rv_create: Assertion `attr->mem' failed.
Aborted (core dumped)
```
The core file is as follows.
[coredump.zip](https://github.com/sysprog21/rv32emu/files/15452006/coredump.zip)

By conducting the postmortem analysis, I realized it was because I did not have sufficient memory for `mmap`, and the `MEM_SIZE` is hard-coded to `2^32 - 1` (4GB).

tony2037:
#449 is just an idea. Perhaps allowing user to config accordingly is the most intuitive way to solve this problem.

vacantron:
Maybe modifying the setting of the overcommit fix this problem? Try `echo 1 | sudo tee /proc/sys/vm/overcommit_memory`.

jserv:
@qwe661234, Check the necessary and desirable memory region required by mmap.

jserv:
Alternatively, we could adopt the approach used by the [mmap_allocator](https://github.com/LinZhihao-723/mmap_allocator), a memory allocator based on mmap that allows users to allocate dynamic memory in the mmap region. This could particularly benefit processes with limited swap space, such as those running in restricted containers or virtual machines, by utilizing kernel-managed file cache for swapping.

Check its `mmap_maptemp` function.

PR #449, opened 2024-05-27
Fix insufficient memory on some platforms

tony2037:
[Problem]
The MEM_SIZE is hard-corded, but on some platforms, memory is insufficient.
Perhaps we should allow user to config accordingly

[Solution]
Expose an environment variable, CONFIG_MEM_SIZE, to allow user config the memory size with it

[Test]
* before $> make
$> ./build/rv32emu ./build/hello.elf
rv32emu: src/riscv.c:199: rv_create: Assertion `attr->mem' failed. Aborted (core dumped)

* after $> CONFIG_MEM_SIZE=0x40000000 make
$> ./build/rv32emu ./build/hello.elf
Hello World!
Hello World!
Hello World!
Hello World!
Hello World!
inferior exit code 0

[Reference]
[#488](https://github.com/sysprog21/rv32emu/issues/448)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,11 @@ LDFLAGS += -flto=thin
 endif
 endif
 
+# Allow user to config the memory size accordingly
+ifdef CONFIG_MEM_SIZE
+    CFLAGS += -DMEM_SIZE=$(CONFIG_MEM_SIZE)
+endif
+
 # Disable Intel's Control-flow Enforcement Technology (CET)
 CFLAGS += $(CFLAGS_NO_CET)
 


jserv:
I don't think the proposed change resolves the insufficient memory issue because it just relies on compile-time configurations instead of an adaptive approach to acquire the necessary system resources based on the requirements of RISC-V programs during instruction simulation.

@qwe661234, can you comment this?

qwe661234:
This issue arises when the memory size is insufficient, especially when running rv32emu on a VM. However, arbitrarily setting a MEM_SIZE to pass specific tests can cause segmentation faults in other tests. A better solution is to revert to the original virtual memory implementation when memory size is insufficient, allocating memory only as needed for read or write operations.

qwe661234:
After investigation, you can set MEM_SIZE to `0x80000` to run the benchmarks in directory tests. However, if you want to run some video games like doom or quake, the 4G memory is required.

jserv:
Reworked in commit feca1dea732e110a8b08926f3266d2db732147b0

PR #450, opened 2024-06-03
Remove libc subsituation

qwe661234:
Based on the performance evaluation, the checking overhead of libc subsituation is larger than its performance improvement in most cases. Its effectiveness only displays in the specific cases, and it cannot be further optimized though JIT. Therefore, we decided to remove this mechanism.


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: b8bb4b4edfd2037d91efe8c995721b28662b5b4a | Previous: ded6f042ef2a294d668e7d2dc54f808644a6278e | Ratio |
|-|-|-|-|
| `Dhrystone` | `4.44` Average DMIPS over 10 runs | `4.37` Average DMIPS over 10 runs | `0.98` |

</details>

PR #457, opened 2024-06-13
jit: Introduce background compilation

### LINKED ISSUE #239: jit: Background compilation thread
In tiered compilation, the interpreter monitors the frequency of jumps to specific code addresses. Once a pre-set threshold is reached, it initiates JIT compilation. Concurrently, the emulator can continue operating with the interpreter, while JIT compilation is processed in a background thread, seamlessly switching to the JIT-compiled code once it is ready.

This emulator utilizes a JIT compiler for executing all RISC-V instructions. When execution begins, each emulator thread is equipped to compile new code. This implementation allows JIT compilation to occur in a background thread while the interpreter continues its execution until compilation is complete. If the emulator encounters a function that hasn't been compiled yet, it acquires a lock on the JIT code backend and attempts to compile the entire function into the JIT backend before resuming execution. Importantly, this lock only restricts other threads from adding new code to the JIT backend during compilation, without impeding their ability to use the JIT backend. Essentially, this means that one thread compiling new code has a minimal impact on other threads, resulting in a lock that imposes little overhead.

qwe661234:
Given the significant runtime compilation overhead associated with performing aggressive optimizations, we have implemented a background compilation mechanism to mitigate this issue. When the runtime profiler identifies a strong hotspot, it adds a T2C compilation request to the wait queue. A background thread, which continuously monitors this queue, triggers T2C to process the requests and notifies the main thread upon completion by updating a flag.

This mechanism defers the execution of T2C-generated machine code, leading to more frequent use of T1C-generated code. Despite this, the approach effectively minimizes runtime compilation delays by eliminating the need for the main thread to wait for T2C compilation to complete, thereby improving overall performance.

* The workflow of background compilation.
image

* The performance comparison with and without the background compialtion.
image

Close: #239

### DIFF
--- a/src/feature.h
+++ b/src/feature.h
@@ -57,5 +57,11 @@
 #define RV32_FEATURE_T2C 0
 #endif
 
+/* T2C depends on JIT configuration */
+#if !RV32_FEATURE_JIT
+#undef RV32_FEATURE_T2C
+#define RV32_FEATURE_T2C 0
+#endif
+
 /* Feature test macro */
 #define RV32_HAS(x) RV32_FEATURE_##x
--- a/src/riscv.c
+++ b/src/riscv.c
@@ -28,6 +28,9 @@
 #include "riscv_private.h"
 #include "utils.h"
 #if RV32_HAS(JIT)
+#if RV32_HAS(T2C)
+#include <pthread.h>
+#endif
 #include "cache.h"
 #include "jit.h"
 #define CODE_CACHE_SIZE (4 * 1024 * 1024)
@@ -184,6 +187,27 @@ IO_HANDLER_IMPL(byte, write_b, W)
 #undef R
 #undef W
 
+#if RV32_HAS(T2C)
+static pthread_t t2c_thread;
+static void *t2c_runloop(void *arg)
+{
+    riscv_t *rv = (riscv_t *) arg;
+    while (rv->quit) {
+        if (!list_empty(&rv->wait_queue)) {
+            queue_entry_t *entry =
+                list_last_entry(&rv->wait_queue, queue_entry_t, list);
+            pthread_mutex_lock(&rv->wait_queue_lock);
+            list_del_init(&entry->list);
+            pthread_mutex_unlock(&rv->wait_queue_lock);
+            t2c_compile(entry->block,
+                        (uint64_t) ((memory_t *) PRIV(rv)->mem)->mem_base);
+            free(entry);
+        }
+    }
+    return NULL;
+}
+#endif
+
 riscv_t *rv_create(riscv_user_t rv_attr)
 {
     assert(rv_attr);
@@ -269,6 +293,14 @@ riscv_t *rv_create(riscv_user_t rv_attr)
     rv->jit_state = jit_state_init(CODE_CACHE_SIZE);
     rv->block_cache = cache_create(BLOCK_MAP_CAPACITY_BITS);
     assert(rv->block_cache);
+#if RV32_HAS(T2C)
+    rv->quit = false;
+    /* prepare wait queue. */
+    pthread_mutex_init(&rv->wait_queue_lock, NULL);
+    INIT_LIST_HEAD(&rv->wait_queue);
+    /* activate the background compilation thread. */
+    pthread_create(&t2c_thread, NULL, t2c_runloop, rv);
+#endif
 #endif
 
     return rv;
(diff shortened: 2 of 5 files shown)


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: 51ce53be52b626176e58bde93664c0d86989377a | Previous: 6b462c03dc8c64863ae146a94fe7abffe4da19bb | Ratio |
|-|-|-|-|
| `Dhrystone` | `5.22` Average DMIPS over 10 runs | `4.88` Average DMIPS over 10 runs | `0.93` |

</details>

src/riscv.c
```diff
@@ -184,6 +186,31 @@ IO_HANDLER_IMPL(byte, write_b, W)
 #undef R
 #undef W
 
+#if RV32_HAS(JIT)
+static pthread_t t2c_thread;
+void *t2c_routine(void *arg)
```
jserv: Should be `static`.

src/riscv_private.h
```diff
@@ -132,8 +140,11 @@ struct riscv_internal {
 #if !RV32_HAS(JIT)
     block_map_t block_map; /**< basic block map */
 #else
+    bool exit /**< Determine the main thread is exit or not */;
```
jserv: Rename to `quit` in order to distinguish from the state of being exited from the execution or program being terminated. In this case, using `quit` is much clearer.

src/riscv.c
```diff
@@ -269,6 +296,10 @@ riscv_t *rv_create(riscv_user_t rv_attr)
     rv->jit_state = jit_state_init(CODE_CACHE_SIZE);
     rv->block_cache = cache_create(BLOCK_MAP_CAPACITY_BITS);
     assert(rv->block_cache);
+    rv->exit = false;
+    pthread_mutex_init(&rv->wait_queue_lock, NULL);
+    INIT_LIST_HEAD(&rv->wait_queue);
+    pthread_create(&t2c_thread, NULL, t2c_routine, rv);
```
jserv: Add some comments here. E.g., activate the background compilation thread.

src/riscv.c
```diff
@@ -184,6 +186,31 @@ IO_HANDLER_IMPL(byte, write_b, W)
 #undef R
 #undef W
 
+#if RV32_HAS(JIT) && RV32_HAS(T2C)
+static pthread_t t2c_thread;
+static void *t2c_routine(void *arg)
```
jserv: Rename to `t2c_runloop` to emphasize its behavior in background thread.

jserv:

src/riscv.c
```diff
@@ -28,6 +28,8 @@
 #include "riscv_private.h"
 #include "utils.h"
 #if RV32_HAS(JIT)
+#include <pthread.h>
```
jserv, line 32: If T2C is not enabled, is pthread required?

src/riscv.c
```diff
@@ -184,6 +186,31 @@ IO_HANDLER_IMPL(byte, write_b, W)
 #undef R
 #undef W
 
+#if RV32_HAS(JIT) && RV32_HAS(T2C)
```
jserv: We can centralize the dependency in `src/feature.h` to ensure T2C depends on JIT, eliminating the need to write `#if RV32_HAS(JIT) && RV32_HAS(T2C)`.

src/feature.h
```diff
@@ -50,12 +50,11 @@
 /* Experimental just-in-time compiler */
 #ifndef RV32_FEATURE_JIT
 #define RV32_FEATURE_JIT 0
-#endif
-
 /* Experimental tier-2 just-in-time compiler */
 #ifndef RV32_FEATURE_T2C
 #define RV32_FEATURE_T2C 0
 #endif
+#endif
```
jserv, line 64: Don't do that. Consider the following:
```c
#ifndef RV32_FEATURE_JIT
#define RV32_FEATURE_JIT 0
#endif

#ifndef RV32_FEATURE_T2C
#define RV32_FEATURE_T2C 0
#endif

/* T2C depends on JIT configuration */
#if !RV32_FEATURE_JIT
#undef RV32_FEATURE_T2C
#define RV32_FEATURE_T2C 0
#endif
```

src/riscv.c
```diff
@@ -184,6 +187,31 @@ IO_HANDLER_IMPL(byte, write_b, W)
...
+            queue_entry_t *entry =
+                list_last_entry(&rv->wait_queue, queue_entry_t, list);
+            pthread_mutex_lock(&rv->wait_queue_lock);
+            list_del_init(&entry->list);
+            pthread_mutex_unlock(&rv->wait_queue_lock);
+            t2c_compile(entry->block,
+                        (uint64_t) ((memory_t *) PRIV(rv)->mem)->mem_base);
+            free(entry);
+        }
+        /* Instead of writing while(rv->quit), placing the code here prevents
```
jserv: Does `volatile` help for such purpose?

qwe661234: Yes, it works.

jserv:
Finally, T2C is landed!

PR #462, opened 2024-06-17
Initialize ir->branch_table->PC with safe value

### LINKED ISSUE #461: A segmentation fault occurs when running a simple ELF file
I ran `hello.elf` and it is fine.
But when I tried to run my own mul.elf, segmentation fault happened.
I tried to do some simple gdb debugging to figure out why but just gave up because my unfamiliarity with the emulator.
Can someone explain the cause?

Emulator build configuration:
Every ENABLE_* in Makefile are disabled.
```shell
$ cc --version      
cc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-44)     
Copyright (C) 2015 Free Software Foundation, Inc.    
This is free software; see the source for copying conditions.  There is NO    
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.     
```

`hello.elf` result:
```
$ ./rv32emu hello.elf 
Hello World!
Hello World!
Hello World!
Hello World!
Hello World!
inferior exit code 0
```

`mul.elf` result:
```shell
./rv32emu mul.elf 
Segmentation fault (core dumped)`
```

The objdump of mul.elf:
[mul_objdump.txt](https://github.com/user-attachments/files/15860983/mul_objdump.txt)

c source code:
```c
int mul(int a, int b)
{
    a = 0x12345678;
    b = 0x87654321;
    return a * b;
}
```

It is just an integer multiply function.

ChinYikMing:
If the ra(return address) is 0x0, the
LOOKUP_OR_UPDATE_BRANCH_HISTORY_TABLE will bahave abnormally since calloc initialize ir->branch_table->PC[i] to 0x0. The 0x0 address might be not yet translated to a valid block, thus ir->branch_table->target[i] might be NULL, calling a NULL function pointer cause segmentation fault. It can be solved by initializing ir->branch_table->PC will other value than 0x0. Here, I choose unsigned integer of -1.

Close #461

### DIFF
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -617,6 +617,8 @@ static void block_translate(riscv_t *rv, block_t *block)
             ) {
                 ir->branch_table = calloc(1, sizeof(branch_history_table_t));
                 assert(ir->branch_table);
+                memset(ir->branch_table->PC, -1,
+                       sizeof(uint32_t) * HISTORY_SIZE);
             }
             break;
         }


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: b65d24c0adb4f2e222a8a70639ce5af36819c15f | Previous: 1e4ef3c8802143795e7f4f4e2fee04f995412793 | Ratio |
|-|-|-|-|
| `Dhrystone` | `5` Average DMIPS over 10 runs | `5` Average DMIPS over 10 runs | `1` |

</details>

qwe661234 (approved):

PR #467, opened 2024-07-09
Upgrade CI ubuntu version to 24.04

henrybear327:
Required dependencies are all updated, such as
- clang-format-18
- LLVM 18
- fix an issue with the pip3 install
- fix an issue with testing gdbstub (notice that the riscv-gnu-toolchain releases are built using Ubuntu 22.04, thus, we need
to install libpython3.10 for the CI pipeline to work)

Reference:
- https://stackoverflow.com/questions/75608323/how-do-i-solve-error-externally-managed-environment-every-time-i-use-pip-3

### DIFF
--- a/.ci/check-format.sh
+++ b/.ci/check-format.sh
@@ -8,7 +8,7 @@ set -x
 
 for file in ${SOURCES};
 do
-    clang-format-12 ${file} > expected-format
+    clang-format-18 ${file} > expected-format
     diff -u -p --label="${file}" --label="expected coding style" ${file} expected-format
 done
-exit $(clang-format-12 --output-replacements-xml ${SOURCES} | egrep -c "</replacement>")
+exit $(clang-format-18 --output-replacements-xml ${SOURCES} | egrep -c "</replacement>")
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   detect-code-related-file-changes:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     outputs:
       has_code_related_changes: ${{ steps.set_has_code_related_changes.outputs.has_code_related_changes }}
     steps:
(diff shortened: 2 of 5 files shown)


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: f2e41a8d50722d4166a26d4457269ea7a82a79f8 | Previous: 66dec82a1bc9fe792dacace199aada34324f62b6 | Ratio |
|-|-|-|-|
| `Dhrystone` | `1589.12` Average DMIPS over 10 runs | `4.12` Average DMIPS over 10 runs | `0.0025926298832057997` |
| `Coremark` | `1402.631` Average iterations/sec over 10 runs | `0.004` Average iterations/sec over 10 runs | `0.000002851783541073882` |

</details>

jserv:
The introduction of clang-format-18 does change the program's indentation. I am a bit worried about the inconsistency.

henrybear327:
> The introduction of clang-format-18 does change the program's indentation. I am a bit worried about the inconsistency.

Would we like to stick with clang-format-12? :)

jserv:
> > The introduction of clang-format-18 does change the program's indentation. I am a bit worried about the inconsistency.
> 
> Would we like to stick with clang-format-12? :)

The better way is to ensure the `.clang-format` configurations compatible with latest clang-format.

henrybear327:
> > > The introduction of clang-format-18 does change the program's indentation. I am a bit worried about the inconsistency.
> > 
> > 
> > Would we like to stick with clang-format-12? :)
> 
> The better way is to ensure the `.clang-format` configurations compatible with the latest clang-format.

I did a quick pass on this [documentation](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) and can't quite find out what rules that are introduced in clang-format-18 that we are violating. 

Is there a way for clang-format to print out what rules we have violated that caused the format error to be raised?

jserv:
> I did a quick pass on this [documentation](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) and can't quite find out what rules that are introduced in clang-format-18 that we are violating.
> Is there a way for clang-format to print out what rules we have violated that caused the format error to be raised?

[CONTRIBUTING.md](https://github.com/sysprog21/rv32emu/blob/master/CONTRIBUTING.md) states the supported clang-format version is 12 or later. It is vital to ensure the compatibility for CI pipeline.

jserv:
With [clang-format lint action](https://github.com/DoozyX/clang-format-lint-action), we can flexibly specify the versions of clang-format.

jserv (changes requested):

jserv:
Close in favor of #497

PR #469, opened 2024-07-10
Reintroduce benchmark.yml

henrybear327:
There is an issue with accessing `github-token` from `mail.yml`, see https://github.com/sysprog21/rv32emu/actions/runs/9873045064/job/27264462124?pr=467.

The error is `Error: 'auto-push' is enabled but 'github-token' is not set. Please give API token to push GitHub pages branch to remote`. This wasn't present when the benchmark was in its dedicated file.

The proposal is to move the modified benchmark pipeline back to its original file.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,50 @@
+name: Benchmark
+
+on: [push, pull_request_target, workflow_dispatch]
+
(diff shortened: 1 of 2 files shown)


henrybear327:
@jserv Good news! https://sysprog21.github.io/rv32emu-bench/

PR #474, opened 2024-08-17
Automate prebuilt executable updates via GitHub Actions

### LINKED ISSUE #451: Keep prebuilt ELF files up to date
Due to https://github.com/sysprog21/rv32emu/pull/444 , we need to update the binary with the new toolchain.

vacantron:
To keep the executables up to date for testing and benchmarking, this pull request introduces GitHub CI/CD pipeline to release the tarball of the latest executables when the files under `tests/` be changed.

Close #451


.github/workflows/build-testbenches.yml
```diff
@@ -0,0 +1,61 @@
...
+
+on:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+
+env:
+  CROSS_CC_PREFIX: riscv-none-elf-
+  CROSS_CC_URL: https://github.com/xpack-dev-tools/riscv-none-elf-gcc-xpack/releases/download/v14.2.0-1/xpack-riscv-none-elf-gcc-14.2.0-1-linux-x64.tar.gz
```
jserv: Let's maintain the rules to fetch and extract prebuilt GNU Toolchain via the scripts in `.ci` directory.

.github/workflows/build-testbenches.yml
```diff
@@ -0,0 +1,61 @@
+name: Build testbenches
```
jserv: Rename the filename from `.github/workflows/build-testbenches.yml` to `.github/workflows/build-Artifact.yml`, which implies all benchmarks and test programs.

Makefile
```diff
@@ -182,7 +182,7 @@ $(OUT)/emulate.o: CFLAGS += -foptimize-sibling-calls -fomit-frame-pointer -fno-s
 .DEFAULT_GOAL := all
 
 include mk/external.mk
-
+include mk/benchmark.mk
```
jserv: Shorten as `mk/bench.mk`.

README.md
```diff
@@ -327,46 +327,36 @@ Use of this source code is governed by a MIT license that can be found in the [L
 
 ## External sources
 
-In `rv32emu` repository, there are some prebuilt ELF files for testing purpose.
-* `aes.elf` : See [tests/aes.c](tests/aes.c)
-* `captcha.elf` : See [tests/captcha.c](tests/captcha.c)
+There are some prebuilt executables in [rv32emu-prebuilt](https://github.com/sysprog21/rv32emu-prebuilt) for testing and benchmarking purpose.
+The prebuilt executables are built from GNU/GCC v9 and [xPack GCC v14.2.0-1](https://github.com/xpack-dev-tools/gcc-xpack) under Ubuntu-20.04.
```
jserv: Why did you mention GCC-9?

vacantron: The version of the pre-installed GCC of Ubuntu-20.04 is v9. Maybe upgrade it to v14 through the APT?

jserv: Most prebuilt ELF files are dedicated to RV32 (bare-metal), and the only exception is `tests/arch-test-target/sail_cSim/riscv_sim_RV32` (x86-64/Linux ABI), which serves as the reference simulator for [RISC-V Architecture Test](https://github.com/riscv-non-isa/riscv-arch-test). You shall explain the rationale.

mk/benchmark.mk
```diff
@@ -0,0 +1,60 @@
...
+
+.PHONY: build-testbenches benchmark
+
+# TODO: generate results automatically
+benchmark: build-testbenches
+
+build-testbenches:
+ifeq ($(USE_PREBUILT),1)
+	@echo "Fetching prebuilt executables in \"rv32emu-prebuilt\"..."
+	@wget -O - https://github.com/sysprog21/rv32emu-prebuilt/releases/download/$(LATEST_RELEASE)/rv32emu-prebuilt.tar.gz | tar zx -C build
```
jserv: Add `-q` (aka `--quiet`) option to `wget`.

.github/workflows/build-testbenches.yml
```diff
@@ -0,0 +1,61 @@
...
+
+on:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+
+env:
+  CROSS_CC_PREFIX: riscv-none-elf-
```
jserv: Honor `CROSS_COMPILE` env. Consult `mk` directory for details.

jserv:
Why did you remove the file `tests/aes.c`?

vacantron:
> Why did you remove the file `tests/aes.c`?

`aes.c` is from `rv8-bench` submodule, and `stream.c` is from `ansibench`.

.github/workflows/build-testbenches.yml
vacantron: Use the uppercase A for `build-Artifact`?

jserv:
It is a bit confusing that `mk/external.mk` uses `curl`, while the proposed change opts for `wget`. Could you explain the reasons behind this preference?

jserv:
> `aes.c` is from `rv8-bench` submodule, and `stream.c` is from `ansibench`.

No, the original `tests/aes.c` came from [micro-AES](https://github.com/polfosol/micro-AES), which is independent from [rv8-bench/src/aes.c](https://github.com/sysprog21/rv8-bench/blob/master/src/aes.c).

.github/workflows/build-testbenches.yml
jserv: > Use the uppercase A for `build-Artifact`?

No, simply name it `build-artifact`.

README.md
vacantron: The tarball which is going to be released contains the binaries for both amd64 and rv32 in order to automate the benchmark.

.ci/gdbstub-test.sh
```diff
@@ -21,7 +21,7 @@ if [ -z ${GDB} ]; then
     exit 1
 fi
 
-build/rv32emu -g build/puzzle.elf &
+build/rv32emu -g build/bin/riscv32/puzzle &
```
jserv: It is boring to name the directory as `build/bin` since [binary blob](https://en.wikipedia.org/wiki/Binary_blob) locates in `build` directory. You can reconstruct as `build/riscv32` and `build/linux-x64`.

README.md
jserv: > The tarball which is going to be released contains the binaries for both amd64 and rv32 in order to automate the benchmark.

Yes, as discussed previously, the prebuilt binary files are provided as artifacts for evaluation and to automate architectural testing.

Makefile
```diff
@@ -229,36 +229,35 @@ tool: $(TOOLS_BIN)
...
 	)
 
-EXPECTED_aes_sha1 = 1242a6757c8aef23e50b5264f5941a2f4b4a347e  -
-misalign: $(BIN)
-	$(Q)$(PRINTF) "Running aes.elf ... ";
-	$(Q)if [ "$(shell $(BIN) -m $(OUT)/aes.elf | $(SHA1SUM))" = "$(EXPECTED_aes_sha1)" ]; then \
+EXPECTED_aes_sha1 = f9924635666d3d58d5b60c0bde8b986a2a99effb  -
+misalign: $(BIN) build-testbenches
+	$(Q)$(PRINTF) "Running aes ... ";
+	$(Q)if [ "$(shell $(BIN) -m $(OUT)/bin/riscv32/aes | $(SHA1SUM))" = "$(EXPECTED_aes_sha1)" ]; then \
```
jserv: In order to differentiate the ELF files built from micro-AES and rv8-bench, we can rename the filenames to `microaes` or `uaes`.

README.md
```diff
@@ -327,46 +327,36 @@ Use of this source code is governed by a MIT license that can be found in the [L
 
 ## External sources
 
-In `rv32emu` repository, there are some prebuilt ELF files for testing purpose.
-* `aes.elf` : See [tests/aes.c](tests/aes.c)
-* `captcha.elf` : See [tests/captcha.c](tests/captcha.c)
+There are some prebuilt executables in [rv32emu-prebuilt](https://github.com/sysprog21/rv32emu-prebuilt) for testing and benchmarking purpose.
```
jserv: The prebuilt files were not limited to benchmarking. Explain more.

mk/benchmark.mk
```diff
@@ -0,0 +1,60 @@
+USE_PREBUILT ?= 1
+
+CC ?= gcc
+CROSS_CC ?= riscv-none-elf-gcc
```
jserv: Honor `CROSS_COMPILE` env, following Linux kernel's manner.

Makefile
jserv: For consistency, let's rename `mk/benchmark.mk` to `mk/artifact.mk`, which more accurately reflects its role in generating build artifacts or binary blobs required for [RISC-V Architecture Test](https://github.com/riscv-non-isa/riscv-arch-test) and further evaluation.

README.md
```diff
@@ -327,46 +327,36 @@ Use of this source code is governed by a MIT license that can be found in the [L
...
+Or, to compile the executables from the source, run:
+
+```shell
+$ make build-testbenches USE_PREBUILT=0 CROSS_CC=<riscv-cross-compiler>
+```
+
+Notice that the RISC-V cross-compiler is required to build the binary.
+
+In default, `rv32emu` will download the tarball from GitHub and extract the executables into `build/bin/`.
+
```
ChinYikMing: Since the executables are currently located in `build/`, we can use `build/` directly and do not need to create a separate `build/bin/` directory. The description should be rephrased.

.github/workflows/build-testbenches.yml
```diff
@@ -0,0 +1,61 @@
...
+          echo "/opt/riscv-gnu-toolchain/bin" >> $GITHUB_PATH
+          echo "CROSS_CC=/opt/riscv-gnu-toolchain/bin/$CROSS_CC_PREFIX""gcc" >> $GITHUB_ENV
+      - name: Build RISC-V executables
+        if: ${{ steps.file-changed.outputs.any_changed == 'true' || github.event_name == 'workflow_dispatch' }}
+        run: |
+          CROSS_CC=$CROSS_CC USE_PREBUILT=0 make build-testbenches
+      - name: Create tarballs
+        if: ${{ steps.file-changed.outputs.any_changed == 'true' || github.event_name == 'workflow_dispatch' }}
+        run: |
+          tar -zcf rv32emu-prebuilt.tar.gz -C build bin
```
ChinYikMing: If we are agreed to put all of the executables inside `build`, then the `bin` is not needed.

README.md
```diff
@@ -327,46 +327,36 @@ Use of this source code is governed by a MIT license that can be found in the [L
...
+To fetch the prebuilt executables, run the following command:
+
+```shell
+$ make build-testbenches
+```
+
+Or, to compile the executables from the source, run:
+
+```shell
+$ make build-testbenches USE_PREBUILT=0 CROSS_CC=<riscv-cross-compiler>
```
jserv: I would expect more consistent naming scheme. That is,
* `make artifact`: fetch the prebuilt files by default.
* `make artifact ENABLE_PREBUILT=0 CROSS_COMPILE=riscv-none-eabi-`: build everything from scratch.

jserv:
If we are about to run `make arch-test`, the prebuilt simulator binary is required for referencing purpose.

visitorckw (commented):
I would like to know the motivation, benefits, and reasons for switching from curl to wget?

vacantron:
> If we are about to run `make arch-test`, the prebuilt simulator binary is required for referencing purpose.

Yes, the dependency will be added later.

tests/spirograph.c
```diff
@@ -9,6 +9,7 @@
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <unistd.h>
```
visitorckw, line 12: Could you please clarify what issue this patch addresses, such as compile warnings or errors?

vacantron:
> I would like to know the motivation, benefits, and reasons for switching from curl to wget?

Use `wget` is more consistent with other workflows, and also, `curl` is not pre-installed in Ubuntu distro.

visitorckw:
> Use `wget` is more consistent with other workflows, and also, `curl` is not pre-installed in Ubuntu distro.

Perhaps it would be helpful to include these reasons in the commit message. When I first saw this patch, I was confused about why this change was needed.

vacantron:
> Perhaps it would be helpful to include these reasons in the commit message. When I first saw this patch, I was confused about why this change was needed.

It is an internal commit and is going to be squashed before merging.

visitorckw:
> > Perhaps it would be helpful to include these reasons in the commit message. When I first saw this patch, I was confused about why this change was needed.
> 
> It is an internal commit and is going to be squashed before merging.

Got it.

README.md
```diff
@@ -327,46 +327,36 @@ Use of this source code is governed by a MIT license that can be found in the [L
...
+
+```shell
+$ make build-artifact USE_PREBUILT=0
+```
+
+Notice that the RISC-V cross-compiler is required to build the binary.
+
+In default, `rv32emu` will download the tarball from GitHub and extract the executables into `build/`.
+
+There are still some prebuilt ELF files for testing purpose under `build/`.
```
jserv: Create a file `docs/prebuilt.md` detailing the prebuilt ELF files with a comprehensive listing about their origins.

Makefile
```diff
@@ -229,25 +229,24 @@ tool: $(TOOLS_BIN)
 include mk/riscv-arch-test.mk
 include mk/tests.mk
 
-CHECK_ELF_FILES := \
-	hello \
-	puzzle \
-	fcalc
+# the prebuilt executables are built for "rv32im"
```
jserv, line 232: If rv32emu is configured as RV32I-only, the target `make check` should validate `hello.elf`.

jserv:
The importance of this task is threefold:
1. to reduce the size of the tarball for the upcoming v1.0 release,
2. to separate the validation of the ISA simulator from the generation of complementary ELF files, and
3. to make rv32emu more versatile, serving as both an ISA simulator and a system emulator with on-demand test suites.

tests/spirograph.c
vacantron: GCC complains without including `unistd.h`

```shell
tests/spirograph.c:177:5: warning: implicit declaration of function ‘usleep’ [-Wimplicit-function-declaration]
```

vacantron:
> It is a bit confusing that `mk/external.mk` uses `curl`, while the proposed change opts for `wget`. Could you explain the reasons behind this preference?

The most workflow in rv32emu use `wget` and it is pre-installed in Ubuntu distro. Should we replace the existing `curl` with `wget`?

jserv:
> The most workflow in rv32emu use `wget` and it is pre-installed in Ubuntu distro. Should we replace the existing `curl` with `wget`?

Yes, proceed with unifying to `wget`.

docs/prebuilt.md
```diff
@@ -0,0 +1,32 @@
...
+
+```shell
+$ make artifact USE_PREBUILT=0 [CROSS_COMPILE=<COMPILER_PREFIX>]
+```
+
+The compiler prefix varies according to the used toolchain, such as `riscv-none-elf-`, `riscv32-unknwon-elf-`, etc.
+
+---
+
+There are still some standalone RISC-V binaries under `build/` directory for testing purpose.
```
jserv: You should mention the following:
* uaes
* captcha
* cc
* chacha20
* donut
* fcalc
* hamilton
* jit-bf
* lena
* line
* maj2random
* mandelbrot
* nqueens
* nyancat
* pi
* qrcode
* readelf
* richards
* rvsim
* smolnes
* spirograph

And then make a list of the benchmark programs provided by rv8-bench and ansibench.

mk/artifact.mk
```diff
@@ -0,0 +1,57 @@
+USE_PREBUILT ?= 1
+
+CC ?= gcc
+CROSS_COMPILE ?= riscv-none-elf-
+
+BINDIR := $(abspath $(OUT))
+
+TEST_SUITES += \
+    ansibench \
+	rv8-bench
```
jserv, line 10: Indent.

mk/artifact.mk
```diff
@@ -0,0 +1,57 @@
...
+
+# "ieee754" needs F extension
+# "smolnes", "ticks" have inline assembly and only work in riscv
+TESTBENCHES += \
+    captcha \
+    donut \
+    fcalc \
+    hamilton \
+    $(ieee754) \
```
jserv: Use capitalized identifiers such as `PROG_ieee754` or `PROG_IEEE754`

mk/artifact.mk
```diff
@@ -0,0 +1,57 @@
...
+    puzzle \
+    qrcode \
+    richards \
+    rvsim \
+    $(smolnes) \
+    spirograph \
+    $(ticks) \
+    uaes
+
+ifeq ($(USE_PREBUILT),1)
```
jserv: Check `mk/common.mk` and use `$(call has, ...)`.

vacantron: The `has` macro is to determine the enabled option of rv32emu such as `ENABLE_GDB`, `ENABLE_JIT`, etc,

mk/artifact.mk
vacantron: The testbenches quoted with `$(...)` are provided by `rv32emu` repository and are just a memo for the migration in the future. Should we simply remove them?

mk/artifact.mk
jserv: > The `has` macro is to determine the enabled option of rv32emu such as `ENABLE_GDB`, `ENABLE_JIT`, etc,

No, check the following:
```
POSITIVE_WORDS = 1 true yes                                                                                                                                   
define has
$(if $(filter $(firstword $(ENABLE_$(strip $1))), $(POSITIVE_WORDS)),1,0)
endef
```

Using `ENABLE_PREBUILT` consistently will simplify reusing this macro.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+- `nyancat` : See [tests/nyancat.c](tests/nyancat.c)
+- `pi` : See [tests/pi.c](tests/pi.c)
+- `puzzle` : See [tests/puzzle.c](tests/puzzle.c)
+- `qrcode` : See [tests/qrcode.c](tests/qrcode.c)
+- `richards` : See [tests/richards.c](tests/richards.c)
+- `rvsim` : See [tests/rvsim.c](tests/rvsim.c)
+- `spirograph` : See [tests/spirograph.c](tests/spirograph.c)
+- `uaes` : See [tests/uaes.c](tests/uaes.c)
+
+---
```
jserv: Remove this line.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
+# Prebuilt Binaries
+
+There are some prebuilt binaries placed in [rv32emu-prebuilt](https://github.com/sysprog21/rv32emu-prebuilt).
```
jserv: Rewrite this paragraph as following:
> The prebuilt binaries for rv32emu are prepared primarily because the [RISC-V Sail Model](https://github.com/riscv/sail-riscv) executable is required for the [RISC-V Architecture Test](https://github.com/riscv-non-isa/riscv-arch-test), and selected RISC-V ELF files are useful for ISA simulation validation and testing. Some of these prebuilt binaries are stored in [rv32emu-prebuilt](https://github.com/sysprog21/rv32emu-prebuilt). During testing or benchmarking, these binaries are automatically downloaded into the `build/linux-x64/` and `build/riscv32/` directories by default. The RISC-V binaries are compiled using the [xPack RISC-V GCC toolchain](https://github.com/xpack-dev-tools/riscv-none-elf-gcc-xpack) with the options `-march=rv32im -mabi=ilp32`.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+
+```shell
+$ make artifact USE_PREBUILT=0 [CROSS_COMPILE=<COMPILER_PREFIX>]
+```
+
+The compiler prefix varies according to the used toolchain, such as `riscv-none-elf-`, `riscv32-unknwon-elf-`, etc.
+
+The prebuilt binaries in `rv32emu-prebuilt` are built from the following repositories and resources:
+
+- [ansibench](https://github.com/nfinit/ansibench)
```
jserv: Use the forked repositories.

mk/riscv-arch-test.mk
```diff
@@ -9,7 +9,7 @@ ifeq ($(RISCV_DEVICE),FCZicsr)
 ARCH_TEST_SUITE := tests/rv32fc-test-suite
 endif
 
-arch-test: $(BIN)
+arch-test: $(BIN) artifact
```
jserv, line 12: Verify the availability of `tests/arch-test-target/sail_cSim/riscv_sim_RV32` before launching the `riscv_sim_RV32` executable with RISCOF.

mk/artifact.mk
```diff
@@ -0,0 +1,57 @@
...
+	@echo "Fetching prebuilt executables in \"rv32emu-prebuilt\"..."
+	@wget -q --show-progress https://github.com/sysprog21/rv32emu-prebuilt/releases/download/$(LATEST_RELEASE)/rv32emu-prebuilt.tar.gz -O- | tar -C build -xz
+else
+	@$(foreach tb,$(TEST_SUITES), \
+	    git submodule update --init ./tests/$(tb) &&) true
+	@$(foreach tb,$(TEST_SUITES), \
+		$(MAKE) -C ./tests/$(tb) all BINDIR=$(BINDIR) &&) true
+	@$(foreach tb,$(TESTBENCHES), \
+	    $(CC) -m32 -O2 -Wno-unused-result -o $(BINDIR)/linux-x64/$(tb) tests/$(tb).c -lm &&) true
```
jserv: Is `-m32` required? The binary is placed in `build/linux-x64` directory.

vacantron: If we want to get the more precise result of the performance evaluation between the native execution and the simulation, `-m32` option is required because x86-64 ABI cannot disable the SIMD instruction of the FPU

jserv: > If we want to get the more precise result of the performance evaluation between the native execution and the simulation, `-m32` option is required because x86-64 ABI cannot disable the SIMD instruction of the FPU

Then, you should clarify the directory name as `linux-x86`. Once a file is built with `-m32`, it would be recognized as the following:
> ELF 32-bit LSB shared object, Intel 80386, version 1 (SYSV)

vacantron: To make it clearer, I'm going to rename it as `linux-x86-soft-float`.

jserv: Per GCC's naming convention, "softfp" can be applied. That is, linux-x86-softfp.

.github/workflows/build-artifact.yml
```diff
@@ -0,0 +1,69 @@
...
+          fetch_additional_submodule_history: 'true'
+          files: |
+            tests/ansibench/*
+            tests/rv8-bench/*
+            tests/*.c
+      - name: Set alias
+        id: has_files_changed
+        run: |
+          if [[ ${{ steps.files-changed.outputs.any_changed }} == true ]]; then
+            echo "has_files_changed=true" >> $GITHUB_OUTPUT
```
ChinYikMing: Description of `any_changed` event, quote from [README](https://github.com/tj-actions/changed-files#output_any_changed):
> Returns `true` when any of the filenames provided using the `files*` or `files_ignore*` inputs have changed. This defaults to `true` when no patterns are specified. i.e. includes a combination of all added, copied, modified and renamed files (ACMR).

Description of `any_modified` event, quote from [README](https://github.com/tj-actions/changed-files#output_any_modified):
>  Returns `true` when any of the filenames provided using the `files*` or `files_ignore*` inputs have been modified. This defaults to `true` when no patterns are specified. i.e. includes a combination of all added, copied, modified, renamed, and deleted files (ACMRD).

I would prefer `any_modified` event instead as it supports deletion. If any test suites are not needed anymore, we can rely on the CI to remove it from `sysprog21/rv32emu-prebuilt`.

.github/workflows/build-artifact.yml
```diff
@@ -0,0 +1,69 @@
...
+          submodules: 'true'
+      - name: Test files changed
+        id: files-changed
+        uses: tj-actions/changed-files@v44
+        with:
+          fetch_additional_submodule_history: 'true'
+          files: |
+            tests/ansibench/*
+            tests/rv8-bench/*
+            tests/*.c
```
ChinYikMing, line 25: May I know if we add a directory of test suite, can these syntax detect a new directory and any subdirectories of it? E.g., `tests/system`

vacantron: If the new directory is added, this rule should be updated too.

ChinYikMing: My initial intention is to ensure that if the system emulation is merged in the future, additional test suites will be included as well. To distinguish these from the ISA simulation test suite, I'd like to add a new directory (e.g., `tests/system`).

Thus, we shall support detection of a new test suite directory and any subdirectories of it.

vacantron: From the usage [section](https://github.com/tj-actions/changed-files?tab=readme-ov-file#usage-),

> Matching Files and Folders: To match all files and folders under a directory, this requires a globstar pattern e.g. dir_name/** which matches any number of subdirectories and files.

I think updating the rules with `tests/system/**` is fine.

ChinYikMing: > From the usage [section](https://github.com/tj-actions/changed-files?tab=readme-ov-file#usage-),
> 
> > Matching Files and Folders: To match all files and folders under a directory, this requires a globstar pattern e.g. dir_name/** which matches any number of subdirectories and files.
> 
> I think updating the rules with `tests/system/**` is fine.

Thanks for the information. I think it is not appropriate to change the rule in this PR. When system emulation is integrated, I will consider the change.

mk/artifact.mk
```diff
@@ -0,0 +1,68 @@
+ENABLE_PREBUILT ?= 1
+
+CC ?= gcc
+CROSS_COMPILE ?= riscv-none-elf-
+
+BINDIR := $(abspath $(OUT))
+
```
ChinYikMing, line 7: Use `BIN_DIR` instead for consistency.

mk/artifact.mk
```diff
@@ -0,0 +1,68 @@
...
+TEST_SUITES += \
+	ansibench \
+	rv8-bench
+
+# "ieee754" needs F extension
+# "smolnes", "ticks" have inline assembly and only work in riscv
+TESTBENCHES += \
+	captcha \
```
ChinYikMing, line 15: Ditto.

mk/artifact.mk
```diff
@@ -0,0 +1,68 @@
...
+
+BINDIR := $(abspath $(OUT))
+
+TEST_SUITES += \
+	ansibench \
+	rv8-bench
+
+# "ieee754" needs F extension
+# "smolnes", "ticks" have inline assembly and only work in riscv
+TESTBENCHES += \
```
ChinYikMing: Ditto.

mk/artifact.mk
```diff
@@ -0,0 +1,68 @@
...
+	puzzle \
+	qrcode \
+	richards \
+	rvsim \
+	spirograph \
+	uaes
+
+SHELL_HACK := $(shell mkdir -p $(BINDIR)/linux-x86-softfp $(BINDIR)/riscv32)
+
+ifeq ($(call has, PREBUILT), 1)
```
jserv, line 36: `ENABLE_PREBUILT` should be set explicitly in advance.

vacantron: It has been set at the first line and is enabled in default.

.ci/riscv-toolchain-install.sh
```diff
@@ -8,10 +8,15 @@ check_platform
 
 mkdir -p toolchain
 
-# GNU Toolchain for RISC-V
-GCC_VER=2024.04.12
-TOOLCHAIN_REPO=https://github.com/riscv-collab/riscv-gnu-toolchain/releases
+if [[ "$#" = 0 ]] || [[ "$1" != "riscv-collab" ]]; then
```
ChinYikMing: `"$#" = 0` is missing one '=' ?

vacantron: No, but the quotation marks are missing here. In bash, both single and double equals sign (`=`) are used to string comparison, or use `-eq` for simplicity.

ChinYikMing: > No, but the quotation marks are missing here. In bash, both single and double equals sign (`=`) are used to string comparison, or use `-eq` for simplicity.

Consider `"$#" = "0"` or `$# == 0`. I tried both, both have the same effect.

vacantron: > Consider "$#" = "0" or $# == 0. I tried both, both have the same effect.

After testing, the original bash script seems to work too.

.github/workflows/build-artifact.yml
```diff
@@ -0,0 +1,87 @@
...
+          sudo apt-get install -q -y gcc-multilib g++-multilib
+          sudo apt-get install -q -y opam build-essential libgmp-dev z3 pkg-config zlib1g-dev
+          .ci/riscv-toolchain-install.sh
+          echo "$PWD/toolchain/bin" >> $GITHUB_PATH
+      - name: Build binaries
+        run: |
+          make artifact ENABLE_PREBUILT=0
+          mkdir -p /tmp/rv32emu-prebuilt
+          mv build/linux-x86-softfp build/riscv32 /tmp/rv32emu-prebuilt
+      - name: Buile Sail model
```
jserv: Fix typo. Should be "Build"

Makefile
```diff
@@ -244,10 +244,19 @@ EXPECTED_puzzle = success in 2005 trials
...
 
-check: $(BIN)
+check-hello: $(BIN)
+	$(Q)$(PRINTF) "Running hello.elf ..."; \
+	    if [ "$(shell $(BIN) $(OUT)/hello.elf | uniq)" = "$(strip $(EXPECTED_hello)) inferior exit code 0" ]; then \
+	        $(call notice, [OK]); \
+	        else \
+	        $(PRINTF) "Failed.\n"; \
+	        exit 1; \
+	        fi;
```
jserv: Indent.

mk/artifact.mk
```diff
@@ -0,0 +1,68 @@
...
+	rvsim \
+	spirograph \
+	uaes
+
+SHELL_HACK := $(shell mkdir -p $(BIN_DIR)/linux-x86-softfp $(BIN_DIR)/riscv32)
+
+ifeq ($(call has, PREBUILT), 1)
+  LATEST_RELEASE := $(shell wget -q https://api.github.com/repos/sysprog21/rv32emu-prebuilt/releases/latest -O- | grep -Po '(?<="tag_name": ").+(?=",)')
+else
+  CFLAGS := -m32 -mno-sse -mno-sse2 -msoft-float -O2 -L$(BIN_DIR)
```
jserv, line 41: Add comments to emphasize the reason why linux-x86 target with soft floating point is built here.

mk/artifact.mk
```diff
@@ -0,0 +1,70 @@
...
+	qrcode \
+	richards \
+	rvsim \
+	spirograph \
+	uaes
+
+SHELL_HACK := $(shell mkdir -p $(BIN_DIR)/linux-x86-softfp $(BIN_DIR)/riscv32)
+
+ifeq ($(call has, PREBUILT), 1)
+  LATEST_RELEASE := $(shell wget -q https://api.github.com/repos/sysprog21/rv32emu-prebuilt/releases/latest -O- | grep -Po '(?<="tag_name": ").+(?=",)')
```
ChinYikMing: The `-P` option of `grep` is not supported on macOS, the reason is that the BSD version uses the POSIX Compatible Regular Expressions and the GNU version uses the Perl Compatible Regular Expressions (PCRE). Although my BSD version said it is compatible to GNU version ( output of `grep -V`: `grep (BSD grep, GNU compatible) 2.6.0-FreeBSD` ), but it does not support PCRE.

Consider this instead: `wget -q https://api.github.com/repos/sysprog21/rv32emu-prebuilt/releases/latest -O- | grep '"tag_name"' | sed -E 's/.*"tag_name": "([^"]+)".*/\1/'`. This works on both Linux and macOS.

vacantron: Thanks for the information!

jserv:
I defer to @ChinYikMing for confirmation.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+$ make artifact
+```
+
+Or build the binaries from scratch (the RISC-V cross-compiler is required):
+
+```shell
+$ make artifact ENABLE_PREBUILT=0 [CROSS_COMPILE=<COMPILER_PREFIX>]
+```
+
+The compiler prefix varies according to the used toolchain, such as `riscv-none-elf-`, `riscv32-unknwon-elf-`, etc.
```
ChinYikMing: There is a typo in `riscv32-unknwon-elf-`.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+    - nbench
+- [rv8-bench](https://github.com/sysprog21/rv8-bench)
+    - aes
+    - dhrystone
+    - miniz
+    - norx
+    - primes
+    - qsort
+    - sha512
+- `captcha` : See [tests/captcha.c](tests/captcha.c)
```
ChinYikMing: Add `/` to the beginning of the link because the document is not located in the root of the repository, which causes the link to be broken.

Some description of the link has `tests/` prefix and some do not. Align all of them for consistency.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+- `donut` : See [donut.c](tests/donut.c)
```
ChinYikMing: Ditto.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+- `fcalc` : See [fcalc.c](tests/fcalc.c)
```
ChinYikMing: Ditto.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+- `hamilton` : See [hamilton.c](tests/hamilton.c)
```
ChinYikMing: Ditto.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+- `jit` : See [tests/jit.c](tests/jit.c)
```
ChinYikMing: Ditto.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+- `lena`: See [tests/lena.c](tests/lena.c)
```
ChinYikMing: Ditto.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+- `line` : See [tests/line.c](tests/line.c)
```
ChinYikMing: Ditto.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+- `maj2random` : See [tests/maj2random.c](tests/maj2random.c)
```
ChinYikMing: Ditto.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+- `mandelbrot` : See [tests/mandelbrot.c](tests/mandelbrot.c)
```
ChinYikMing: Ditto.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+- `nqueens` : See [tests/nqueens.c](tests/nqueens.c)
```
ChinYikMing: Ditto.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+- `nyancat` : See [tests/nyancat.c](tests/nyancat.c)
```
ChinYikMing: Ditto.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+- `pi` : See [tests/pi.c](tests/pi.c)
```
ChinYikMing: Ditto.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+- `puzzle` : See [tests/puzzle.c](tests/puzzle.c)
```
ChinYikMing: Ditto.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+- `qrcode` : See [tests/qrcode.c](tests/qrcode.c)
```
ChinYikMing: Ditto.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+- `richards` : See [tests/richards.c](tests/richards.c)
```
ChinYikMing: Ditto.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+- `rvsim` : See [tests/rvsim.c](tests/rvsim.c)
```
ChinYikMing: Ditto.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+- `spirograph` : See [tests/spirograph.c](tests/spirograph.c)
```
ChinYikMing: Ditto.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+- `uaes` : See [tests/uaes.c](tests/uaes.c)
```
ChinYikMing: Ditto.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+
+There are still some prebuilt standalone RISC-V binaries under `build/` directory only for testing purpose:
+
+- `hello.elf` : See [tests/asm-hello](tests/asm-hello)
```
ChinYikMing: Ditto.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+- `cc.elf` : See [tests/cc](tests/cc)
```
ChinYikMing: Ditto.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+- `chacha20.elf` : See [tests/chacha20](tests/chacha20)
```
ChinYikMing: Ditto.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+- `doom.elf` : See [sysprog21/doom_riscv](https://github.com/sysprog21/doom_riscv) [RV32M]
+- `ieee754.elf` : See [tests/ieee754.c](tests/ieee754.c) [RV32F]
```
ChinYikMing: Ditto.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+- `jit-bf.elf` : See [ezaki-k/xkon_beta](https://github.com/ezaki-k/xkon_beta)
+- `quake.elf` : See [sysprog21/quake-embedded](https://github.com/sysprog21/quake-embedded) [RV32F]
+- `readelf.elf` : See [tests/readelf](tests/readelf)
```
ChinYikMing: Ditto.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+- `scimark2.elf` : See [tests/scimark2](tests/scimark2) [RV32MF]
```
ChinYikMing: Ditto.

docs/prebuilt.md
```diff
@@ -0,0 +1,67 @@
...
+- `smolnes.elf` : See [tests/smolnes](tests/smolnes.c) [RV32M]
```
ChinYikMing: Ditto.

.github/workflows/build-artifact.yml
```diff
@@ -0,0 +1,87 @@
...
+  detect-files-changed:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          submodules: 'true'
+      - name: Test files changed
+        id: files-changed
+        uses: tj-actions/changed-files@v44
```
ChinYikMing: v45 just released. Bump to latest.

docs/prebuilt.md
vacantron: > Add `/` to the beginning of the link because the document is not located in the root of the repository, which causes the link to be broken.

The single slash (`/`) seems not to redirect to the repository root. Or just remove the hyperlink for simplicity?

ChinYikMing: > > Add `/` to the beginning of the link because the document is not located in the root of the repository, which causes the link to be broken.
> 
> The single slash (`/`) seems not to redirect to the repository root. Or just remove the hyperlink for simplicity?

Did you try `[tests/captcha.c](/tests/captcha.c)` ?

vacantron: > Did you try `[tests/captcha.c](/tests/captcha.c)` ?

So weird, it works. I remember it reported 404 a couple of months ago.

ChinYikMing: > > Did you try `[tests/captcha.c](/tests/captcha.c)` ?
> 
> So weird, it works. I remember it reported 404 a couple of months ago.

Let's proceed with this.

ChinYikMing (approved):
LGTM.

jserv:

PR #480, opened 2024-08-28
Automate doom, quake, and scimark2 builds

vacantron:


docs/prebuilt.md
```diff
@@ -48,8 +49,10 @@ The prebuilt binaries in `rv32emu-prebuilt` are built from the following reposit
 - `pi` : See [tests/pi.c](/tests/pi.c)
 - `puzzle` : See [tests/puzzle.c](/tests/puzzle.c)
 - `qrcode` : See [tests/qrcode.c](/tests/qrcode.c)
+- `quake` : See [sysprog21/quake-embedded](https://github.com/sysprog21/quake-embedded)
```
jserv: Mention RV32F explicitly.

docs/prebuilt.md
```diff
@@ -36,6 +36,7 @@ The prebuilt binaries in `rv32emu-prebuilt` are built from the following reposit
     - sha512
 - `captcha` : See [tests/captcha.c](/tests/captcha.c)
 - `donut` : See [tests/donut.c](/tests/donut.c)
+- `doom` : See [sysprog21/doom_riscv](https://github.com/sysprog21/doom_riscv)
```
jserv, line 39: Mention RV32M for Doom.

docs/prebuilt.md
vacantron: The current prebuilt binaries are built with `-march=rv32im` option. I'd like to keep `quake` been built without F extension for the consistency.

jserv: SciMark2 should be built in RV32F since it reports a composite score in approximate Mflops (Millions of floating point operations per second).

jserv:
CI reported:
```
GPG error: https://apt.llvm.org/jammy llvm-toolchain-jammy-17 InRelease:
The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 15CF4D18AF4F7421
E: The repository 'http://apt.llvm.org/jammy llvm-toolchain-jammy-17 InRelease' is not signed.
```

See [Signatures couldn't be verified because the public key is not available: NO_PUBKEY](https://github.com/carla-simulator/carla/issues/6445).

docs/prebuilt.md
vacantron: > SciMark2 should be built in RV32F since it reports a composite score in approximate Mflops (Millions of floating point operations per second).

Got it. Should we also build `quake` with `rv32f`? I feel the execution is smooth when building with `rv32im`.

jserv: > Got it. Should we also build `quake` with `rv32f`? I feel the execution is smooth when building with `rv32im`.

Quake needs to be compiled with RV32IMF. Otherwise, it is unlikely to function properly.

vacantron:
> CI reported:
> ...
> See [Signatures couldn't be verified because the public key is not available: NO_PUBKEY](https://github.com/carla-simulator/carla/issues/6445).

I think this error was caused by the script from https://apt.llvm.org/llvm.sh. In line 160: `wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key | tee /etc/apt/trusted.gpg.d/apt.llvm.org.asc`. It seems to fail to retrieve the GPG key (maybe the Internet issue?).

docs/prebuilt.md
vacantron: I found `quake` is already built with RV32IMF within its Makefile. I'll update the document later.

jserv:

PR #483, opened 2024-08-30
Enhance precision and safety in time calculations

visitorckw:
Introduces a new inline function, mult_frac(), to handle fractional multiplication with improved precision and safety by computing x * n / d while avoiding common issues like overflow and precision loss. Additionally, the get_time_info() function has been refactored to use mult_frac() for time-related calculations, including nanosecond conversions from mach_absolute_time() and clock() in low-resolution timer mode. These updates collectively enhance the accuracy and reliability of time computations throughout the codebase.

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -26,6 +26,19 @@
 
 #define MAX_PATH_LEN 1024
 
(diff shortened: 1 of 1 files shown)


jserv:

PR #487, opened 2024-09-01
Bring wasm service up

### LINKED ISSUE #486: Wasm service is down and a generator for the ELF executables selection list should be introduced
Wasm service down after merging #474. The reason is that multiple ELF executables are being placed in the `build/linux-x86-softfp` and `build/riscv32` directories, whereas wasm service expects the ELF executables to be located in the `build/` directory.

ChinYikMing:
After merging https://github.com/sysprog21/rv32emu/pull/474, multiple executables were moved to other location (build/riscv32), which caused the wasm service to fail. The intuitive solution is to embed build/riscv32 into the wasm. Also, executables that stored in rv32emu-prebuilt must be pulled before embedding into wasm, so emcc_deps should depend on the artifact target.

Additionally, the Makefile's distclean target does not fully clean wasm artifacts, as CC=emcc might not be set during make distclean. To fix this, the declaration of the WEB_FILES and DEMO_DIR wasm-related variables has been reordered.

As the number of ELF executables may increase over time, manually updating the HTML ELF list becomes tedious. Therefore, an ELF list generator has been introduced. When new ELF executables are added to the build or build/riscv32 directories, the list will be generated dynamically.

Close: #486


jserv (changes requested):
Clarify the git commit message: if this pull request is able to resolve #486, you should write `Close #486` explicitly.

ChinYikMing:
> Clarify the git commit message: if this pull request is able to resolve #486, you should write `Close #486` explicitly.

I would close #486 if the ELF selection list generator is being merged. Alternatively, I could update the issue title to simplify tracking.

ChinYikMing:
I changed to title in #486. The ELF selection list generator is in progress and may take some time. To bring up the WASM service quickly, this PR could be merged first.

ChinYikMing:
Close: #486

mk/wasm.mk
```diff
@@ -28,15 +29,19 @@ CFLAGS_emcc += -sINITIAL_MEMORY=2GB \
...
+	       --embed-file build/riscv32@/riscv32 \
 	       --embed-file build/timidity@/etc/timidity \
 	       -DMEM_SIZE=0x40000000 \
 	       -DCYCLE_PER_STEP=2000000 \
 	       --pre-js $(WEB_JS_RESOURCES)/pre.js \
 	       -O3 \
 	       -w
 
+gen_elf_list_js:
+	$(Q)tools/gen-elf-list-js.py > $(DEMO_DIR)/elf_list.js
```
jserv: Consider the following:
```shell
$(DEMO_DIR)/elf_list.js: tools/gen-elf-list-js.py
    $(Q)tools/gen-elf-list-js.py > $@
```

jserv (changes requested):
Rebase the latest `master` branch for passing the architecture tests.

jserv:

PR #488, opened 2024-09-02
CI: Fix missing file in WebAssembly deployment

ChinYikMing:
After merging #487, the elf_list.js file is generated dynamically, and the wasm service requires this file to run. However, during CI deployment, the file is not pushed to the rv32emu-demo repo, causing the GitHub Pages to malfunction. To resolve this, pushing the file to the rv32emu-demo repo is the solution.

The 'demo' directory is only created when the WASM service runs locally. To address this, the elf_list.js file should be placed in the 'build' directory instead, as the CI deployment does not invoke the Makefile's start-web target, so the 'demo' directory will not be generated.


ChinYikMing:
The WebAssembly CI needs to be triggered on our own.

PR #490, opened 2024-09-02
Refactor mt19937_extract function

huaxinliao:
This commit is to improve maintainability.


jserv:
The proposed change does not really help. Read https://en.wikipedia.org/wiki/Code_refactoring carefully.

PR #494, opened 2024-09-14
Fix gdbstub dependency and enable parallel compilation for faster CI

visitorckw:
Fix a compilation error related to the missing gdbstub dependency in the Makefile when building with ENABLE_GDBSTUB=1, ensuring the gdbstub library is correctly compiled before the dependent source files to prevent missing header file issues. Additionally, it introduces parallel compilation using -j$(nproc) for both the default build and gdbstub tests, utilizing all available CPU cores to accelerate the build and testing processes, reducing CI execution time.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -49,7 +49,7 @@ jobs:
             sudo ./llvm.sh 17
       shell: bash
     - name: default build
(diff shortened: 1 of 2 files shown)


jserv:

PR #499, opened 2024-10-10
CI: Exclude external files for CodeQL

jserv:


jserv:
Use standard CodeQL scanning.

Issue #103, opened 2023-01-02
Generate RISC-V instruction decoder from ISA descriptor


jserv:
There is some relevant documentation included with the current RISC-V instructions decoding implementation. The maintenance and verification, however, are not straightforward. Instead, we may describe how RISC-V instructions are encoded in human readable form; a code generator will then convert this information into C code.
See [make_decoder.py](https://github.com/badlydrawnrod/arviss/blob/develop/make_decoder.py) from [arviss](https://github.com/badlydrawnrod/arviss) and [HiSimu](https://github.com/xzhong86/hisimu) for reference.

Expected output:
1. Create `src/instructions.in` which contains the following:
```
# format of a line in this file:
# <instruction name> <args> <opcode>
#
# <opcode> is given by specifying one or more range/value pairs:
# hi..lo=value or bit=value or arg=value (e.g. 6..2=0x45 10=1 rd=0)
#
# <args> is one of rd, rs1, rs2, rs3, imm20, imm12, imm12lo, imm12hi,
# shamtw, shamt, rm
# rv32i
beq     bimm12hi rs1 rs2 bimm12lo 14..12=0 6..2=0x18 1..0=3
bne     bimm12hi rs1 rs2 bimm12lo 14..12=1 6..2=0x18 1..0=3
blt     bimm12hi rs1 rs2 bimm12lo 14..12=4 6..2=0x18 1..0=3
bge     bimm12hi rs1 rs2 bimm12lo 14..12=5 6..2=0x18 1..0=3
bltu    bimm12hi rs1 rs2 bimm12lo 14..12=6 6..2=0x18 1..0=3
bgeu    bimm12hi rs1 rs2 bimm12lo 14..12=7 6..2=0x18 1..0=3
```
2. Prepare `scripts/gen-decoder.py` (other scripting languages are acceptable.) which can convert from the above into the corresponding C implementation.
3. Modify build system and `src/decode.c` to be aware of the above changes.
4. Create an entry in directory `docs` which describe the high level idea and the way to describe more extensions.

jserv:
[rvjit](https://github.com/latte-c/rvjit) provides similar code generation. See its [scripts](https://github.com/latte-c/rvjit/tree/main/script).

jserv:
Google's [mpact-riscv](https://github.com/google/mpact-riscv) offers ISA description for the RV32/RV64 architecture. See `riscv/*.isa` for details.

jserv:
[...]

PR #501, opened 2024-10-17
Add decoder for RVV instructions

howjmay:
relate to #504


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: 7554c3b3812ec6bbe380ae07da92cd862d391883 | Previous: 584625f09fcd744eacbc25111f9ddb54e8207c4f | Ratio |
|-|-|-|-|
| `Dhrystone` | `1336` Average DMIPS over 10 runs | `1336` Average DMIPS over 10 runs | `1` |
| `Coremark` | `948.638` Average iterations/sec over 10 runs | `957.189` Average iterations/sec over 10 runs | `1.01` |

</details>

jserv:
How can you test the RVV instructions?

howjmay:
A test set own by chipsalliance is here. https://github.com/chipsalliance/riscv-vector-tests/tree/main

I think we can test instructions' behavior with this repo. However, testing decoder is harder. I think I will Spike the verify the behavior, or generating the test cases

src/emulate.c
```diff
@@ -370,6 +370,9 @@ static bool has_loops = false;
     }
 
 #include "rv32_template.c"
+#if RV32_HAS(EXT_RVV)
+#include "rv32_rvv_template.c"
```
jserv: Rename it to `rv32_vector.c`

src/encoding.h
```diff
@@ -0,0 +1,4467 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+
+/* Copyright (c) 2023 RISC-V International */
+
+/*
+ * This file is auto-generated by running 'make' in
+ * https://github.com/riscv/riscv-opcodes (9226b0d)
+ * and it is temporarily used by RVV only
+ */
```
jserv, line 9: Related: https://github.com/sysprog21/rv32emu/issues/103

Do you think it makes sense to replace the existing hand-written RISC-V instruction decoder with the generated one?

howjmay: I think the current one is good for education, especially the comments are valuable.
However I think having a table to clearly store the instructions in human-readable form is good for the future development.

howjmay: But this encoding.h is only a temporary header file for me to decode rvv instructions.

jserv:
Close in favor of #545

Issue #506, opened 2024-10-27
Failed to run arch-test on arm64 machine


howjmay:
Prebuilt binary of Sail doesn't exist.
I was running test on Macbook M1, commit `164da62b64fa050569b22910ea207525884802ba`.
```
$make arch-test
  CC    build/map.o
  CC    build/utils.o
  CC    build/decode.o
  CC    build/io.o
  CC    build/syscall.o
  CC    build/emulate.o
  CC    build/riscv.o
  CC    build/elf.o
  CC    build/cache.o
  CC    build/mpool.o
  CC    build/main.o
  LD    build/rv32emu
Fetching SHA-1 of prebuilt binaries ... [OK]
Checking SHA-1 of prebuilt binaries ... [OK]
git submodule update --init tests/
cp: build/sail_cSim/riscv_sim_RV32: No such file or directory
make: *** [arch-test] Error 1
```

ChinYikMing:
I was running on an M2 without any errors (but having some dependency issue, i.e., `riscof` not found). Could you provide more details on how to reproduce this?

howjmay:
But I did nothing. I have even run `make clean` to ensure I didnt change anything in the codebase.
`make arch-test` is the only thing I ran

jserv:
> Prebuilt binary of Sail doesn't exist. 

@vacantron, If including a darwin-arm64 (i.e., macOS ARM 64-bit binary) build of `sail_cSim`, what modifications would be necessary for rv32emu-prebuilt?

vacantron:
The executable has been added to `rv32emu-prebuilt` .

See [Releases](https://github.com/sysprog21/rv32emu-prebuilt/releases) .

jserv:
@howjmay, can you confirm with the latest `master` branch?

vacantron:
> [@howjmay](https://github.com/howjmay), can you confirm with the latest `master` branch?

The executable has been added, however, the Aarch64 support of architecture test is currently in progress in #547 .

vacantron:
The executable for Arm64/MacOS has been added to `rv32emu-prebuilt` . Rerun `.ci/riscv-tests.sh` on master branch and it should work as expected

See [Releases](https://github.com/sysprog21/rv32emu-prebuilt/releases/tag/2025.01.15-sail) .

jserv:
[...]

PR #507, opened 2024-10-27
Ask to install RISCOF before running arch-test locally

ChinYikMing:
A specific version of RISCOF is used in CI, so requesting for that version as well.

### DIFF
--- a/mk/riscv-arch-test.mk
+++ b/mk/riscv-arch-test.mk
@@ -1,3 +1,9 @@
+riscof-check:
+	$(Q)if [ "$(shell pip show riscof 2>&1 | head -n 1 | cut -d' ' -f1)" = "WARNING:" ]; then \
+	$(PRINTF) "Run 'pip3 install git+https://github.com/riscv/riscof.git@d38859f85fe407bcacddd2efcd355ada4683aee4' to install RISCOF\n"; \
+	exit 1; \
+	fi;
+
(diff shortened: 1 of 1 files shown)


vacantron (approved):

mk/riscv-arch-test.mk
```diff
@@ -1,3 +1,9 @@
+riscof-check:
+	$(Q)if [ "$(shell pip show riscof 2>&1 | head -n 1 | cut -d' ' -f1)" = "WARNING:" ]; then \
+	$(PRINTF) "Run 'pip3 install git+https://github.com/riscv/riscof.git@d38859f85fe407bcacddd2efcd355ada4683aee4' to install RISCOF\n"; \
```
jserv, line 3: Is this workaround still making sense?

ChinYikMing: > Is this workaround still making sense?

This may not be the solution for issue #506, but if someone attempts to run `make arch-test` instead of executing the script ".ci/riscv-tests.sh", the target will bail because riscof is not the default Python module. This alert can easily clarify any confusion.

jserv: Got it. Mention the above rationale in git commit messages.

ChinYikMing: Done.

jserv:

PR #509, opened 2024-11-02
CI: Bump Clang to 18

ChinYikMing:


.github/workflows/main.yml
```diff
@@ -143,16 +143,17 @@ jobs:
     - name: set up scan-build
       run: |
             sudo apt-get update -q -y
-            sudo apt-get install -q -y clang clang-tools libsdl2-dev libsdl2-mixer-dev
+            sudo apt-get install -q -y libsdl2-dev libsdl2-mixer-dev
             wget https://apt.llvm.org/llvm.sh
             chmod +x ./llvm.sh
             sudo ./llvm.sh 18
+            sudo apt-get install -q -y clang-18 clang-tools-18
```
visitorckw, line 150: Is there any dependency that requires us to install clang-18 and clang-tools-18 after running llvm.sh?

ChinYikMing: > Is there any dependency that requires us to install clang-18 and clang-tools-18 after running llvm.sh?

Maybe `sudo ./llvm.sh 18` is enough since it includes both `clang-18` and `clang-tools-18`.

ChinYikMing: The llvm.sh script seems like only adding LLVM sources list to the system. It does have the option to install all packages (e.g., `sudo ./llvm.sh all`). But, we only need `clang-18` and  `clang-tools-18` during static analysis. Additionally, the system might not include LLVM 18 in its sources list, so be sure to add the LLVM 18 source list before installing them.

Please correct me if anything wrong. @jserv

jserv:

PR #125, opened 2023-04-20
Implement LFU as default cache along with memory pool

qwe661234:
We have implemented an adaptive replacement cache to reduce memory usage. However, the complexity of the replacement policy of ARC can lead to a decrease in overall performance. As a result, we implemented the LFU cache, which performed better, but we still retained the ARC for future use. Additionally, we imported the memory pool we developed to limit the memory usage of both caches.

The statistics below illustrate the performance of the LFU cache and ARC while running the CoreMark benchmark, indicating that the LFU cache outperforms ARC.

|Test| CoreMark(Iterations/Sec)  |
|--------| -------- | 
| ARC | 1123.776 |
| LFU Cache  |  1155.174  |

PR #518, opened 2024-12-02
jit: Replace LFU with LRU cache replacement policy

vacantron:
The LFU (least-frequently-used) cache replacement policy would be inefficient when the cache is not big enough to hold all translated blocks (IRs). For example, there are two new entries going to be inserted to cache. They would evict each other because the other one has the least used times. This could cause the inefficiency in cache and make the cache-hit ratio fallen.

The LRU (least-recently-used) cache replacement policy is more suitable for our needs. However, when the least used cache is evicted from the cache, the counter which is used to trigger JIT compilation is going to be lost.

To address this issue, this patch introduces the degenerated adaptive replacement cache (ARC) which has only LRU part and its ghost list. The evicted entries will be temporarily preserved in the ghost list as the history. If the key of inserted entry matches the one in the ghost list, the history will be freed, and the frequency mentioned above would be inherited by the new entry.

The performance difference between the original implementation and this patch is shown below:

* 1024 entries (10 bits)

| Metric    |    Original    |     Patched    |
|-----------|---------------:|---------------:|
| dhrystone |  17720 DMIPS   |  17660 DMIPS   |
| coremark  |   5540 iters/s |   5550 iters/s |
| aes       |  9.164 s       |  9.084 s       |
| nqueens   |  1.551 s       |  1.493 s       |
| hamilton  | 12.917 s       | 12.565 s       |

* 256 entries (8 bits)

| Metric    |    Original    |     Patched    |
|-----------|---------------:|---------------:|
| dhrystone |  17420 DMIPS   |  18025 DMIPS   |
| coremark  |     48 iters/s |   5575 iters/s |
| aes       |  8.904 s       |  8.834 s       |
| nqueens   |  6.416 s       |  1.348 s       |
| hamilton  |   3400 s       | 13.004 s       |

* 64 entries (6 bits)

| Metric    |    Original    |     Patched    |
|-----------|---------------:|---------------:|
| dhrystone |  17720 DMIPS   |  17850 DMIPS   |
| coremark  |   (timeout)    |    215 iters/s |
| aes       |    342 s       |  8.882 s       |
| nqueens   |    680 s       |  1.506 s       |
| hamilton  |   (timeout)    | 13.724 s       |

* Experimental Linux Kernel booting: 126s (original) -> 21s (patched)


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: 5ffb157eeaecdfd165f26949914fd8e7a326e217 | Previous: 448f434b8628c5dfb72a4ba64b531f7734109c5b | Ratio |
|-|-|-|-|
| `Dhrystone` | `1512` Average DMIPS over 10 runs | `1556` Average DMIPS over 10 runs | `1.03` |
| `Coremark` | `1399.459` Average iterations/sec over 10 runs | `1400.104` Average iterations/sec over 10 runs | `1.00` |

</details>

vacantron:
@jserv , can we suppress the warnings from scan-build?

https://github.com/sysprog21/rv32emu/actions/runs/12116509397/job/33777032051?pr=518

jserv:
> can we suppress the warnings from scan-build?
> https://github.com/sysprog21/rv32emu/actions/runs/12116509397/job/33777032051?pr=518

Since both 'list' and 'hlist' are heavily tested, we can suppress the potential false alarm raised by LLVM static analysis.

Makefile
```diff
@@ -141,35 +141,39 @@ endif
...
-endif
-
-ifneq ($(processor),$(filter $(processor),x86_64 aarch64 arm64))
-$(error JIT mode only supports for x64 and arm64 target currently.)
-endif
+    OBJS_EXT += jit.o
+    ENABLE_T2C ?= 1
+    $(call set-feature, T2C)
+    ifeq ($(call has, T2C), 1)
+        # tier-2 JIT compiler powered LLVM
```
jserv: Refine the comment:
> tier-2 JIT compiler is powered by LLVM.

Makefile
```diff
@@ -141,35 +141,39 @@ endif
...
+        endif
+        ifeq ("$(findstring -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS, "$(shell $(LLVM_CONFIG) --cflags)")", "")
+            $(error No llvm-config-18 installed. Check llvm-config-18 installation in advance, or use "ENABLE_T2C=0" to disable tier-2 LLVM compiler)
+        endif
+		CHECK_LLVM_LIBS := $(shell $(LLVM_CONFIG) --libs 2>/dev/null 1>&2; echo $$?)
+		ifeq ("$(CHECK_LLVM_LIBS)", "0")
+            OBJS_EXT += t2c.o
+            CFLAGS += -g $(shell $(LLVM_CONFIG) --cflags)
+            LDFLAGS += $(shell $(LLVM_CONFIG) --libs)
+		else
```
jserv: Indent these lines.

src/cache.c
```diff
@@ -107,102 +113,156 @@ static inline void hlist_del_init(struct hlist_node *n)
...
+                              ht_list, cache_entry_t)
+#endif
+    {
+        if (entry->alive)
+            continue;
+        /* TODO: determine which history should be removed precisely */
+        if (entry->frequency >= THRESHOLD)
+            continue;
+        if (entry->last_used > cache->counter - N_CACHE_HISTORY)
+            continue;
```
jserv: Any heuristic algorithm available to determine?

src/emulate.c
```diff
@@ -832,11 +832,19 @@ static block_t *block_find_or_translate(riscv_t *rv)
 #else
     list_add(&next_blk->list, &rv->block_list);
 
+#if RV32_HAS(T2C)
+    pthread_mutex_lock(&rv->cache_lock);
+#endif
+
     /* insert the block into block cache */
     block_t *replaced_blk = cache_put(rv->block_cache, rv->PC, &(*next_blk));
```
ChinYikMing: `&(*next_blk)` can be simplified as `next_blk`.

Makefile
```diff
@@ -141,35 +141,39 @@ endif
...
+            $(error No llvm-config-18 installed. Check llvm-config-18 installation in advance, or use "ENABLE_T2C=0" to disable tier-2 LLVM compiler)
+        endif
+    endif
+    ifneq ($(processor),$(filter $(processor),x86_64 aarch64 arm64))
+        $(error JIT mode only supports for x64 and arm64 target currently.)
+    endif
```
jserv, line 176: Properly indent with space characters.

visitorckw:
> The LFU (least-frequently-used) replacement policy would encounter the performance issue when the cache is not big enough to hold all translated blocks (IRs). The LRU (least-recently-used) replacement policy is more suitable for our needs. However, when the least-used cache is dropped, the counter which is used to trigger JIT compilation is going to be cleared.
> 
> To address this issue, we store the "live" cache in both list and hash map, and manage the history of the replaced cache in hash map. Since the out-of-use cache is added again, it can retrieve the previous counter. However, the hash map might become very big if we don't trim the unused history in it. We use the distance with the current counter and the last accessed time to determine the history should be dropped or not.

The PR description and the first patch mention performance issues but provide no benchmark numbers. They also state that LRU fits our needs better than LFU without explaining the workload or why it is more suitable. This contradicts the conclusion of commit bdc5348 ("Implement LFU as default cache along with memory pool (#125)") without providing an explanation.

Could you update the descriptions to address these issues?

vacantron:
> Could you update the descriptions to address these issues?

Sure, I will provide some benchmarks to indicate the performance difference on the upcoming changes.

src/cache.c
```diff
@@ -107,102 +115,158 @@ static inline void hlist_del_init(struct hlist_node *n)
...
      * THRESHOLD, the block is dispatched to the code generator to generate C
      * code. The generated C code is then compiled into machine code by the
      * target compiler.
      */
-    if (update && entry->frequency < THRESHOLD) {
-        list_del_init(&entry->list);
-        list_add(&entry->list, cache->lists[entry->frequency++]);
+    if (update) {
+        entry->last_used = cache->counter;
+        entry->frequency++;
```
visitorckw: Would it be safe to assume the frequency number never overflows?

src/cache.c
vacantron: > Any heuristic algorithm available to determine?

Yes, this part of code has been updated for managing the evicted cache more precisely. The degenerated adaptive replacement cache (ARC) is used here.

src/cache.c
vacantron: > Would it be safe to assume the frequency number never overflows?

No, however, since the frequency is greater than `THRESHOLD`, the tiered compilation is triggered. After the blocks are compiled and marked as `hot`, the value of frequency has no other effect, even it might overflow from `UINT32_MAX`.

src/cache.c
```diff
@@ -30,20 +29,24 @@ struct hlist_node {
...
 typedef struct {
     struct hlist_head *ht_list_head;
 } hashtable_t;
 
 typedef struct cache {
-    struct list_head *lists[THRESHOLD];
-    uint32_t list_size;
-    hashtable_t *map;
+    struct list_head list;
+    struct list_head ghost_list;
```
jserv: Document the use of ghost_list, optionally referring adaptive replacement cache (ARC).

src/cache.c
```diff
@@ -30,20 +29,24 @@ struct hlist_node {
...
+    hashtable_t map;
+    uint32_t size;
+    uint32_t ghost_list_size;
```
jserv, line 60: Ditto.

src/cache.c
```diff
@@ -107,102 +112,139 @@ static inline void hlist_del_init(struct hlist_node *n)
...
+        return;
+
+    cache_entry_t *entry =
+        list_last_entry(&cache->ghost_list, cache_entry_t, list);
+    assert(!entry->alive);
+    hlist_del_init(&entry->ht_list);
+    list_del_init(&entry->list);
+    cache->ghost_list_size--;
+    free(entry);
+}
```
jserv: Ditto. Since it was not typical LRU, explain its behavior explicitly.

jserv (changes requested):
Clarify the variant of LRU.

tests/cache/test-cache.c
```diff
@@ -5,12 +5,12 @@
 
 #include "cache.h"
 
-static void print_value(int *val)
+static void print_value(int *val, int freq)
```
ChinYikMing, line 8: Dereference `val` then pass it to  the `print_value` is preferable since no value updating in the implementation of `print_value`.

tests/cache/test-cache.c
```diff
@@ -40,14 +44,15 @@ int main(int argc, char *argv[])
     char *line = NULL, *ptr = NULL;
     size_t len = 0;
     struct cache *cache = NULL;
-    int key, *ans, *val;
+    int key, freq, *ans, *val;
     while (getline(&line, &len, fp) != -1) {
         char *arr[3];
         split(arr, line, " ");
         if (!strcmp(arr[0], "GET")) {
             key = (int) strtol(arr[1], &ptr, 10);
```
ChinYikMing: Simply pass `NULL` to replace `&ptr` since the remaining string is not interested.

tests/cache/test-cache.c
```diff
@@ -40,14 +44,15 @@ int main(int argc, char *argv[])
...
             ans = cache_get(cache, key, true);
-            print_value(ans);
+            freq = cache_freq(cache, key);
+            print_value(ans, freq);
         } else if (!strcmp(arr[0], "PUT")) {
             key = (int) strtol(arr[1], &ptr, 10);
```
ChinYikMing: Ditto.

src/cache.c
```diff
@@ -107,102 +120,139 @@ static inline void hlist_del_init(struct hlist_node *n)
...
      * target compiler.
      */
-    if (update && entry->frequency < THRESHOLD) {
-        list_del_init(&entry->list);
-        list_add(&entry->list, cache->lists[entry->frequency++]);
-    }
+    if (update)
+        entry->frequency++;
 
     /* return NULL if cache miss */
```
ChinYikMing: This line of comment should be stated in line 192 is more appropriate? The wording has a bit misleading. Mention a hit here is more ituitive.

src/cache.c
```diff
@@ -214,66 +264,77 @@ void *cache_put(cache_t *cache, uint32_t key, void *value)
...
-        hlist_del_init(&replaced_entry->ht_list);
-        mpool_free(cache_mp, replaced_entry);
-        cache->list_size--;
+        cache->size--;
+        list_add(&replaced_entry->list, &cache->ghost_list);
+        cache->ghost_list_size++;
     }
 
-    lfu_entry_t *new_entry = mpool_alloc(cache_mp);
+    cache_entry_t *new_entry = calloc(1, sizeof(cache_entry_t));
```
ChinYikMing, line 287: Add assertion here.

src/cache.c
```diff
@@ -107,102 +120,139 @@ static inline void hlist_del_init(struct hlist_node *n)
...
          pos = hlist_entry_safe((pos)->member.next, type, member))
+
+#define hlist_for_each_entry_safe(pos, n, head, member, type) \
+    for (pos = hlist_entry_safe((head)->first, type, member); \
+         pos && ({ n = pos->member.next; 1; });               \
+         pos = hlist_entry_safe(n, type, member))
 #endif
+/* clang-format on */
 
 cache_t *cache_create(int size_bits)
```
ChinYikMing: Use unsigned instead since `size_bits` should not be signed.

src/cache.c
```diff
@@ -214,66 +264,77 @@ void *cache_put(cache_t *cache, uint32_t key, void *value)
...
     new_entry->key = key;
     new_entry->value = value;
-    new_entry->frequency = 0;
-    list_add(&new_entry->list, cache->lists[new_entry->frequency++]);
+    new_entry->alive = true;
+
+    if (!revived_entry) {
+        new_entry->frequency = 1;
+    } else {
+        new_entry->frequency = revived_entry->frequency + 1;
```
ChinYikMing: nit: use `revived_entry->frequency++` for shorter.

src/cache.c
```diff
@@ -214,66 +264,77 @@ void *cache_put(cache_t *cache, uint32_t key, void *value)
...
-    free(cache->map);
+    free(cache->map.ht_list_head);
     free(cache);
 }
 
 uint32_t cache_freq(const struct cache *cache, uint32_t key)
 {
-    if (!cache->capacity ||
-        hlist_empty(&cache->map->ht_list_head[cache_hash(key)]))
+    if (unlikely(!cache->capacity))
```
ChinYikMing, line 327: Shouldn't it be checking the `size` since the `capacity` is fixed after initialization but `size` changes during runtime?

src/cache.c
```diff
@@ -282,45 +343,60 @@ uint32_t cache_freq(const struct cache *cache, uint32_t key)
 #if RV32_HAS(JIT)
 bool cache_hot(const struct cache *cache, uint32_t key)
 {
-    if (!cache->capacity ||
-        hlist_empty(&cache->map->ht_list_head[cache_hash(key)]))
+    if (unlikely(!cache->capacity))
```
ChinYikMing, line 351: Ditto.

ChinYikMing (changes requested):
1. Reorder the commit order such that squashing `Update cache tests` with `Replace LFU with LRU cache replacement policy` or arrange them as consecutive commit because these two commits are highly relevant . This ensures that the commit log remains clear and easy to understand.

2. Commit message typo in first commit: `Matric` -> `Metric`.

tests/cache/test-cache.c
vacantron: The pointer `val` might be NULL since there is no cache replacement occuring. So that we cannot simply derefence it to get its value.

src/cache.c
vacantron: Thess two line are not logically equivalent. I think this may be somewhat misleading.

src/cache.c
vacantron: If the cache capacity is set to zero, we could and should alway early return with 0 for all cases.

ChinYikMing: > If the cache capacity is set to zero, we could and should alway early return with 0 for all cases.

Got it. The `size` can be checked by `hlist_empty`.

src/cache.c
ChinYikMing: > Thess two line are not logically equivalent. I think this may be somewhat misleading.

My bad. It should be `++revived_entry->frequency`. Pre-increment has same effect of `revived_entry->frequency + 1`.

Let me clarify the only changes is the rhs.

vacantron: The `revived_entry` would be read-only and freed after the stored information is inherited by the new entry.

src/cache.c
```diff
@@ -107,102 +120,139 @@ static inline void hlist_del_init(struct hlist_node *n)
...
+ */
 void *cache_put(cache_t *cache, uint32_t key, void *value)
 {
-    assert(cache->list_size <= cache->capacity);
+    assert(cache->size <= cache->capacity);
 
-    lfu_entry_t *replaced_entry = NULL, *entry;
-    hlist_for_each_entry (entry, &cache->map->ht_list_head[cache_hash(key)],
-                          ht_list) {
+    cache_entry_t *replaced_entry = NULL, *revived_entry = NULL, *entry;
```
jserv: How about renaming them to `replaced` and `revived` for short?

src/cache.c
```diff
@@ -30,20 +29,32 @@ struct hlist_node {
 
 typedef struct {
     void *value;
+    bool alive; /* indicates whether this cache is alive or a history of evicted
+                   cache in hash map */
     uint32_t key;
     uint32_t frequency;
```
jserv: Use shorter `freq` for consistency.

src/cache.c
```diff
@@ -30,20 +29,32 @@ struct hlist_node {
...
     struct hlist_node ht_list;
-} lfu_entry_t;
+} cache_entry_t;
 
 typedef struct {
     struct hlist_head *ht_list_head;
 } hashtable_t;
 
+/*
+ * The cache utilizes the degenerated adaptive replacement cache (ARC), which
```
jserv, line 45: Denote the considerations of degenerated ARC introduced here.

jserv:

PR #530, opened 2024-12-27
Use instruction-based emulation for debug mode

RinHizakura:
In debug mode(gdbstub), we'll need one instruction/cycle per step for trace. For such usecase, using the block-based emulation will be too expensive which slow down the execution. Switch to instruction-based emulation for debug mode.

### DIFF
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -862,12 +862,9 @@ static block_t *block_find_or_translate(riscv_t *rv)
     block_translate(rv, next_blk);
 
     optimize_constant(rv, next_blk);
-#if RV32_HAS(GDBSTUB)
-    if (likely(!rv->debug_mode))
-#endif
 #if RV32_HAS(MOP_FUSION)
-        /* macro operation fusion */
-        match_pattern(rv, next_blk);
+    /* macro operation fusion */
+    match_pattern(rv, next_blk);
 #endif
 
 #if !RV32_HAS(JIT)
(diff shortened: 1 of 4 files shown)


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: ecb20c82197edf5429109a3f906d735b2ba687db | Previous: 158597a1ad1266a073877db9e47ab6a70a859aad | Ratio |
|-|-|-|-|
| `Dhrystone` | `1346` Average DMIPS over 10 runs | `1329` Average DMIPS over 10 runs | `0.99` |
| `Coremark` | `950.448` Average iterations/sec over 10 runs | `973.118` Average iterations/sec over 10 runs | `1.02` |

</details>

src/emulate.c
```diff
@@ -958,6 +955,47 @@ static bool rv_has_plic_trap(riscv_t *rv)
     return ((rv->csr_sstatus & SSTATUS_SIE || !rv->priv_mode) &&
             (rv->csr_sip & rv->csr_sie));
 }
+
+static void rv_check_interrupt(riscv_t *rv)
+{
+    vm_attr_t *attr = PRIV(rv);
+    /* check for any interrupt after every block/instruction emulation */
```
ChinYikMing: Move this comment to the beginning of the while loop in the `rv_step` since `rv_step_debug` call it step by step.

src/emulate.c
```diff
@@ -1111,6 +1116,41 @@ void rv_step(void *arg)
 #endif
 }
 
+void rv_step_debug(void *arg)
+{
+    assert(arg);
+    riscv_t *rv = arg;
+
+    rv_check_interrupt(rv);
```
ChinYikMing, line 1122: How about modify here like following?
```
#if RV32_HAS(SYSTEM) && !RV32_HAS(ELF_LOADER)
    rv_check_interrupt(rv);
#endif
```

In this way, the dummy `rv_check_interrupt` can be removed.

vacantron: The debug mode of system simulation is not supported yet. Should we remove this statement to avoid ambiguity?

RinHizakura: What does "debug mode of system simulation is not supported" mean here? The debug/gdbstub mode should be supported naturally because there's no big difference between program or system emulation. I already checked that we could read/write the register, set the breakpoint and step instructions on system emulation.

vacantron: Would `rv32emu` support debugging with booting Linux Kernel now? I failed to run `file build/linux-image/Image` in `riscv32-unknown-elf-gdb` to launch the remote GDB since it is not the RISC-V executable.

RinHizakura: Yes. Please use the `vmlinux` file, `Image` is not the right one. By the way, you can skip the step to link GDB to an ELF file if you don't need any debug symbol(i.e. just run `target remote :1234` and start debugging).

RinHizakura: You can check like this
1. Open a bash shell and run `./build/rv32emu -k build/linux-image/Image -i build/linux-image/rootfs.cpio -b build/minimal.dtb -g`
2. Open another shell and run `riscv32-unknown-elf-gdb -ex "target remote :1234" -ex "cont"`
3. You can see the emulator start executing in the first shell

vacantron: Oh, it works. Thank you to point it out. Can you also update the document for this? (maybe in another PR)

vacantron (approved):
LGTM

jserv:

PR #531, opened 2024-12-28
Avoid to wrongly allocate memory in map_file()

RinHizakura:
In the case that `HAVE_MMAP` is 0, the emulator crashes.

```
$ make ENABLE_SYSTEM=1 system
rv32emu: src/emulate.c:626: block_translate: Assertion `insn' failed.
Aborted
```

The `map_file()` logic when `HAVE_MMAP == 0` is weird as it allocates extra resources on the emulated memory. After this change, the aborting issue is gone, so I believe the original is not the correct implementation.

### DIFF
--- a/src/riscv.c
+++ b/src/riscv.c
@@ -239,10 +239,6 @@ static void map_file(char **ram_loc, const char *name)
     if (*ram_loc == MAP_FAILED)
         goto cleanup;
 #else
-    /* calloc and load data to a memory region */
-    *ram_loc = calloc(st.st_size, sizeof(uint8_t));
-    if (!*ram_loc)
-        goto cleanup;
     if (read(fd, *ram_loc, st.st_size) != st.st_size) {
         free(*ram_loc);
         goto cleanup;


RinHizakura:
BTW, I guess that we don't have coverage for `HAVE_MMAP = 0`. Do we need to have it?

jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: f554a005f401de86ae7de59339bdc20d63a234de | Previous: cb5eb1068f7121e0426a74d2ed03e25e89980045 | Ratio |
|-|-|-|-|
| `Dhrystone` | `1287` Average DMIPS over 10 runs | `1328` Average DMIPS over 10 runs | `1.03` |
| `Coremark` | `962.761` Average iterations/sec over 10 runs | `964.938` Average iterations/sec over 10 runs | `1.00` |

</details>

vacantron (changes requested):
Please refine the commit message that might contain "avoid reallocating memory" (or something like that) to indicate your change.

jserv:
> BTW, I guess that we don't have coverage for `HAVE_MMAP = 0`. Do we need to have it?

The condition `HAVE_MMAP=0` is useful when profiling, as it is not straightforward to measure the system memory consumption when `mmap` is heavily used during emulation.

ChinYikMing:
@RinHizakura Thanks for the fix. It indeed an implementation bug.

RinHizakura:
> The condition HAVE_MMAP=0 is useful when profiling, as it is not straightforward to measure the system memory consumption when mmap is heavily used during emulation.

Sorry for the unclear question, I was asking about the CI/CD coverage in fact.

jserv:
> I was asking about the CI/CD coverage in fact.

You can simply skip the condition `HAVE_MMAP=0` in CI/CD pipeline.

jserv:
Breakage reported by CI:
```
src/jit.c:1760:17: warning: ‘do_sret’ defined but not used [-Wunused-function]
 1760 |     static void do_##inst(struct jit_state *state UNUSED, riscv_t *rv UNUSED, \
      |                 ^~~
src/rv32_jit.c:355:1: note: in expansion of macro ‘GEN’
  355 | GEN(sret, {assert(NULL);
      | ^~~
```

RinHizakura:
> Breakage reported by CI:

This is not related to this change. The same warning can be found in other CI/CD jobs https://github.com/sysprog21/rv32emu/actions/runs/12540405349/job/34967656388 even if it passes. Not sure if any issues on the CI/CD justify this as a fail.

jserv:

Issue #510, opened 2024-11-03
Trap-and-emulate guestOS SDL-based applications


ChinYikMing:
Currently, the guest OS primarily uses its own syscall table. To enable it to run SDL-based applications, we may need to trap and emulate the SDL instruction, which would be handled by the rv32emu SDL backend.

Goal: at least able to run Doom, Quake and smolnes.

A simple demo is opening a white SDL screen.
![image](https://github.com/user-attachments/assets/80aa57f3-5130-4437-acd5-071675fd60d5)

ChinYikMing:
Any chance to run [mado](https://github.com/sysprog21/mado)?

jserv:
> Any chance to run [mado](https://github.com/sysprog21/mado)?

Hopefully, Mado is designed as an advanced yet lean window system for resource-constrained environments, making it suitable for system emulation validation.

ChinYikMing:
A quick demo to run Doom in guest OS, showing the possibility and capability of the guest OS to handle display and audio :

https://github.com/user-attachments/assets/579297d8-0fac-480d-a674-bdad0c7e070e

BTW, some technical issues have to be resolved:
- Quake seems has to be increase the memory pool size by changing the source code. (Doom is not necessary)
- How to reap music thread after the application exit (Ctrl  + c)? If this is not handled, the background music might play forever.
- Sometime the bidirectional event queues is not responding. (guessing the application getting context switch by the guest OS)
- Sometime getting double free or malloc corrupting error from the host OS (could be investigated by `valgrind` or address sanitizer)
- Destroy the previous SDL window after exiting. If this is not handled, the previous SDL window might be reused by the next application, which could cause issues since not all applications have the same window size.

PR #533, opened 2025-01-03
Unify MMU translation APIs

ChinYikMing:
To enable guest OS to run SDL applications using bidirectional queues, the VMM or system emulator must access the guest OS's virtual memory, as the queues are defined there. This requires MMU translation. In addition, the queues event data size likely larger than a word (maybe a page), thus the existing rv->io.mem_{read,write}_w do not sufficient for such memory manipulation.

Export rv->io.mem_translate() to perform gVA2gPA translation. Unifying MMU translation logic by using 
rv->io.mem_translate(), eliminating redundant flows: mmu_walk -> check_pg_fault -> check_signal -> get_ppn_and_offset.
Also, this interface allowing src/syscall_sdl.c to handle virtual memory by first translating the gVA2gPA, followed by memory_{read/write} the chunk of data at once.

TODO: dTLB can be introduced in rv->io.mem_trans() to cache the gVA2gPA translation.

See: #310, #510

### DIFF
--- a/src/riscv.h
+++ b/src/riscv.h
@@ -404,6 +404,20 @@ typedef void (*riscv_mem_write_s)(riscv_t *rv,
 typedef void (*riscv_mem_write_b)(riscv_t *rv,
                                   riscv_word_t addr,
                                   riscv_byte_t data);
+#if RV32_HAS(SYSTEM)
+/*
+ * VA2PA handler
+ * The MMU walkers and fault checkers are defined in system.c
+ * Thus, exporting this handler through function pointer
+ * preserves the encapsulation of MMU translation.
+ *
+ * ifetch do not leverage this translation because basic block
+ * might be retranslated and the corresponding PTE is NULL.
+ */
+typedef riscv_word_t (*riscv_mem_translate_t)(riscv_t *rv,
+                                              riscv_word_t vaddr,
+                                              bool rw);
+#endif
 
 /* system instruction handlers */
 typedef void (*riscv_on_ecall)(riscv_t *rv);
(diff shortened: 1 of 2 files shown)


src/riscv.h
```diff
@@ -404,6 +404,20 @@ typedef void (*riscv_mem_write_s)(riscv_t *rv,
...
+/*
+ * VA2PA handler
+ * The MMU walkers and fault checkers are defined in system.c
+ * Thus, exporting this handler through function pointer
+ * preserves the encapsulation of MMU translation.
+ *
+ * ifetch do not leverage this translation because basic block
+ * might be retranslated and the corresponding PTE is NULL.
+ */
+typedef riscv_word_t (*riscv_mem_trans)(riscv_t *rv,
```
jserv: Specify `_t` suffix for typedef.

jserv: I am not sure if `riscv_mem_trans` is meaningful enough. How about `mmu_translate_t` or `mem_translate_t`?

ChinYikMing: > Specify `_t` suffix for typedef.

For only this one? What about the others?

ChinYikMing: > I am not sure if `riscv_mem_trans` is meaningful enough. How about `mmu_translate_t` or `mem_translate_t`?

Use `riscv` prefix for consistency with others. I would go with `riscv_mem_translate_t`, is it OK?

jserv: > Use `riscv` prefix for consistency with others. I would go with `riscv_mem_translate_t`, is it OK?

It looks good.

The reason not to use "trans" is that it is an existing adjective which might not be intended to mean "translate." See https://dictionary.cambridge.org/dictionary/english/trans

jserv:

Issue #527, opened 2024-12-25
Customize device tree for system emulation


RinHizakura:
Currently, I want to try adding a `-append` option similar to QEMU to specify kernel boot args more conveniently. I want to discuss which implementation may be a better choice for the future development of rv32emu.

In more detail, system simulation is currently implemented using the `-b` option to provide a device tree describing the platform. However, because the device tree is coupled with the hardware (emulation) design of rv32emu itself, it means that even if the user wants to customize the device tree, he/she still needs to be based on a template like `minimal.dts`. Therefore, I think it might be more appropriate to design only parts of the device tree that provide specific options for tweaking.

For the requirement to customize bootargs, there are several possible options as far as I know:
1. Use libfdt to dynamically create a device tree(similar to qemu)
2. Use `minimal.dts` as the template, and then provide the option to apply dt overlay
3. Represent dts as a formatted string, and then fill in the customized part when executing the emulator

Please let me know what you think about this.

jserv:
For system-level emulation, rv32emu (and semu) takes a different approach compared to QEMU. The core idea behind rv32emu is simplicity, avoiding the need to provide generic solutions upfront. Regarding device tree generation, I suggest leveraging an existing minimal DTS as a template combined with Python scripts (or other scripting tools) to generate peripherals and devices. Employing libfdt for this purpose might be unnecessarily complex.

RinHizakura:
What do you think if we generate the DTS like [`make_dtb()`](https://github.com/RinHizakura/riscv-emulator/blob/main/src/emu.c#L17) and compile it in runtime? We can consider using `#embed` to embed the existing `minimal.dts` with some placeholder to specify the bootargs.

jserv:
> What do you think if we generate the DTS like [`make_dtb()`](https://github.com/RinHizakura/riscv-emulator/blob/main/src/emu.c#L17) and compile it in runtime?

It looks great to process text for DTB generation. However, `dtc` is now becoming a runtime dependency.
If using libfdt, do you think the code for generation could be as concise as [kvm-host](https://github.com/sysprog21/kvm-host/blob/master/src/arch/arm64/vm.c)?

RinHizakura:
> It looks great to process text for DTB generation. However, dtc is now becoming a runtime dependency.

From `system.mk` I think we already have `dtc` dependency to build the minimalist DTS, so only if we consider the user may use the default prebuilt DTB only, the `dtc` should always be required.

[...]

PR #534, opened 2025-01-05
Customize bootargs for system emulation 

RinHizakura:
As discussed in #527: The device tree is coupled with the hardware design of rv32emu itself, so if the user wants to customize the device tree, he/she still needs to be based on a template. Therefore, embedding the DT template in the emulator can be more appropriate, and we can allow the user to tweak parts of the device tree from specific options.

To accomplish this, we first use [bin_to_c](https://github.com/bitbank2/bin_to_c.git) to include DTB in the emulator binary itself and use [libfdt](https://github.com/qemu/dtc/tree/v1.7.0) to modify the property.

You can verify the effect of this change by running the following command to see the difference.
```
$ build/rv32emu -k build/linux-image/Image -i build/linux-image/rootfs.cpio -b "earlycon console=ttyS0 printk.time=0"
$ build/rv32emu -k build/linux-image/Image -i build/linux-image/rootfs.cpio -b "earlycon console=ttyS0 printk.time=1"
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -31,3 +31,7 @@
 	path = tests/quake
 	url = https://github.com/sysprog21/quake-embedded
 	shallow = true
+[submodule "src/dtc"]
+	path = src/dtc
+	url = https://git.kernel.org/pub/scm/utils/dtc/dtc.git
+	shallow = true
--- a/mk/system.mk
+++ b/mk/system.mk
@@ -4,10 +4,18 @@ ifeq ($(call has, SYSTEM), 1)
 DEV_SRC := src/devices
 
 DTC ?= dtc
-$(OUT)/minimal.dtb: $(DEV_SRC)/minimal.dts
+BUILD_DTB := $(OUT)/minimal.dtb
+$(BUILD_DTB): $(DEV_SRC)/minimal.dts
 	$(VECHO) " DTC\t$@\n"
 	$(Q)$(DTC) $^ -o $@
-BUILD_DTB := $(OUT)/minimal.dtb
+
+BIN_TO_C := $(OUT)/bin2c
+$(BIN_TO_C): tools/bin2c.c
+	$(Q)$(CC) -Wall -o $@ $^
+
+BUILD_DTB2C := src/minimal_dtb.h
+$(BUILD_DTB2C): $(BIN_TO_C) $(BUILD_DTB)
+	$(BIN_TO_C) $(BUILD_DTB) > $(BUILD_DTB2C)
 
 $(DEV_OUT)/%.o: $(DEV_SRC)/%.c $(deps_emcc)
 	$(Q)mkdir -p $(DEV_OUT)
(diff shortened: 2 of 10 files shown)


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: 4c73b1aefe58c3dc5cb2330d9990573da94b2d06 | Previous: 08146d93d1bd6fbd9d8bde28b39a500a05045710 | Ratio |
|-|-|-|-|
| `Dhrystone` | `1336` Average DMIPS over 10 runs | `1329` Average DMIPS over 10 runs | `0.99` |
| `Coremark` | `942.606` Average iterations/sec over 10 runs | `942.861` Average iterations/sec over 10 runs | `1.00` |

</details>

src/riscv.c
```diff
@@ -259,6 +262,62 @@ static void map_file(char **ram_loc, const char *name)
     exit(EXIT_FAILURE);
 }
 
+#define ALIGN_FDT(x) (((x) + (FDT_TAGSIZE) - 1) & ~((FDT_TAGSIZE) - 1))
+static char *realloc_property(char *fdt,
+                              int nodeoffset,
+                              const char *name,
+                              int newlen)
+{
+    /* FIXME: assume RAM is enough */
```
RinHizakura: I'm unsure if we need to consider the case of the preallocated memory being smaller than the required size to extend the DTB. Ideally, we can assume this won't happen.

src/riscv.c
```diff
@@ -259,6 +260,62 @@ static void map_file(char **ram_loc, const char *name)
...
+    int totalsize;
+
+    memcpy(blob, minimal, sizeof(minimal));
+
+    if (bootargs) {
+        node = fdt_path_offset(blob, "/chosen");
+        assert(node > 0);
+
+        len = strlen(bootargs) + 1;
+        buf = malloc(len);
```
ChinYikMing, line 303: Include an assertion or verify the return value of malloc, as it may fail.

.gitmodules
```diff
@@ -31,3 +31,9 @@
 	path = tests/quake
 	url = https://github.com/sysprog21/quake-embedded
 	shallow = true
+[submodule "tools/bin_to_c"]
+	path = tools/bin_to_c
+	url = https://github.com/bitbank2/bin_to_c.git
```
ChinYikMing: May I know any reason not to use `shallow = true` like other submodule? Shallow mode could speed up the download.

.gitmodules
```diff
@@ -31,3 +31,9 @@
...
+[submodule "src/dtc"]
+	path = src/dtc
+	url = https://git.kernel.org/pub/scm/utils/dtc/dtc.git
```
ChinYikMing, line 36: Ditto.

.gitmodules
jserv: Alternatively, you can simply copy the source code from `bin_2_c` since it is released under the MIT License and is unmaintained. A repository is worth adding as a submodule only when it is actively maintained.

tools/bin_to_c.c
```diff
@@ -0,0 +1,143 @@
+//
+// bin_to_c - convert binary data into c-compatible data tables
+//
+// Written by Larry Bank
+// Copyright (c) 2009 BitBank Software, Inc.
+// Change history
+// 2/1/09 - Started the project
+// 11/6/15 - converted to Linux
+// 7/29/20 - tailored to Arduino FLASH data output
+//
```
jserv: * Use C-style comments.
* Put the original copyright notice: https://github.com/bitbank2/bin_to_c/blob/master/LICENSE and keep "Written by Larry Bank" and "Copyright (c) 2015 BitBank Software, Inc."
* Remove the change history for short.

mk/system.mk
```diff
@@ -1,13 +1,23 @@
...
-$(OUT)/minimal.dtb: $(DEV_SRC)/minimal.dts
+BUILD_DTB := $(OUT)/minimal.dtb
+$(BUILD_DTB): $(DEV_SRC)/minimal.dts
 	$(VECHO) " DTC\t$@\n"
 	$(Q)$(DTC) $^ -o $@
-BUILD_DTB := $(OUT)/minimal.dtb
+
+BIN_TO_C := $(OUT)/bin_to_c
+$(BIN_TO_C): tools/bin_to_c.c
+	$(Q)$(CC) -Wall -o $@ $^
```
jserv: Rename from `bin_to_c` to `bin2c` for consistency.

jserv:

PR #535, opened 2025-01-06
Fix build failures and expand test coverage

eleanorLYJ:
# Fix build issues and enhance build system tests
Improve our build system testing and fix various build issues discovered in the process

### CI Improvements
- Add compiler matrix to test with both gcc-latest and clang-latest
- Set warnings as errors to catch potential issues early
- Add diverse build configurations testing

### Build Fixes
- Fix linker error when using Clang by adding missing `$(LDFLAGS)`
- Resolve multiple compiler warning, including:
    - Enabling JIT with specific features disabled, such as Zba, Zbb, Zbc, Zbs, EXT_M, or Zicsr
    - Disabling Zicsr 
    - Disabling MOP_FUSIO
- Resolve multiple Clang-specific build errors, including:
    - Enabling JIT when EXT_M or EXT_C disabled
    - Disabling BLOCK_CHAINING
    - Disabling MOP FUSION
    - Disabling Zicsr

## Testing Done
- Test builds with various feature flags disabled
- Confirm CI pipeline runs successfully with all new test cases

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -82,6 +82,7 @@ jobs:
             make distclean && make ENABLE_SDL=0 check -j$(nproc)
             make distclean && make ENABLE_Zicsr=0 check -j$(nproc)
             make distclean && make ENABLE_MOP_FUSION=0 check -j$(nproc)
+            make distclean && make ENABLE_BLOCK_CHAINING=0 check -j$(nproc)
     - name: misalignment test in block emulation
       env:
         CC: ${{ steps.install_cc.outputs.cc }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -106,6 +106,7 @@ jobs:
             make ENABLE_JIT=1 clean && make ENABLE_EXT_F=0 ENABLE_JIT=1 check -j$(nproc)
             make ENABLE_JIT=1 clean && make ENABLE_EXT_C=0 ENABLE_JIT=1 check -j$(nproc)
             make ENABLE_JIT=1 clean && make ENABLE_Zba=0 ENABLE_JIT=1 check -j$(nproc)
+            make ENABLE_JIT=1 clean && make ENABLE_Zbb=0 ENABLE_JIT=1 check -j$(nproc)
     - name: undefined behavior test
       run: |
             make clean && make ENABLE_UBSAN=1 check -j$(nproc)
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -80,6 +80,7 @@ jobs:
             make distclean && make ENABLE_EXT_F=0 check -j$(nproc)
             make distclean && make ENABLE_EXT_C=0 check -j$(nproc)
             make distclean && make ENABLE_SDL=0 check -j$(nproc)
+            make distclean && make ENABLE_Zicsr=0 check -j$(nproc)
     - name: misalignment test in block emulation
       env:
         CC: ${{ steps.install_cc.outputs.cc }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -105,6 +105,7 @@ jobs:
             make ENABLE_JIT=1 clean && make ENABLE_EXT_A=0 ENABLE_JIT=1 check -j$(nproc)
             make ENABLE_JIT=1 clean && make ENABLE_EXT_F=0 ENABLE_JIT=1 check -j$(nproc)
             make ENABLE_JIT=1 clean && make ENABLE_EXT_C=0 ENABLE_JIT=1 check -j$(nproc)
+            make ENABLE_JIT=1 clean && make ENABLE_Zba=0 ENABLE_JIT=1 check -j$(nproc)
     - name: undefined behavior test
       run: |
             make clean && make ENABLE_UBSAN=1 check -j$(nproc)
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -87,6 +87,7 @@ jobs:
             make distclean && make ENABLE_Zbb=0 check -j$(nproc)
             make distclean && make ENABLE_Zbc=0 check -j$(nproc)
             make distclean && make ENABLE_Zbs=0 check -j$(nproc)
+            make distclean && make ENABLE_Zifencei=0 check -j$(nproc)
     - name: misalignment test in block emulation
       env:
         CC: ${{ steps.install_cc.outputs.cc }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -86,6 +86,7 @@ jobs:
             make distclean && make ENABLE_Zba=0 check -j$(nproc)
             make distclean && make ENABLE_Zbb=0 check -j$(nproc)
             make distclean && make ENABLE_Zbc=0 check -j$(nproc)
+            make distclean && make ENABLE_Zbs=0 check -j$(nproc)
     - name: misalignment test in block emulation
       env:
         CC: ${{ steps.install_cc.outputs.cc }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -120,6 +120,7 @@ jobs:
             make ENABLE_JIT=1 clean && make ENABLE_Zbs=0 ENABLE_JIT=1 check -j$(nproc)
             make ENABLE_JIT=1 clean && make ENABLE_Zicsr=0 ENABLE_JIT=1 check -j$(nproc)
             make ENABLE_JIT=1 clean && make ENABLE_Zifencei=0 ENABLE_JIT=1 check -j$(nproc)
+            make ENABLE_JIT=1 clean && make ENABLE_MOP_FUSION=0 ENABLE_JIT=1 check -j$(nproc)
     - name: undefined behavior test
       run: |
             make clean && make ENABLE_UBSAN=1 check -j$(nproc)
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -107,6 +107,7 @@ jobs:
             make ENABLE_JIT=1 clean && make ENABLE_EXT_C=0 ENABLE_JIT=1 check -j$(nproc)
             make ENABLE_JIT=1 clean && make ENABLE_Zba=0 ENABLE_JIT=1 check -j$(nproc)
             make ENABLE_JIT=1 clean && make ENABLE_Zbb=0 ENABLE_JIT=1 check -j$(nproc)
+            make ENABLE_JIT=1 clean && make ENABLE_Zbc=0 ENABLE_JIT=1 check -j$(nproc)
     - name: undefined behavior test
       run: |
             make clean && make ENABLE_UBSAN=1 check -j$(nproc)
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -119,6 +119,7 @@ jobs:
             make ENABLE_JIT=1 clean && make ENABLE_Zbc=0 ENABLE_JIT=1 check -j$(nproc)
             make ENABLE_JIT=1 clean && make ENABLE_Zbs=0 ENABLE_JIT=1 check -j$(nproc)
             make ENABLE_JIT=1 clean && make ENABLE_Zicsr=0 ENABLE_JIT=1 check -j$(nproc)
+            make ENABLE_JIT=1 clean && make ENABLE_Zifencei=0 ENABLE_JIT=1 check -j$(nproc)
     - name: undefined behavior test
       run: |
             make clean && make ENABLE_UBSAN=1 check -j$(nproc)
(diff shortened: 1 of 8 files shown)


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: 99c52b9f4e89afb58a185ac9309a96cfee3ac17f | Previous: 3f37f514e33891e2d35bc614fb8d33035f2a4e24 | Ratio |
|-|-|-|-|
| `Dhrystone` | `1327` Average DMIPS over 10 runs | `1317` Average DMIPS over 10 runs | `0.99` |
| `Coremark` | `954.362` Average iterations/sec over 10 runs | `956.978` Average iterations/sec over 10 runs | `1.00` |

</details>

.github/workflows/main.yml
```diff
@@ -49,43 +53,77 @@ jobs:
...
       run: |
             make ENABLE_JIT=1 clean && make ENABLE_JIT=1 check -j$(nproc)
             make ENABLE_JIT=1 clean && make ENABLE_EXT_A=0 ENABLE_JIT=1 check -j$(nproc)
             make ENABLE_JIT=1 clean && make ENABLE_EXT_F=0 ENABLE_JIT=1 check -j$(nproc)
             make ENABLE_JIT=1 clean && make ENABLE_EXT_C=0 ENABLE_JIT=1 check -j$(nproc)
+            make ENABLE_JIT=1 clean && make ENABLE_EXT_M=0 ENABLE_JIT=1 check -j$(nproc)
+            make ENABLE_JIT=1 clean && make ENABLE_Zba=0 ENABLE_JIT=1 check -j$(nproc)
+            make ENABLE_JIT=1 clean && make ENABLE_Zbb=0 ENABLE_JIT=1 check -j$(nproc)
+            make ENABLE_JIT=1 clean && make ENABLE_Zbc=0 ENABLE_JIT=1 check -j$(nproc)
+            make ENABLE_JIT=1 clean && make ENABLE_Zbs=0 ENABLE_JIT=1 check -j$(nproc)
```
visitorckw, line 126: How about repeating all diverse configuration tests with JIT enabled?

.github/workflows/main.yml
```diff
@@ -49,43 +53,77 @@ jobs:
...
             make distclean && make ENABLE_EXT_F=0 check -j$(nproc)
             make distclean && make ENABLE_EXT_C=0 check -j$(nproc)
             make distclean && make ENABLE_SDL=0 check -j$(nproc)
+            make distclean && make ENABLE_Zicsr=0 check -j$(nproc)
+            make distclean && make ENABLE_MOP_FUSION=0 check -j$(nproc)
+            make distclean && make ENABLE_BLOCK_CHAINING=0 check -j$(nproc)
+            make distclean && make ENABLE_Zba=0 check -j$(nproc)
+            make distclean && make ENABLE_Zbb=0 check -j$(nproc)
+            make distclean && make ENABLE_Zbc=0 check -j$(nproc)
+            make distclean && make ENABLE_Zbs=0 check -j$(nproc)
```
visitorckw, line 91: Maybe also add:
`make distclean && make ENABLE_Zifencei=0 check -j$(nproc)`

visitorckw:
Hmmm, it seems CI fails due to insufficient memory. Any ideas on how we can reduce the memory usage?

jserv:
> Hmmm, it seems CI fails due to insufficient memory. Any ideas on how we can reduce the memory usage?

@henrybear327, your turn for tweaking memory usage.

Makefile
```diff
@@ -7,7 +7,7 @@ BIN := $(OUT)/rv32emu
 CONFIG_FILE := $(OUT)/.config
 -include $(CONFIG_FILE)
 
-CFLAGS = -std=gnu99 -O2 -Wall -Wextra
```
henrybear327, line 10: Since we are modifying this, do we also want to add `-Wshadow` to catch more problems :) ?

.github/workflows/main.yml
```diff
@@ -36,6 +36,10 @@ jobs:
   host-x64:
     needs: [detect-code-related-file-changes]
     if: needs.detect-code-related-file-changes.outputs.has_code_related_changes == 'true'
+    strategy:
+      fail-fast: false
+      matrix:
+        compiler: [gcc-latest, clang-latest]
```
jserv: Any reason why ordinary `gcc` and `clang` fail to work?

Makefile
visitorckw: I think this depends on whether our coding style aims to forbid local variables from shadowing other variables?

visitorckw:
> Hmmm, it seems CI fails due to insufficient memory. Any ideas on how we can reduce the memory usage?

Sorry for the misleading description. The CI failure is due to the addition of more tests, causing excessive http requests to github.com during the fetch-checksum, which results in a 403 too many requests error. Here’s the error log:

```
--2025-01-10 17:19:46--  https://api.github.com/repos/sysprog21/rv32emu-prebuilt/releases
Resolving api.github.com (api.github.com)... 140.82.114.6
Connecting to api.github.com (api.github.com)|140.82.114.6|:443... connected.
HTTP request sent, awaiting response... 403 rate limit exceeded
2025-01-10 17:19:46 ERROR 403: rate limit exceeded.
```

eleanorLYJ:
Thank you for your help. I will fix it.

eleanorLYJ:
> Sorry for the misleading description. The CI failure is due to the addition of more tests, causing excessive http requests to github.com during the fetch-checksum, which results in a 403 too many requests error. Here’s the error log:
> 
> ```
> --2025-01-10 17:19:46--  https://api.github.com/repos/sysprog21/rv32emu-prebuilt/releases
> Resolving api.github.com (api.github.com)... 140.82.114.6
> Connecting to api.github.com (api.github.com)|140.82.114.6|:443... connected.
> HTTP request sent, awaiting response... 403 rate limit exceeded
> 2025-01-10 17:19:46 ERROR 403: rate limit exceeded.
> ```

I noticed that we make frequent requests to fetch the checksum file and store it in $(BIN_DIR)/sha1sum-linux-image.

To reduce the number of requests, we could modify the code to check if the checksum file already exists and is less than a day old. If it’s up-to-date, we can skip fetching it again. This way, we minimize unnecessary requests.

Any thoughts?

jserv:
> I noticed that we make frequent requests to fetch the checksum file and store it in $(BIN_DIR)/sha1sum-linux-image.
> 
> To reduce the number of requests, we could modify the code to check if the checksum file already exists and is less than a day old. If it’s up-to-date, we can skip fetching it again. This way, we minimize unnecessary requests.

Yes, create an issue for tracking.
Cc. @vacantron

ChinYikMing:
I'll help review the pull request this week. Thanks.

visitorckw (approved):
Except for the occasional rate limit exceed issue when fetching checksums in CI, everything else LGTM!

I'm surprised there are so many compilation issues. Thanks for your work!

.github/workflows/main.yml
jserv: ?

eleanorLYJ: Are you asking if it's better to avoid using the latest gcc/clang and instead use older, more stable version?

jserv: > Are you asking if it's better to avoid using the latest gcc/clang and instead use older, more stable version?

Yes, using the clang/gcc compilers from official Linux distribution packages makes it easier to track and report bugs, as we can identify build failures specific to the distribution versions we use in our CI pipeline.

eleanorLYJ: Do you have a preference for a specific version of gcc or clang?

jserv: > Do you have a preference for a specific version of gcc or clang?

Use the versions of gcc/clang provided by selected Linux distributions.

src/emulate.c
```diff
@@ -688,6 +708,7 @@ static void block_translate(riscv_t *rv, block_t *block)
         remove_next_nth_ir(rv, ir, block, count - 1);             \
     }
 
+#if RV32_HAS(MOP_FUSION)
```
ChinYikMing: Move this to line 686, as `COMBINE_MEM_OPS` macro is only utilized when `MOP_FUSION` is enabled and for readability, such that:
```c
#if RV32_HAS(MOP_FUSION)
#define COMBINE_MEM_OPS(RW)
...
```

src/emulate.c
```diff
@@ -702,7 +723,9 @@ static inline void remove_next_nth_ir(const riscv_t *rv,
         block->ir_tail = ir;
     block->n_insn -= n;
 }
+#endif
```
ChinYikMing: Remove this since line 826 cover the condition.

src/emulate.c
```diff
@@ -702,7 +723,9 @@ static inline void remove_next_nth_ir(const riscv_t *rv,
...
 
+#if RV32_HAS(MOP_FUSION)
```
ChinYikMing: Remove this since line 686 cover this.

src/emulate.c
```diff
@@ -79,12 +81,14 @@ static void __trap_handler(riscv_t *rv);
...
 
 /* FIXME: use more precise methods for updating time, e.g., RTC */
+#if RV32_HAS(Zicsr)
 static uint64_t ctr = 0;
 static inline void update_time(riscv_t *rv)
 {
     rv->csr_time[0] = ctr & 0xFFFFFFFF;
     rv->csr_time[1] = ctr >> 32;
 }
+#endif
```
ChinYikMing: Remove this since line 84 covers the condition.

src/emulate.c
```diff
@@ -79,12 +81,14 @@ static void __trap_handler(riscv_t *rv);
...
 
 #if RV32_HAS(Zicsr)
```
ChinYikMing: Ditto.

ChinYikMing:
May I know if the remaining tests will still be executed in case any single test fails?

src/emulate.c
eleanorLYJ: Thank you for your comment.
I noticed that line 686 isn’t protected by the macro (if RV32_HAS(MOP_FUSION)). Perhaps you mean the code starting from line 711 is protected?

src/emulate.c
eleanorLYJ: Ack

src/emulate.c
eleanorLYJ: Ack

src/emulate.c
eleanorLYJ: Ack

eleanorLYJ:
> May I know if the remaining tests will still be executed in case any single test fails?

Based on the current configuration, if any single test fails, the remaining tests will not be executed. Should we consider allowing the other tests to continue running even if one test fails?

src/emulate.c
ChinYikMing: > Thank you for your comment. I noticed that line 686 isn’t protected by the macro (if RV32_HAS(MOP_FUSION)). Perhaps you mean the code starting from line 711 is protected?

Check this:
```c
#if RV32_HAS(MOP_FUSION)
#define COMBINE_MEM_OPS(RW)
...
```

ChinYikMing:
> > May I know if the remaining tests will still be executed in case any single test fails?
> 
> Based on the current configuration, if any single test fails, the remaining tests will not be executed. Should we consider allowing the other tests to continue running even if one test fails?

Yes, I think revealing all errors or warnings at once for each build flag is a better approach. Moreover, the validation process for both compilers only takes around 3 to 5 minutes based on previous actions log.

src/emulate.c
eleanorLYJ: You're right. The macro should start covering from line 686 instead, as 'remove_next_nth_ir' is used inside the '#define COMBINE_MEM_OPS(RW)'.

ChinYikMing:
In 'mk/toolchain.mk', we have `emcc` (Emscripten). @eleanorLYJ Can you also add the `emcc` to the CI? The reason is that 'deploy-wasm.yml' is only executed when specific files are changed or manually triggered. By detecting warnings or errors early, we can address them before the actual execution of 'deploy-wasm.yml'.

eleanorLYJ:
> In 'mk/toolchain.mk', we have `emcc` (Emscripten). @eleanorLYJ Can you also add the `emcc` to the CI? The reason is that 'deploy-wasm.yml' is only executed when specific files are changed or manually triggered. By detecting warnings or errors early, we can address them before the actual execution of 'deploy-wasm.yml'.

The current action we are using, `rlalik/setup-cpp-compiler`, does not support installing `emcc` at the moment. I plan to address this limitation in the next PR by implementing functionality to include `emcc` in the CI setup.

ChinYikMing:
> > In 'mk/toolchain.mk', we have `emcc` (Emscripten). @eleanorLYJ Can you also add the `emcc` to the CI? The reason is that 'deploy-wasm.yml' is only executed when specific files are changed or manually triggered. By detecting warnings or errors early, we can address them before the actual execution of 'deploy-wasm.yml'.
> 
> The current action we are using, `rlalik/setup-cpp-compiler`, does not support installing `emcc` at the moment. I plan to address this limitation in the next PR by implementing functionality to include `emcc` in the CI setup.

Got it. Looking forward to that!

ChinYikMing:
LGTM.

ChinYikMing (approved):
LGTM.

jserv:

PR #543, opened 2025-01-24
Code cleanup

ChinYikMing:
src/riscv.h contains repeated definitions (might be due to the rebase), thus removing them.


jserv:

PR #545, opened 2025-01-24
Support partial vector extension instructions

### LINKED ISSUE #504: Implement RISC-V Vector Extension
RISC-V "V" Vector Extension (shorten as `RVV`) is the instruction extension that introduces vector process capabilities to RISCV. 

The staged goal of this issue is to explore the possibility to run RVV instructions with rv32emu.
* Configure `vtype` and `vl` CSRs with vsetvli/vsetivli/vsetvl. This is required because RVV vector register is variable-legnth.
* [Unit-Stride store/load instructions](https://github.com/riscv/riscv-isa-manual/blob/ae98787c8d718260ecde6345d8468ed96b1a843c/src/v-st-ext.adoc#vector-unit-stride-instructions). This can help us to build the minimum POC.
* Basic integer ALU instructions ( vadd, vsub, vmin, vmax, vand, vor, vxor) and permutation instructions (vrgather, vslideup, vslidedown)
 
To achieve the goals we need to 
1. Implement the instruction decoder (in `decode.c`). This decoder should be able to decode all three instructions formats (valu, vcfg, vmem)
2. Add the corresponding CSRs (`vl`, `vtype`)
3. Implement RVV instructions into custom IRs. Some of RVV may be implemented as a loop of basic instructions (like `vadd` from `add`). To maintain the readability of throughout the entire codebase, adding a new src file `rvv_template.c` for the RVV's IRs may be a better idea.
4. Study the further implementations (exception handling, rounding mode, float-points, comparison, mask instructions, etc.)

We can refer the following resource 
1. RVV spec https://github.com/riscv/riscv-isa-manual/blob/main/src/v-st-ext.adoc
2. instruction tables https://github.com/riscv/riscv-isa-manual/blob/main/src/images/wavedrom/v-inst-table.edn
3. alu instructions formats https://github.com/riscv/riscv-isa-manual/blob/main/src/images/wavedrom/valu-format.edn
5. vector configuration instructions format https://github.com/riscv/riscv-isa-manual/blob/main/src/images/wavedrom/vcfg-format.edn
6. Spike's bit masks for decoding instructions https://github.com/riscv-software-src/riscv-isa-sim/blob/master/riscv/encoding.h

vestata:
Add support for the RISC-V "V" Vector Extension. This pull request implements decoding for 585 out of 616 version 1.0 spec vector instructions, with partial interpreter implementation. 

The decoding method for vector instructions, including vector configuration and load/store instructions, follows the approach used in `rv32emu`. The new `rvv_jumptable` is introduced to handle remaining arithmetic instructions.

The interpreter implementation is tested using the [riscv-vector-tests repository](https://github.com/chipsalliance/riscv-vector-tests), with current limitations, as outlined in the repo. Included partial support for vector load/store instructions and single-width arithmetic instructions. The architecture now supports different settings for `sew`, `lmul`, and vector masking.

Vector instructions passing the tests include:
- `vle8.v`, `vle16.v`, `vle32.v`
- `vse8.v`, `vse16.v`, `vse32.v`
- `vadd.vv`, `vadd.vx`, `vadd.vi`
- `vsub.vv`, `vsub.vx`, `vsub.vi`,
- `vand.vv`, `vand.vx`, `vand.vi`
- `vor.vv`, `vor.vx`, `vor.vi`
- `vxor.vv`, `vxor.vx`, `vxor.vi`
- `vsll.vv`, `vsll.vx`, `vsll.vi`
- `vmul.vv`, `vmul.vx`, `vmul.vi`

Close #504 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request implements extensive support for the RISC-V Vector Extension, decoding 585 out of 616 vector instructions. It enhances vector load/store operations, arithmetic instructions, and configuration management, while introducing a new jump table for remaining instructions. Many operations are still placeholders, indicating areas for future development.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
</div>


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: c3374eaa45548cd84ec8a0903b9e99019c063640 | Previous: 4ef61b8966268a765455d5e30c5e0bdb18eca93e | Ratio |
|-|-|-|-|
| `Dhrystone` | `1256` Average DMIPS over 10 runs | `1284` Average DMIPS over 10 runs | `1.02` |
| `Coremark` | `946.035` Average iterations/sec over 10 runs | `972.508` Average iterations/sec over 10 runs | `1.03` |

</details>

src/decode.c
```diff
@@ -3,10 +3,10 @@
  * "LICENSE" for information on usage and redistribution of this file.
  */
 
+#include "decode.h"
```
jserv: Don't change this.

src/decode.c
```diff
@@ -1971,73 +1937,2326 @@ static inline bool op_cfsw(rv_insn_t *ir, const uint32_t insn)
 #define op_cflwsp OP_UNIMP
 #endif /* RV32_HAS(EXT_C) && RV32_HAS(EXT_F) */
 
-/* handler for all unimplemented opcodes */
-static inline bool op_unimp(rv_insn_t *ir UNUSED, uint32_t insn UNUSED)
+
+#if RV32_HAS(EXT_V) /* (RV32_HAS(EXT_V) */
+
```
jserv: Remove unnecessary blank lines like this.

src/decode.c
```diff
@@ -1971,73 +1937,2326 @@ static inline bool op_cfsw(rv_insn_t *ir, const uint32_t insn)
...
+    default: /* illegal instruction */
+        return false;
     }
-#endif
+    return true;
+}
 
-    /* standard uncompressed instruction */
-    const uint32_t index = (insn & INSN_6_2) >> 2;
+static inline bool op_000000(rv_insn_t *ir, const uint32_t insn)
```
jserv, line 2503: `op_000000` looks misleading. Can you improve its naming scheme?

jserv (changes requested):
Check [CONTRIBUTING.md](https://github.com/sysprog21/rv32emu/blob/master/CONTRIBUTING.md) carefully.

src/riscv.h
```diff
@@ -289,6 +289,12 @@ typedef uint32_t riscv_exception_t;
 #if RV32_HAS(EXT_F)
 typedef softfloat_float32_t riscv_float_t;
 #endif
+#if RV32_HAS(EXT_V)
+#ifndef VLEN
```
ChinYikMing: This check is redundant since you have checked in Makefile.

vacantron:
> The interpreter implementation is tested using the [riscv-vector-tests repository](https://github.com/chipsalliance/riscv-vector-tests)

Could we create an CI like using [ROSCOF](https://github.com/riscv-software-src/riscof) for this?

src/decode.c
vestata: The naming scheme is based on the `function6` field listed in [riscv-v-spec/inst-table.adoc](https://github.com/riscvarchive/riscv-v-spec/blob/master/inst-table.adoc). Since each `function6` may include `OPI`, `OPM`, or `OPF` functions, often corresponding to unrelated operations. I chose to name them directly based on the `function6` for consistency.

This might seem unclear without additional context. To improve clarity, I could add comments explaining the naming convention for each `op_function6`. Would this address your concern?

src/decode.c
vestata: Reverted this change in the latest update. Thanks!

src/decode.c
vestata: Unnecessary blank lines in `src/decode.c` have been removed in the latest update. Thanks!

src/riscv.h
vestata: Redundant `VLEN` check removed in the latest update. Thank!

vestata:
> > The interpreter implementation is tested using the [riscv-vector-tests repository](https://github.com/chipsalliance/riscv-vector-tests)
> 
> Could we create an CI like using [ROSCOF](https://github.com/riscv-software-src/riscof) for this?

I'm not familiar with ROSCOF, but I'll look into it and give it a try.

eleanorLYJ (changes requested):
You don't need to include everything you've done in the commit message.
For example, a message like: "Apply clang-format and clean up code." is unnecessary.

src/riscv_private.h
```diff
@@ -208,6 +217,25 @@ struct riscv_internal {
...
+#if RV32_HAS(EXT_V)    
+    vector128_t V[N_RV_REGS];
+
+    /* Fixme:Temporary way to set configuration for vectors. */
+    bool vma;
+    bool vta;
+    uint8_t vsew;
+    uint8_t vlmul;
+
+    /* ???: Figure out how to control them. */
```
eleanorLYJ: Consider using FIXME or TODO instead of ??? to clearly indicate that this is a task or issue that needs to be addressed.

src/emulate.c
```diff
@@ -655,7 +655,9 @@ static void block_translate(riscv_t *rv, block_t *block)
             break;
         }
         ir->impl = dispatch_table[ir->opcode];
-        ir->pc = block->pc_end; /* compute the end of pc */
```
eleanorLYJ: No need to move the comment's position.

eleanorLYJ:
Suggest using `git rebase -i` to squash the commit into the previous one instead of adding a new commit.

src/rv32_template.c
```diff
@@ -2987,3 +2987,5181 @@ RVOP(
...
+RVOP(
+    vfwnmsac_vf,
+    {
+        for (int i = 0; i < 4; i++) {
+            rv->V[rv_reg_zero][i] = 0;
+        }
+    },
+    GEN({/* no operation */}))
+
+#endif
```
visitorckw: When using git for version control, always add a newline at the end of files.

See: https://medium.com/@alexey.inkin/how-to-force-newline-at-end-of-files-and-why-you-should-do-it-fdf76d1d090e

.gitignore
```diff
@@ -29,6 +29,7 @@ tests/**/*.elf
 tests/arch-test-target/config.ini
 tests/arch-test-target/sail_cSim/riscv_sim_RV32
 tests/scimark2/
+tests/riscv-vector-tests/
```
visitorckw: Hmm, under what circumstances is this file generated, making it necessary to add it to `.gitignore`?

src/decode.c
```diff
@@ -1971,73 +1934,2315 @@ static inline bool op_cfsw(rv_insn_t *ir, const uint32_t insn)
...
+#endif
+
+    /* decode instruction */
+    const decode_t op = rv_jump_table[index];
+    assert(op);
+    return op(ir, insn);
+
+#undef OP_UNIMP
+#undef OP
+}
```
visitorckw: Ditto.

src/decode.h
```diff
@@ -231,7 +231,603 @@ enum op_field {
...
             _(cfsw, 0, 2, 1, ENC(rs1, rs2))            \
         )                                              \
+    )                                                  \
+    /* Zve32x Standard Extension */                    \
+    IIF(RV32_HAS(EXT_V))(                              \
+        /* Configuration-setting Instructions */       \
+        _(vsetvli, 0, 4, 0, ENC(rs1, rd))              \
+        _(vsetivli, 0, 4, 0, ENC(rs1, rd))             \
+        _(vsetvl, 0, 4, 0, ENC(rs1, rd))               \
+        /* Vectore Load */                             \
```
ChinYikMing: Typo.

src/decode.c
```diff
@@ -1971,73 +1934,2315 @@ static inline bool op_cfsw(rv_insn_t *ir, const uint32_t insn)
...
+
+/* STORE-FP: S-type
+ *  31       25 24   20 19   15 14   12 11       7 6      0
+ * | imm[11:5] |  rs2  |  rs1  | width | imm[4:0] | opcode |
+ */
+static inline bool op_store_fp(rv_insn_t *ir, const uint32_t insn)
+{
+#if RV32_HAS(EXT_V)
+    /* Fixme: The implementation now is just using switch statement, since there
+     * are multiple duplicate elements in vectore load/store instruction. I'm
```
ChinYikMing: Ditto.

src/decode.h
```diff
@@ -231,7 +231,603 @@ enum op_field {
...
+        _(vloxseg7ei32_v, 0, 4, 0, ENC(rs1, vd))       \
+        _(vloxseg8ei32_v, 0, 4, 0, ENC(rs1, vd))       \
+        _(vloxseg2ei64_v, 0, 4, 0, ENC(rs1, vd))       \
+        _(vloxseg3ei64_v, 0, 4, 0, ENC(rs1, vd))       \
+        _(vloxseg4ei64_v, 0, 4, 0, ENC(rs1, vd))       \
+        _(vloxseg5ei64_v, 0, 4, 0, ENC(rs1, vd))       \
+        _(vloxseg6ei64_v, 0, 4, 0, ENC(rs1, vd))       \
+        _(vloxseg7ei64_v, 0, 4, 0, ENC(rs1, vd))       \
+        _(vloxseg8ei64_v, 0, 4, 0, ENC(rs1, vd))       \
+        /* Vectore store */                            \
```
ChinYikMing: Ditto.

vestata:
Thank you all for your feedback and suggestions! I will fix the typos, add a newline at the end of files, and remove any unnecessary elements. I also noticed that the current code does not fully meet the contributing guidelines, so I will make sure to address those issues. In addition, I will add more detailed comments in `src/decode.c` and `src/rv32_template.c` and ensure the formatting is correct.

Since some of the code was misplaced from the beginning, and as @eleanorLYJ mentioned, there are non-compliant comments in an early commit, I’m considering `git rebase -i` everything from the start. Do you have any suggestions or concerns about that approach?

I’d appreciate your guidance. Thank you!

src/riscv_private.h
vestata: Sure, I'll fix it.

src/emulate.c
vestata: Ok, I'll fix it.

src/rv32_template.c
vestata: Ok, I'll fix it. Thanks for the information.

.gitignore
vestata: This is just a file I used for testing purposes. I will remove it. Thanks for pointing it out.

visitorckw:
> Thank you all for your feedback and suggestions! I will fix the typos, add a newline at the end of files, and remove any unnecessary elements. I also noticed that the current code does not fully meet the contributing guidelines, so I will make sure to address those issues. In addition, I will add more detailed comments in `src/decode.c` and `src/rv32_template.c` and ensure the formatting is correct.
> 
> Since some of the code was misplaced from the beginning, and as @eleanorLYJ mentioned, there are non-compliant comments in an early commit, I’m considering `git rebase -i` everything from the start. Do you have any suggestions or concerns about that approach?
> 
Ack. Please go ahead.

src/rv32_template.c
```diff
@@ -2987,3 +2987,5181 @@ RVOP(
...
+            break;                                                           \
+        case 32:                                                             \
+            sew_32b_handler(des, op1, op2, op, op_type);                     \
+            break;                                                           \
+        default:                                                             \
+            break;                                                           \
+    }                                                                        \
+}
+
+#define VMV_LOOP(des, op1, op2, op, SHIFT, MASK, i, j, itr, vm)             \
```
howjmay: may I ask where this is one used?

vestata: The `VMV_LOOP` macro is used in the implementation of `vmv_v_i`(at `src/rv32_template.c`, line 6366), as the [riscv-vector-tests](https://github.com/chipsalliance/riscv-vector-tests) frequently utilize `vmv_v_i` to clear bits in vector registers during each test. This serves as a quick implementation for the `vmv_v_i` instruction.

Additionally, I noticed that the implementations of `vmv_v_*` (representing `vmv_v_v`, `vmv_v_x`, and `vmv_v_i`) can be refactored to reuse existing macros such as `VV_LOOP`, `VX_LOOP`, `VI_LOOP`, and their `_LEFT` variants (collectively referred to as `V*_LOOP` and `V*_LOOP_LEFT`). I will remove the `VMV_LOOP` and related `_LEFT` macros accordingly. Thank you for pointing this out!

jserv (changes requested):
1. Rebase the latest `master` branch to resolve Arm64 host breakage on CI.
2. Use `git rebase -i` to refine your Git commits, ensuring it accurately reflects your work and omits unnecessary details.
3. Specify your name in Pinyin for each git commit. See https://git-scm.com/docs/user-manual

src/decode.c
```diff
@@ -1199,6 +1359,145 @@ static inline bool op_amo(rv_insn_t *ir, const uint32_t insn)
...
+     *              ├── 000
+     *              │   └── vloxei<eew>.v
+     *              ├── 001
+     *              │   └── vloxseg<2>ei<eew>.v
+     *              ├── ...
+     *              └── 111
+     *                  └── vloxseg<8>ei<eew>.v
+     */
+    if (decode_funct3(insn) != 0b010) {
+        uint8_t eew = decode_eew(insn);
```
visitorckw, line 1431: `decode_eew()` returns an `int32_t` and may return `-1`, but it is assigned to a `uint8_t` variable, causing an implicit type conversion, which seems odd. Should we check if the return value is `-1` and raise an illegal instruction exception in that case?

vestata: Sorry, I made a mistake here. In `decode_eew()`, the only possible values in the switch statement are `0b000`, `0b101`, `0b110`, and `0b111`. Since there are no other possibilities, I replaced `return -1` with `__UNREACHABLE`. Thanks for pointing it out.

vestata:
In `src/rv32_template.c`, the implementation of vector instruction operations using `OPT()` relies heavily on macros, including multiple-use parameters within the macro. To address this issue, I am currently refactoring them into functions while maintaining the same logic.

src/decode.c
```diff
@@ -306,6 +306,87 @@ static inline uint16_t c_decode_cbtype_imm(const uint16_t insn)
 }
 #endif /* RV32_HAS(EXT_C) */
 
+#if RV32_HAS(EXT_V) /* RV32_HAS(EXT_V) */
```
jserv: Replace the comment `/* RV32_HAS(EXT_V) */` with something like "Vector extension."

src/decode.c
```diff
@@ -1971,67 +2384,2039 @@ static inline bool op_cfsw(rv_insn_t *ir, const uint32_t insn)
 #define op_cflwsp OP_UNIMP
 #endif /* RV32_HAS(EXT_C) && RV32_HAS(EXT_F) */
 
-/* handler for all unimplemented opcodes */
-static inline bool op_unimp(rv_insn_t *ir UNUSED, uint32_t insn UNUSED)
+#if RV32_HAS(EXT_V) /* RV32_HAS(EXT_V) */
```
jserv: Ditto. Refine the comment.

src/rv32_template.c
```diff
@@ -2988,3 +2988,5129 @@ RVOP(
     }))
 
 #endif
+
+#if RV32_HAS(EXT_V)
+#define LEN ((VLEN) >> (5))
```
jserv: Carefully define `LEN`, which may lead to redefinitions.

src/rv32_template.c
```diff
@@ -2988,3 +2988,5129 @@ RVOP(
...
+            rv->csr_vtype = rv->X[ir->rs2];
+        } else {
+            if (!ir->rd) {
+                rv->csr_vtype = rv->X[ir->rs2];
+            } else {
+                rv->csr_vl = vlmax;
+                rv->csr_vtype = rv->X[ir->rs2];
+            }
+        }
+        rv->X[ir->rd] = rv->X[ir->rs1];
```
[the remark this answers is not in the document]
vestata: You're right. In the `vset{i}vl{i}` instructions, the destination register `rv->X[ir->rd]` should be updated with `rv->csr_vl` , not with the `rv->X[ir->rs1]`.

src/rv32_template.c
```diff
@@ -2988,3 +2988,5129 @@ RVOP(
...
+            rv->csr_vtype = ir->zimm;
+        } else {
+            if (!ir->rd) {
+                rv->csr_vtype = ir->zimm;
+            } else {
+                rv->csr_vl = vlmax;
+                rv->csr_vtype = ir->zimm;
+            }
+        }
+        rv->X[ir->rd] = ir->rs1;
```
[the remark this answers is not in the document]
vestata: Ditto.

src/decode.c
```diff
@@ -306,6 +306,94 @@ static inline uint16_t c_decode_cbtype_imm(const uint16_t insn)
 }
 #endif /* RV32_HAS(EXT_C) */
 
+#if RV32_HAS(EXT_V) /* (RV32_HAS(EXT_V) */
+/* Sign-extened vector immediate */
+static inline int32_t decode_v_imm(const uint32_t insn)
+{
+    return ((int32_t) ((insn << 12) & FR4_RS3)) >> 27;
+}
```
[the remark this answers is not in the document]
vestata: This flag is a false positive. The `decode_v_imm` function correctly implements sign extension.

src/decode.c
```diff
@@ -1971,67 +2391,2037 @@ static inline bool op_cfsw(rv_insn_t *ir, const uint32_t insn)
...
+ *  31    26  25  24   20 19   15 14    12 11 7 6       0
+ * | funct6 | vm |  vs2  |  vs1  | funct3 | vd | 1010111 |
+ */
+static inline void decode_vvtype(rv_insn_t *ir, const uint32_t insn)
+{
+    ir->vs2 = decode_rs2(insn);
+    ir->vs1 = decode_rs1(insn);
+    ir->vd = decode_rd(insn);
+    ir->vm = decode_vm(insn);
+}
```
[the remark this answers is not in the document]
vestata: I don't think we need explicit vector register validation in `decode_vvtype`. GCC already performs these checks during compilation.  Validation is necessary for the interpreter, but redundant at the decode stage.

src/decode.c
```diff
@@ -1971,67 +2391,2037 @@ static inline bool op_cfsw(rv_insn_t *ir, const uint32_t insn)
...
+ *  31    26  25  24   20 19   15 14    12 11 7 6       0
+ * | funct6 | vm |  vs2  |  rs1  | funct3 | vd | 1010111 |
+ */
+static inline void decode_vxtype(rv_insn_t *ir, const uint32_t insn)
+{
+    ir->vs2 = decode_rs2(insn);
+    ir->rs1 = decode_rs1(insn);
+    ir->vd = decode_rd(insn);
+    ir->vm = decode_vm(insn);
+}
```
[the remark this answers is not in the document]
vestata: Ditto.

src/decode.c
```diff
@@ -1971,67 +2384,2039 @@ static inline bool op_cfsw(rv_insn_t *ir, const uint32_t insn)
...
-typedef bool (*decode_t)(rv_insn_t *ir, uint32_t insn);
+/* decode vsetivli
+ *  31  30 29        20 19    15 14  12 11   7 6      0
+ * | 1 |1 | zimm[11:0] |  rs1   | 111  |  rd  | opcode |
+ */
+static inline void decode_vsetivli(rv_insn_t *ir, const uint32_t insn)
+{
+    ir->zimm = decode_vsetivli_zimm(insn);
+    ir->rs1 = decode_rs1(insn);
+    ir->rd = decode_rd(insn);
```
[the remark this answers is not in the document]
vestata: Ditto. Vector register validation at the decode stage is redundant.

jserv (changes requested):
Rebase the latest `matser` branch, which resolves the recent CI failures.

vestata:
Here’s the current status:

Done:
1. Rebased onto the latest `master`.
2. Consolidated the repeated vector zero-initialization in `src/rv32_template.c`.
3. Renamed `LEN` to `VREG_U32_COUNT`.

Pending:
1. Fix the code that did not pass the CI tests.
2. Refactor the `OPT` related macros in `src/rv32_template.c`.

jserv (changes requested):
Rebase the latest `master` branch, which resolves all known CI regressions.

src/rv32_template.c
```diff
@@ -2988,3 +2988,4082 @@ RVOP(
...
+#define op_add(a, b) ((a) + (b))
+RVOP(vadd_vv, {OPT(ir->vd, ir->vs2, ir->vs1, add, VV)}, GEN({
+         assert; /* FIXME: Implement */
+     }))
+RVOP(vadd_vx, {OPT(ir->vd, ir->vs2, ir->rs1, add, VX)}, GEN({
+         assert; /* FIXME: Implement */
+     }))
+RVOP(vadd_vi, {OPT(ir->vd, ir->vs2, ir->imm, add, VI)}, GEN({
+         assert; /* FIXME: Implement */
+     }))
```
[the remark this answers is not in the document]
vestata: Invalid issue.

src/decode.c
```diff
@@ -1971,67 +2384,2039 @@ static inline bool op_cfsw(rv_insn_t *ir, const uint32_t insn)
...
-        const uint16_t c_index = (insn & FC_FUNC3) >> 11 | (insn & FC_OPCODE);
+/* decode vector mask instructions with single vector operand
+ *  31    26  25  24   20 19   15 14    12 11 7 6       0
+ * | funct6 | vm |  vs2  |  rs1  | funct3 | rd | 1010111 |
+ */
+static inline void decode_mtype(rv_insn_t *ir, const uint32_t insn)
+{
+    ir->vs2 = decode_rs2(insn);
+    ir->rd = decode_rd(insn);
+    ir->vm = decode_vm(insn);
```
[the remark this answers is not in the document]
vestata: Invalid issue.

src/rv32_template.c
```diff
@@ -2988,3 +2988,4078 @@ RVOP(
...
+        uint8_t v_lmul = ir->zimm & 0b111;
+        uint8_t v_sew = (ir->zimm >> 3) & 0b111;
+
+        if (v_lmul == 4 || v_sew >= 4) {
+            /* Illegal setting */
+            rv->csr_vl = 0;
+            rv->csr_vtype = 0x80000000;
+            return true;
+        }
+        uint16_t vlmax = (v_lmul < 4)
```
howjmay, line 3149: Maybe some comment for this part?

src/rv32_template.c
```diff
@@ -2988,3 +2988,4078 @@ RVOP(
...
+ * which 32-bit word within the vector register is processed; when it reaches
+ * the end of the current VLEN vector register, __j increments to move on to the
+ * next vector register, and __i resets.
+ *
+ * The current approach supports only SEW values of 8, 16, and 32 bits.
+ */
+
+/* clang-format off */
+
+#define OPT(des, op1, op2, op, op_type) {                                    \
```
howjmay: Not a problem, but on thing I concern is some more complex RVV instructions can't be simply implemented by this macro. 
These kinds of instructions include but not limit to vmerge, nclip, etc.
Not sure what is your plan for these instructions. 
And maybe rename or add some comment to identify the limitation of `OPT()`.

howjmay:
In general is an amazing PR. You made some incredible contribution

src/decode.c
```diff
@@ -1199,6 +1352,145 @@ static inline bool op_amo(rv_insn_t *ir, const uint32_t insn)
...
+     *              ├── 001
+     *              │   └── vloxseg<2>ei<eew>.v
+     *              ├── ...
+     *              └── 111
+     *                  └── vloxseg<8>ei<eew>.v
+     */
+    if (decode_funct3(insn) != 0b010) {
+        uint8_t eew = decode_eew(insn);
+        ir->eew = 8 << eew;
+        uint8_t nf = decode_nf(insn);
```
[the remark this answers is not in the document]
vestata: Invalid issue.

src/decode.c
```diff
@@ -1971,67 +2384,2039 @@ static inline bool op_cfsw(rv_insn_t *ir, const uint32_t insn)
...
+/* decode vsetvli
+ *  31  30        20 19   15 14  12 11   7 6      0
+ * | 0 | zimm[11:0] |  rs1  | 111  |  rd  | opcode |
+ */
+static inline void decode_vsetvli(rv_insn_t *ir, const uint32_t insn)
 {
-    return false;
+    ir->zimm = decode_vsetvli_zimm(insn);
+    ir->rs1 = decode_rs1(insn);
+    ir->rd = decode_rd(insn);
```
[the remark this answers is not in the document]
vestata: Invalid issue.

src/rv32_template.c
```diff
@@ -2988,3 +2988,4078 @@ RVOP(
...
+                rv->csr_vl = vlmax;
+                rv->csr_vtype = ir->zimm;
+            }
+        }
+        rv->X[ir->rd] = rv->csr_vl;
+    },
+    GEN({
+        assert; /* FIXME: Implement */
+    }))
+RVOP(
```
jserv, line 3137: Separate with one blank line for each `RVOP` definition.

src/rv32_template.c
```diff
@@ -2988,3 +2988,4078 @@ RVOP(
...
+ * for each sub-element, and MASK filters out the bits that belong to a single
+ * sub-element, such as 0xFF for 8-bit or 0xFFFF for 16-bit. Index __i tracks
+ * which 32-bit word within the vector register is processed; when it reaches
+ * the end of the current VLEN vector register, __j increments to move on to the
+ * next vector register, and __i resets.
+ *
+ * The current approach supports only SEW values of 8, 16, and 32 bits.
+ */
+
+/* clang-format off */
```
jserv: Remove `clang-format off`. That is, indent the macro.

src/rv32_template.c
jserv: The name `OPT` is not precise. It is a matter of `sew` variant.

src/rv32_template.c
```diff
@@ -2988,3 +2988,4078 @@ RVOP(
...
+        __k += 1;                                                           \
+        __i++;                                                              \
+        if (!(__k % (VREG_U32_COUNT))) {                                    \
+            __j++;                                                          \
+            __i = 0;                                                        \
+        }                                                                   \
+        vm >>= 1;                                                           \
+    }                                                                       \
+}
+/* clang-format on */
```
jserv: Avoid marking `clang-format off` for long code listing.

src/rv32_template.c
```diff
@@ -2988,3 +2988,4078 @@ RVOP(
     }))
 
 #endif
+
+#if RV32_HAS(EXT_V)
+#define NO_IMP                       \
```
jserv: Rename to `V_NOP`.

src/rv32_template.c
```diff
@@ -2988,3 +2988,4078 @@ RVOP(
...
+    { NO_IMP; },
+    GEN({
+        assert; /* FIXME: Implement */
+    }))
+RVOP(
+    vfwnmsac_vf,
+    { NO_IMP; },
+    GEN({
+        assert; /* FIXME: Implement */
+    }))
```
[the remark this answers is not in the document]
vestata: Invalid issue.

jserv:
Reported by [Clang Static Analyzer](https://clang-analyzer.llvm.org/):
```
src/decode.c:348:23: error: unused function 'decode_mop' [-Werror,-Wunused-function]
  348 | static inline uint8_t decode_mop(const uint32_t insn)
      |                       ^~~~~~~~~~
src/decode.c:356:23: error: unused function 'decode_eew' [-Werror,-Wunused-function]
  356 | static inline uint8_t decode_eew(const uint32_t insn)
      |                       ^~~~~~~~~~
src/decode.c:376:23: error: unused function 'decode_nf' [-Werror,-Wunused-function]
  376 | static inline uint8_t decode_nf(const uint32_t insn)
      |                       ^~~~~~~~~
src/decode.c:384:23: error: unused function 'decode_24_[20](https://github.com/sysprog21/rv32emu/actions/runs/13313707728/job/37318856634?pr=545#step:4:21)' [-Werror,-Wunused-function]
  384 | static inline uint8_t decode_24_20(const uint32_t insn)
      |                       ^~~~~~~~~~~~
src/decode.c:474:20: error: unused function 'decode_VL' [-Werror,-Wunused-function]
  474 | static inline void decode_VL(rv_insn_t *ir, const uint32_t insn)
      |                    ^~~~~~~~~
src/decode.c:485:20: error: unused function 'decode_VLS' [-Werror,-Wunused-function]
  485 | static inline void decode_VLS(rv_insn_t *ir, const uint32_t insn)
      |                    ^~~~~~~~~~
src/decode.c:497:20: error: unused function 'decode_VLX' [-Werror,-Wunused-function]
  497 | static inline void decode_VLX(rv_insn_t *ir, const uint[32](https://github.com/sysprog21/rv32emu/actions/runs/13313707728/job/37318856634?pr=545#step:4:33)_t insn)
      |                    ^~~~~~~~~~
src/decode.c:509:20: error: unused function 'decode_VS' [-Werror,-Wunused-function]
  509 | static inline void decode_VS(rv_insn_t *ir, const uint32_t insn)
      |                    ^~~~~~~~~
src/decode.c:520:20: error: unused function 'decode_VSS' [-Werror,-Wunused-function]
  520 | static inline void decode_VSS(rv_insn_t *ir, const uint32_t insn)
      |                    ^~~~~~~~~~
src/decode.c:532:20: error: unused function 'decode_VSX' [-Werror,-Wunused-function]
  532 | static inline void decode_VSX(rv_insn_t *ir, const uint32_t insn)
      |                    ^~~~~~~~~~
10 errors generated.
```

src/rv32_template.c
vestata: Got it.

src/rv32_template.c
vestata: Got it.

src/rv32_template.c
vestata: Got it. I'll add some detail to the `vlmax` calculation.

src/rv32_template.c
vestata: >Not a problem, but on thing I concern is some more complex RVV instructions can't be simply implemented by this macro.
These kinds of instructions include but not limit to vmerge, nclip, etc.
Not sure what is your plan for these instructions.
And maybe rename or add some comment to identify the limitation of OPT().

You're right. The implemetation right now can't handle wideing/narrowing vector instructions, also `vmerge` or `vnclip`. The original plan is to handle these instructions in by introducing WV_LOOP

vestata: >Not a problem, but on thing I concern is some more complex RVV instructions can't be simply implemented by this macro.
These kinds of instructions include but not limit to vmerge, nclip, etc.
Not sure what is your plan for these instructions.
And maybe rename or add some comment to identify the limitation of OPT().

You're right. The current implementation doesn't support widening/narrowing vector instructions, as well as `vmerge` and `vnclip`. The original plan is to introduce variants, such as `WV_LOOP`, to handle these cases. However, this PR focuses on the basic implementation of vector instructions, so I'll leave this for future work.

vestata: > And maybe rename or add some comment to identify the limitation of `OPT()`.

> The name `OPT` is not precise. It is a matter of sew variant.

I'll rename `OPT` to `VECTOR_DISPATCH`, but let me know if you have a better suggestion.

visitorckw:
> > > The interpreter implementation is tested using the [riscv-vector-tests repository](https://github.com/chipsalliance/riscv-vector-tests)
> > 
> > 
> > Could we create an CI like using [ROSCOF](https://github.com/riscv-software-src/riscof) for this?
> 
> I'm not familiar with ROSCOF, but I'll look into it and give it a try.

The current [riscv-arch-test](https://github.com/riscv-non-isa/riscv-arch-test) does not include a test suite for vector extension. However, we could explore using [riscv-ctg](https://github.com/riscv-software-src/riscv-ctg) to generate a suitable test suite.

Makefile
```diff
@@ -142,6 +142,16 @@ $(call set-feature, EXT_C)
 ENABLE_RV32E ?= 0
 $(call set-feature, RV32E)
 
+# Vector extension instructions
+ENABLE_EXT_V ?= 0
+$(call set-feature, EXT_V)
+VLEN ?= 128 # Default VLEN is 128
```
ChinYikMing, line 148: Shall this moved into the conditional block of `ifeq ($(call has, EXT_V), 1)` ?

src/rv32_template.c
```diff
@@ -3031,3 +3031,4644 @@ RVOP(
...
+            break;                                       \
+        case 32:                                         \
+            sew_32b_handler(des, op1, op2, op, op_type); \
+            break;                                       \
+        default:                                         \
+            break;                                       \
+        }                                                \
+    }
+
+#define VI_LOOP(des, op1, op2, op, SHIFT, MASK, i, j, itr, vm)                 \
```
ChinYikMing, line 3247: The `MASK` parameter has same name with `MASK()` function-like macro. I am concern with the naming scheme. Can you pass `MASK_BIT` as the parameter and leverage the `MASK()` function-like macro?

P.S. The compiler should smart enough to do constant propagation for the `MASK()` function-like macro during compile time.

Makefile
vestata: You're right. I'll move `VLEN ?= 128` into the conditional block. Thanks for pointing that out.

src/rv32_template.c
vestata: I'll rename `MASK` to `MASK_BIT`. Thanks for your suggestion and the information!

jserv:
Can you provide tools to validate RVV compliance? As I know, there are few projects below:
* [RISC-V Vector benchmark](https://github.com/camel-cdr/rvv-bench)
* [RISC-V Vector Tests Generator](https://github.com/chipsalliance/riscv-vector-tests)

vestata:
@jserv I am using [RISC-V Vector Tests Generator](https://github.com/chipsalliance/riscv-vector-tests) to validate the implemented rvv instructions. 
Following [build-and-test.yml](https://github.com/chipsalliance/riscv-vector-tests/blob/main/.github/workflows/build-and-test.yml) to generate binary test cases with `VLEN=128` and `XLEN=32`.

 ```make all --environment-overrides VLEN=128 XLEN=32```

By using `riscv32-unknown-elf-objdump` to check the binaries in `riscv-vector-tests/out/v128x32machine/bin/stage2/` we can see the following pass and fail code sections in every test case.

```
8001eed4 <fail>:
8001eed4:	0ff0000f          	fence
8001eed8:	00018063          	beqz	gp,8001eed8 <fail+0x4>
8001eedc:	0186                	slli	gp,gp,0x1
8001eede:	0011e193          	ori	gp,gp,1
8001eee2:	05d00893          	li	a7,93
8001eee6:	850e                	mv	a0,gp
8001eee8:	00000073          	ecall

8001eeec <pass>:
8001eeec:	0ff0000f          	fence
8001eef0:	4185                	li	gp,1
8001eef2:	05d00893          	li	a7,93
8001eef6:	4501                	li	a0,0
8001eef8:	00000073          	ecall
```

I ran the tests using `build/rv32emu -t` and check the trace output. If the implementation is correct, the trace log includes a `pass` entry, `false` on the other hand. I simply wrote a Python script to go though the directory,and so far,  94/1412 test cases passed.

src/riscv_private.h
```diff
@@ -227,6 +236,18 @@ struct riscv_internal {
      */
     uint32_t last_csr_sepc;
 #endif
+
+#if RV32_HAS(EXT_V)
+    vreg_t V[N_RV_REGS];
+
+    uint32_t csr_vstart; /* Vector start position */
```
howjmay, line 243: This is not very important either. Not sure whether this project will support RV64. 
 If supporting 64 bits architecture is on TODO list, then maybe we need to mention about the register width here, after all not all the register is `XLEN` wide

@jserv do you have a plan about this?

Address | Privilege | Name | Width (bits) | Description
-- | -- | -- | -- | --
0x008 | URW | vstart | XLEN | Vector start position
0x009 | URW | vxsat | 32 | Fixed-Point Saturate Flag
0x00A | URW | vxrm | 32 | Fixed-Point Rounding Mode
0x00F | URW | vcsr | XLEN | Vector control and status register
0xC20 | URO | vl | XLEN | Vector length
0xC21 | URO | vtype | XLEN | Vector data type register
0xC22 | URO | vlenb | XLEN | VLEN/8 (vector register length in bytes)

jserv: > This is not very important either. Not sure whether this project will support RV64. If supporting 64 bits architecture is on TODO list, then maybe we need to mention about the register width here, after all not all the register is `XLEN` wide
> @jserv do you have a plan about this?

Yes, I do expect to support RV64 once the system emulation for RV32 is consolidated.

howjmay: Then I suggest to add comment here, otherwise the contributor may not notice this one in the future

vestata:
Hi, I'm working on building the CI for this PR. As @visitorckw  and  @vacantron  suggested, using [riscv-ctg](https://github.com/riscv-software-src/riscv-ctg) to generate test suite for [ROSCOF](https://github.com/riscv-software-src/riscof) would be the ideal approach for this project. However, I’ve found it quite challenging to create a test suite for RVV.

Meanwhile, I noticed that in [sail-riscv](https://github.com/riscv/sail-riscv) they [execute test binaries](https://github.com/riscv/sail-riscv?tab=readme-ov-file#executing-test-binaries) with [riscv-tests](https://github.com/riscv/riscv-tests). Since [riscv-vector-tests](https://github.com/chipsalliance/riscv-vector-tests) is structured similarly to riscv-tests — and this PR originally used riscv-vector-tests — I’m considering integrating the test suite generated by riscv-vector-tests to the CI. Even though we can’t currently run it via RISCOF, we can still implement rv32emu as the DUT and the sail-riscv simulator as the reference model, using riscv-vector-tests to validate both of them. This approach could serve as an alternative comparison framework similar to RISCOF.

Please let me know if there’s a better way to handle RVV testing.

src/rv32_template.c
```diff
@@ -3031,3 +3031,4644 @@ RVOP(
...
+ * which 32-bit word within the vector register is processed; when it reaches
+ * the end of the current VLEN vector register, __j increments to move on to the
+ * next vector register, and __i resets.
+ *
+ * The current approach supports only SEW values of 8, 16, and 32 bits.
+ */
+
+#define VECTOR_DISPATCH(des, op1, op2, op, op_type)      \
+    {                                                    \
+        switch (8 << ((rv->csr_vtype >> 3) & 0b111)) {   \
```
howjmay, line 3232: BTW I am curious about left shit 8 here, is it necessary? 
I believe it may help readability, but not sure whether there are some other concern

vestata: You are correct. The left shift operation is primarily for enhancing readability and does not serve any other specific purpose.

jserv:
Close in favor of #733

PR #546, opened 2025-01-25
CI: Build check with various optimization levels

ChinYikMing:
The submodules are built locally, which may lead to build failures. Specifically, the dtc submodule fails to build when using GCC versions 11.4.0 or 12.3.0 with optimization levels set to -O0 or with debug flag -g. This is a critical issue, as it can disable the debug mode. To address this, optimization levels and debug level build are introduced to stabilize the pull request changes.

The compiled error:
/usr/bin/ld: /tmp/cc2eaHdM.ltrans0.ltrans.o: in function fdt_del_node': <artificial>:(.text+0x279f0): undefined reference to fdt_node_end_offset_' collect2: error: ld returned 1 exit status

Note: I have only used gcc 11.4.0 and gcc 12.3.0 to test this out. Github runners should test other compilers.

After bisecting, the original cause was found after #534.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -58,16 +58,107 @@ jobs:
       uses: rlalik/setup-cpp-compiler@master
       with:
         compiler: ${{ matrix.compiler }}
-    - name: default build
+    - name: default build with -g
       env:
         CC: ${{ steps.install_cc.outputs.cc }}
-      run: make -j$(nproc)
+      run: make OPT_LEVEL=-g -j$(nproc)
+      if: ${{ always() }}
+    - name: default build with -Og
+      env:
+        CC: ${{ steps.install_cc.outputs.cc }}
+      run: |
+            make distclean
+            make OPT_LEVEL=-Og -j$(nproc)
+      if: ${{ always() }}
+    - name: default build with -O0
+      env:
+        CC: ${{ steps.install_cc.outputs.cc }}
+      run: |
+            make distclean
+            make OPT_LEVEL=-O0 -j$(nproc)
+      if: ${{ always() }}
+    - name: default build with -O1
+      env:
+        CC: ${{ steps.install_cc.outputs.cc }}
+      run: |
+            make distclean
+            make OPT_LEVEL=-O1 -j$(nproc)
+      if: ${{ always() }}
+    - name: default build with -O2
+      env:
+        CC: ${{ steps.install_cc.outputs.cc }}
+      run: |
+            make distclean
+            make OPT_LEVEL=-O2 -j$(nproc)
+      if: ${{ always() }}
+    - name: default build with -O3
+      env:
+        CC: ${{ steps.install_cc.outputs.cc }}
+      run: |
+            make distclean
+            make OPT_LEVEL=-O3 -j$(nproc)
+      if: ${{ always() }}
+    - name: default build with -Ofast
+      env:
+        CC: ${{ steps.install_cc.outputs.cc }}
+      run: |
+            make distclean
+            make OPT_LEVEL=-Ofast -j$(nproc)
+      if: ${{ always() }}
+    - name: default build for system emulation with -g
+      env:
+        CC: ${{ steps.install_cc.outputs.cc }}
+      run: |
+            make distclean
+            make OPT_LEVEL=-g ENABLE_SYSTEM=1 -j$(nproc)
+      if: ${{ always() }}
+    - name: default build for system emulation with -Og
+      env:
+        CC: ${{ steps.install_cc.outputs.cc }}
+      run: |
+            make distclean
+            make OPT_LEVEL=-Og ENABLE_SYSTEM=1 -j$(nproc)
+      if: ${{ always() }}
+    - name: default build for system emulation with -O0
+      env:
+        CC: ${{ steps.install_cc.outputs.cc }}
+      run: |
+            make distclean
+            make OPT_LEVEL=-O0 ENABLE_SYSTEM=1 -j$(nproc)
+      if: ${{ always() }}
+    - name: default build for system emulation with -O1
+      env:
+        CC: ${{ steps.install_cc.outputs.cc }}
+      run: |
+            make distclean
+            make OPT_LEVEL=-O1 ENABLE_SYSTEM=1 -j$(nproc)
+      if: ${{ always() }}
+    - name: default build for system emulation with -O2
+      env:
+        CC: ${{ steps.install_cc.outputs.cc }}
+      run: |
+            make distclean
+            make OPT_LEVEL=-O2 ENABLE_SYSTEM=1 -j$(nproc)
+      if: ${{ always() }}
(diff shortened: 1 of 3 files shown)


ChinYikMing:
See [action](https://github.com/sysprog21/rv32emu/actions/runs/12965405707/job/36165100731) for more information.

Clang has no problem with various optimization build. Interesting.

RinHizakura:
This is not the right way to fix it. The issue could happen because I only compile the object that I think we need, but maybe more objects are required for `-g` + `-O0`. However, this can't explain the different results on Clang and GCC that you mentioned. I also have no idea here.
 
Can you check if you can also fix this with the following change?

```diff
--- a/mk/system.mk
+++ b/mk/system.mk
@@ -28,7 +28,7 @@ DEV_OBJS := $(patsubst $(DEV_SRC)/%.c, $(DEV_OUT)/%.o, $(wildcard $(DEV_SRC)/*.c
 deps := $(DEV_OBJS:%.o=%.o.d)

 OBJS_EXT += system.o
-OBJS_EXT += dtc/libfdt/fdt.o dtc/libfdt/fdt_ro.o dtc/libfdt/fdt_rw.o
+OBJS_EXT += dtc/libfdt/fdt.o dtc/libfdt/fdt_ro.o dtc/libfdt/fdt_rw.o dtc/libfdt/fdt_wip.o
```

RinHizakura:
Oh sorry, not aware that you are just trying to robust the CI/CD here instead of fixing anything. We can accept the test first and fix it in another PR.

.github/workflows/main.yml
```diff
@@ -58,16 +58,107 @@ jobs:
...
+      run: |
+            make distclean
+            make OPT_LEVEL=-O3 ENABLE_SYSTEM=1 -j$(nproc)
+      if: ${{ always() }}
+    - name: default build for system emulation with -Ofast
+      env:
+        CC: ${{ steps.install_cc.outputs.cc }}
+      run: |
+            make distclean
+            make OPT_LEVEL=-Ofast ENABLE_SYSTEM=1 -j$(nproc)
```
RinHizakura, line 155: Not sure if we need to take care of `-Os` or even just retain the presentative -O level only, otherwise the possible options may be too many. Will let @jserv decide this.

ChinYikMing:
> Oh sorry, not aware that you are just trying to robust the CI/CD here instead of fixing anything. We can accept the test first and fix it in another PR.

Agree. Thanks! Since with default `-O2` build, the problem does not exist. I revealed this problem when trying to do some debugging.

.github/workflows/main.yml
ChinYikMing: One more `-Oz` can be the candidate. It is related to size just like `-Os`. Thus, not embracing them at first.

ChinYikMing:
> This is not the right way to fix it. The issue could happen because I only compile the object that I think we need, but maybe more objects are required for `-g` + `-O0`. However, this can't explain the different results on Clang and GCC that you mentioned. I also have no idea here.
> 
> Can you check if you can also fix this with the following change?
> 
> ```diff
> --- a/mk/system.mk
> +++ b/mk/system.mk
> @@ -28,7 +28,7 @@ DEV_OBJS := $(patsubst $(DEV_SRC)/%.c, $(DEV_OUT)/%.o, $(wildcard $(DEV_SRC)/*.c
>  deps := $(DEV_OBJS:%.o=%.o.d)
> 
>  OBJS_EXT += system.o
> -OBJS_EXT += dtc/libfdt/fdt.o dtc/libfdt/fdt_ro.o dtc/libfdt/fdt_rw.o
> +OBJS_EXT += dtc/libfdt/fdt.o dtc/libfdt/fdt_ro.o dtc/libfdt/fdt_rw.o dtc/libfdt/fdt_wip.o
> ```

Tested it out, fixed it! Another commit will be push later.

ChinYikMing:
@RinHizakura  One more thing, I noticed the `dtc` submodule has integrated the `dtc` compiler, shall we use it? So that, the new user does not need to install the `dtc` on their own.

ChinYikMing:
The action that passed the various optimization level build after apply commit 139af81: [link](https://github.com/sysprog21/rv32emu/actions/runs/12965695115).

otteryc:
Link-Time Optimization (LTO) may be the culprit when building with `-O0` flag. GCC worked fine after adding the `-fno-lto` flag.

ChinYikMing:
> Link-Time Optimization (LTO) may be the culprit when building with `-O0` flag. GCC worked fine after adding the `-fno-lto` flag.

Without modifying the lto configuration, adding this object `dtc/libfdt/fdt_wip.o` solves it.

RinHizakura:
> One more thing, I noticed the dtc submodule has integrated the dtc compiler, shall we use it? So that, the new user does not need to install the dtc on their own.

I think we don't need to introduce such complications because building dtc may require other dependencies.

jserv (changes requested):
Rebase the latest `master` branch to resolve Arm64 host breakage on CI.

jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: 1a2bcc9cd1b1c18c5d9b7d46a74fba63b075cf0a | Previous: 64ea138c99ecfa1a91a4a26e23226867ac857cb5 | Ratio |
|-|-|-|-|
| `Dhrystone` | `1313` Average DMIPS over 10 runs | `1324` Average DMIPS over 10 runs | `1.01` |
| `Coremark` | `962.001` Average iterations/sec over 10 runs | `969.943` Average iterations/sec over 10 runs | `1.01` |

</details>

jserv:
I am concerned that the proposed changes increase the CI pipeline completion time. Could we rework this for two types of builds: one without optimizations (using `-O0 -g`) and one with optimizations (using either `-O2` or `-O3`)?"

ChinYikMing:
> I am concerned that the proposed changes increase the CI pipeline completion time. Could we rework this for two types of builds: one without optimizations (using `-O0 -g`) and one with optimizations (using either `-O2` or `-O3`)?"

action-build-time

When reviewing the build times (see the figure), each takes less than 10 seconds to complete, making it reasonable to include all types. Furthermore, the sum of these additional build time is significantly shorter than the running time on the Arm64 host.

jserv:
> When reviewing the build times (see the figure), each takes less than 10 seconds to complete, making it reasonable to include all types. Furthermore, the sum of these additional build time is significantly shorter than the running time on the Arm64 host.

It sounds great for x86-64 host. Let's avoid performing the checks with different optimization levels for Arm64 host though.

ChinYikMing:
> > When reviewing the build times (see the figure), each takes less than 10 seconds to complete, making it reasonable to include all types. Furthermore, the sum of these additional build time is significantly shorter than the running time on the Arm64 host.
> 
> It sounds great for x86-64 host. Let's avoid performing the checks with different optimization levels for Arm64 host though.

The current proposed changes only apply to x86-64 host.

jserv:

PR #552, opened 2025-02-03
Standardize logging utility

ChinYikMing:
Previously, messages were printed using both printf and fprintf,
causing the information to be mixed between stdout and stderr. To
address this, the log.[ch] was integrated to standardize the logging.
The log.[ch]'s API are encapsulated one more layer with prefix 'rv',
and included in src/common.h.

The logging API uses color to differentiate messages at different
levels, ensuring that logging all information to the same stdout stream
does not cause confusion. The color feature is controlled by
ENABLE_LOG_COLOR and is enabled by default.

Note that the logging stdout stream is registered during
rv_remap_stdstream() as the new stdout stream could be remapped at
there.

Since the newly introduced logging APIs generate logs during runtime,
the check recipe must filter out this log information before validation.

Additionally, refine the check recipe into a new target, check-test,
which serves as a template to enhance readability. 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR implements a unified logging system that replaces scattered printf/fprintf calls with a standardized API supporting multiple severity levels and color-coded output. The system is integrated across the RISC-V emulator core, device emulation, and system calls, with the build system updated to handle filtered logging output and improved test recipes.
<br>
<br>
<br>
<br>
<b>Estimated effort to review (1-5, lower is better)</b>: 5
</div>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,10 @@ $(call set-feature, MOP_FUSION)
 ENABLE_BLOCK_CHAINING ?= 1
 $(call set-feature, BLOCK_CHAINING)
 
+# Enable logging with color
+ENABLE_LOG_COLOR ?= 1
+$(call set-feature, LOG_COLOR)
+
 # Enable system emulation
 ENABLE_SYSTEM ?= 0
 $(call set-feature, SYSTEM)
@@ -293,6 +297,7 @@ OBJS := \
 	syscall.o \
 	emulate.o \
 	riscv.o \
+	log.o \
 	elf.o \
 	cache.o \
 	mpool.o \
--- a/Makefile
+++ b/Makefile
@@ -352,54 +352,48 @@ EXPECTED_puzzle = success in 2005 trials
 EXPECTED_fcalc = Performed 12 tests, 0 failures, 100% success rate.
 EXPECTED_pi = 3.141592653589793238462643383279502884197169399375105820974944592307816406286208998628034825342117067982148086
 
+LOG_FILTER=sed -E '/^[0-9]{2}:[0-9]{2}:[0-9]{2} /d'
+
+define exec
+$(eval OUTPUT_FILE := $(shell mktemp))
+$(eval _ := $(shell LC_ALL=C $(BIN) $(1) $(2) > $(OUTPUT_FILE)))
+$(eval RC := $(.SHELLSTATUS))
+endef
+
+# $(1): rv32emu's extra CLI parameter
+# $(2): ELF executable
+# $(3): ELF executable name
+# $(4): extra command in the pipeline
+# $(5): expected output
+define check-test
+$(call exec, $(1), $(2))
+$(Q)$(PRINTF) "Running $(3) ... "; \
+if [ 0 -eq $(RC) ] && [ "$(strip $(shell cat $(OUTPUT_FILE) | $(LOG_FILTER) | $(4)))" = "$(strip $(5))" ]; then \
+    $(call notice, [OK]); \
+else \
+    $(PRINTF) "Failed.\n"; \
+    exit 1; \
+fi; \
+$(RM) $(OUTPUT_FILE)
+endef
+
 check-hello: $(BIN)
-	$(Q)$(PRINTF) "Running hello.elf ... "; \
-	    if [ "$(shell LC_ALL=C $(BIN) $(OUT)/hello.elf | uniq)" = "$(strip $(EXPECTED_hello)) inferior exit code 0" ]; then \
-	    $(call notice, [OK]); \
-	    else \
-	    $(PRINTF) "Failed.\n"; \
-	    exit 1; \
-	    fi;
+	$(call check-test, , $(OUT)/hello.elf, hello.elf, uniq, $(EXPECTED_hello))
 
 check: $(BIN) check-hello artifact
-	$(Q)$(foreach e,$(CHECK_ELF_FILES),\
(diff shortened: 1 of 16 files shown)


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: 85701c40e26bcfd033bd8fb3e3b511ba317f19fc | Previous: 9e07773593d8de2472f0af4d249e0cd21a3b4788 | Ratio |
|-|-|-|-|
| `Dhrystone` | `1307` Average DMIPS over 10 runs | `1330` Average DMIPS over 10 runs | `1.02` |
| `Coremark` | `969.939` Average iterations/sec over 10 runs | `962.043` Average iterations/sec over 10 runs | `0.99` |

</details>

ChinYikMing:
Some basic information are logged. Below are simple examples:

process VM:
```shell
01:12:10 INFO  src/riscv.c:419: Log level: TRACE
01:12:10 INFO  src/riscv.c:435: build/hello.elf ELF loaded
01:12:10 INFO  src/main.c:282: RISC-V emulator is created and ready to run
Hello World!
Hello World!
Hello World!
Hello World!
Hello World!
01:12:10 INFO  src/main.c:296: RISC-V emulator is destroyed
```

system VM:
```shell
01:16:13 INFO  src/riscv.c:419: Log level: TRACE
01:16:13 INFO  src/riscv.c:489: Kernel loaded
01:16:13 INFO  src/riscv.c:494: DTB loaded
01:16:13 INFO  src/riscv.c:503: Rootfs loaded
01:16:13 INFO  src/main.c:282: RISC-V emulator is created and ready to run
[    0.000000] Linux version 6.1.119 (mirv32@node11) (riscv32-buildroot-linux-gnu-gcc.br_real (Buildroot 2024.08.3) 14.2.0, GNU ld (GNU Binutils) 2.42) #1 Sat Dec 14 22:54:23 CST 2024
[    0.000000] Machine model: rv32emu
[    0.000000] earlycon: ns16550 at MMIO 0xf4000000 (options '')
...
Welcome to Buildroot
buildroot login: root
# poweroff
# Stopping network: OK
Stopping klogd: OK
Stopping syslogd: stopped /sbin/syslogd (pid 35)
OK
umount: devtmpfs busy - remounted read-only
umount: can't unmount /: Invalid argument
The system is going down NOW!
Sent SIGTERM to all processes
Sent SIGKILL to all processes
Requesting system poweroff
[   10.463788] reboot: Power down
01:16:24 INFO  src/syscall.c:465: System reset: type=0, reason=0
01:16:24 INFO  src/main.c:296: RISC-V emulator is destroyed
```

src/log.c
```diff
@@ -0,0 +1,179 @@
...
+    void *udata;
+    log_lock_func_t lock;
+    int level;
+    bool quiet;
+#if RV32_HAS(LOG_CALLBACK)
+    callback_t callbacks[MAX_CALLBACKS];
+#endif /* RV32_HAS(LOG_CALLBACK) */
+} L;
+
+static const char *level_strings[] = {"TRACE", "DEBUG", "INFO",
```
jserv: Add a comma (`,`) right after "FATAL" to make `clang-format` happier.

src/log.c
```diff
@@ -0,0 +1,179 @@
...
+                                      "WARN",  "ERROR", "FATAL"};
+
+#if RV32_HAS(LOG_COLOR)
+static const char *level_colors[] = {"\x1b[94m", "\x1b[36m", "\x1b[32m",
```
jserv: Ditto. Make `clang-format` happier.

src/main.c
```diff
@@ -61,31 +61,31 @@ static char *opt_virtio_blk_img;
 
 static void print_usage(const char *filename)
 {
-    fprintf(stderr,
-            "RV32I[MAFC] Emulator which loads an ELF file to execute.\n"
-            "Usage: %s [options] [filename] [arguments]\n"
-            "Options:\n"
+    rv_log_info(
```
jserv: This message should always appear, so that INFO is not sufficient.

Makefile
```diff
@@ -27,6 +27,14 @@ $(call set-feature, MOP_FUSION)
 # Enable block chaining, easier for ablation study
 ENABLE_BLOCK_CHAINING ?= 1
 $(call set-feature, BLOCK_CHAINING)
+#
+# Enable logging with color
+ENABLE_LOG_COLOR ?= 1
+$(call set-feature, LOG_COLOR)
+
+# Enable adding callback for logging utility
+ENABLE_LOG_CALLBACK ?= 0
```
jserv: If the callback is never used, we can safely remove the related code.

jserv (changes requested):
Since the logging feature is quite useful, we can promote the functions as public APIs. That is, include `log.h` in the header `riscv.h`.

src/main.c
ChinYikMing: > This message should always appear, so that INFO is not sufficient.

Use TRACE instead.

jserv: > Use TRACE instead.

How about `ERROR` instead? While dumping the usage, the program exits with non-zero code.

ChinYikMing: > > Use TRACE instead.
> 
> How about `ERROR` instead? While dumping the usage, the program exits with non-zero code.

I chose wrong level (TRACE) which is the lowest log level. I originally planned to choose highest log level (FATAL), but ERROR works as well, so I'll go with that.

jserv:

PR #553, opened 2025-02-04
CI: Specify larger INITRD_SIZE for larger rootfs

ChinYikMing:
Since 515e011 enable the guestOS to run SDL-oriented applications, the artifacts for these applications must be prepared in the rootfs to ensure they can run, making the rootfs larger than before which exceeds the default 8 MiB INITRD_SIZE allocated for loading the rootfs.


jserv:

Issue #519, opened 2024-12-15
CI: integrate macOS


ChinYikMing:
The current CI pipeline utilizes Linux-based x86-64 and Aarch64 machines for verification and validation (V&V). To enhance the portability of rv32emu, it is both reasonable and feasible to integrate macOS into the CI pipeline.

ChinYikMing:
A simple CI pipeline for system emulation:
```diff
...
+macOS:
+    runs-on: macos-latest
+   steps:
+    - uses: actions/checkout@v4
+    - name: install-dependencies
+      run: |
+            brew install make dtc expect
+    - name: boot Linux kernel test
+      run: |
+            make clean && make ENABLE_SYSTEM=1 && make ENABLE_SYSTEM=1 artifact -j$(sysctl -n hw.logicalcpu)
+            .ci/boot-linux.sh
+            make ENABLE_SYSTEM=1 clean
 ...
```

ChinYikMing:
The fetching of the checksum occasionally fails in previous actions([1](https://github.com/sysprog21/rv32emu/actions/runs/12333005640/job/34423878298), [2](https://github.com/sysprog21/rv32emu/actions/runs/12332623010/job/34420668188). [3](https://github.com/sysprog21/rv32emu/actions/runs/12332106397/job/34419526096), ...).

To fully close this issue, the failure should not occur too frequently and the macOS pipeline should be capable of performing V&V for both user and system emulation.

Found at #508

ChinYikMing:
The occasional failure could be fixed by #536.

ChinYikMing:
CI breakage on macOS/M1 when running JIT test: [CI link](https://github.com/ChinYikMing/rv32emu/actions/runs/13268504435/job/37042018292)

It obviously shown that the function is not called ever. @vacantron Can you check it?

vacantron:
It seems like it's been a long time since the last test on Arm64/MacOS. I'll fix it later.

ChinYikMing:
I will try to integrate the minimal runnable CI for macOS first.

ChinYikMing:
Shall we support macOS/Intel? The CI [1] breaks since it expects llvm18 library installed at path `/usr/local/opt/` but Makefile detects at path `/opt/homebrew/opt/` which is the default path of M series chips, so macOS/M1 runners do not suffer this when running JIT test suite.

[...]

PR #556, opened 2025-02-06
Allow macOS build

jserv:
Commit 2ab8a68 failed to detect macOS/Arm64 properly, and commit 2e87a75 relied on the `.SHELLSTATUS` feature introduced in GNU make 4.2 (released on 2016-05-22). However, macOS ships with GNU make 3.81 by default, lacking of `.SHELLSTATUS`.

This commit fixes the macOS/Arm64 detection issue and reworks the build commands to run everything in a single shell invocation, capturing both the exit code and program output without relying on `.SHELLSTATUS`. 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR enhances macOS build compatibility through two main changes: implementing ARM64 architecture detection and refactoring Makefile command execution for better compatibility with macOS GNU make 3.81. The changes focus on maintaining existing functionality while improving cross-platform support.
<br>
<br>
<br>
<br>
<b>Estimated effort to review (1-5, lower is better)</b>: 1
</div>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -357,46 +357,42 @@ EXPECTED_pi = 3.1415926535897932384626433832795028841971693993751058209749445923
 
 LOG_FILTER=sed -E '/^[0-9]{2}:[0-9]{2}:[0-9]{2} /d'
 
-define exec
-$(eval OUTPUT_FILE := $(shell mktemp))
-$(eval _ := $(shell LC_ALL=C $(BIN) $(1) $(2) > $(OUTPUT_FILE)))
-$(eval RC := $(.SHELLSTATUS))
-endef
-
 # $(1): rv32emu's extra CLI parameter
 # $(2): ELF executable
 # $(3): ELF executable name
 # $(4): extra command in the pipeline
 # $(5): expected output
 define check-test
-$(call exec, $(1), $(2))
-$(Q)$(PRINTF) "Running $(3) ... "; \
-if [ 0 -eq $(RC) ] && [ "$(strip $(shell cat $(OUTPUT_FILE) | $(LOG_FILTER) | $(4)))" = "$(strip $(5))" ]; then \
+$(Q)true; \
+$(PRINTF) "Running $(3) ... "; \
+OUTPUT_FILE="$$(mktemp)"; \
+if (LC_ALL=C $(BIN) $(1) $(2) > "$$OUTPUT_FILE") && \
+   [ "$$(cat "$$OUTPUT_FILE" | $(LOG_FILTER) | $(4))" = "$(5)" ]; then \
     $(call notice, [OK]); \
 else \
     $(PRINTF) "Failed.\n"; \
     exit 1; \
 fi; \
-$(RM) $(OUTPUT_FILE)
+$(RM) "$$OUTPUT_FILE"
 endef
 
 check-hello: $(BIN)
-	$(call check-test, , $(OUT)/hello.elf, hello.elf, uniq, $(EXPECTED_hello))
+	$(call check-test, , $(OUT)/hello.elf, hello.elf, uniq,$(EXPECTED_hello))
 
 check: $(BIN) check-hello artifact
-	$(Q)$(foreach e, $(CHECK_ELF_FILES), $(call check-test, , $(OUT)/riscv32/$(e), $(e), uniq, $(EXPECTED_$(e))))
+	$(Q)$(foreach e, $(CHECK_ELF_FILES), $(call check-test, , $(OUT)/riscv32/$(e), $(e), uniq,$(EXPECTED_$(e))))
 
 EXPECTED_aes_sha1 = 89169ec034bec1c6bb2c556b26728a736d350ca3  -
 misalign: $(BIN) artifact
-	$(call check-test, -m, $(OUT)/riscv32/uaes, uaes.elf, $(SHA1SUM), $(EXPECTED_aes_sha1))
+	$(call check-test, -m, $(OUT)/riscv32/uaes, uaes.elf, $(SHA1SUM),$(EXPECTED_aes_sha1))
 
 EXPECTED_misalign = MISALIGNED INSTRUCTION FETCH TEST PASSED!
 misalign-in-blk-emu: $(BIN)
-	$(call check-test, , tests/system/alignment/misalign.elf, misalign.elf, tail -n 1, $(EXPECTED_misalign))
+	$(call check-test, , tests/system/alignment/misalign.elf, misalign.elf, tail -n 1,$(EXPECTED_misalign))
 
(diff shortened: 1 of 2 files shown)


mk/common.mk
```diff
@@ -10,6 +10,8 @@ ifeq ($(UNAME_M),x86_64)
     HOST_PLATFORM := x86
 else ifeq ($(UNAME_M),aarch64)
     HOST_PLATFORM := aarch64
+else ifeq ($(UNAME_M),arm64) # macOS
+    HOST_PLATFORM := aarch64
```
jserv: @vacantron, I am not sure for this since it is related to the manipulation in `mk/artifact.mk`. Can you check?

ChinYikMing:
Sorry for the regression of macOS build. 

I do not have macOS machine right now for verifying the changes. Shall we partially introduce build CI on macOS without triggering any `fetch-checksum` (related to #519) which similar to the build check in #546.

jserv:
> Shall we partially introduce build CI on macOS without triggering any `fetch-checksum` (related to #519) which similar to the build check in #546.

Hopefully, yes. I do suffer from repeated CI breakage related to checksum.

mk/common.mk
vacantron: We only run the architecture test on the x86-64 and the Aarch64 because of the lack of the Sail model . I think it is better to use `Arm64` to distinguish with `Aarch64` targets for fetching the model in the future.

Makefile
```diff
@@ -357,27 +357,23 @@ EXPECTED_pi = 3.1415926535897932384626433832795028841971693993751058209749445923
...
 # $(5): expected output
 define check-test
-$(call exec, $(1), $(2))
-$(Q)$(PRINTF) "Running $(3) ... "; \
-if [ 0 -eq $(RC) ] && [ "$(strip $(shell cat $(OUTPUT_FILE) | $(LOG_FILTER) | $(4)))" = "$(strip $(5))" ]; then \
+$(Q)true; \
+$(PRINTF) "Running $(3) ... "; \
+OUTPUT_FILE="$$(mktemp)"; \
+if (LC_ALL=C $(BIN) $(1) $(2) > "$$OUTPUT_FILE") && \
+   [ "$$(cat "$$OUTPUT_FILE" | $(LOG_FILTER) | $(4))" = "$(strip $(5))" ]; then \
```
ChinYikMing: No need `$(strip)`? Since it truncates the whitespace in the sha1sum output of the misalign test when running uaes.elf, causing the test to fail."

ChinYikMing: If `$(strip)` is removed, the last argument of line 380, 383,  387, 391, 394 should be refined by removing  the whitespace after the comma to pass the test.

ChinYikMing:
```diff
diff --git a/Makefile b/Makefile
index d2ab6f2..a0224cd 100644
--- a/Makefile
+++ b/Makefile
@@ -367,7 +367,7 @@ $(Q)true; \
 $(PRINTF) "Running $(3) ... "; \
 OUTPUT_FILE="$$(mktemp)"; \
 if (LC_ALL=C $(BIN) $(1) $(2) > "$$OUTPUT_FILE") && \
-   [ "$$(cat "$$OUTPUT_FILE" | $(LOG_FILTER) | $(4))" = "$(strip $(5))" ]; then \
+   [ "$$(cat "$$OUTPUT_FILE" | $(LOG_FILTER) | $(4))" = "$(5)" ]; then \
     $(call notice, [OK]); \
 else \
     $(PRINTF) "Failed.\n"; \
@@ -377,22 +377,22 @@ $(RM) "$$OUTPUT_FILE"
 endef
 
 check-hello: $(BIN)
-	$(call check-test, , $(OUT)/hello.elf, hello.elf, uniq, $(EXPECTED_hello))
+	$(call check-test, , $(OUT)/hello.elf, hello.elf, uniq,$(EXPECTED_hello))
 
 check: $(BIN) check-hello artifact
-	$(Q)$(foreach e, $(CHECK_ELF_FILES), $(call check-test, , $(OUT)/riscv32/$(e), $(e), uniq, $(EXPECTED_$(e))))
+	$(Q)$(foreach e, $(CHECK_ELF_FILES), $(call check-test, , $(OUT)/riscv32/$(e), $(e), uniq,$(EXPECTED_$(e))))
 
 EXPECTED_aes_sha1 = 89169ec034bec1c6bb2c556b26728a736d350ca3  -
 misalign: $(BIN) artifact
-	$(call check-test, -m, $(OUT)/riscv32/uaes, uaes.elf, $(SHA1SUM), $(EXPECTED_aes_sha1))
+	$(call check-test, -m, $(OUT)/riscv32/uaes, uaes.elf, $(SHA1SUM),$(EXPECTED_aes_sha1))
 
 EXPECTED_misalign = MISALIGNED INSTRUCTION FETCH TEST PASSED!
 misalign-in-blk-emu: $(BIN)
-	$(call check-test, , tests/system/alignment/misalign.elf, misalign.elf, tail -n 1, $(EXPECTED_misalign))
+	$(call check-test, , tests/system/alignment/misalign.elf, misalign.elf, tail -n 1,$(EXPECTED_misalign))
 
 EXPECTED_mmu = STORE PAGE FAULT TEST PASSED!
 mmu-test: $(BIN)
-	$(call check-test, , tests/system/mmu/vm.elf, vm.elf, tail -n 1, $(EXPECTED_mmu))
+	$(call check-test, , tests/system/mmu/vm.elf, vm.elf, tail -n 1,$(EXPECTED_mmu))
 
 # Non-trivial demonstration programs
 ifeq ($(call has, SDL), 1)

```

ChinYikMing: After applying the proposed changes to allow build on macOS and the above diff of Makefile, the default tests passed on both host-x64 and macOS-arm64, check [CI](https://github.com/ChinYikMing/rv32emu/actions/runs/13268742926/job/37042827874). Note that the macOS CI currently excludes the `gdbstub-test` and `arch-test` because the RISC-V toolchain cannot be installed using the existing scripts (will try to fix it).

mk/common.mk
```diff
@@ -10,6 +10,8 @@ ifeq ($(UNAME_M),x86_64)
     HOST_PLATFORM := x86
 else ifeq ($(UNAME_M),aarch64)
     HOST_PLATFORM := aarch64
+else ifeq ($(UNAME_M),arm64) # macOS
+    HOST_PLATFORM := arm64
```
ChinYikMing, line 14: There is no Sail model with suffix `arm64` currently (Check [sail model in rv32emu-prebuilt](https://github.com/sysprog21/rv32emu-prebuilt/releases/tag/2025.01.15-sail)). I would defer the integration of arch-test on masOS CI.

jserv:

Issue #536, opened 2025-01-11
Reduce HTTP requests for checksum file in fetch-checksum


eleanorLYJ:
After adding more test cases, the CI pipeline fails due to excessive HTTP requests to GitHub during the fetch-checksum step. This triggers rate-limiting errors (403: rate limit exceeded).

## Proposed Solution:
Check if the checksum file exists and is less than one day old before fetching it. This will reduce unnecessary requests while ensuring the file is up-to-date.

jserv:
Network quality issues may sometimes cause unexpected disconnections when downloading from certain remote sites, requiring retries.

vacantron:
It seems to be too many requests in a short period that `make artifact` is called when running `make all` . Maybe maintain the local copies of `sha1sum` files which use the recently commit hash as the postfix of them to reduce the refetching?

vacantron:
I think the main reason causing this is

https://github.com/sysprog21/rv32emu/blob/c9dfd362f6ef08818a4f8f541afa28c4baadbf72/mk/artifact.mk#L39-L47

Every time we run `make`, the GitHub API is called. We can reduce the number of HTTP requests by checking the prebuilt files is existing or not.

However, in CI, we always run `make distclean` before building. The method mentioned above would not work.

https://github.com/sysprog21/rv32emu/blob/c9dfd362f6ef08818a4f8f541afa28c4baadbf72/Makefile#L418-L421

PR #560, opened 2025-02-09
CI: Reduce HTTP requests when possible

jserv:
Due to frequent CI failures during checksum verification, this commit bypasses integrity checks by assuming all downloaded files are complete, removing the requirement to download and verify checksums. 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR enhances the CI build process by optimizing HTTP requests and file download handling through the introduction of a fetch-releases-tag function. The changes include improved file existence checks to prevent redundant downloads, streamlined build processes with better checksum verification, and enhanced cleanup procedures in the Makefile.
<br>
<br>
<br>
<br>
</div>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -58,10 +58,20 @@ jobs:
       uses: rlalik/setup-cpp-compiler@master
       with:
         compiler: ${{ matrix.compiler }}
+    - name: fetch artifact first to reduce HTTP requests
+      env:
+        CC: ${{ steps.install_cc.outputs.cc }}
+      run: |
+            make artifact
+            make ENABLE_SYSTEM=1 artifact
+            make ENABLE_ARCH_TEST=1 artifact
+      if: ${{ always() }}
     - name: default build with -g
       env:
         CC: ${{ steps.install_cc.outputs.cc }}
-      run: make OPT_LEVEL=-g -j$(nproc)
+      run: |
+            make distclean
+            make OPT_LEVEL=-g -j$(nproc)
       if: ${{ always() }}
     - name: default build with -Og
       env:
@@ -268,6 +278,7 @@ jobs:
           ./llvm.sh 18
         # Append custom commands here
         run: |
+          make artifact
           make -j$(nproc)
           make check -j$(nproc)
           make ENABLE_JIT=1 clean && make ENABLE_JIT=1 check -j$(nproc)
--- a/mk/artifact.mk
+++ b/mk/artifact.mk
@@ -36,15 +36,33 @@ SCIMARK2_SHA1 := de278c5b8cef84ab6dda41855052c7bfef919e36
 
 SHELL_HACK := $(shell mkdir -p $(BIN_DIR)/linux-x86-softfp $(BIN_DIR)/riscv32 $(BIN_DIR)/linux-image)
 
+# $(1): tag of GitHub releases
+# $(2): name of GitHub releases
+# $(3): name showing in terminal
+define fetch-releases-tag
+    $(if $(wildcard $(BIN_DIR)/$(2)), \
+        $(info $(3) is found. Skipping downloading.), \
+        $(eval LATEST_RELEASE := $(shell wget -q https://api.github.com/repos/sysprog21/rv32emu-prebuilt/releases -O- \
+                                    | grep '"tag_name"' \
+                                    | grep "$(1)" \
+                                    | head -n 1 \
+                                    | sed -E 's/.*"tag_name": "([^"]+)".*/\1/')) \
+        $(if $(LATEST_RELEASE),, \
+            $(error Fetching tag of latest releases failed) \
+        ) \
+    )
+endef
+
 ifeq ($(call has, PREBUILT), 1)
-ifeq ($(call has, SYSTEM), 1)
-  LATEST_RELEASE := $(shell wget -q https://api.github.com/repos/sysprog21/rv32emu-prebuilt/releases -O- | grep '"tag_name"' | grep "Linux-Image" | head -n 1 | sed -E 's/.*"tag_name": "([^"]+)".*/\1/')
-else ifeq ($(call has, ARCH_TEST), 1)
-  LATEST_RELEASE := $(shell wget -q https://api.github.com/repos/sysprog21/rv32emu-prebuilt/releases -O- | grep '"tag_name"' | grep "sail" | head -n 1 | sed -E 's/.*"tag_name": "([^"]+)".*/\1/')
-else
-  LATEST_RELEASE := $(shell wget -q https://api.github.com/repos/sysprog21/rv32emu-prebuilt/releases -O- | grep '"tag_name"' | grep "ELF" | head -n 1 | sed -E 's/.*"tag_name": "([^"]+)".*/\1/')
-endif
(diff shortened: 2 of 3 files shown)


jserv:
@vacantron, I tried to bypass checksum validation, but it did not work as intended. Can you help me fix this?

mk/artifact.mk
```diff
@@ -60,25 +60,35 @@ endif
 artifact: fetch-checksum ieeelib scimark2
 ifeq ($(call has, PREBUILT), 1)
 	$(Q)$(PRINTF) "Checking SHA-1 of prebuilt binaries ... "
+ifeq ($(SKIP_CHECKSUM),)
 	$(Q)$(eval RES := 0)
+else
+	# Assume every downloaded file is verified.
+	$(Q)$(eval RES := 1)
```
otteryc: Setting `RES` to 1 here actually indicates an abnormal event, i.e., the SHA-1 checksum verification has failed or the file does not exist.
Originally, the variable `RES` is passed into the macro `verify` located in `mk/external.mk`. However, the `verify` process is bypassed when `SKIP_CHECKSUM ` is set. Consequently, the `RES` variable remains unchanged, resulting in the perpetual re-download of the prebuilt file whenever the target `artifact` is triggered.

mk/artifact.mk
```diff
@@ -107,6 +118,9 @@ else
 	    $(call notice, [OK]); \
 	fi
 endif
+ifneq ($(SKIP_CHECKSUM),)
+	$(Q)$(eval RES := 0)
```
otteryc: My apologies for any confusion.
Setting `RES` to `1` wrongly makes that the error-checking statements at lines 107 and 114 are always true. This implies that it is incorrect to set `RES` at a previous position.

Furthermore, omitting the entire `fetch-checksum` process may lead to a predicament, as the filenames, which  should be downloaded, are retrieved from the checksum file. Additionally, we cannot verify the file’s existence by the tarball, as it is directly piped into the `tar` utility. Consequently, it is not possible to cover the scenario that the files do not exist in current implementation.

One potential solution is to store the prebuilt tarball instead of piping it, thereby ensuring its verification.
For reference, please refer to the following commit: https://github.com/otteryc/rv32emu/commit/5cf1e728e46a7070325e6eca32e79e81c6b4a4f7. This test run verifies the existence of the tarball, although the CI test (clang) still fails due to exceeding the rate limit while downloading the prebuilt tarball.

jserv:
I defer to @vacantron to determine if we should proceed with the proposed changes.

jserv:
Append `Close #536` at the end of git commit messages.

jserv:
Mention Arm64 regressions in git commit messages to enable earlier pull request merging despite ongoing CI pipeline issues.

mk/artifact.mk
```diff
@@ -37,14 +37,27 @@ SCIMARK2_SHA1 := de278c5b8cef84ab6dda41855052c7bfef919e36
...
-  LATEST_RELEASE := $(shell wget -q https://api.github.com/repos/sysprog21/rv32emu-prebuilt/releases -O- | grep '"tag_name"' | grep "Linux-Image" | head -n 1 | sed -E 's/.*"tag_name": "([^"]+)".*/\1/')
-else ifeq ($(call has, ARCH_TEST), 1)
-  LATEST_RELEASE := $(shell wget -q https://api.github.com/repos/sysprog21/rv32emu-prebuilt/releases -O- | grep '"tag_name"' | grep "sail" | head -n 1 | sed -E 's/.*"tag_name": "([^"]+)".*/\1/')
-else
-  LATEST_RELEASE := $(shell wget -q https://api.github.com/repos/sysprog21/rv32emu-prebuilt/releases -O- | grep '"tag_name"' | grep "ELF" | head -n 1 | sed -E 's/.*"tag_name": "([^"]+)".*/\1/')
-endif
-  PREBUILT_BLOB_URL = https://github.com/sysprog21/rv32emu-prebuilt/releases/download/$(LATEST_RELEASE)
+    ifeq ($(call has, SYSTEM), 1)
+        ifeq ($(wildcard $(BIN_DIR)/rv32emu-linux-image-prebuilt.tar.gz),)
+            LATEST_RELEASE := $(shell wget -q https://api.github.com/repos/sysprog21/rv32emu-prebuilt/releases -O- | grep '"tag_name"' | grep "Linux-Image" | head -n 1 | sed -E 's/.*"tag_name": "([^"]+)".*/\1/')
```
jserv: Rewrite as following:
```shell
wget -qO- https://api.github.com/repos/sysprog21/rv32emu-prebuilt/releases \
  | grep -m1 '"tag_name".*Linux-Image' \
  | sed -E 's/.*"tag_name": "([^"]+)".*/\1/'
```
and consider to use `define` for code reusability.

mk/artifact.mk
```diff
@@ -149,14 +163,26 @@ fetch-checksum:
...
+    else
+		$(info Downloading skipped.)
+    endif
 else
-	$(Q)wget -q -O $(BIN_DIR)/sha1sum-linux-x86-softfp $(PREBUILT_BLOB_URL)/sha1sum-linux-x86-softfp
-	$(Q)wget -q -O $(BIN_DIR)/sha1sum-riscv32 $(PREBUILT_BLOB_URL)/sha1sum-riscv32
-	$(Q)$(call notice, [OK])
+    ifeq ($(wildcard $(BIN_DIR)/rv32emu-prebuilt.tar.gz),)
+		wget -q -O $(BIN_DIR)/sha1sum-linux-x86-softfp $(PREBUILT_BLOB_URL)/sha1sum-linux-x86-softfp
+		wget -q -O $(BIN_DIR)/sha1sum-riscv32 $(PREBUILT_BLOB_URL)/sha1sum-riscv32
```
jserv: Later, we may consider to introduce [function calls](https://www.gnu.org/software/make/manual/html_node/Call-Function.html) to wrap `wget` in another pull request.

.github/workflows/main.yml
```diff
@@ -58,10 +58,20 @@ jobs:
       uses: rlalik/setup-cpp-compiler@master
       with:
         compiler: ${{ matrix.compiler }}
+    - name: fetch artifact first to reduce HTTP requests
+      env:
+        CC: ${{ steps.install_cc.outputs.cc }}
+      run: |
+            make artifact
+            make ENABLE_SYSTEM=1 artifact
+            make ENABLE_ARCH_TEST=1 artifact
```
ChinYikMing, line 67: Since `LATEST_RELEASE` is guarded to the tarball, I am not sure if fetching all artifacts in a single job is a good choice.

Would not it make more sense to fetch the required artifact in a relevant job instead? If it is not, the line 242 should be refined such that removing the redundant artifact fetching.

vacantron: This is just moving up the artifacts fetching. We do all userspace test, system simulation test, and architecture test below.

The redundant `make artifact` could not cause any problem. Beccause once the tarball is downloaded, we check that tarball to decide triggering downloading or not.

mk/artifact.mk
vacantron: Sure. Moreover, consolidating the file name of sha1sum and prebuilts, such as `rv32emu-prebuilt-*.sha` would make it easier.

vacantron:
The new patch to solve the Arm64 CI failure will be opened after this pull request merged.

https://github.com/sysprog21/rv32emu/actions/runs/13257429033/job/37006598258?pr=560

jserv:

Issue #435, opened 2024-05-10
Calling close() system call with fd < 3


visitorckw:
Currently, if the close() system call is invoked with fd < 3, close() won't take any action and won't return an error code of -1. Some programs may need to close stdout or stderr to avoid excessive output. Therefore, the current implementation may lead to users getting unexpected results. However, rv32emu also uses stdout and stderr to output some information. Therefore, directly closing stdout and stderr would result in the loss of relevant output information from rv32emu. Here are two possible ways to address this:

1. close() performs no action but returns an error code of -1.
2. Close stdout and stderr as usual and allow rv32emu to lose output information.

banglday:
## Conflict in This Issue

1. Some user programs may need to close `stdout` or `stderr`.
2. rv32emu itself relies on `stdout` and `stderr` for its own output.

Because of these two needs, there is a conflict about how `close()` should behave when `fd < 3`.


## Analysis of the Solutions

### Solution 1: Do not actually close, but return an error code (-1)

**Advantages:**

- `stdout` and `stderr` remain open, so rv32emu can continue printing messages or logs.
- User programs get an error code, so they know the close operation was not successful. This avoids them thinking they successfully closed the output channels.
- After seeing the error, user programs can redirect output or do other error handling. This prevents confusion like “I thought it was closed, but it’s still printing.”

**Disadvantages:**

- This does not match real system behavior exactly. In a typical Unix-like system, calling `close(1)` or `close(2)` is allowed. Here, the emulator denies it to keep its own output working.

---

### Solution 2: Allow normal closing of `stdout` and `stderr`

**Advantages:**

- This is closer to real Unix behavior (where closing file descriptor 1 or 2 is allowed).
- If a user program really needs to close `stdout` or `stderr`, it will behave like on a real system (no further output on those descriptors).

**Disadvantages:**

- rv32emu’s own output (errors, debug messages, etc.) will be “lost,” because `stdout` and `stderr` will actually be closed. This makes debugging or analysis more difficult.
- If rv32emu wants to keep printing messages, it must handle that differently (for example, by opening a new file descriptor or redirecting output), or else lose its log messages.

#### Option: Redirect rv32emu’s Output Inside the Emulator

How to keep rv32emu’s output even after `stdout` and `stderr` are closed?
- At startup, rv32emu could redirect its own logging to other file descriptors (for example, 100 or 101). Then all of rv32emu’s messages go to those custom descriptors.
[...]

PR #566, opened 2025-02-12
Return error code -1 when closing file descriptors < 3 (stdin, stdout, stderr)

banglday:
closes Issue #435

## Conflict in This Issue

1. Some user programs may need to close `stdout` or `stderr`.
2. rv32emu itself relies on `stdout` and `stderr` for its own output.

Because of these two needs, there is a conflict about how `close()` should behave when `fd < 3`.

### Solution : Do not actually close, but return an error code (-1)

**Advantages:**

- `stdout` and `stderr` remain open, so rv32emu can continue printing messages or logs.
- User programs get an error code, so they know the close operation was not successful. This avoids them thinking they successfully closed the output channels.
- After seeing the error, user programs can redirect output or do other error handling. This prevents confusion like “I thought it was closed, but it’s still printing.”

**Disadvantages:**

- This does not match real system behavior exactly. In a typical Unix-like system, calling `close(1)` or `close(2)` is allowed. Here, the emulator denies it to keep its own output working.

---

## Conclusion

To prevent a user program from thinking it closed `stdout` or `stderr` when it really did not, the simplest approach is **Solution **: do nothing but return an error code (`-1`). 

- This tells the user program that closing `fd < 3` failed, so it will not assume the channels are closed. 
- For most needs, where preserving emulator output is important (especially for debugging or analysis), this is a good compromise.

Since rv32emu is often used in teaching and research, being able to see the emulator’s own outputs is very helpful. Thus, returning `-1` is recommended by default. If needed, you can provide parameters or redirection mechanisms to allow truly closing the file descriptors in special cases. This maintains a good user experience while still letting people enable the more realistic behavior if they want.

PS: I save all details of implementation in this [document](https://hackmd.io/@KenHoryuKen/BJoTW9dYye). 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request enhances the RISC-V emulator by modifying the syscall_close function to return an error code for standard file descriptors, preventing user programs from incorrectly assuming closure. Additional improvements to related functions support this behavior, enhancing the overall user experience.
<br>
<br>
<br>
<br>
</div>

### DIFF
--- a/src/syscall.c
+++ b/src/syscall.c
@@ -230,6 +230,22 @@ static void syscall_close(riscv_t *rv)
     /* _close(fd); */
     uint32_t fd = rv_get_reg(rv, rv_reg_a0);
 
+#if !RV32_HAS(SYSTEM)
+    /*
+     * The crt0 closes standard file descriptor(0, 1, 2) when
+     * the process exits. Thus, the operations by the crt0
+     * should not considered as error.
+     */
+    if (fd < 3 && !PRIV(rv)->on_exit) {
+        rv_set_reg(rv, rv_reg_a0, -1);
+        rv_log_error(
+            "Attempted to close a file descriptor < 3 (fd=%u). Operation "
+            "not supported.",
+            fd);
+        return;
+    }
+#endif
+
     if (fd >= 3) { /* lookup the file descriptor */
         map_iter_t it;
         map_find(attr->fd_map, &it, &fd);
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -1078,11 +1078,17 @@ void rv_step(void *arg)
         assert(block->satp == rv->csr_satp);
 #endif
 
-        /* After emulating the previous block, it is determined whether the
-         * branch is taken or not. The IR array of the current block is then
-         * assigned to either the branch_taken or branch_untaken pointer of
-         * the previous block.
-         */
+#if !RV32_HAS(SYSTEM)
+        /* on exit */
+        if (unlikely(block->ir_head->pc == PRIV(rv)->exit_addr))
+            PRIV(rv)->on_exit = true;
+#endif
+
+            /* After emulating the previous block, it is determined whether the
+             * branch is taken or not. The IR array of the current block is then
+             * assigned to either the branch_taken or branch_untaken pointer of
+             * the previous block.
+             */
 
 #if RV32_HAS(BLOCK_CHAINING)
         if (prev
--- a/src/riscv.c
+++ b/src/riscv.c
@@ -434,6 +434,15 @@ riscv_t *rv_create(riscv_user_t rv_attr)
     if ((end = elf_get_symbol(elf, "_end")))
         attr->break_addr = end->st_value;
 
+#if !RV32_HAS(SYSTEM)
+    /* set not exiting */
+    attr->on_exit = false;
+
+    const struct Elf32_Sym *exit;
+    if ((exit = elf_get_symbol(elf, "exit")))
+        attr->exit_addr = exit->st_value;
+#endif
+
     assert(elf_load(elf, attr->mem));
 
     /* set the entry pc */
--- a/src/riscv.h
+++ b/src/riscv.h
@@ -548,6 +548,14 @@ typedef struct {
     /* the data segment break address */
     riscv_word_t break_addr;
 
+#if !RV32_HAS(SYSTEM)
+    /* the exit entry address */
+    riscv_word_t exit_addr;
+
+    /* flag to determine if the emulator exits the target program */
+    bool on_exit;
+#endif
+
     /* SBI timer */
     uint64_t timer;
 } vm_attr_t;


jserv (changes requested):
1. Read https://cbea.ms/git-commit/ carefully and refine the git commit messages.
2. Append `Close #435` at the end of git commit messages.

banglday:
> 1. Read https://cbea.ms/git-commit/ carefully and refine the git commit messages.
> 2. Append `Close #435` at the end of git commit messages.

Finished refining the commit message.

src/syscall.c
```diff
@@ -244,10 +244,10 @@ static void syscall_close(riscv_t *rv)
             /* success */
             rv_set_reg(rv, rv_reg_a0, 0);
         }
+    } else {
+        /* error */
```
ChinYikMing: Add some `rv_log_error()` here for better clarity?

jserv:
> Finished refining the commit message.

Read https://cbea.ms/git-commit/ carefully and enforce the rules!

src/syscall.c
banglday: When the emulator attempts to close a file descriptor less than 3, error logs are generated, resulting in the following output.

```
ubuntu@riscv-playground:~/projects/rv32emu$ build/rv32emu ~/projects/my-riscv-tests/hello.elf 
12:47:54 INFO  src/riscv.c:418: Log level: TRACE
12:47:54 INFO  src/riscv.c:431: /home/ubuntu/projects/my-riscv-tests/hello.elf ELF loaded
12:47:54 INFO  src/main.c:286: RISC-V emulator is created and ready to run
hello riscv
12:47:54 ERROR src/syscall.c:249: Attempted to close a file descriptor < 3 (fd=0). Operation not supported.
12:47:54 ERROR src/syscall.c:249: Attempted to close a file descriptor < 3 (fd=1). Operation not supported.
12:47:54 ERROR src/syscall.c:249: Attempted to close a file descriptor < 3 (fd=2). Operation not supported.
12:47:54 INFO  src/main.c:300: RISC-V emulator is destroyed
```

Source code of `hello.elf` doesn't close any file descriptor.  This is likely because the operating system automatically closes all file descriptors belonging to a program when it terminates. As a result, close(0), close(1), and close(2) are executed, causing the error message to be printed multiple times.

I believe that outputting an error message in the syscall_close implementation may mislead users. I am currently considering a better way to display the error message.

src/syscall.c
```diff
@@ -244,10 +244,10 @@ static void syscall_close(riscv_t *rv)
...
+        rv_set_reg(rv, rv_reg_a0, -1);
```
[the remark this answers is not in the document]
visitorckw: I agree with robot's point. While it's a matter of taste, CONTRIBUTING.md explicitly advises avoiding nested if-else statements when possible.

src/syscall.c
ChinYikMing: We can utilize the built-in ELF loader to obtain the program counter (PC) of the exit entry in crt0. This allows us to determine whether the standard streams are closed by the process itself or by crt0, as crt0 follows the sequence: `exit()` → `close(fd)` when closing standard streams.

Please apply the patch if it skips the logging of closing standard streams by the crt0:
```diff
From 0fe2ac16114701fc53b0b2f90d49d0d77848df6e Mon Sep 17 00:00:00 2001
From: ChinYikMing <yikming2222@gmail.com>
Date: Fri, 14 Feb 2025 01:21:27 +0800
Subject: [PATCH] Deny closing standard stream by the process

---
 src/emulate.c | 12 +++++++++---
 src/riscv.c   |  9 +++++++++
 src/riscv.h   |  8 ++++++++
 src/syscall.c | 15 +++++++++++++++
 4 files changed, 41 insertions(+), 3 deletions(-)

diff --git a/src/emulate.c b/src/emulate.c
index 70ec291..ec6292c 100644
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -1053,13 +1053,19 @@ void rv_step(void *arg)
         /* by now, a block should be available */
         assert(block);
 
-        /* After emulating the previous block, it is determined whether the
+#if !RV32_HAS(SYSTEM)
+        /* on exit */
+        if (unlikely(block->ir_head->pc == PRIV(rv)->exit_addr))
+            PRIV(rv)->on_exit = true;
+#endif
+
+#if RV32_HAS(BLOCK_CHAINING)
+        /*
+         * After emulating the previous block, it is determined whether the
          * branch is taken or not. The IR array of the current block is then
          * assigned to either the branch_taken or branch_untaken pointer of
          * the previous block.
          */
-
-#if RV32_HAS(BLOCK_CHAINING)
         if (prev) {
             rv_insn_t *last_ir = prev->ir_tail;
             /* chain block */
diff --git a/src/riscv.c b/src/riscv.c
index d277618..61d95d2 100644
--- a/src/riscv.c
+++ b/src/riscv.c
@@ -434,6 +434,15 @@ riscv_t *rv_create(riscv_user_t rv_attr)
     if ((end = elf_get_symbol(elf, "_end")))
         attr->break_addr = end->st_value;
 
+#if !RV32_HAS(SYSTEM)
+    /* set not exiting */
+    attr->on_exit = false;
+
+    const struct Elf32_Sym *exit;
+    if ((exit = elf_get_symbol(elf, "exit")))
+        attr->exit_addr = exit->st_value;
+#endif
+
     assert(elf_load(elf, attr->mem));
 
     /* set the entry pc */
diff --git a/src/riscv.h b/src/riscv.h
index bc343e8..88eceef 100644
--- a/src/riscv.h
+++ b/src/riscv.h
@@ -548,6 +548,14 @@ typedef struct {
     /* the data segment break address */
     riscv_word_t break_addr;
 
+#if !RV32_HAS(SYSTEM)
+    /* the exit entry address */
+    riscv_word_t exit_addr;
+
+    /* flag to determine if the emulator exits the target program */
+    bool on_exit;
+#endif
+
     /* SBI timer */
     uint64_t timer;
 } vm_attr_t;
diff --git a/src/syscall.c b/src/syscall.c
index 09eab56..403fabe 100644
--- a/src/syscall.c
+++ b/src/syscall.c
@@ -230,6 +230,21 @@ static void syscall_close(riscv_t *rv)
     /* _close(fd); */
     uint32_t fd = rv_get_reg(rv, rv_reg_a0);
 
+#if !RV32_HAS(SYSTEM)
+    /*
+     * The crt0 closes standard file descriptor(0, 1, 2) when
+     * the process exits. Thus, the operations by the crt0
+     * should not considered as error.
+     */
+    if (fd < 3 && !PRIV(rv)->on_exit) {
+        rv_log_error(
+            "Attempted to close a file descriptor < 3 (fd=%u). Operation "
+            "not supported.",
+            fd);
+        return;
+    }
+#endif
+
     if (fd >= 3) { /* lookup the file descriptor */
         map_iter_t it;
         map_find(attr->fd_map, &it, &fd);
-- 
2.34.1

```

jserv (changes requested):
Rebase the latest `master` branch, which resolves all known CI regressions.

src/syscall.c
ChinYikMing: I forgot this: `rv_set_reg(rv, rv_reg_a0, -1);` :)

banglday: I have a concern about the newly added attributes, `riscv_word_t exit_addr` and `bool on_exit`.

Do these attributes have any other use cases? In other words, is marking whether the emulator is currently in the `exit()` phase useful in other scenarios? Does it bring any additional benefits to the overall emulator functionality?

If these attributes are added only to fix this issue, I worry that they might affect the overall simplicity of the code.

ChinYikMing: > I have a concern about the newly added attributes, `riscv_word_t exit_addr` and `bool on_exit`.
> 
> Do these attributes have any other use cases? In other words, is marking whether the emulator is currently in the `exit()` phase useful in other scenarios? Does it bring any additional benefits to the overall emulator functionality?
> 
> If these attributes are added only to fix this issue, I worry that they might affect the overall simplicity of the code.

IMO, adding these two attributes neither causes the emulator to regress nor increases its complexity. Additionally, returning -1 from the system call does not provide a clear error message to the user, as we do not maintain the [errno](https://man7.org/linux/man-pages/man3/errno.3.html) mechanism for specifying the type of error that occurred. Therefore, logging the error seems to be the simplest solution for now. @jserv Can you comment this?

banglday: @ChinYikMing  I totally agree that we should provide clear error message to user when syscall return -1. Just wondering if there is any advantage to add on_exit flag. Maybe it can be a new feature based on it.

ChinYikMing: > @ChinYikMing I totally agree that we should provide clear error message to user when syscall return -1. Just wondering if there is any advantage to add on_exit flag. Maybe it can be a new feature based on it.

Oh, I miss the first question. I think no other usage for now. BTW, how would you know the emulator is exiting the ELF process without the `on_exit` flag?

banglday: `on_exit` flag is necessary to know the emulator is exiting the ELF process. Just wondering is there any other usage in the future. I will add this patch to this PR.

ChinYikMing: Consider rename `on_exit` to `is_exit` to avoid naming conflict with libc function [on_exit](https://man7.org/linux/man-pages/man3/on_exit.3.html).

banglday: @ChinYikMing Applied the patch and removed the condition `if (fd>=3)`. Please help review the code. Thanks.

jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: f706e107c6945ee581a67238d6a11cd288dc8038 | Previous: c475d34cfff8d360c9d6ab0ee84b6e26972c65dc | Ratio |
|-|-|-|-|
| `Dhrystone` | `1203` Average DMIPS over 10 runs | `1257` Average DMIPS over 10 runs | `1.04` |
| `Coremark` | `962.856` Average iterations/sec over 10 runs | `971.038` Average iterations/sec over 10 runs | `1.01` |

</details>

ChinYikMing:
@banglday  Could you add `Co-authored-by: ChinYikMing <yikming2222@gmail.com>` to the end of the commit message which gives me some credit, see [here](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors)? Thanks!

ChinYikMing (commented):
Run `clang-format -i src/*.{c,h}` in the top level of rv32emu working directory to solve the coding style failed in CI.

banglday:
> @banglday Could you add `Co-authored-by: ChinYikMing <yikming2222@gmail.com>` to the end of the commit message which gives me some credit, see [here](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors)? Thanks!

Sure! I forgot to add it. Thanks for reminding.

ChinYikMing:
@banglday  I think you have to use `clang-format-18` or above.

src/rv32_template.c
```diff
@@ -3030,4 +3030,4 @@ RVOP(
         assert; /* FIXME: Implement */
     }))
 
-#endif
+#endif
```
ChinYikMing: Do not remove the newline. Check [https://medium.com/@alexey.inkin/how-to-force-newline-at-end-of-files-and-why-you-should-do-it-fdf76d1d090e](https://medium.com/@alexey.inkin/how-to-force-newline-at-end-of-files-and-why-you-should-do-it-fdf76d1d090e).

banglday: Got it. It is changed by `clang-format-18`. I am wondering when you run `clang-format-18 -i src/*.{c, h}` command, will `clang-format` remove this newline or not.

src/syscall.c
```diff
@@ -230,24 +230,35 @@ static void syscall_close(riscv_t *rv)
     /* _close(fd); */
     uint32_t fd = rv_get_reg(rv, rv_reg_a0);
 
-    if (fd >= 3) { /* lookup the file descriptor */
```
ChinYikMing, line 233: Why removing this?

We keep the standard stream open because some logs may be generated after the ELF process ends. Additionally, the emulator's crt0 will handle closing the standard stream.

src/rv32_template.c
banglday: Finished adding a newline in the end.

ChinYikMing:
LGTM. Thanks! 

Let's wait for the CI to pass and see if anyone has other ideas.

ChinYikMing (approved):
LGTM. Thanks!

Let's wait for the CI to pass and see if anyone has other ideas.

jserv:
Thank @banglday and @ChinYikMing for contributing!

PR #572, opened 2025-02-20
Enable VirtIO block to access hostOS /dev/ block devices

ChinYikMing:
The user may not always have a disk image but might have a /dev/x block device, such as a USB drive that they want to share with the guest OS. So, allowing this type of virtio-blk source is intuitive. To support this, ioctl is used to retrieve the actual size of the /dev/x block device. This implementation supports both Apple and Linux platforms.

### Testing
1. Clone the repo:
```
$ git clone https://github.com/ChinYikMing/rv32emu.git -b virtio-blk-ioctl --depth 1 && cd rv32emu
```

2. Get the prebuilt images and build:
```
$ make ENABLE_SYSTEM=1 INITRD_SIZE=32 -j100 && make ENABLE_SYSTEM=1 artifact
```

3. Prepare /dev/x device:

macOS:
```bash
# Might need to install file system utility
$ brew install e2fsprogs

# Create a disk image
$ dd if=/dev/zero of=disk.img bs=4M count=32
$ $(brew --prefix e2fsprogs)/sbin/mkfs.ext4 disk.img

# Setup a block device with the disk image
$ hdiutil attach -nomount disk.img              # this would output the <block-device>
$ sudo build/rv32emu -k build/linux-image/Image -i build/linux-image/rootfs.cpio -x vblk:<block-device>
```

Linux
```bash
# Create a disk image
$ dd if=/dev/zero of=disk.img bs=4M count=32
$ mkfs.ext4 disk.img

# Setup a loop back device with the disk image
$ sudo losetup -f              # this would output the <first-unused-loop-device>
$ sudo losetup <first-unused-loop-device> disk.img 
$ sudo build/rv32emu -k build/linux-image/Image -i build/linux-image/rootfs.cpio -x vblk:<first-unused-loop-device>
```

4. Check the kernel boot log which should have something as the following:
```
...
...
[    0.583109] virtio_blk virtio0: 1/0/0 default/read/poll queues
[    0.584873] virtio_blk virtio0: [vda] 262144 512-byte logical blocks (134 MB/128 MiB)
...
...
```

5. Mount the virtio block device and manipulate the device:
```bash
# mkdir mnt
# mount /dev/vda mnt
# echo "rv32emu" > mnt/emu.txt
# ls mnt
```
Might see:
```
emu.txt     lost+found
```

6. Detach newly created block device:

macOS
```
$ hdiutil detach <block-device>
```

Linux
```
$ losetup -d <first-unused-loop-device>
```

### Known Issue:
On macOS/arm64, OS v15.3.1, the /dev/ block device cannot be `mmap` with error log: `ERROR src/devices/virtio-blk.c:475: Could not map disk <block-device>`.

After some debugging, the errno is set to `EINVAL`. Then, reading the man page of `mmap` in the system:
```
[EINVAL]           MAP_FIXED was specified and the addr argument was not page aligned, or part of the desired address space resides out of the valid address space
                        for a user process.

[EINVAL]           flags does not include either MAP_PRIVATE or MAP_SHARED.

[EINVAL]           flags includes bits that are not part of any valid flags value.

[EINVAL]           The len argument was negative or zero. Historically, the system call would not return an error if the argument was zero.  See other potential
                        additional restrictions in the COMPATIBILITY section below.

[EINVAL]           The offset argument was not page-aligned based on the page size as returned by getpagesize(3).
```
The page size in my system is 16KiB, which the block device size aligned of (128MiB % 16KiB == 0). The rest of the requirements appear to be met, so this seems a bit strange to me. If anyone has a macOS/arm64 machine, maybe can give it a try. 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request significantly enhances the virtio-blk driver, enabling guest OS access to hostOS /dev/ block devices like USB drives. It implements ioctl calls for accurate device size retrieval, improves compatibility for Apple and Linux, and includes better error handling, logging, and updated documentation. Unit tests have been added to ensure reliability.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
</div>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -65,7 +65,14 @@ jobs:
         actions-cache-folder: 'emsdk-cache'
     - name: Set parallel jobs variable
       run: |
-            echo "PARALLEL=-j$(nproc)" >> $GITHUB_ENV
+            echo "PARALLEL=-j$(nproc)" >> "$GITHUB_ENV"
+            echo "BOOT_LINUX_TEST=TMP_FILE=\$(mktemp "$RUNNER_TEMP/tmpfile.XXXXXX"); \
+                                  sudo env TMP_FILE=\${TMP_FILE} .ci/boot-linux-prepare.sh setup; \
+                                  . \${TMP_FILE}; \
+                                  .ci/boot-linux.sh; \
+                                  EXIT_CODE=\$?; \
+                                  sudo env TMP_FILE=\${TMP_FILE} BLK_DEV=\${BLK_DEV} .ci/boot-linux-prepare.sh cleanup; \
+                                  exit \${EXIT_CODE};" >> "$GITHUB_ENV"
     - name: fetch artifact first to reduce HTTP requests
       env:
         CC: ${{ steps.install_cc.outputs.cc }}
@@ -265,7 +272,7 @@ jobs:
         CC: ${{ steps.install_cc.outputs.cc }}
       run: |
             make distclean && make INITRD_SIZE=32 ENABLE_SYSTEM=1 $PARALLEL && make ENABLE_SYSTEM=1 artifact $PARALLEL
-            .ci/boot-linux.sh
+            bash -c "${BOOT_LINUX_TEST}"
             make ENABLE_SYSTEM=1 clean
       if: ${{ always() }}
     - name: boot Linux kernel test (JIT)
@@ -273,7 +280,7 @@ jobs:
         CC: ${{ steps.install_cc.outputs.cc }}
       run: |
             make distclean && make INITRD_SIZE=32 ENABLE_SYSTEM=1 ENABLE_JIT=1 ENABLE_T2C=0 ENABLE_MOP_FUSION=0 $PARALLEL && make ENABLE_SYSTEM=1 artifact $PARALLEL
-            .ci/boot-linux.sh
+            bash -c "${BOOT_LINUX_TEST}"
             make ENABLE_SYSTEM=1 ENABLE_JIT=1 ENABLE_T2C=0 ENABLE_MOP_FUSION=0 clean
       if: ${{ always() }}
     - name: Architecture test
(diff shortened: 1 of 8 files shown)


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: e2e87ee26385a3602bdbee4629d58bc85e986fa1 | Previous: d1743ff59790b50e9fb142022455dccd48637ac0 | Ratio |
|-|-|-|-|
| `Dhrystone` | `1318` Average DMIPS over 10 runs | `1315` Average DMIPS over 10 runs | `1.00` |
| `Coremark` | `916.239` Average iterations/sec over 10 runs | `913.984` Average iterations/sec over 10 runs | `1.00` |

</details>

ChinYikMing:
On Apple platform, if `mmap()` /dev/ block devices fails, fallback to `malloc()`. In this way, both Apple and Linux platform are supported to access hostOS /dev/ block devices.

src/devices/virtio-blk.c
```diff
@@ -403,29 +424,81 @@ uint32_t *virtio_blk_init(virtio_blk_state_t *vblk,
...
-    fstat(disk_fd, &st);
-    VBLK_PRIV(vblk)->disk_size = st.st_size;
+    const char *disk_file_dirname = dirname(disk_file);
+    if (!disk_file_dirname) {
+        rv_log_error("Fail dirname disk_file: %s", disk_file);
+        goto disk_size_fail;
+    }
+    /* Get the disk size */
+    uint64_t disk_size;
+    if (strcmp(disk_file_dirname, "/dev") ==
```
jserv: Simply rewrite as `!strcmp(...)`.

ChinYikMing:
37cf3f3 refines the Linux boot script to validate all new system emulation features, excluding SDL.

.github/workflows/main.yml
```diff
@@ -265,15 +265,15 @@ jobs:
         CC: ${{ steps.install_cc.outputs.cc }}
       run: |
             make distclean && make INITRD_SIZE=32 ENABLE_SYSTEM=1 $PARALLEL && make ENABLE_SYSTEM=1 artifact $PARALLEL
-            .ci/boot-linux.sh
+            sudo .ci/boot-linux.sh
```
jserv: Instead of running with superuser, can you relax the access permission of loop device and then run the emulator in non-root user?

ChinYikMing: > Instead of running with superuser, can you relax the access permission of loop device and then run the emulator in non-root user?

The system validation flow in CI is abstracted into a variable called BOOT_LINUX_TEST which uses sudo only for data preparation and does not use sudo to run .ci/boot-linux.sh, enhancing security when launching the system VM.

jserv:

PR #574, opened 2025-02-22
Eliminate Linux SUBLEVEL bump and decouple Linux image build CI

ChinYikMing:
Add SHA256SUM as a new checksum verifier for downloading Linux tarballs since the latest Linux tarball is available at https://cdn.kernel.org/pub/linux/kernel/v6.x/, and the official SHA256 values can be found at https://cdn.kernel.org/pub/linux/kernel/v6.x/sha256sums.asc. To maintain compatibility with SHA1SUM, the verify function in mk/external.mk has been modified to support multiple SHA algorithms. With these changes, the Linux kernel SUBLEVEL bump is no longer necessary and builds can be triggered via a workflow dispatch event. Afterwards, version updates of Linux will only apply to the VERSION and PATCHLEVEL. 

Separate the Linux image build process from the build-artifact CI. The Linux image is now majorly built via a workflow dispatch event while ELF artifacts do not need to be built in the same event, thus decoupling these processes ensures that ELF artifacts are not unnecessarily rebuilt. This modular approach improves the flexibility of the artifact building workflow. 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request enhances the Linux image build process by replacing SHA1SUM with SHA256SUM for improved security. It modifies the verify function to support multiple SHA algorithms and decouples the build processes for Linux images and ELF artifacts, streamlining workflows and enhancing verification accuracy.
<br>
<br>
<br>
<br>
</div>

### DIFF
--- a/.github/workflows/build-linux-artifacts.yml
+++ b/.github/workflows/build-linux-artifacts.yml
@@ -1,0 +1,78 @@
+name: Build Linux artifact
+
+on:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+
+jobs:
+  detect-file-change:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          submodules: 'true'
+      - name: Test file change of Linux image
+        id: test-linux-image-version-change
+        uses: tj-actions/changed-files@v45
+        with:
+          files: |
+            mk/external.mk
+      - name: Set alias
+        id: has_changed_files
+        run: |
+          if [[ ${{ steps.test-linux-image-version-change.outputs.any_modified }} == true ]]; then
+            # Determine if the changes are from Buildroot or the Linux version (The Linux might have several patches, so also need to check the SHA value)
+            echo -n $(git --no-pager diff HEAD^ HEAD | grep -e "+BUILDROOT_VERSION" -e "+LINUX_VERSION" -e "+LINUX_PATCHLEVEL") >> linux-image-version-change
+            if [[ -s linux-image-version-change ]]; then
+                echo "has_changed_linux_image_version=true" >> $GITHUB_OUTPUT
+            else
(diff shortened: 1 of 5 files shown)


ChinYikMing:
The workflow_dispatch event should be manually triggered periodically to build the latest Linux image. After these are merged, only the `VERSION` and `PATCHLEVEL` of the Linux kernel and the VERSION of Buildroot are needed to be bumped.

ChinYikMing:
The Doom and Quake patches shall be applied when building rootfs.cpio by Buildroot. The patches could stored at [rv32emu-prebuilt](https://github.com/sysprog21/rv32emu-prebuilt). Will do in future PR.

.github/workflows/build-linux-artifacts.yml
```diff
@@ -0,0 +1,78 @@
...
+          else
+            echo "has_changed_linux_image_version=false" >> $GITHUB_OUTPUT
+          fi
+    outputs:
+      has_changed_linux_image_version: ${{ steps.has_changed_files.outputs.has_changed_linux_image_version }}
+
+  build-linux-image-artifact:
+    needs: [detect-file-change]
+    if: ${{ needs.detect-file-change.outputs.has_changed_linux_image_version == 'true' || github.event_name == 'workflow_dispatch' }}
+    runs-on: ubuntu-22.04
```
jserv, line 43: Why not Ubuntu 24.04?

ChinYikMing: Actually, we can use `ubuntu-latest` for all Linux workflow? In this way, no more old version concern  (hopefully `ubuntu-latest` remains stable).

jserv: > Actually, we can use `ubuntu-latest` for all Linux workflow? In this way, no more old version concern (hopefully `ubuntu-latest` remains stable).

No, I wanted to specify the versions for reproducible builds.

vacantron (approved):
LGTM

jserv:

PR #576, opened 2025-02-24
Allow emcc build and validate in CI

ChinYikMing:
### What changes?
Allow emcc build for both ISA or system emulation. Additionally, add emcc build check to CI.

### How to reproduce emcc build fail for system emulation?
```
$ source ~/emsdk/emsdk_env.sh
$ make CC=emcc ENABLE_SYSTEM=1 -j8
```
Will see the error like the following:
```
BIN2C	src/minimal_dtb.h
/bin/sh: 1: build/bin2c: Permission denied
make: *** [mk/system.mk:22: src/minimal_dtb.h] Error 126
```

The reason is that the `bin2c` is compiled to Wasm by `emcc`  instead of native executable which cannot build the device tree blob to C application. 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request enhances the build system by introducing support for emcc, enabling ISA and system emulation. It includes checks for GCC or Clang usage when emcc is unsuitable and implements error handling for unsupported compilers, improving robustness.
<br>
<br>
<br>
<br>
</div>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -58,6 +58,11 @@ jobs:
       uses: rlalik/setup-cpp-compiler@master
       with:
         compiler: ${{ matrix.compiler }}
+    - name: Setup emsdk
+      uses: mymindstorm/setup-emsdk@v14
+      with:
+        version: 3.1.51
+        actions-cache-folder: 'emsdk-cache'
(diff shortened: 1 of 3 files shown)


jserv:

PR #580, opened 2025-03-01
Fix Docker image build

ChinYikMing:
Commit a40fd0b adds objects file that built with the lib of SDL2 and SDL2_mixer for linker when building rv_histogram, the current Docker image lacks of SDL2 and SDL2_mixer shared objects, so installing them to fix the missing shared objects.

The final Docker image size changes from around 4 MiB -> 28 MiB. 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request enhances the Dockerfile for the rv_histogram application by upgrading to Alpine 3.21 and adding SDL2 libraries, addressing missing shared objects and increasing the image size from 4 MiB to 28 MiB.
<br>
<br>
<br>
<br>
<b>Estimated effort to review (1-5, lower is better)</b>: 2 - The changes are straightforward and primarily involve dependency updates, making the review process relatively simple.
</div>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,13 @@
 FROM alpine:3.19 AS base_gcc
 
-RUN apk add --update alpine-sdk git wget
+RUN apk add --update alpine-sdk git wget sdl2-dev sdl2_mixer-dev
 
 # copy in the source code
 WORKDIR /home/root/rv32emu
 COPY . .
 
 # generate execution file for rv32emu and rv_histogram
-RUN make ENABLE_SDL=0
+RUN make
 RUN make tool
 
 FROM alpine:3.19 AS final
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,21 +1,24 @@
 FROM alpine:3.19 AS base_gcc
```
jserv: End of support of Alpine Linux v3.19.x is on Nov 1 this year. We should migrate to 3.20 or 3.21 later.

Dockerfile
```diff
@@ -1,21 +1,24 @@
...
 
-RUN apk add --update alpine-sdk git wget
+RUN apk add --update alpine-sdk git wget sdl2-dev sdl2_mixer-dev
 
 # copy in the source code
 WORKDIR /home/root/rv32emu
 COPY . .
 
+
```
jserv: Don't change this.

Dockerfile
```diff
@@ -1,21 +1,24 @@
 FROM alpine:3.19 AS base_gcc
 
-RUN apk add --update alpine-sdk git wget
+RUN apk add --update alpine-sdk git wget sdl2-dev sdl2_mixer-dev
```
jserv, line 3: The primary reason for excluding SDL2 packages from the current Alpine Linux image is to maintain a smaller image size, which enables quicker startup and emulation through [blink](https://github.com/jart/blink) or similar emulators. Including SDL2 in the Alpine Linux image would create challenges when attempting to run graphics with the emulator.

jserv:
I defer to @henrybear327 for confirmation.

jserv:

Issue #575, opened 2025-02-23
`make arch-test` fails unless `ENABLE_ARCH_TEST` is explicitly set


vestata:
I encountered an issue on an `x86_64` machine(Ubuntu 24.04) when running `make arch-test` directly. The error output is as follows:

```
$ make arch-test 
  AR    build/softfloat/softfloat.a
  CC    build/map.o
  CC    build/utils.o
  CC    build/decode.o
  CC    build/io.o
  CC    build/syscall.o
  CC    build/emulate.o
  CC    build/riscv.o
  CC    build/log.o
  CC    build/elf.o
  CC    build/cache.o
  CC    build/mpool.o
  CC    build/syscall_sdl.o
  CC    build/main.o
  LD    build/rv32emu
Fetching SHA-1 of prebuilt binaries ... [OK]
Checking SHA-1 of prebuilt binaries ... [OK]
git submodule update --init tests/
cp: cannot stat 'build/rv32emu-prebuilt-sail-x86': No such file or directory
make: *** [mk/riscv-arch-test.mk:23: arch-test] Error 1
```

PR #547 introduced the `ARCH_TEST` feature.
In CI environment (`.ci/riscv-tests.sh`), this problem doesn't occur because `ENABLE_ARCH_TEST` is explicitly set to `1` before running `make arch-test`. However, the default `ENABLE_ARCH_TEST` is set to `0`. As a result, if the repository is cloned or `make distclean` is executed and the `rv32emu-prebuilt-sail-$(HOST_PLATFORM)` file has not been downloaded previously, users who manually run `make arch-test` will encounter the error shown above.

When `ENABLE_ARCH_TEST` is set to `0`, the condition for fetching `rv32emu-prebuilt-sail-$(HOST_PLATFORM)` is not met, and therefore the necessary file is not fetched.
The relevant snippet in `mk/artifact.mk`:
```Makefile
ifeq ($(call has, PREBUILT), 1)
    ifeq ($(call has, SYSTEM), 1)
        $(call fetch-releases-tag,Linux-Image,rv32emu-linux-image-prebuilt.tar.gz,Linux image)
    else ifeq ($(call has, ARCH_TEST), 1)
        $(call fetch-releases-tag,sail,rv32emu-prebuilt-sail-$(HOST_PLATFORM),Sail model)
    else
        $(call fetch-releases-tag,ELF,rv32emu-prebuilt.tar.gz,Prebuilt blob)
    endif

[...]

PR #583, opened 2025-03-03
Update README to clarify make arch-test usage

vestata:
Running make arch-test without explicitly setting ENABLE_ARCH_TEST=1 may fail due to missing prebuilt binaries. This issue arises because the default value of ENABLE_ARCH_TEST is 0, stopping Sail model from being fetched.

Additionally, if ENABLE_SYSTEM=1 was previously set, the build system fetches the kernel image instead of the Sail model. To ensure the correct prebuilt binary is fetched, it is recommended to run make distclean beforehand.

This serves as a temporary solution to #575.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,8 @@ $ brew install dtc
 ```
 
 #### Build and run system emulation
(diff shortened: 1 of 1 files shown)


vacantron (approved):

jserv:
Thank @vestata for contributing!

PR #592, opened 2025-04-28
Advance timer properly during JIT execution

vacantron:
In the current implementation, the timer would not be updated until the control flow exits the JIT-ed code. This commit updates the timer correctly after every instruction has been dispatched. 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request enhances the timer functionality in the JIT execution context by replacing the static counter with a dynamic timer variable. It introduces new methods for timer management in JIT compilation and T2C generation, improving the accuracy of time-related operations.
<br>
<br>
<br>
<br>
<b>Estimated effort to review (1-5, lower is better)</b>: 2 - The changes are straightforward and well-defined, making the review process relatively simple.
</div>

### DIFF
--- a/src/jit.c
+++ b/src/jit.c
@@ -211,6 +211,7 @@ enum operand_size {
     S8,
     S16,
     S32,
+    S64,
 };
 
 #if defined(__x86_64__)
@@ -756,7 +757,7 @@ static inline void emit_alu64(struct jit_state *state, int op, int src, int dst)
 #endif
 }
 
-#if RV32_HAS(EXT_M)
+#if RV32_HAS(EXT_M) || defined(__aarch64__)
 static inline void emit_alu64_imm8(struct jit_state *state,
                                    int op,
                                    int src UNUSED,
@@ -771,6 +772,9 @@ static inline void emit_alu64_imm8(struct jit_state *state,
     if (op == 0xc1) {
         emit_load_imm(state, R10, imm);
         emit_dataproc_2source(state, true, DP2_LSRV, dst, dst, R10);
+    } else if (src == 0) {
+        emit_load_imm(state, R10, imm);
+        emit_addsub_register(state, true, AS_ADD, dst, dst, R10);
     }
 #endif
 }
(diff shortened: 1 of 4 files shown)


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: a27625bbbb672ea690717b12f0c13378fd462d69 | Previous: 5001cfd1645c8be9caeb3986fd1cfeab814f5ebf | Ratio |
|-|-|-|-|
| `Dhrystone` | `1285` Average DMIPS over 10 runs | `1087` Average DMIPS over 10 runs | `0.85` |
| `Coremark` | `915.126` Average iterations/sec over 10 runs | `915.504` Average iterations/sec over 10 runs | `1.00` |

</details>

src/jit.c
```diff
@@ -2026,10 +2036,28 @@ void parse_branch_history_table(struct jit_state *state,
...
+#elif defined(__aarch64__)
+    emit_load(state, S64, parameter_reg[0], temp_reg, offsetof(riscv_t, timer));
+    emit_alu64_imm8(state, 0, 0, temp_reg, 1);
+    emit_store(state, S64, temp_reg, parameter_reg[0],
+               offsetof(riscv_t, timer));
+#endif
+}
+
 #define GEN(inst, code)                                                       \
     static void do_##inst(struct jit_state *state UNUSED, riscv_t *rv UNUSED, \
```
ChinYikMing, line 2057: nit: Since the timer is updated for every JIT instruction, the `state` is actively used for each instruction. Consider removing the `UNUSED` attribute to improve readability and avoid potential confusion.

vacantron: For the unsupported operations in `src/rv32_jit.c`, we simply make the simulation crash. Removing this statement would make the compiler raise warning.

ChinYikMing (approved):

src/jit.c
```diff
@@ -2026,10 +2036,28 @@ void parse_branch_history_table(struct jit_state *state,
     }
 }
 
+void emit_jit_update_timer(struct jit_state *state)
```
jserv: How about renaming this function to `emit_jit_inc_timer`? It emphasizes the increment operation, which is the core action (incrementing the timer).

jserv (changes requested):
Fix typo: "untill"

src/jit.c
vacantron: In some other implementations, it might only be updated at the end of a basic block, but we update it every time for convenience here. So I would prefer to keep it that way.

jserv:

PR #600, opened 2025-06-10
Enhance CI formatting checks for Python files

Lzzz666:
- Update `.ci/check-format.sh` to include checks for Python files using Black.
- Modify `.github/workflows/main.yml` to install Python dependencies for formatting.
- Add Python-specific settings to `.editorconfig` to enforce consistent formatting rules.

This ensures that both shell scripts and Python files adhere to defined coding standards. 
 <div id='description'>
    <a href="https://bito.ai#summarystart"></a>
<h3>Summary by Bito</h3>
This pull request enhances CI formatting checks by integrating support for Python files using the Black formatter. It updates CI scripts and GitHub Actions to include Python-specific checks and dependencies, while refactoring existing Python code for improved readability and maintainability. Additionally, the JIT template generation logic has been refactored for enhanced clarity.

</div>

### DIFF
--- a/.ci/check-format.sh
+++ b/.ci/check-format.sh
@@ -21,4 +21,14 @@ for file in ${SH_SOURCES}; do
 done
 SH_MISMATCH_FILE_CNT=$(shfmt -l ${SH_SOURCES})
 
-exit $((C_MISMATCH_LINE_CNT + SH_MISMATCH_FILE_CNT))
+PY_SOURCES=$(find "${REPO_ROOT}" | egrep "\.py$")
+for file in ${PY_SOURCES}; do
+    echo "Checking Python file: ${file}"
+    black --diff "${file}"
+done
+PY_MISMATCH_FILE_CNT=0
+if [ -n "${PY_SOURCES}" ]; then
+    PY_MISMATCH_FILE_CNT=$(echo "$(black --check ${PY_SOURCES} 2>&1)" | grep -c "^would reformat ")
+fi
+
+exit $((C_MISMATCH_LINE_CNT + SH_MISMATCH_FILE_CNT + PY_MISMATCH_FILE_CNT))
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,6 +39,7 @@ However, participation requires adherence to fundamental ground rules:
   While there is some flexibility in basic style, it is crucial to stick to the current coding standards.
   Complex algorithmic constructs without proper comments will not be accepted.
 * Shell scripts must be formatted before submission. Use consistent flags across the project to ensure uniform formatting.
+* Python scripts must be formatted before submission. Use consistent flags across the project to ensure uniform formatting.
 * External pull requests should include thorough documentation in the pull request comments for consideration.
 * When composing documentation, code comments, and other materials in English,
   please adhere to the American English (`en_US`) dialect.
@@ -48,9 +49,12 @@ However, participation requires adherence to fundamental ground rules:
 Software requirement:
 * [clang-format](https://clang.llvm.org/docs/ClangFormat.html) version 18 or later.
 * [shfmt](https://github.com/mvdan/sh).
+* [black](https://github.com/psf/black) version 25.1.0.
 
-This repository consistently contains an up-to-date `.clang-format` file with rules that match the explained ones and uses shell script formatting supported by `shfmt`.
-For maintaining a uniform coding style, execute the command `clang-format -i *.{c,h}` and `shfmt -w $(find . -type f -name "*.sh")`.
+To maintain a uniform style across languages, run:
+* `clang-format -i *.{c,h}` to apply the project’s C/C++ formatting rules from the up-to-date .clang-format file.
+* `shfmt -w $(find . -type f -name "*.sh")` to clean and standardize all shell scripts.
+* `black .` to enforce a consistent, idiomatic layout for Python code.
 
 ## Coding Style for Shell Script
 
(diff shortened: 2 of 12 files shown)


jserv (changes requested):
Check `CONTRIBUTING.md` carefully and enforce the guidelines.

.editorconfig
```diff
@@ -11,4 +11,4 @@ insert_final_newline = true
 function_next_line = true
 switch_case_indent = true
 space_redirects = true
-binary_next_line = true
+binary_next_line = true
```
jserv: Be aware of newline.

CONTRIBUTING.md
```diff
@@ -65,6 +66,53 @@ Shell scripts must be clean, consistent, and portable. The following `shfmt` rul
...
+extend-exclude = '''
+/(
+  # directories
+  \.git
+  | build
+)/
+'''
+```
+
+
```
jserv: Avoid unnecessary blank lines.

jserv (changes requested):
Use `git rebase -i` to squash and rework git commit messages.
Enforce the rules mentioned in https://cbea.ms/git-commit/ .

CONTRIBUTING.md
```diff
@@ -48,9 +48,10 @@ However, participation requires adherence to fundamental ground rules:
 Software requirement:
 * [clang-format](https://clang.llvm.org/docs/ClangFormat.html) version 18 or later.
 * [shfmt](https://github.com/mvdan/sh).
+* [black](https://github.com/psf/black) for Python code formatting.
```
ChinYikMing: Simply mention [black](https://github.com/psf/black) is informative.

CONTRIBUTING.md
```diff
@@ -48,9 +48,10 @@ However, participation requires adherence to fundamental ground rules:
...
 
-This repository consistently contains an up-to-date `.clang-format` file with rules that match the explained ones and uses shell script formatting supported by `shfmt`.
-For maintaining a uniform coding style, execute the command `clang-format -i *.{c,h}` and `shfmt -w $(find . -type f -name "*.sh")`.
+This repository consistently contains an up-to-date `.clang-format` file with rules that match the explained ones, uses shell script formatting supported by `shfmt`, and Python code formatting with `black`.
+For maintaining a uniform coding style, execute the command `clang-format -i *.{c,h}`, `shfmt -w $(find . -type f -name "*.sh")`, and `black .` for Python files.
```
ChinYikMing: Refine the sentence to emphasize the goal of each command for different target programming languages.

.ci/check-format.sh
```diff
@@ -21,4 +21,15 @@ for file in ${SH_SOURCES}; do
...
-exit $((C_MISMATCH_LINE_CNT + SH_MISMATCH_FILE_CNT))
+PY_SOURCES=$(find "${REPO_ROOT}" | egrep "\.py$")
+for file in ${PY_SOURCES}; do
+    echo "Checking Python file: ${file}"
+    black --diff "${file}"
+done
+PY_MISMATCH_FILE_CNT=0
+if [ -n "${PY_SOURCES}" ]; then
+    PY_MISMATCH_OUTPUT=$(black --check ${PY_SOURCES} 2>&1)
+    PY_MISMATCH_FILE_CNT=$(echo "${PY_MISMATCH_OUTPUT}" | grep -c "^would reformat ")
```
ChinYikMing: Combine two statements into single for simplicity:
```
PY_MISMATCH_FILE_CNT=$(echo "$(black --check ${PY_SOURCES} 2>&1)" | grep -c "^would reformat ")
```

ChinYikMing (commented):
Add a description about Python code formatting in the following section: https://github.com/sysprog21/rv32emu/blob/88ddb73f1db5de08b779804a4aba08b4a417d6fe/CONTRIBUTING.md?plain=1#L34

Lzzz666:
@ChinYikMing  I’ve pushed the updates, please check!

CONTRIBUTING.md
```diff
@@ -65,6 +71,51 @@ Shell scripts must be clean, consistent, and portable. The following `shfmt` rul
 * Add spaces around redirection operators (e.g., `>`, `>>`).
 * Place binary operators (e.g., `&&`, `|`) on the next line when breaking lines.
 
+## Coding Style for Python
+
+Python scripts must be clean, consistent, and adhere to modern Python best practices. The following formatting rules are enforced project-wide using `black`:
+
+* Use 4 spaces for indentation (never tabs).
+* Limit lines to 88 characters maximum.
```
visitorckw: Since the CONTRIBUTING.md specifies an 80 characters per line limit for C code, I noticed that 88 characters are used here for Python. I'm curious - how was the number 88 chosen? Should we consider aligning the limit across languages?

visitorckw:
I noticed that your commit message mentions several different changes in a single patch. Would it be better to split them into separate patches? For example, one patch could fix the Python code style, and another could add the CI check - instead of combining them all in one.

CONTRIBUTING.md
Lzzz666: I agree with your suggestion—having the same limit across languages makes sense. I initially chose 88 characters simply because that’s Black’s default.

pyproject.toml
```diff
@@ -0,0 +1,11 @@
+[tool.black]
+line-length = 80
+target-version = ['py37', 'py38', 'py39', 'py310', 'py311']
```
ChinYikMing: Why not including `py312`? Just for curiosity, any reason for that (e.g., parser compatibility)?

pyproject.toml
```diff
@@ -0,0 +1,11 @@
...
+include = '\.pyi?$'
+extend-exclude = '''
```
ChinYikMing: Under what circumstances do the `.git` and `build` directories generate Python files? If no, we can eliminate the rules in the configuration file for simplicity.

@jserv @visitorckw , can you confirm this?

pyproject.toml
Lzzz666: I noticed that using `Black v24.8.0` will issue an error when used with `Python 3.12.5`, due to an upstream memory safety issue. If we decide to support `py312`, should we also include `py313` ? Additionally, should we pin Black to the latest v25.1.0 when installing? Since `Black v24.10.0`, it immediately errors in a `Python 3.12.5` environment, and it requires at least `Python 3.9`.

pyproject.toml
jserv: > Under what circumstances do the `.git` and `build` directories generate Python files? If no, we can eliminate the rules in the configuration file for simplicity.

The use of Python scripts is dedicated to code generation at first glance. The formatting of Python code is not really a serious concern since it is used as one-shot tooling.

Lzzz666: Thanks for the suggestion, I’ve removed the extend-exclude rules from Black configuration.

ChinYikMing:
LGTM.

jserv:
Thank @Lzzz666 for contributing!

PR #389, opened 2024-03-27
CI: deploy wasm

ChinYikMing:
Since wasm is deployed on Github Pages, thus it is reasonable to use CI to deploy automatically.

The CI supports two events: merged pull request and workflow_dispatch. The former only run after merging and latter is used to deploy by maintainer manually. It also verifies whether any associated files have changed. If so, carry out the CI; if not, stop.

Related: #75

PR #602, opened 2025-07-04
Enable System Emulation in Web Browsers

ChinYikMing:
# Demo Link
[link](https://chinyikming.github.io/rv32emu-demo/system/). Press 'Run Linux' button to boot Linux. After booting, use it as usual. Experience graphics and sounds by running doom-riscv, quake, smolnes.

# Demo Video
https://github.com/user-attachments/assets/7cc1f507-4102-4193-884f-a9e8d6b99054

# Description
User-space emulation has been supported and deployable in WebAssembly
since https://github.com/sysprog21/rv32emu/pull/389, but system emulation was not yet available. With https://github.com/sysprog21/rv32emu/pull/508,
system emulation was introduced, and later https://github.com/sysprog21/rv32emu/pull/551 added support for
trap-and-emulate of guest Linux SDL syscalls, enabling offloading to the
host’s SDL backend. This commit bridges these components together to
enable full system emulation in the browser.

xterm.js is leveraged as the frontend terminal, bridged with the backend
VM shell through a custom buffer management mechanism. This mechanism
provides a shell experience in the browser that closely resembles a real
terminal.

To handle terminal input without blocking the browser’s cooperative
multitasking model, the original approach of mapping the read()
system call to window.prompt() is avoided. Blocking behavior would
freeze the browser’s event loop and degrade responsiveness. Instead,
input from xterm.js is stored in a shared input buffer. The rv32emu
periodically checks this buffer when handling external interrupts, and
if input is available, it is read and consumed by the guest OS shell.

The SDL graphic and sound backend are also supported. After booting the
guest Linux system, users can run graphical applications such as
doom-riscv, quake, or smolnes. These programs can be exited using
Ctrl+C or their built-in exit funtionality.

To reduce the size of the WebAssembly build and for the sake of the
modularity, the project is now separated into user and system targets.
As a result, two dedicated HTML files and corresponding preload
JavaScript files are introduced:
- user.html with user-pre.js
- system.html with system-pre.js

Navigation buttons are added to the index pages of both user and system
demos, allowing users to switch easily between the two modes.
Note that these navigation buttons are only functional when both user
and system demos are deployed together, otherwise, the target pages may
not be reachable.
To improve usability, logic is implemented to disable and enable the
"Run" button at appropriate times, preventing accidental re-execution
when the process is already running.

Additional improvements:
- Ensure xterm.js uses \r\n instead of \n when logging to correctly move
the cursor to the beginning of the line.
- Add a new src/em_runtime.h,c to store Emscripten-related variables and
function declarations and implementations for better management.

This implementation has been tested on the latest versions of Chrome,
Firefox, and Safari.

Updates the CI to deploy both user and system WebAssembly
demos to the rv32emu-demo repository, resulting in the following file
structure:
```
.
|-- coi-serviceworker.min.js
|-- elf_list.js
|-- index.html
|-- rv32emu.js
|-- rv32emu.wasm
|-- rv32emu.worker.js
|-- system
    |-- coi-serviceworker.min.js
    |-- index.html
    |-- rv32emu.js
    |-- rv32emu.wasm
    |-- rv32emu.worker.js
```

The top-level files serve the user-space demo, while the system/
subdirectory hosts the system emulation demo. This structure allows both
pages to coexist and be navigated independently in the same deployment.

Improvements:
- The release artifacts in the rv32emu-prebuilt repository include
  either a user-space executable (e.g., RISC-V ELF binaries) or a Linux
  image for system emulation. To distinguish between these two types of
  releases and trigger only the necessary deployment workflow, two
  separate dispatch event types are introduced (see [here](https://github.com/sysprog21/rv32emu-prebuilt/blob/master/.github/workflows/main.yml)):
  - deploy_user_wasm for user-space emulation WebAssembly deployment.
  - deploy_system_wasm for system emulation WebAssembly deployment.
- Add needs and always() to ensure proper sequencing and execution of
  dependent jobs when both targets are deployed.
-  Add src/em_runtime.c as a newly monitored file.
- Change the source of the shareware Doom artifact:
  Downloading directly from the original site often results in 403
  Forbidden errors on GitHub runners recently. The artifact is now
  hosted in our own repository (rv32emu-prebuilt) for more reliable
  access. Error:
  Resolving [www.doomworld.com](http://www.doomworld.com/) ([www.doomworld.com](http://www.doomworld.com/))... 172.67.171.63, 104.21.29.17, 2606:4700:3037::ac43:ab3f, ...
  Connecting to [www.doomworld.com](http://www.doomworld.com/) ([www.doomworld.com)|172.67.171.63|:443](http://www.doomworld.com)%7C172.67.171.63%7C:443/)... connected.
  HTTP request sent, awaiting response... 403 Forbidden

# Reproduce
1. Clone:
```
$ git clone https://github.com/ChinYikMing/rv32emu.git -b sys-emu-wasm --depth 1 && cd rv32emu
```

2. Source your emsdk's environment:
```
$ source ~/emsdk/emsdk_env.sh
```

3.1. To serve user space emulation index page:
```
$ make start-web CC=emcc ENABLE_SDL=1 -j8
```

3.2. To serve system emulation index page:
```
$ make start-web CC=emcc ENABLE_SYSTEM=1 ENABLE_SDL=1 INITRD_SIZE=32 -j8
```
Once the server is running, open your browser (Chrome, Firefox, or Safari) and navigate to `http://127.0.0.1:8000/` to test it. The expected result should match to [link](https://chinyikming.github.io/rv32emu-demo). 
 <div id='description'>
    <a href="https://bito.ai#summarystart"></a>
<h3>Summary by Bito</h3>
This pull request significantly enhances system emulation in web browsers using WebAssembly, enabling users to run Linux and graphical applications like doom-riscv and quake through a new HTML interface. Key improvements include optimized memory management, a modular build structure, and enhanced input handling, along with updated documentation for user guidance.

</div>

### DIFF
--- a/.github/workflows/deploy-wasm.yml
+++ b/.github/workflows/deploy-wasm.yml
@@ -10,13 +10,108 @@ on:
     branches:
       - master
   repository_dispatch:   # listening to rv32emu-prebuilt events
-    types: [deploy_wasm]
+    types: [deploy_user_wasm, deploy_system_wasm]
 
 jobs:
-  wasm-deploy:
+  wasm-system-deploy:
     if: github.event.pull_request.merged == true ||
         github.event_name == 'workflow_dispatch' ||
-        github.event_name == 'repository_dispatch'
+        github.event_name == 'repository_dispatch' && github.event.action == 'deploy_system_wasm'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v4
+      - name: install-dependencies
+        run: |
+              sudo apt-get update -q=2
+              sudo apt-get install -q=2 device-tree-compiler
+      - name: Verify if the JS or HTML files has been modified
+        id: changed-files
+        uses: tj-actions/changed-files@v46
+        with:
+          files: |
+              assets/wasm/html/system.html
+              assets/wasm/js/system-pre.js
+              src/em_runtime.c
+              # Files below may have a potential performance impact (reference from benchmark.yml)
+              src/devices/*.c
+              src/system.c
+              src/riscv.c
+              src/decode.c
+              src/emulate.c
+              src/rv32_template.c
+              src/rv32_constopt.c
+      - name: install emcc
+        if: ${{ steps.changed-files.outputs.any_modified == 'true' ||
+            github.event_name == 'workflow_dispatch' ||
+            (github.event_name == 'repository_dispatch' && github.event.action == 'deploy_system_wasm') }}
+        run: |
+            git clone https://github.com/emscripten-core/emsdk -b 3.1.51
+            cd emsdk
+            ./emsdk install latest
+            ./emsdk activate latest
+            source ./emsdk_env.sh
+            echo "$PATH" >> $GITHUB_PATH
+        shell: bash
+      - name: fetch artifact
+        run: |
+              make artifact
+              # get from rv32emu-prebuilt
+              wget -O build/shareware_doom_iwad.zip "https://raw.githubusercontent.com/sysprog21/rv32emu-prebuilt/doom-artifact/shareware_doom_iwad.zip"
+              unzip -d build/ build/shareware_doom_iwad.zip
+      - name: build with emcc and move application files to /tmp
+        if: ${{ steps.changed-files.outputs.any_modified == 'true' ||
+            github.event_name == 'workflow_dispatch' ||
+            (github.event_name == 'repository_dispatch' && github.event.action == 'deploy_system_wasm') }}
+        run: |
+            make CC=emcc ENABLE_SYSTEM=1 ENABLE_SDL=1 INITRD_SIZE=32 -j
+            mkdir /tmp/rv32emu-system-demo
+            mv assets/wasm/html/system.html /tmp/rv32emu-system-demo/index.html
+            mv assets/wasm/js/coi-serviceworker.min.js /tmp/rv32emu-system-demo
+            mv build/rv32emu.js /tmp/rv32emu-system-demo
+            mv build/rv32emu.wasm /tmp/rv32emu-system-demo
+            mv build/rv32emu.worker.js /tmp/rv32emu-system-demo
+            ls -al /tmp/rv32emu-system-demo
+      - name: Check out the rv32emu-system-demo repo
+        if: ${{ steps.changed-files.outputs.any_modified == 'true' ||
+            github.event_name == 'workflow_dispatch' ||
+            (github.event_name == 'repository_dispatch' && github.event.action == 'deploy_system_wasm') }}
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false # otherwise, the token used is the GITHUB_TOKEN, instead of your personal access token.
+          repository: sysprog21/rv32emu-demo
+      - name: Create local changes
+        if: ${{ steps.changed-files.outputs.any_modified == 'true' ||
+            github.event_name == 'workflow_dispatch' ||
+            (github.event_name == 'repository_dispatch' && github.event.action == 'deploy_system_wasm') }}
+        run: |
+            mkdir -p system
+            mv /tmp/rv32emu-system-demo/index.html ./system
+            mv /tmp/rv32emu-system-demo/coi-serviceworker.min.js ./system
+            mv /tmp/rv32emu-system-demo/rv32emu.js ./system
+            mv /tmp/rv32emu-system-demo/rv32emu.wasm ./system
+            mv /tmp/rv32emu-system-demo/rv32emu.worker.js ./system
+      - name: Commit files
+        if: ${{ steps.changed-files.outputs.any_modified == 'true' ||
+            github.event_name == 'workflow_dispatch' ||
+            (github.event_name == 'repository_dispatch' && github.event.action == 'deploy_system_wasm') }}
+        run: |
+            git config --local user.email "github-actions[bot]@users.noreply.github.com"
+            git config --local user.name "github-actions[bot]"
+            git add system/
+            git commit -m "Add changes to system emulation"
+      - name: Push changes
+        if: ${{ steps.changed-files.outputs.any_modified == 'true' ||
+            github.event_name == 'workflow_dispatch' ||
+            (github.event_name == 'repository_dispatch' && github.event.action == 'deploy_system_wasm') }}
+        uses: ad-m/github-push-action@master
+        with:
+          repository: sysprog21/rv32emu-demo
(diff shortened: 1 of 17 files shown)


ChinYikMing:
Marked as draft since separating the deployment of user and system emulation for WebAssembly may require further discussion.

If we all agree to proceed with the separation, a similar repo like [rv32emu-demo](https://github.com/sysprog21/rv32emu-demo) (might be called `rv32emu-demo-system`) should be created.

ChinYikMing:
The visualization of the running architecture:
image

jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: 39ff299b34daee1339aee693cb4bc9b3626ead5e | Previous: a5afe4dae2caad120cf732f1d046d59ebbfc3d69 | Ratio |
|-|-|-|-|
| `Dhrystone` | `1294` Average DMIPS over 10 runs | `1308` Average DMIPS over 10 runs | `1.01` |
| `Coremark` | `925.371` Average iterations/sec over 10 runs | `913.009` Average iterations/sec over 10 runs | `0.99` |

</details>

jserv:
> If we all agree to proceed with the separation,, a similar repo like [rv32emu-demo](https://github.com/sysprog21/rv32emu-demo) (might be called `rv32emu-demo-system`) should be created.

Can the system emulation be part of `rv32emu-demo`?

ChinYikMing:
> > If we all agree to proceed with the separation,, a similar repo like [rv32emu-demo](https://github.com/sysprog21/rv32emu-demo) (might be called `rv32emu-demo-system`) should be created.
> 
> Can the system emulation be part of `rv32emu-demo`?

Yes, we can. Take this as an example: [https://chinyikming.github.io/rv32emu-demo/](https://chinyikming.github.io/rv32emu-demo/) for user space emulation, and [https://chinyikming.github.io/rv32emu-demo/system](https://chinyikming.github.io/rv32emu-demo/system) for system emulation.

src/riscv.c
```diff
@@ -234,7 +234,9 @@ static void map_file(char **ram_loc, const char *name)
     struct stat st;
     fstat(fd, &st);
 
-#if HAVE_MMAP
+// EMSCRIPTEN: We don't currently support location hints for the address of the mapping
+// https://github.com/emscripten-core/emscripten/blob/52bc455316b2f44d3a94104776a335a5861ad73b/system/lib/libc/emscripten_mmap.c#L105
```
visitorckw: In this project, we use `/* ... */` style comments instead of `// ...`.

src/emulate.c
```diff
@@ -1003,12 +1003,19 @@ static bool rv_has_plic_trap(riscv_t *rv)
             (rv->csr_sip & rv->csr_sie));
 }
 
+#if defined(__EMSCRIPTEN__)
+extern bool is_input_buf_avail;
```
visitorckw: Is there a more appropriate header where this `extern` declaration could be placed?

ChinYikMing:
> > > If we all agree to proceed with the separation,, a similar repo like [rv32emu-demo](https://github.com/sysprog21/rv32emu-demo) (might be called `rv32emu-demo-system`) should be created.
> > 
> > 
> > Can the system emulation be part of `rv32emu-demo`?
> 
> Yes, we can. Take this as an example: https://chinyikming.github.io/rv32emu-demo/ for user space emulation, and https://chinyikming.github.io/rv32emu-demo/system for system emulation.

The index page of the demo links have been updated. Both user-space and system emulation can now be navigated back and forth using the newly added navigation buttons.

src/emulate.c
ChinYikMing: > Is there a more appropriate header where this `extern` declaration could be placed?

Introduce `src/em_runtime.h,c` for the concern.

.github/workflows/deploy-wasm.yml
```diff
@@ -10,13 +10,109 @@ on:
...
+              src/decode.c
+              src/emulate.c
+              src/rv32_template.c
+              src/rv32_constopt.c
+      - name: install emcc
+        if: ${{ steps.changed-files.outputs.any_modified == 'true' ||
+            github.event_name == 'workflow_dispatch' ||
+            (github.event_name == 'repository_dispatch' && github.event.action == 'deploy_system_wasm') }}
+        run: |
+            git clone https://github.com/emscripten-core/emsdk.git
```
jserv: Use `https://github.com/emscripten-core/emsdk` instead.

.github/workflows/deploy-wasm.yml
```diff
@@ -10,13 +10,109 @@ on:
...
+            cd emsdk
+            git pull
+            git checkout 3.1.51
```
jserv: Why version 3.1.51 ?

.github/workflows/deploy-wasm.yml
jserv: You can specify revision/branch during git cloning via `-b` option.

.github/workflows/deploy-wasm.yml
ChinYikMing: > Why version 3.1.51 ?

Quote from previous discussion at [here](https://github.com/sysprog21/rv32emu/issues/75#issuecomment-2001347020) :
> I tried to use emcc version 3.1.55 for porting and the video game sound effects are working but background musics do not. After some debugging, I found that the latest emcc will fetch the latest SDL2_mixer library for porting. After switching version back to 3.1.51 (the previous demonstration using this emcc version), both are working. As a result, I would like to lock emcc version to 3.1.51 until we figure out what the root cause is.

Even after updating to version 4.0.x, the same issue persists when running system emulation. Therefore, the version is locked to 3.1.51. In [here](https://github.com/sysprog21/rv32emu/blob/56d20a26b27368cb324b29fac9a979388ed05184/mk/toolchain.mk#L13), if the version is not 3.1.51, the compilation using emcc will fail.

src/main.c
```diff
@@ -11,6 +11,19 @@
...
+#if RV32_HAS(SYSTEM)
+EM_JS(void, disable_run_button, (), {
+    document.getElementById('runSysButton').disabled = true;
+});
+#else
+EM_JS(void, disable_run_button, (), {
+    document.getElementById('runButton').disabled = true;
+});
+#endif
+#endif
```
jserv: It is better to wrap Emscripten specific function in a dedicated header.

jserv:
The name `emsc` is quite confusing. How about `em-runtime`?

jserv:
By the way, would you use [Binaryen](https://github.com/WebAssembly/binaryen) for more optimizations?

jserv:

PR #604, opened 2025-07-18
Update README

ChinYikMing:
- the customizable bootargs feature was introduced in #534, but the README was not updated at that time.

- build-linux-img should be build-linux-image. See mk/tools.mk.

- highlight WebAssembly support in feature list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ Features:
 * Newlib-compatible system-call layer for standalone programs
 * Minimal system emulation capable of booting an RV32 Linux kernel and running user-space binaries
 * Experimental SDL-based display/event/audio system calls for running video games
+* WebAssembly build for user-mode and system emulation with SDL graphics and audio in modern browsers
 * Remote debugging through the GDB Remote Serial Protocol
 * Tiered JIT compilation for performance boost while maintaining a small footprint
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,10 +26,11 @@ Features:
 * Fast interpreter for executing the RV32 ISA
 * Fast interpreter that faithfully executes the complete RV32 instruction set
 * Full coverage of RV32I / RV32E plus the M (integer multiply–divide), A (atomics), F (single-precision floating-point), C (compressed), and Zba/Zbb/Zbc/Zbs bit-manipulation extensions
-* Built-in ELF loader for user-mode emulation
+* Built-in ELF loader for user-space emulation
```
jserv: “Built‑in ELF loader for user‑mode emulation” is the clearer, more conventional wording.

QEMU and most academic papers distinguish “user‑mode emulation” (running a single user process on the host kernel) from “system‑mode emulation” (full guest OS). Using the same term makes your simulator’s capabilities immediately recognizable to anyone familiar with QEMU.

“Mode” pairs naturally with other RISC‑V privilege levels (M‑mode, S‑mode, U‑mode). “User‑space” is correct but can blur the distinction between CPU privilege levels and the broader concept of an OS’s address‑space separation.

README.md
```diff
@@ -26,10 +26,11 @@ Features:
...
 * Newlib-compatible system-call layer for standalone programs
 * Minimal system emulation capable of booting an RV32 Linux kernel and running user-space binaries
 * Experimental SDL-based display/event/audio system calls for running video games
+* WebAssembly build for user-space and system emulation with SDL graphics and audio in modern browsers
```
ChinYikMing: @jserv Shall we use `user-mode` here as well for consistency? For the [WebAssembly](https://github.com/sysprog21/rv32emu?tab=readme-ov-file#webassembly-translation) section, should we also align with the `user-mode` terminology?

jserv: > Shall we use `user-mode` here as well for consistency? 

Yes, please do.

jserv:

PR #606, opened 2025-08-31
Enhance docker image

### LINKED ISSUE #581: Enhancing the Docker Image
The current Docker image includes only the ELF files stored in the build/ directory. However, the [rv32emu-prebuilt](https://github.com/sysprog21/rv32emu-prebuilt) repository contains a broader set of ELF files that could also be included. Alternatively, to minimize the Docker image size, the documentation could be updated with instructions on how to download the additional prebuilt ELF files separately.

Additionally, since rv32emu supports system emulation, it would be beneficial to build a dedicated executable for this purpose. This executable could be named `rv32emu-system`, while the existing ELF target one executable could be renamed `rv32emu-user`, similar with QEMU’s naming convention. The prebuilt Linux image is also available in the [rv32emu-prebuilt](https://github.com/sysprog21/rv32emu-prebuilt) repository. Similarly, to keep the Docker image lightweight, the documentation could be updated with guidance on downloading the prebuilt Linux image separately.

kaiii708:
This commit updates the Docker image to include the following executables:
- rv32emu-user: for user-mode emulation
- rv32emu-system: for system-mode emulation
- rv_histogram: for statistical analysis

To keep the image compact, build artifacts are excluded from the final image. Instructions for retrieving prebuilt binaries are added to the README.

Close #581

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,18 @@
 FROM alpine:3.21 AS base_gcc
 
-RUN apk add --update alpine-sdk git wget sdl2-dev sdl2_mixer-dev
+RUN apk add --update alpine-sdk git wget sdl2-dev sdl2_mixer-dev dtc
 
 # copy in the source code
 WORKDIR /home/root/rv32emu
 COPY . .
 
-# generate execution file for rv32emu and rv_histogram
-RUN make
-RUN make tool
+# build and rename executables: rv32emu-user, rv32emu-system, and rv32emu-histogram
+RUN make -j"$(nproc)"
+RUN make tool -j"$(nproc)"
+RUN mv ./build/rv32emu ./build/rv32emu-user
+RUN mv ./build/rv_histogram ./build/rv32emu-histogram
+RUN make distclean
+RUN make ENABLE_SYSTEM=1 INITRD_SIZE=32 -j"$(nproc)"
 
 FROM alpine:3.21 AS final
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -9,17 +9,22 @@ COPY . .
 # generate execution file for rv32emu and rv_histogram
 RUN make
 RUN make tool
+RUN mv ./build/rv32emu ./build/rv32emu-user
+RUN mv ./build/rv_histogram ./build/tmp_rv_histogram
+RUN make distclean
+RUN make ENABLE_SYSTEM=1 INITRD_SIZE=32
 
-FROM alpine:3.19 AS final
+FROM alpine:3.21 AS final
```
visitorckw, line 17: Which parts of this change rely on Alpine 3.21 and wouldn't work on 3.19?

visitorckw:
This patch updates the Dockerfile but also makes large README changes.
I’d prefer splitting them into separate commits so each does one thing and review is easier.

visitorckw:
Also, I'd prefer using the imperative mood when describing changes [1].
E.g. This commit updates ... -> Update ...

Link: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/submitting-patches.rst?h=v6.17-rc3#n94 [1]

Dockerfile
kaiii708: Commit 37c9838 mentioned that the end of support for Alpine Linux v3.19.x is on 1 Nov 2025, so I upgraded it to 3.21.

kaiii708:
Split into 3 commits. Please take another look. Thanks!

kaiii708:
The CI build has succeeded, and the Docker image is available here:  
https://github.com/kaiii708/rv32emu/actions/runs/17394344573  
Test it with:  
```bash
docker run --rm -it kaiii708/rv32emu
```

Dockerfile
```diff
@@ -9,17 +9,22 @@ COPY . .
 # generate execution file for rv32emu and rv_histogram
 RUN make
 RUN make tool
+RUN mv ./build/rv32emu ./build/rv32emu-user
+RUN mv ./build/rv_histogram ./build/tmp_rv_histogram
```
visitorckw: Renaming rv_histogram to tmp_rv_histogram and then back looks a bit hacky and may confuse readers.
Since we already renamed rv32emu to rv32emu-user/rv32emu-system, how about also giving rv_histogram an rv32emu- prefix?
That way all three binaries can be renamed to their final names right after build, and the later COPY steps become cleaner and easier to follow.

README.md
```diff
@@ -143,6 +143,40 @@ The usage and limitations of Doom and Quake demo are listed in [docs/demo.md](do
...
+
+$ tar -xzvf rv32emu-linux-image-prebuilt.tar.gz
+```
+To run `rv32emu-system`, use the following:
+```shell
+$ build/rv32emu-system \
+  -k rv32emu-linux-image-prebuilt/linux-image/Image \
+  -i rv32emu-linux-image-prebuilt/linux-image/rootfs.cpio
+```
+
```
visitorckw, line 179: Since we renamed rv_histogram, could we also update the README's Docker section to document how to run the histogram tool in addition to the user/system mode emulators?

kaiii708: Done — thanks for the suggestion!

CI for the parallel build version passed: https://github.com/kaiii708/rv32emu/actions/runs/17405012825

jserv:
Thank @kaiii708 for contributing!

PR #614, opened 2025-09-21
Attempt to fix JIT instruction cache coherency on Arm64

jserv:
The JIT compiler was experiencing intermittent failures on Arm64 / Apple Silicon due to missing instruction cache invalidation after patching branch instructions. When update_branch_imm() modified branch targets in JIT-compiled code, the CPU's instruction cache was not being invalidated, causing it to execute stale cached instructions instead of the newly patched ones.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -177,20 +177,33 @@ static bool parse_args(int argc, char **args)
         assert(getcwd(cwd_path, PATH_MAX));
 
         char rel_path[PATH_MAX] = {0};
-        memcpy(rel_path, args[0], strlen(args[0]) - 7 /* strlen("rv32emu") */);
+        size_t args0_len = strlen(args[0]);
+        /* Ensure args[0] is long enough before subtracting */
+        if (args0_len > 7) { /* strlen("rv32emu") */
+            size_t copy_len = args0_len - 7;
+            if (copy_len >= PATH_MAX)
+                copy_len = PATH_MAX - 1;
+            memcpy(rel_path, args[0], copy_len);
+            rel_path[copy_len] = '\0';
+        }
 
         char *prog_basename = basename(opt_prog_name);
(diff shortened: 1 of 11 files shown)


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: ef57d84484b240cd0c0b04f7ad7c8f6d433ffc50 | Previous: 9ba555370ad56ec73f725afedf863ff9454f6c05 | Ratio |
|-|-|-|-|
| `Dhrystone` | `1333` Average DMIPS over 10 runs | `1323` Average DMIPS over 10 runs | `0.99` |
| `Coremark` | `957.852` Average iterations/sec over 10 runs | `906.292` Average iterations/sec over 10 runs | `0.95` |

</details>

src/main.c
```diff
@@ -184,8 +184,10 @@ static bool parse_args(int argc, char **args)
                                strlen(prog_basename) + 5 + 1);
         assert(prof_out_file);
 
-        sprintf(prof_out_file, "%s/%s%s.prof", cwd_path, rel_path,
-                prog_basename);
+        snprintf(prof_out_file,
+                 strlen(cwd_path) + 1 + strlen(rel_path) +
+                     strlen(prog_basename) + 5 + 1,
+                 "%s/%s%s.prof", cwd_path, rel_path, prog_basename);
```
visitorckw: Since we already allocate exactly the required memory for prof_out_file, I don't think there's an actual security issue here, and using snprintf might even be slower. If the goal is just to resolve the macOS build warning, that could be a reasonable justification. However, this approach causes strlen() to be called twice on cwd_path, rel_path, and prog_basename. Maybe we could store the buffer length in an extra variable to avoid relying on the compiler to optimize away the duplicate strlen() calls?

PR #617, opened 2025-10-01
Fix critical memory safety issues

jserv:
This adds explicit NULL checks for all malloc/calloc/mpool_alloc calls that previously relied solely on assert(), which compiles out with `-DNDEBUG`. It prevents NULL pointer dereferences in production builds.
    

---

## Summary by cubic
Adds NULL checks and graceful fallbacks for critical allocations across the emulator, JIT, cache, IO, and syscalls. Prevents crashes when assert() is disabled and degrades behavior safely on OOM.

- **Bug Fixes**
  - Block/IR/branch-table allocations now checked; translation can fail cleanly.
  - Fused operation generation falls back to non-fused ops if malloc fails.
  - JIT init and queue entry allocations validated; cleanup on failure; skip tier-2 enqueue on OOM.
  - Cache put restores the replaced entry if new allocation fails.
  - IO, syscall, and riscv init guard malloc/calloc; return NULL or -1 instead of crashing.

### DIFF
--- a/src/cache.c
+++ b/src/cache.c
@@ -147,6 +147,10 @@ static inline void hlist_del_init(struct hlist_node *n)
 
 cache_t *cache_create(uint32_t size_bits)
 {
+    /* Prevent integer overflow in 1 << size_bits */
+    if (size_bits >= 32)
+        return NULL;
+
     cache_t *cache = malloc(sizeof(cache_t));
     if (!cache)
         return NULL;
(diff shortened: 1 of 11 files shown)


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: adab7d7fc616d69e09049d78e2c1501c35cce163 | Previous: 9220ffb41505adcf874741bd55fd4d2a2bb2f9e0 | Ratio |
|-|-|-|-|
| `Dhrystone` | `1302` Average DMIPS over 10 runs | `1267` Average DMIPS over 10 runs | `0.97` |
| `Coremark` | `952.843` Average iterations/sec over 10 runs | `958.558` Average iterations/sec over 10 runs | `1.01` |

</details>

jserv:
@cubic-dev-ai Review

visitorckw:
Perhaps we can consider introducing a fault injection framework down the road to improve test coverage for error handling cases such as this one.

PR #636, opened 2025-12-11
Adopt preserve_none attribute for MUST_TAIL optimization

### LINKED ISSUE #601: Suggest using preserve_none to improve tail call optimization in interpreter
I’ve been exploring rv32emu's interpreter mode and found that while the RVOP macro is designed to support MUST_TAIL tail calls via:

```c
MUST_TAIL return next->impl(rv, next, cycle, PC);
```

In practice, the compiler often emits full stack frames and callee-saved register spills, which breaks the intended tail call optimization (TCO). This happens even though the control flow would otherwise be eligible for a proper tail jump.

For example, in a simple handler like do_addi, the compiler emits a proper tail call via jmp *%rax, and avoids saving any callee-saved registers:

```
af38:   ff e0    jmp *%rax   # tail call to next->impl
```

However, in do_lw, due to a non-tail call (e.g., memory access through a function pointer), the compiler emits callee-saved register preservation and a full frame:

```
aaa0:   55        push   %rbp
aaa1:   41 57     push   %r15
...
ab75:   c3        ret

```

This overhead defeats the benefits of using MUST_TAIL, especially in a hot loop of instruction dispatch.

##Suggestion

Clang provides the __attribute__((preserve_none)) attribute (available since 19.1.0) which modifies the calling convention to avoid saving any registers. It’s suitable when the function is only ever tail-called and doesn’t return via ret.

This attribute allows better register usage and completely avoids unnecessary stack setup for tail calls — perfect for interpreter dispatch functions.

References

[Clang docs: preserve_none](https://clang.llvm.org/docs/AttributeReference.html#preserve-none)

[A Tail Calling Interpreter For Python (And Other Updates)](https://blog.reverberate.org/2025/02/10/tail-call-updates.html)

visitorckw:
Despite using __attribute__((musttail)), the compiler may still emit full stack frames and spill callee-saved registers in complex handlers (e.g., do_lw), undermining the benefits of the threaded interpreter.

Integrate Clang's __attribute__((preserve_none)) (available since Clang 19.1.0) into the MUST_TAIL macro. This attribute modifies the calling convention to avoid preserving registers, thereby minimizing stack setup and register pressure during instruction dispatch.

This ensures more consistent tail call optimization and reduces overhead in the interpreter loop.

Closes: https://github.com/sysprog21/rv32emu/issues/601

---
## Summary by cubic
Extend MUST_TAIL to include preserve_none when available, so tail-call dispatch avoids stack frames and callee-saved spills. This makes tail calls more reliable and reduces interpreter loop overhead, especially in complex handlers.

<sup>Written for commit d8382ab75e1873d939e418af346af0fcd59562a7. Summary will update automatically on new commits.</sup>

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -191,7 +191,11 @@ static inline uint8_t ilog2(uint32_t x)
  * even without optimizations enabled.
  */
 #if defined(__has_attribute) && __has_attribute(musttail)
+#if __has_attribute(preserve_none)
+#define MUST_TAIL __attribute__((musttail)) __attribute__((preserve_none))
+#else
 #define MUST_TAIL __attribute__((musttail))
+#endif
 #else
 #define MUST_TAIL
 #endif


jserv:

PR #639, opened 2025-12-25
Fix preserve_none attribute usage for macOS

jserv:
The preserve_none calling convention attribute must be applied to function declarations, not return statements. The previous commit (d8382ab) incorrectly added it to the MUST_TAIL macro which is used on return statements, causing compilation errors.

Additionally, macOS clang falsely reports __has_attribute(preserve_none) as true but doesn't actually support the attribute, causing 20+ errors.

The optimization now works on Linux (Clang 19.1.0+) while building cleanly on macOS.

---
## Summary by cubic
Fixes misuse of the preserve_none attribute to restore builds on macOS while keeping the calling-convention optimization on Linux (Clang 19.1+). The attribute is now applied to function declarations via a new PRESERVE_NONE macro and disabled on Apple toolchains.

- **Bug Fixes**
  - Removed preserve_none from MUST_TAIL and introduced PRESERVE_NONE for function declarations.
  - Disabled PRESERVE_NONE on macOS due to clang falsely advertising support.
  - Annotated the interpreter op function pointer and do_* handlers with PRESERVE_NONE to retain the optimization on supported compilers.

<sup>Written for commit aa023f04706fa4595dc631768db3a3c57a39819f. Summary will update automatically on new commits.</sup>

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -191,13 +191,20 @@ static inline uint8_t ilog2(uint32_t x)
  * even without optimizations enabled.
  */
 #if defined(__has_attribute) && __has_attribute(musttail)
-#if __has_attribute(preserve_none)
-#define MUST_TAIL __attribute__((musttail)) __attribute__((preserve_none))
-#else
 #define MUST_TAIL __attribute__((musttail))
-#endif
 #else
 #define MUST_TAIL
+#endif
+
+/* The preserve_none calling convention minimizes register preservation overhead
+ * in the interpreter's threaded dispatch. It must be applied to function
+ * declarations, not return statements. macOS clang falsely reports support.
+ */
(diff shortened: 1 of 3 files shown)


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: aa023f04706fa4595dc631768db3a3c57a39819f | Previous: ac608b2ec48578e0f29f5dbed9415218bcb1f0c7 | Ratio |
|-|-|-|-|
| `Dhrystone` | `1337` Average DMIPS over 10 runs | `1274` Average DMIPS over 10 runs | `0.95` |
| `Coremark` | `967.957` Average iterations/sec over 10 runs | `959.277` Average iterations/sec over 10 runs | `0.99` |

</details>

PR #642, opened 2025-12-25
Fix broken Timidity download URL

jserv:
The [original URL](http://www.libsdl.org/projects/mixer/timidity/timidity.tar.gz) now returns 404. Migrate to self-hosted asset in `rv32emu-prebuilt` repository and correct the SHA1 checksum for directory verification.

---
## Summary by cubic
Updated the Timidity data download URL to https://www.libsdl.org/projects/old/SDL_mixer/timidity/timidity.tar.gz to replace the 404 link. This restores successful download of the SDL2_mixer Timidity configuration during builds.

<sup>Written for commit 4e06513453ee7531761cba58ab4a62d009d825ac. Summary will update automatically on new commits.</sup>

### DIFF
--- a/mk/external.mk
+++ b/mk/external.mk
@@ -107,7 +107,7 @@ QUAKE_DATA_SHA = 36b42dc7b6313fd9cabc0be8b9e9864840929735
 QUAKE_DATA_SHA_CMD = $(SHA1SUM)
 
 # Timidity software synthesizer configuration for SDL2_mixer
-TIMIDITY_DATA_URL = http://www.libsdl.org/projects/mixer/timidity/timidity.tar.gz
+TIMIDITY_DATA_URL = https://www.libsdl.org/projects/old/SDL_mixer/timidity/timidity.tar.gz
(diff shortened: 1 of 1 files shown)


ChinYikMing:
The link is changed to this: `https://www.libsdl.org/projects/old/SDL_mixer/timidity/timidity.tar.gz`. I prefer to use this link since the self-hosted link contains `git` keyword, which violates the download mechanism: https://github.com/sysprog21/rv32emu/blob/c769053d67cc65017d25e91f1d34db4eda4734dd/mk/external.mk#L27

Also, no need to modify the SHA1 value (calculated via `find build/timidity/ -type f -not -path '*/.git/*' -print0 | sort -z | xargs -0 sha1sum | sort | sha1sum`).

ChinYikMing:
Refine the commit message since the self-hosted link is not used anymore.

ChinYikMing (approved):
LGTM.

jserv:
Thank @ChinYikMing for review!

PR #646, opened 2025-12-25
Restore wall-clock time source for user mode

### LINKED ISSUE #640: Test case `nyancat` doesn't work well on the latest rv32emu.
Recently, I checked out the latest commit (40fcbfef), and I found that the `nyancat` test case doesn't work as well as it used to, both in JIT and non-JIT mode. It’s possible that some instructions implementation or refinement is impacting this test case.

Reproducer
```
make ENABLE_JIT=0 clean all
./build/rv32emu ./build/riscv32/nyancat
```
The test case displays a static image without any animation.

visitorckw:
Commit a6ecb0c ("Bring up Linux kernel") refactored the time update mechanism to use a static cycle counter. However, that counter only increments when the SYSTEM feature is enabled.

In user mode, the counter remains at zero, causing the timer CSRs to freeze. This regression breaks time-dependent applications, such as nyancat, resulting in a static image.

Restore the use of gettimeofday for non-SYSTEM builds to ensure the timer reflects actual time passage.

Fixes: a6ecb0c ("Bring up Linux kernel")
Closes: https://github.com/sysprog21/rv32emu/issues/640

---
## Summary by cubic
Restore wall-clock time in user mode by using gettimeofday when SYSTEM is disabled. Fixes frozen timer CSRs so time-dependent apps (like nyancat) run correctly.

<sup>Written for commit f04d93a53e8abf0ef23f4a00c5cf5cff6efbfc75. Summary will update automatically on new commits.</sup>

### DIFF
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -102,6 +102,12 @@ static inline void check_tohost_write(riscv_t *rv,
 #if RV32_HAS(Zicsr)
 static inline void update_time(riscv_t *rv)
 {
+#if !RV32_HAS(SYSTEM)
+    struct timeval tv;
+
+    rv_gettimeofday(&tv);
+    rv->timer = (uint64_t) tv.tv_sec * 1e6 + (uint32_t) tv.tv_usec;
+#endif
     rv->csr_time[0] = rv->timer & 0xFFFFFFFF;
     rv->csr_time[1] = rv->timer >> 32;
 }


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: f04d93a53e8abf0ef23f4a00c5cf5cff6efbfc75 | Previous: 3623c9d27ad5c02d6b8dea2e92c5de71557c719f | Ratio |
|-|-|-|-|
| `Dhrystone` | `1306` Average DMIPS over 10 runs | `1065` Average DMIPS over 10 runs | `0.82` |
| `Coremark` | `942.154` Average iterations/sec over 10 runs | `955.705` Average iterations/sec over 10 runs | `1.01` |

</details>

ChinYikMing:
Can we add CI to guard these time-dependent applications?

visitorckw:
> Can we add CI to guard these time-dependent applications?

Directly verifying the visual output of applications like nyancat in CI is challenging. While a frozen animation is obvious to a human observer, detecting this state programmatically in a headless environment is not straightforward.

Do you have any suggestions for guarding this type of time-dependent application?

ChinYikMing:
Agree that GUI might be not intuitive (like doom, quake, smolnes) to be tested in headless env (maybe can file an issue to track it in the future). 

BTW, this is source code of nyancat: [link](https://github.com/sysprog21/rv32emu/blob/master/tests/nyancat.c). Actually, it is a tui (terminal user interface) application, we shall see the animation in headless github runner terminal when running it. Thus, we can leverage existing `expect` CI mechanism to match certain pattern. cc @visitorckw

visitorckw:
IIUC, we could capture the terminal output and match frames 0-11 with a timeout. However, explicitly matching specific frames feels a bit hardcoded. Is this considered an ideal approach?

ChinYikMing:
> IIUC, we could capture the terminal output and match frames 0-11 with a timeout. However, explicitly matching specific frames feels a bit hardcoded. Is this considered an ideal approach?

Maybe comparing few frames and using hash value would be better and cleaner. BTW, it should be in other PR if need to embrace such TUI application in CI.

ChinYikMing (approved):
LGTM.

jserv:

PR #660, opened 2025-12-31
Add TLB for MMU and fix trap handling

### LINKED ISSUE #500: jit: Improve MMU efficiency in system simulation
In the current flow of the JIT compilation, the simulation will be reverted back to the interpreter mode to deal with the translation of the virtual address to the physical address when reaching the memory operation (`lw`/`sw`) in the system simulation. This is going to have a significant impact to the performance due to the mode switching.

The one of the possible solutions is to implement the translation lookaside buffer (TLB). However, the situation mentioned above will still happen if the cache misses.

jserv:
This implements TLB for faster address translation:
- 64-entry direct-mapped TLB indexed by VPN for O(1) lookup
- Caches VPN, PPN, permissions (R/W/X/U), and validity
- Superpages handled by caching each 4KB slice with computed PPN
- Flushed on SATP changes and SFENCE.VMA instruction
- Block cache invalidated on SFENCE.VMA for JIT/PTE consistency

Privilege and permission checking:
- U-mode can only access pages with PTE_U bit set
- S-mode accessing U-page requires SUM=1 in sstatus
- MXR bit handling for executable page reads

ELF_LOADER+SYSTEM mode fixes:
- Fix stack alignment: use `&= -16` (not `&= 16`) for 16-byte boundary
- Add retry-after-trap in mmu_ifetch/mmu_translate: after page fault handler returns, retry page walk to check if page was mapped
- Add page fault trap generation for out-of-bounds memory access
- Remove duplicate DTB dependencies from Makefile

Misaligned instruction fetch trap handler fix:
- Use `sret` instead of `ret` to properly exit trap mode
- Set `sepc` to return address before `sret`
- Matches behavior of misaligned load/store handlers

Close #500

### DIFF
--- a/src/cache.c
+++ b/src/cache.c
@@ -12,6 +12,8 @@
 
 #include "cache.h"
 #include "mpool.h"
+#include "riscv.h"
+#include "riscv_private.h"
 #include "utils.h"
 
 static uint32_t cache_size, cache_size_bits;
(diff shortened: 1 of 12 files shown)


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: 40449dc6d9647319399655dd876979d864a5f8e0 | Previous: 72d712bdd161f14a50259bde8bbc434fea7c29dd | Ratio |
|-|-|-|-|
| `Dhrystone` | `1315` Average DMIPS over 10 runs | `1311` Average DMIPS over 10 runs | `1.00` |
| `Coremark` | `936.776` Average iterations/sec over 10 runs | `939.005` Average iterations/sec over 10 runs | `1.00` |

</details>

PR #668, opened 2026-01-01
Fix gdbstub exit handling for fd close syscalls

jserv:
This adds on_exit detection to `rv_step_debug()` to match the behavior in block-based execution. Without this, `syscall_close()` logs ERROR when the guest program closes fd 0, 1, 2 during exit, causing gdbstub-test to fail on macOS CI.

---
## Summary by cubic
Detect guest exit in rv_step_debug() by setting on_exit when PC == exit_addr to match block-based execution, avoid false ERROR logs when fd 0-2 close on exit, and fix gdbstub-test failures. Update gdbstub-test.sh to handle unset CROSS_COMPILE, skip empty prefixes to avoid host gdb, break on first match, fix quoting in -z checks, and use main() breakpoint addresses that are consistent across platforms.

<sup>Written for commit 0e88a5940692059a3867add50a0e78fcb9b888fe. Summary will update on new commits.</sup>

### DIFF
--- a/.ci/gdbstub-test.sh
+++ b/.ci/gdbstub-test.sh
@@ -5,19 +5,19 @@ set -e -u -o pipefail
 export PATH=$(pwd)/toolchain/bin:$PATH
 
 GDB=
-prefixes=("${CROSS_COMPILE}" "riscv32-unknown-elf-" "riscv-none-elf-")
+prefixes=("${CROSS_COMPILE:-}" "riscv32-unknown-elf-" "riscv-none-elf-")
 for prefix in "${prefixes[@]}"; do
+    # Skip empty prefix to avoid matching host gdb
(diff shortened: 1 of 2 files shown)


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: 0e88a5940692059a3867add50a0e78fcb9b888fe | Previous: 9a877d523f76843ef9dc384d92c2044f16958069 | Ratio |
|-|-|-|-|
| `Dhrystone` | `1295` Average DMIPS over 10 runs | `1318` Average DMIPS over 10 runs | `1.02` |
| `Coremark` | `936.19` Average iterations/sec over 10 runs | `942.76` Average iterations/sec over 10 runs | `1.01` |

</details>

PR #671, opened 2026-01-02
Add instruction fusion patterns and RAM fast-path optimization

### LINKED ISSUE #298: Evaluate a hybrid approach for macro-operation fusion in both the decoding and execution stages
The rv32emu employs block chaining, macro-operation fusion during the decoding stage would be generally more advantageous. This is because the primary benefit of block chaining is to streamline the transition between basic blocks during execution. By fusing operations during decoding, we reduce the overall number of operations and potentially the number of basic blocks, making the block chaining more efficient. Additionally, doing this work upfront during decoding means that the execution phase can remain fast and efficient, which is critical for both interpreters and JIT compilers.

However, the best approach can vary based on specific emulator design goals, the nature of the instruction set being emulated, and the typical workload expected for the emulator. In practice, a hybrid approach might also be feasible, where some fusion decisions are made during decoding and others are deferred until execution, depending on the context and potential optimization gains.

Once macro-operation fusion is implemented through the hybrid instruction decoding and execution stages, more complex scenarios, such as dynamic superinstructions (#282), can be applied in a similar and more concrete manner. Similarly, C routine substitution can be implemented more efficiently in the instruction decoding stage, because optimized routines such as `memcmp` can always be used to recognize the target instruction sequences.

Additional materals:
* [A listing of standard RISC-V pseudoinstructions](https://github.com/riscv-non-isa/riscv-asm-manual/blob/master/riscv-asm.md#pseudoinstructions)
* [ Pseudoinstruction Equivalents](https://www.robertwinkler.com/projects/riscv_book/riscv_book.html#_no_pseudoinstructions_allowed)

jserv:
This implements new macro-operation fusion patterns and memory access optimizations to reduce instruction dispatch overhead.

## Fusion Patterns

**Existing (fuse6-8):**
- **fuse6**: `LI a7, imm + ECALL` → Fused syscall dispatch
- **fuse7**: Multiple consecutive `ADDI` → Batched register updates
- **fuse8**: `LUI + ADDI` → 32-bit constant load (li pseudo-instruction)

**New (fuse9-12):**
- **fuse9**: `LUI + LW` → Absolute/PC-relative word load
  - Fuses upper immediate load with word load for global variable access
- **fuse10**: `LUI + SW` → Absolute/PC-relative word store
  - Fuses upper immediate load with word store for global variable writes
- **fuse11**: `LW + ADDI` (post-increment) → Load with pointer bump
  - Fuses load followed by base register increment, common in memcpy/strlen
  - Constraint: `rd != rs1` prevents base clobbering before address use
- **fuse12**: `ADDI + BNE` → Loop counter decrement-branch
  - Fuses counter decrement with conditional branch for countdown loops
  - Preserves `branch_taken`/`branch_untaken` for block chaining

## RAM Fast-Path Optimization (non-SYSTEM mode)

In userspace emulation, memory accesses always target RAM, making io callback indirection pure overhead. Added inline functions that bypass the function pointer dispatch:
- `ram_read_w/s/b()`: Direct memory read via memcpy (safe for unaligned)
- `ram_write_w/s/b()`: Direct memory write via memcpy
- `MEM_READ/WRITE` macros: Conditional dispatch based on SYSTEM mode

All load/store instructions (lb, lh, lw, lbu, lhu, sb, sh, sw) and their compressed/float variants updated to use `MEM_*` macros.

## Lazy LW/SW Fusion (SYSTEM_MMIO mode)

When running in kernel boot mode, memory ops may target MMIO regions. Hybrid decode/execution fusion verifies addresses are RAM before fusing:
- At decode time: Mark consecutive LW/SW as "fusion candidates"
- At execution time: Verify addresses are RAM via `GUEST_RAM_CONTAINS()`
- If safe: Perform fusion; otherwise mark as failed (never retry)

Safety guards: MMU guard (satp != 0), hotness threshold, base register mutation check, pointer chasing detection.

## Benchmark Results

**Interpreter mode (higher is better):**

| Benchmark | Machine | Arch | Baseline | Improved | Change |
|-----------|---------|------|----------|----------|--------|
| Dhrystone | node11 | x86_64 | 1041 DMIPS | 1398 DMIPS | **+34.3%** |
| Dhrystone | Jetson Orin | aarch64 | 1022 DMIPS | 1208 DMIPS | **+18.2%** |
| CoreMark | node11 | x86_64 | 799.66 iter/s | 843.50 iter/s | **+5.5%** |
| CoreMark | Jetson Orin | aarch64 | 755.11 iter/s | 793.07 iter/s | **+5.0%** |

**JIT mode (Jetson Orin):**

| Benchmark | Baseline | Improved | Change |
|-----------|----------|----------|--------|
| fibonacci | 44.60s | 33.45s | **+25.0%** |
| jit-bf | 0.86s | 0.52s | **+39.5%** |

The RAM fast-path eliminates ~5-10 cycles of indirect call overhead per memory access. Combined with new fusion patterns, this yields significant speedups especially on x86_64 where deeper pipelines benefit more from reduced dispatch overhead.

## Verified Build Configurations

- `make` (default interpreter)
- `make ENABLE_JIT=1`
- `make ENABLE_SYSTEM=1`
- `make ENABLE_SYSTEM=1 ENABLE_JIT=1`

All tests pass in interpreter, JIT, and SYSTEM modes.

Close #298

### DIFF
--- a/mk/artifact.mk
+++ b/mk/artifact.mk
@@ -40,17 +40,16 @@ SHELL_HACK := $(shell mkdir -p $(BIN_DIR)/linux-x86-softfp $(BIN_DIR)/riscv32 $(
 # $(2): name of GitHub releases
 # $(3): name showing in terminal
 define fetch-releases-tag
+    $(eval LATEST_RELEASE := $(shell wget -q https://api.github.com/repos/sysprog21/rv32emu-prebuilt/releases -O- \
+                                     | grep '"tag_name"' \
+                                     | grep "$(1)" \
+                                     | head -n 1 \
+                                     | sed -E 's/.*"tag_name": "([^"]+)".*/\1/')) \
+    $(if $(LATEST_RELEASE),, \
+        $(error Fetching tag of latest releases failed) \
+    ) \
     $(if $(wildcard $(BIN_DIR)/$(2)), \
-        $(info $(call warnx, $(3) is found. Skipping downloading.)), \
-        $(eval LATEST_RELEASE := $(shell wget -q https://api.github.com/repos/sysprog21/rv32emu-prebuilt/releases -O- \
-                                         | grep '"tag_name"' \
-                                         | grep "$(1)" \
-                                         | head -n 1 \
-                                         | sed -E 's/.*"tag_name": "([^"]+)".*/\1/')) \
-        $(if $(LATEST_RELEASE),, \
-            $(error Fetching tag of latest releases failed) \
-        ) \
-    )
+        $(info $(call warnx, $(3) is found. Skipping downloading.)))
 endef
 
 # Verify prebuilt files against a SHA checksum file.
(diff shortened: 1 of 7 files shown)


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: cd9b2460df0c46de2120a5855995b692b6f820d3 | Previous: 566bc0d7defb6298410769175c06ab8316a46820 | Ratio |
|-|-|-|-|
| `Dhrystone` | `1609` Average DMIPS over 10 runs | `1309` Average DMIPS over 10 runs | `0.81` |
| `Coremark` | `1007.933` Average iterations/sec over 10 runs | `927.053` Average iterations/sec over 10 runs | `0.92` |

</details>

PR #672, opened 2026-01-02
Fix verify and bump

ChinYikMing:
- Set locale env for consistent sorting order at verification
- Bump Buildroot to 2025.11
- Remove SHA tmp file at epilogue only 
  - Download artifacts takes time, also user might want to change config
for another build (e.g., Buildroot and Linux)
- Verify and add comments to show the calculation of the SHA1 value 
  - The old SHA1 value of timidity and simplefs are just same as the SHA1
  value after setting the LC_ALL=C.
  - It is easier debug by copy the calculation pipeline from the comment
  instead copying from the verify function.

---
## Summary by cubic
Make checksum verification deterministic and speed up rebuilds by keeping downloaded artifacts. Also bump Buildroot to 2025.11.

- **Bug Fixes**
  - Force LC_ALL=C during file sorting in verify to get stable SHA results.
  - Epilogue now removes only SHA temp files, preserving downloaded sources for reconfig builds.
  - Add commented SHA1 pipelines for timidity/simplefs/buildroot; timidity and simplefs SHAs remain unchanged.

- **Dependencies**
  - Update Buildroot to 2025.11 and refresh its SHA.

<sup>Written for commit 385f1a7a01a3aa58b495f4382e86663847c4644c. Summary will update on new commits.</sup>

### DIFF
--- a/mk/external.mk
+++ b/mk/external.mk
@@ -15,9 +15,9 @@ define prologue
     $(info $(_))
 endef
 
-# $(1), $(2), $(3): files to be deleted
+# $(1), $(2): files to be deleted
 define epilogue
-    $(RM) $(1) $(2) $(3)
+    $(RM) $(1) $(2)
 endef
 
 # $(1): compressed source URL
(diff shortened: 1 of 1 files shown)


ChinYikMing:
Run successfully (Build Linux Image) at [here](https://github.com/ChinYikMing/rv32emu/actions/runs/20660071497/job/59320550474), the failure is the permission denied problem since rv32emu-prebuilt release is only createable from rv32emu upstream.

PR #674, opened 2026-01-02
Fix MMU page fault handling in JIT block-trans

jserv:
The MMU test crashes with JIT enabled because block_translate() uses assert(insn) which fails when mem_ifetch returns 0 due to a page fault. In system emulation mode, page faults during instruction fetch are valid conditions that should trigger trap handling, not assertions. Changes:
- block_translate(): Replace assert(insn) with graceful break when instruction fetch fails. Add assertion to verify trap state is consistent (rv->is_trapped must be true when insn==0 in SYSTEM mode).
- rv_step(): Handle trap-pending condition when block_find_or_translate returns NULL. If rv->is_trapped is set, invoke trap_handler() and continue execution instead of halting with fatal error.
- rv_step_debug(): Add trap state verification and invoke trap_handler() when instruction fetch fails due to page fault.

The fix ensures page faults during block translation are handled by the trap mechanism rather than causing assertion failures.

---
## Summary by cubic
Fix handling of instruction-fetch page faults during JIT block translation. Page faults now trigger the trap handler instead of crashing, fixing the MMU test with JIT enabled.

- **Bug Fixes**
  - block_translate: break on failed fetch; assert trap state is set in SYSTEM mode.
  - rv_step: when translation returns NULL and a trap is pending, run trap_handler and continue instead of halting.
  - rv_step_debug: verify trap state and call trap_handler on fetch failure.

<sup>Written for commit 873320c9ef4e7a275f03ea2c69cf984520841281. Summary will update on new commits.</sup>

### DIFF
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -1104,7 +1104,17 @@ retranslate:
         }
 #endif
 
-        assert(insn);
+        /* If instruction fetch failed due to trap (page fault, etc.), break.
+         * The caller checks rv->is_trapped and invokes trap handler.
+         * Note: insn==0 alone is ambiguous; we verify trap state explicitly.
+         */
+        if (!insn) {
+#if RV32_HAS(SYSTEM)
+            assert(rv->is_trapped &&
+                   "insn fetch returned 0 without setting trap state");
+#endif
+            break;
+        }
 
         /* decode the instruction */
         if (!rv_decode(ir, insn)) {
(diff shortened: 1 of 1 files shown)


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: 873320c9ef4e7a275f03ea2c69cf984520841281 | Previous: 9007d973fdd2c6ab3576e98656104c062b09afec | Ratio |
|-|-|-|-|
| `Dhrystone` | `1651` Average DMIPS over 10 runs | `1650` Average DMIPS over 10 runs | `1.00` |
| `Coremark` | `1015.922` Average iterations/sec over 10 runs | `1012.048` Average iterations/sec over 10 runs | `1.00` |

</details>

PR #670, opened 2026-01-02
Refine build system with Kconfig integration

jserv:
Build System Improvements:
- Add mk/common.mk with reusable define templates (compile-rule, test-framework, make-dir) reducing code duplication
- Add mk/kconfig.mk for Kconfig menu integration with environment detection
- Refactor mk/tests.mk using test-framework template (~50% reduction)
- Refactor mk/external.mk with unified download/verify patterns

Kconfig Enhancements:
- Add BUILD_TARGET choice menu (Native/WebAssembly) in configs/Kconfig
- Add HAVE_EMCC detection via tools/detect-env.py --have-emcc
- Auto-select CC=emcc when CONFIG_BUILD_WASM=y in mk/toolchain.mk
- Add consistency warning when CC override conflicts with BUILD_WASM

Source Code Simplification:
- Simplify preprocessor conditionals in src/ based on Kconfig constraints (JIT requires !EMCC, T2C requires JIT, etc.)
- Replace nested #if with flattened logic where Kconfig guarantees valid combinations

---
## Summary by cubic
Adds Kconfig-driven build target selection and environment checks, auto-configuring Emscripten for WebAssembly. Builds now require an explicit .config and checksum verification for external assets is hardened to prevent silent bypass, improving artifact reliability.

- **New Features**
  - Kconfig adds a Build Target choice: Native or WebAssembly.
  - Detects Emscripten (HAVE_EMCC) and auto-sets CC=emcc when BUILD_WASM is selected.
  - Requires .config for builds; clear guidance is shown to run make defconfig or make config (named *_defconfig targets supported).
  - Reusable Make templates (compile-rule, test-framework, make-dir); tests and external asset rules refactored to use them.
  - Skips expensive dependency checks for clean/distclean; SDL flags computed only when enabled.
  - Source guards simplified with SYSTEM_MMIO and enforced T2C/JIT constraints; flattened conditionals in src/.
  - CI updated: uses wasm_defconfig, tracks mk/*, drops CC=emcc overrides, and adds a CodeQL workflow (manual build; TRAP caching disabled).

- **Migration**
  - For WASM: run “make wasm_defconfig” then “make -j” (do not pass CC=emcc).
  - If CONFIG_BUILD_WASM=y but CC isn’t emcc, a warning is shown; remove CC overrides or rerun “make defconfig”.
  - Replace any custom checks of SYSTEM && !ELF_LOADER with SYSTEM_MMIO.
[...]

PR #686, opened 2026-01-04
Fix make build-linux-image

ChinYikMing:
Errors:
cp: cannot create regular file '/tmp/linux/.config': No such file or directory ./tools/build-linux-image.sh: line 47: pushd: /tmp/linux: No such file or directory

The $(LINUX_DATA) is needed by build-linux-image but it is not set, this patch set it.

Cause by #670 (after bisecting).

---
## Summary by cubic
Fixes make build-linux-image by defining LINUX_DATA in mk/external.mk and ensuring build/linux-image/ exists. Prevents cp/pushd and missing directory errors by letting the script find the kernel tarball and use /tmp/linux.

Defined: LINUX_DATA := $(LINUX_DATA_DEST)/linux-$(LINUX_VERSION).$(LINUX_PATCHLEVEL).%.tar.gz

<sup>Written for commit c5285fc88fad5318476d7c9312b654ddacb46b84. Summary will update on new commits.</sup>


ChinYikMing:
Maybe can create a CI for build linux images but not upload to rv32emu-prebuilt, so can avoid regression.

PR #691, opened 2026-01-08
Fix T2C thread synchronization

jserv:
This addresses reliability issue in JIT parallel mode by fixing race conditions in the T2C.
- Add portable ATOMIC_* macros in common.h with GNU __atomic builtins as primary, C11 stdatomic fallback, and single-threaded fallback
- Replace busy-wait polling with proper pthread_cond_wait in t2c_runloop
- Change queue_entry_t from block pointer to cache key (PC|SATP) to prevent use-after-free when blocks are evicted during compilation
- Add defensive check in t2c_compile to skip already-compiled blocks
- Use atomic operations for shared block fields
- Proper shutdown sequence: signal under lock, join, cleanup queue

---
## Summary by cubic
Fixes T2C JIT race conditions to make background compilation reliable and thread-safe. Replaces busy-waiting with condition variables, adds portable atomics, and prevents use-after-free in the compile queue.

- **Bug Fixes**
  - Replace busy-wait polling with pthread_cond_wait in t2c_runloop; signal on enqueue and during shutdown.
  - Add portable ATOMIC_* macros and use them for shared fields (hot2, compiled, n_invoke, is_interrupted) with acquire/release ordering.
  - Queue cache keys (PC | SATP) instead of block pointers to avoid use-after-free; look up and compile under the cache lock only if the block still exists.
  - Defensive checks to skip already-compiled blocks and atomic increment of n_invoke when T2C is enabled.

- **Dependencies**
  - Build with gnu11 instead of gnu99 to support the atomics fallback path.

<sup>Written for commit c2c6847a3739397b506e06da44a1ca2a007e8bb2. Summary will update on new commits.</sup>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ $(eval $(require-config))
 OUT ?= build
 BIN := $(OUT)/rv32emu
 
-CFLAGS = -std=gnu99 $(KCONFIG_CFLAGS) -Wall -Wextra -Werror
+CFLAGS = -std=gnu11 $(KCONFIG_CFLAGS) -Wall -Wextra -Werror
 CFLAGS += -Wno-unused-label -include src/common.h -Isrc/ $(CFLAGS_NO_CET)
 LDFLAGS += $(KCONFIG_LDFLAGS)
 OBJS_EXT :=
--- a/src/common.h
+++ b/src/common.h
@@ -218,6 +218,98 @@ static inline uint8_t ilog2(uint32_t x)
 #define HAVE_MMAP 1
 #endif
 
+/* Portable atomic operations
+ *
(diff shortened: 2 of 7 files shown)


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: c2c6847a3739397b506e06da44a1ca2a007e8bb2 | Previous: 7dad6676ad06456001eded1c55b42e0571a4ebdb | Ratio |
|-|-|-|-|
| `Dhrystone` | `1657.333` DMIPS | `1657` DMIPS | `1.00` |
| `CoreMark` | `1025.5` iterations/sec | `1017.224` iterations/sec | `0.99` |

</details>

PR #709, opened 2026-01-20
Enable tzdata in Buildroot to support guest timezone configuration

kaiii708:
### Update (2026-02-01): Switch to Buildroot Configuration Approach

Based on the feedback from the reviewers, I have completely refactored this PR.
Instead of modifying the emulator source code to force local time on the RTC, the implementation now follows the standard Linux approach by installing timezone data into the Guest OS.

**Changes in this update:**
1.  **Reverted:** All C code changes (`main.c`, `rtc.c`) regarding the `-x rtc` option have been removed. The RTC emulation remains strictly UTC-compliant.
2.  **Added:** Updated `configs/buildroot.config` to:
    * Enable `BR2_TARGET_TZ_INFO` (Installs `tzdata` to `/usr/share/zoneinfo`).
    * Explicitly set `BR2_TARGET_LOCALTIME` to `"Etc/UTC"` (Ensures default system time matches RTC).

**Impact on Image Size:**
Enabling `tzdata` increases the rootfs size by approximately **1.2 MB** (ignoring the gaming artifacts), which is a reasonable trade-off for standard timezone support.

**How to Test:**
I have attached a pre-built `rootfs.cpio.zip` containing the timezone data for verification.
**Download:** [rootfs.zip](https://github.com/user-attachments/files/24991038/rootfs.zip)

**Verification Steps:**
Boot the emulator with the attached rootfs and run:
```bash
# 1. Check default time (should be UTC)
date

# 2. Change timezone to Taipei (CST)
ln -sf /usr/share/zoneinfo/Asia/Taipei /etc/localtime

# 3. Verify time is now correct local time
date
```

**⚠️ Observation: Simulation Time Drift**
The guest system time runs faster than wall-clock time due to existing simulation tick inaccuracies. This is a known emulator behavior unrelated to this PR and remains an open area for future improvements.

<details> <summary><strong>Click to view Original Description (Archived / Obsolete)</strong></summary>

Introduce a new command-line option `-x rtc:<mode>` to configure the Real-Time Clock (RTC) timezone behavior. This allows the guest system's RTC to synchronize with either the host's UTC time or local time.
- Add `-x rtc:<utc|localtime>` option to select RTC timezone mode                            
- UTC mode (default): RTC uses host UTC time                                                 
- Localtime mode: RTC uses host local time                                                   
- Add CI tests for RTC timezone mode option

---
## Summary by cubic
Enable tzdata in Buildroot so guests can configure their local timezone via /etc/localtime. Default remains UTC to match RTC; no emulator changes.

- **New Features**
  - Enabled BR2_TARGET_TZ_INFO with default zone list to install /usr/share/zoneinfo.
  - Set BR2_TARGET_LOCALTIME to "Etc/UTC" for the default system timezone.

<sup>Written for commit 4fbc1434a71dc5721affc6e08ecbe13d12ff0f64. Summary will update on new commits.</sup>

</details>


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: 78cf51fcb64a404bbc0f66152db3dbcaeea110ec | Previous: 3641924d5798573e9171ffb1b83d477b9890ab1c | Ratio |
|-|-|-|-|
| `Dhrystone` | `1869.667` DMIPS | `1619.667` DMIPS | `0.87` |
| `CoreMark` | `1000.228` iterations/sec | `949.751` iterations/sec | `0.95` |

</details>

visitorckw:
Shouldn't RTC hardware be timezone-agnostic?

visitorckw (changes requested):
From my perspective, it is quite counter-intuitive for hardware behavior to change based on the host's timezone.

We might need more explanation on the intended use cases: Which type of users would require this feature, and in what specific scenarios would it be beneficial? Furthermore, are there any real-world hardware implementations that exhibit similar behavior?

ChinYikMing:
We shall keep UTC time only since Linux guest expects UTC time from reading RTC. The option can be extended if any new use cases are introduced in the future.

But, to enable the Linux guest to modify the timezone setting, we can instead to install timezone info to rootfs (Buildroot option `BR2_TARGET_LOCALTIME` can be toggled on) and the Linux guest can aware of timezone by linking  the timezone config to `/etc/localtime` . E.g.,
```
$ ln -s  /usr/share/zoneinfo/Asia/Taipei /etc/localtime
```

Ref: [localtime](https://man7.org/linux/man-pages/man5/localtime.5.html)

kaiii708:
Thanks for the review!
@visitorckw  I agree with your perspective. Modifying the hardware behavior based on the host's config is indeed counter-intuitive and breaks the abstraction layer.

@ChinYikMing  Your suggestion regarding Buildroot is the correct approach. I have verified that it solves my use case while keeping the RTC simulation clean. I will update the PR to reflect these changes.

ChinYikMing (approved):
LGTM.

jserv:
Thank @kaiii708 for contributing!

PR #715, opened 2026-02-09
Fix errors in instruction encoding documentation

minhsun-c:
Correct the immediate value explanation for the `j -12` example, where it was incorrectly stated as 30. Update the binary representation to accurately reflect the 32-bit two's complement of -12.

Additionally, fix several typographical errors, including:
- Change "Decoding di" to "Decoding of".
- Fix pluralization to "pages 81–83".
- Correct the article "a instruction" to "an instruction".

---
## Summary by cubic
Fix instruction encoding docs by correcting the j -12 example to use -12 as the immediate and the proper 32-bit two’s complement. Also cleaned up minor typos for clarity.

- **Bug Fixes**
  - j -12 example: immediate set to -12; binary updated to 32-bit two’s complement.
  - Typos: "Decoding of", "pages 81–83", and "an instruction".

<sup>Written for commit 6a76f84fbdd9f2de198098b831d7e64eba73c68d. Summary will update on new commits.</sup>

### DIFF
--- a/docs/instruction.md
+++ b/docs/instruction.md
@@ -135,7 +135,7 @@ Identify the subfields of the immediate:
 Re-order the subfields as written on the tables. Typically, the immediate
 encoding is located just below the corresponding instruction's encoding.
 
-| Decoding di `<imm>` |    `-m2-`      |  `imm1`    | `m1` |   `imm2`     | `0` |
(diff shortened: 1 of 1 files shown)


jserv:
Thank @minhsun-c for contributing!

PR #716, opened 2026-02-09
Fix memory leak

fourcolor:
This PR fixes several ASan-reported memory leaks:

- Dispose tk/utk builders returned from instruction handlers in t2c_trace_ebb().
- Dispose messages returned by LLVMGetHostCPUName() / LLVMGetHostCPUFeatures().
- Free branch_tables for all cached blocks before teardown.
- Free fd_map and mem in rv_delete().
- Free all live cache entries in cache_free().

---
## Summary by cubic
Fixes ASan-reported memory leaks in the JIT and teardown for a clean shutdown and lower memory use. Disposes temporary LLVM builders/messages and frees branch tables on eviction/teardown, plus fd_map/mem and cache entries.

- **Bug Fixes**
  - t2c: dispose tk/utk builders; dispose LLVMGetHostCPUName/Features messages.
  - eviction/teardown: free ir->branch_table when replacing/evicting blocks and for all remaining blocks before freeing the cache; delete fd_map and mem in rv_delete().
  - cache: free all live and ghost cache entries in cache_free().

<sup>Written for commit 2138efe3ec8e0f7888610a820cd0db5e26fddd60. Summary will update on new commits.</sup>

### DIFF
--- a/src/cache.c
+++ b/src/cache.c
@@ -396,6 +396,22 @@ void cache_free(cache_t *cache)
         }
     }
 #endif
+    /* Free all live cache entries */
+    cache_entry_t *entry, *safe;
+#ifdef __HAVE_TYPEOF
+    list_for_each_entry_safe (entry, safe, &cache->list, list)
+#else
+    list_for_each_entry_safe (entry, safe, &cache->list, list, cache_entry_t)
+#endif
+        free(entry);
+    /* Free all ghost (evicted history) cache entries */
+#ifdef __HAVE_TYPEOF
+    list_for_each_entry_safe (entry, safe, &cache->ghost_list, list)
+#else
+    list_for_each_entry_safe (entry, safe, &cache->ghost_list, list,
+                              cache_entry_t)
+#endif
+        free(entry);
     free(cache->map.ht_list_head);
     free(cache);
 }
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -1983,7 +1983,7 @@ static block_t *block_find_or_translate(riscv_t *rv)
     for (rv_insn_t *ir = replaced_blk->ir_head, *next_ir; ir != NULL;
          ir = next_ir) {
         next_ir = ir->next;
-
+        free(ir->branch_table);
         if (ir->fuse)
             mpool_free(rv->fuse_mp, ir->fuse);
 
--- a/src/riscv.c
+++ b/src/riscv.c
@@ -1063,12 +1063,20 @@ void rv_set_fromhost_addr(riscv_t *rv, uint32_t addr)
 }
 #endif
 
+#if RV32_HAS(JIT)
+static void free_block_branch_tables(void *block)
+{
+    assert(block);
+    block_t *blk = (block_t *) block;
+    for (rv_insn_t *ir = blk->ir_head; ir; ir = ir->next)
+        free(ir->branch_table);
+}
+#endif
+
 void rv_delete(riscv_t *rv)
 {
     assert(rv);
-#if !RV32_HAS(JIT) || (RV32_HAS(SYSTEM_MMIO))
     vm_attr_t *attr = PRIV(rv);
-#endif
 #if !RV32_HAS(JIT)
     map_delete(attr->fd_map);
     memory_delete(attr->mem);
@@ -1099,11 +1107,15 @@ void rv_delete(riscv_t *rv)
     /* Dispose LLVM engines for all remaining blocks before freeing cache */
     clear_cache_hot(rv->block_cache, t2c_dispose_block_engine);
 #endif
+    /* Free branch tables for all remaining blocks before freeing cache */
+    clear_cache_hot(rv->block_cache, free_block_branch_tables);
     jit_state_exit(rv->jit_state);
     cache_free(rv->block_cache);
     mpool_destroy(rv->block_ir_mp);
     mpool_destroy(rv->block_mp);
     mpool_destroy(rv->fuse_mp);
+    map_delete(attr->fd_map);
+    memory_delete(attr->mem);
 #endif
 #if RV32_HAS(SYSTEM_MMIO)
     u8250_delete(attr->uart);
--- a/src/t2c.c
+++ b/src/t2c.c
@@ -380,6 +380,11 @@ static void t2c_trace_ebb(LLVMBuilderRef *builder,
             LLVMBuildRetVoid(*builder);
         }
     }
+
+    if (tk && tk != *builder)
+        LLVMDisposeBuilder(tk);
+    if (utk && utk != *builder)
+        LLVMDisposeBuilder(utk);
 }
 
 void t2c_compile(riscv_t *rv, block_t *block, pthread_mutex_t *cache_lock)
--- a/src/t2c.c
+++ b/src/t2c.c
@@ -532,9 +532,11 @@ void t2c_compile(riscv_t *rv, block_t *block, pthread_mutex_t *cache_lock)
 #else
     LLVMCodeModel code_model = LLVMCodeModelLarge;
 #endif
-    LLVMTargetMachineRef tm = LLVMCreateTargetMachine(
-        target, triple, LLVMGetHostCPUName(), LLVMGetHostCPUFeatures(),
-        LLVMCodeGenLevelNone, LLVMRelocPIC, code_model);
+    char *cpu_name = LLVMGetHostCPUName();
+    char *cpu_features = LLVMGetHostCPUFeatures();
+    LLVMTargetMachineRef tm =
+        LLVMCreateTargetMachine(target, triple, cpu_name, cpu_features,
+                                LLVMCodeGenLevelNone, LLVMRelocPIC, code_model);
     LLVMPassBuilderOptionsRef pb_option = LLVMCreatePassBuilderOptions();
     /* Run LLVM optimization passes on the generated IR.
      *
@@ -589,6 +591,8 @@ void t2c_compile(riscv_t *rv, block_t *block, pthread_mutex_t *cache_lock)
     LLVMDisposePassBuilderOptions(pb_option);
     LLVMDisposeTargetMachine(tm);
     LLVMDisposeMessage(triple);
+    LLVMDisposeMessage(cpu_name);
+    LLVMDisposeMessage(cpu_features);
 
     /* Reacquire lock to update shared state.
      * All block field writes must happen under lock to avoid data races.
--- a/src/t2c.c
+++ b/src/t2c.c
@@ -610,6 +610,7 @@ void t2c_compile(riscv_t *rv, block_t *block, pthread_mutex_t *cache_lock)
             /* Free IRs that main thread skipped during deferred eviction */
             for (rv_insn_t *ir = block->ir_head, *next_ir; ir; ir = next_ir) {
                 next_ir = ir->next;
+                free(ir->branch_table);
                 if (ir->fuse)
                     mpool_free(rv->fuse_mp, ir->fuse);
                 mpool_free(rv->block_ir_mp, ir);
@@ -631,6 +632,7 @@ void t2c_compile(riscv_t *rv, block_t *block, pthread_mutex_t *cache_lock)
         /* Free IRs that main thread skipped during deferred eviction */
         for (rv_insn_t *ir = block->ir_head, *next_ir; ir; ir = next_ir) {
             next_ir = ir->next;
+            free(ir->branch_table);
             if (ir->fuse)
                 mpool_free(rv->fuse_mp, ir->fuse);
             mpool_free(rv->block_ir_mp, ir);


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: 2138efe3ec8e0f7888610a820cd0db5e26fddd60 | Previous: b11bf0f02802be83091a09d02aee5ce390a62694 | Ratio |
|-|-|-|-|
| `Dhrystone` | `1558.667` DMIPS | `1522` DMIPS | `0.98` |
| `CoreMark` | `1103.567` iterations/sec | `1102.738` iterations/sec | `1.00` |

</details>

jserv (changes requested):
Rebase latest `master` branch

jserv (changes requested):
1. All three commits share the same "Fix memory leak" message. Each should describe the specific leak: "Dispose LLVM builders in t2c_trace_ebb", "Free fd_map/mem in JIT teardown path", etc. Makes bisection useless.
2. Ordering concern in `rv_delete`: `free_block_branch_tables` is called after `jit_state_exit`. This is fine since JIT state doesn't reference branch_tables, but a comment explaining the ordering would help future readers understand why `clear_cache_hot` is called twice.
3. The free_block_branch_tables function signature takes `void *block` to match `clear_func_t`. The cast is necessary but consider adding an assert(block) for defensive programming, consistent with the `clear_cache_hot` function's own assert(func).

src/riscv.c
```diff
@@ -1063,12 +1063,19 @@ void rv_set_fromhost_addr(riscv_t *rv, uint32_t addr)
 }
 #endif
 
+#if RV32_HAS(JIT)
+static void free_block_branch_tables(void *block)
+{
+    block_t *blk = (block_t *) block;
+    for (rv_insn_t *ir = blk->ir_head; ir; ir = ir->next)
+        free(ir->branch_table);
```
jserv, line 1072: `free_block_branch_tables` is correct for what it does, but `clear_cache_hot` only iterates `cache->list` (live entries). Blocks evicted during normal execution have their IRs freed in three separate loops (`emulate.c:1983-1990`, `t2c.c:611-616`, `t2c.c:632-637`), and none of those loops call `free(ir->branch_table)`. This means `branch_table` leaks on every runtime eviction -- the shutdown-time cleanup here only catches whatever is still in cache when the emulator exits.

Consider extracting a shared helper (or just inlining `free(ir->branch_table)` in those three loops) to make the IR teardown consistent across all paths.

src/riscv.c
```diff
@@ -1100,10 +1107,13 @@ void rv_delete(riscv_t *rv)
     clear_cache_hot(rv->block_cache, t2c_dispose_block_engine);
 #endif
     jit_state_exit(rv->jit_state);
+    clear_cache_hot(rv->block_cache, free_block_branch_tables);
     cache_free(rv->block_cache);
     mpool_destroy(rv->block_ir_mp);
     mpool_destroy(rv->block_mp);
     mpool_destroy(rv->fuse_mp);
+    map_delete(attr->fd_map);
```
jserv, line 1117: Good catch. These were only freed under `#if !RV32_HAS(JIT)`, so every JIT build leaked the entire `fd_map` and guest memory on exit.

src/t2c.c
```diff
@@ -380,6 +380,11 @@ static void t2c_trace_ebb(LLVMBuilderRef *builder,
             LLVMBuildRetVoid(*builder);
         }
     }
+
+    if (tk && tk != *builder)
```
jserv, line 384: The `tk != *builder` guard is important here -- without it, the caller's builder would be double-disposed. Since `t2c_trace_ebb` is recursive (called again at lines 333 and 359 with new `taken_builder`/`untaken_builder`), each recursive frame has its own `tk`/`utk` locals, so this is safe.

Minor nit: the commit message says "deposed" -- should be "disposed".

src/t2c.c
```diff
@@ -527,9 +532,11 @@ void t2c_compile(riscv_t *rv, block_t *block, pthread_mutex_t *cache_lock)
 #else
     LLVMCodeModel code_model = LLVMCodeModelLarge;
 #endif
-    LLVMTargetMachineRef tm = LLVMCreateTargetMachine(
-        target, triple, LLVMGetHostCPUName(), LLVMGetHostCPUFeatures(),
-        LLVMCodeGenLevelNone, LLVMRelocPIC, code_model);
+    char *cpu_name = LLVMGetHostCPUName();
```
jserv, line 535: Clean fix. `LLVMGetHostCPUName()` and `LLVMGetHostCPUFeatures()` return `malloc`-allocated strings that must be freed with `LLVMDisposeMessage()`. Previously they were passed inline to `LLVMCreateTargetMachine()` and leaked on every T2C compilation.

src/cache.c
```diff
@@ -396,6 +396,22 @@ void cache_free(cache_t *cache)
...
     }
 #endif
+    /* Free all live cache entries */
+    cache_entry_t *entry, *safe;
+#ifdef __HAVE_TYPEOF
+    list_for_each_entry_safe (entry, safe, &cache->list, list)
+#else
+    list_for_each_entry_safe (entry, safe, &cache->list, list, cache_entry_t)
+#endif
+        free(entry);
```
jserv, line 406: Correct. Note the ordering dependency: any `clear_cache_hot` callbacks (like `t2c_dispose_block_engine` and `free_block_branch_tables`) that dereference `entry->value` must run before this, since this frees the cache entries that hold those value pointers. The current call site in `rv_delete()` gets this right.

jserv (commented):
The five individual fixes are correct -- no double-free, no ordering bugs, no regressions introduced. The LLVM builder/message disposal in `t2c.c`, the `fd_map`/`mem` freeing in the JIT teardown path, and the cache entry cleanup in `cache_free()` all look good.

However, `free_block_branch_tables` via `clear_cache_hot` only covers blocks still in the live cache at shutdown. The same `branch_table` leak exists in three runtime eviction paths that free IR nodes but never call `free(ir->branch_table)`:

1. `src/emulate.c:1983-1990` -- normal cache eviction when cache is full
2. `src/t2c.c:611-616` -- deferred free (T2C compilation failed, block was evicted)
3. `src/t2c.c:632-637` -- deferred free (T2C compilation succeeded, block was evicted)

Compare with the non-JIT teardown at `src/riscv.c:75` which correctly calls `free(ir->branch_table)`. Under workloads with heavy indirect branching (JALR), evicted blocks leak their `branch_table` allocations on every eviction cycle -- unbounded host memory growth proportional to eviction rate.

The fix is a one-liner in each loop, inserted before `mpool_free(rv->block_ir_mp, ir)`:
```c
free(ir->branch_table);
```

Also, all three commits share the identical message "Fix memory leak". Each should describe the specific leak it fixes to make `git bisect` and `git log` actually useful.

jserv (commented):
## Review Summary

All five leak categories are correctly fixed. No double-free, no use-after-free, no race conditions, no regressions. Cross-validated with Gemini and Codex (GPT) — all three reviewers agree the fixes are sound.

### Per-commit assessment

1. **970f619** `Dispose LLVM builders in t2c_trace_ebb` — The `tk != *builder` / `utk != *builder` guard is correct; avoids double-disposing the caller's builder.

2. **72f3e75** `Free LLVMGetHostCPU* return strings` — Textbook fix. LLVM C API returns `malloc`'d strings; `LLVMDisposeMessage` is the right cleanup.

3. **3cbd3f0** `Free fd_map/mem in JIT teardown + cache entries` — `fd_map`/`mem` were only freed in the non-JIT `#else` path. The `cache_free()` entry cleanup is also necessary (list nodes were previously leaked).

4. **2138efe** `Free branch_table on runtime block eviction` — Covers all three eviction loops. `free(NULL)` is a no-op per C standard, so calling unconditionally on every IR node is safe and simple.

### Ordering correctness

The teardown sequence in `rv_delete` is the only correct ordering: `clear_cache_hot(free_block_branch_tables)` → `jit_state_exit` → `cache_free` → `mpool_destroy`. Reversing `free_block_branch_tables` after `mpool_destroy` would be use-after-free on IR nodes. The current code is correct.

### Pre-existing leak (not introduced by this PR)

In `src/t2c_template.c`, the hand-written `cbeqz` (line 1214-1218) and `cbnez` handlers do not call `LLVMDisposeBuilder(builder2/builder3)` in their `else` branches, unlike the macro version (line 343/355) which correctly disposes them. Worth a follow-up fix.

src/t2c.c
```diff
@@ -380,6 +380,11 @@ static void t2c_trace_ebb(LLVMBuilderRef *builder,
             LLVMBuildRetVoid(*builder);
         }
     }
+
+    if (tk && tk != *builder)
+        LLVMDisposeBuilder(tk);
+    if (utk && utk != *builder)
+        LLVMDisposeBuilder(utk);
```
jserv, line 387: Good. The `tk != *builder` and `utk != *builder` guards are necessary and correct — when the branch is taken/untaken within the EBB, `tk`/`utk` alias `*builder` and must not be double-disposed. When they point to distinct builders created by instruction handlers, they are properly cleaned up here.

One subtlety: `tk`/`utk` are initialized to `NULL` and only set by branch instruction handlers. The `tk &&` / `utk &&` null checks handle the case where no branch was encountered. Clean.

src/t2c.c
```diff
@@ -584,6 +591,8 @@ void t2c_compile(riscv_t *rv, block_t *block, pthread_mutex_t *cache_lock)
     LLVMDisposePassBuilderOptions(pb_option);
     LLVMDisposeTargetMachine(tm);
     LLVMDisposeMessage(triple);
+    LLVMDisposeMessage(cpu_name);
+    LLVMDisposeMessage(cpu_features);
```
jserv, line 595: Correct. `LLVMGetHostCPUName()` and `LLVMGetHostCPUFeatures()` return `malloc`'d C strings per LLVM C API contract. `LLVMDisposeMessage` (which calls `free`) is the documented cleanup. Previously these were passed as temporaries directly to `LLVMCreateTargetMachine`, leaking on every T2C compilation.

src/emulate.c
```diff
@@ -1983,7 +1983,7 @@ static block_t *block_find_or_translate(riscv_t *rv)
     for (rv_insn_t *ir = replaced_blk->ir_head, *next_ir; ir != NULL;
          ir = next_ir) {
         next_ir = ir->next;
-
+        free(ir->branch_table);
```
jserv, line 1986: Correct. This is the runtime eviction path in `block_find_or_translate` — when the cache is full and a block is replaced. Without this, blocks with indirect branches (JALR dispatch tables) leaked their `branch_table` allocation on every eviction cycle. `free(NULL)` is a no-op for the majority of IR nodes that have no branch table.

src/riscv.c
```diff
@@ -1099,11 +1107,15 @@ void rv_delete(riscv_t *rv)
     /* Dispose LLVM engines for all remaining blocks before freeing cache */
     clear_cache_hot(rv->block_cache, t2c_dispose_block_engine);
 #endif
+    /* Free branch tables for all remaining blocks before freeing cache */
+    clear_cache_hot(rv->block_cache, free_block_branch_tables);
```
jserv, line 1111: This must remain before `mpool_destroy(rv->block_ir_mp)` at line 1114. `free_block_branch_tables` dereferences `ir->branch_table` through IR nodes allocated from that pool — destroying the pool first would be use-after-free.

Consider adding a brief comment noting this ordering constraint, e.g.:
```c
/* Must precede mpool_destroy(block_ir_mp) -- IR nodes live in that pool */
```

src/riscv.c
```diff
@@ -1099,11 +1107,15 @@ void rv_delete(riscv_t *rv)
...
     jit_state_exit(rv->jit_state);
     cache_free(rv->block_cache);
     mpool_destroy(rv->block_ir_mp);
     mpool_destroy(rv->block_mp);
     mpool_destroy(rv->fuse_mp);
+    map_delete(attr->fd_map);
```
jserv, line 1117: Good catch. These were previously only freed in the `#if !RV32_HAS(JIT)` path (line 1081-1082), meaning every JIT-enabled build leaked `fd_map` and `mem` on exit.

src/cache.c
jserv, line 406: Nit: The `#ifdef __HAVE_TYPEOF` / `#else` pattern appears twice (lines 401-406 for live, 408-414 for ghost). Consider factoring into a helper macro or static function that takes a `struct list_head *` to reduce duplication. Not a blocker.

src/t2c.c
```diff
@@ -601,6 +610,7 @@ void t2c_compile(riscv_t *rv, block_t *block, pthread_mutex_t *cache_lock)
             /* Free IRs that main thread skipped during deferred eviction */
             for (rv_insn_t *ir = block->ir_head, *next_ir; ir; ir = next_ir) {
                 next_ir = ir->next;
+                free(ir->branch_table);
```
jserv, line 613: Correct. This deferred eviction path runs under `cache_lock` (acquired at line 600). The block was evicted by the main thread while T2C was compiling it — now we properly free `branch_table` before returning the IR node to the pool. Same pattern correctly applied at line 635 for the successful-compilation-but-evicted case.

jserv:
Thank @fourcolor for contributing!

PR #720, opened 2026-03-01
Fix JIT jump table overflow in block chaining

jserv:
translate_chained_block() recursively chains translated blocks, each emitting jump targets into a fixed-size jumps[MAX_JUMPS] array. Deep chains (triggered by newer Linux kernel images with longer hot paths) overflow the array, hitting assert(n_jumps < MAX_JUMPS).

This adds a budget guard that checks remaining jump slots before translating each block: n_insn * JUMPS_PER_INSN + 2 accounts for the worst-case per-instruction emissions (5: SYSTEM_MMIO load or branch epilogue) plus the page-terminated block epilogue (2: emit_jmp + emit_exit).

---
## Summary by cubic
Prevents JIT jump table overflow during deep block chaining by adding a jump budget check. Fixes crashes (assert n_jumps < MAX_JUMPS) seen with newer Linux kernel hot paths.

- Bug Fixes
  - Add jump-slot budget check in translate_chained_block: n_insn * JUMPS_PER_INSN + 2 against MAX_JUMPS.
  - Define JUMPS_PER_INSN = 5 based on worst-case SYSTEM_MMIO loads and branch epilogues.
  - Bail out cleanly if the root block isn’t translated to avoid resolve_jumps and cache maintenance on stale/empty code.

- Migration
  - jit_translate now returns bool. Update callers to check the result and only execute the block when true.

<sup>Written for commit 8092c66edef8b97886ebf835094ba37f1cdda62d. Summary will update on new commits.</sup>

### DIFF
--- a/src/emulate.c
+++ b/src/emulate.c
@@ -2290,24 +2290,25 @@ void rv_step(void *arg)
             && runtime_profiler(rv, block)
 #endif
         ) {
-            jit_translate(rv, block);
+            if (jit_translate(rv, block)) {
 #if defined(__aarch64__)
-            /* Ensure instruction cache coherency before executing JIT code */
-            __asm__ volatile("isb" ::: "memory");
+                /* Ensure icache coherency before executing JIT */
+                __asm__ volatile("isb" ::: "memory");
 #endif
-            ((exec_block_func_t) state->buf)(
-                rv, (uintptr_t) (state->buf + block->offset));
-            rv->csr_cycle += block->cycle_cost;
+                ((exec_block_func_t) state->buf)(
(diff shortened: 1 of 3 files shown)


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: 8092c66edef8b97886ebf835094ba37f1cdda62d | Previous: 6dca3b98c6ff43f4583e299179c9657515e5cf2b | Ratio |
|-|-|-|-|
| `Dhrystone` | `1535.333` DMIPS | `1555.667` DMIPS | `1.01` |
| `CoreMark` | `1110.937` iterations/sec | `1108.019` iterations/sec | `1.00` |

</details>

PR #746, opened 2026-05-21
Inline dTLB fast path in JIT with superpage

jserv:
System-mode JIT-translated load/stores previously trampolined into the C slow path on every memory access.  Add an inline TLB probe so hits stay inline, with the slow path reached only on miss.

Probe checks the dTLB entry for: alignment, valid+VPN match, permission, dirty (stores), and that paddr is within mem_size (MMIO falls through to C).  On hit, RAM access is emitted inline with the level-appropriate offset mask (12-bit for 4 KiB, 22-bit for 4 MiB superpages) selected via CSEL (aarch64) / CMOVcc (x86-64).

T1 covers both aarch64 and x86-64 SysV.  T2C emits the equivalent LLVM IR for 4 KiB pages; T2C superpage handling stays on the slow path to avoid an intermittent crash with a select-based variant.

GEN_LOAD/STORE in src/rv32_jit.c set up the allocator state once before the probe (store_back + reset_reg + map_vm_reg(rd)) so the fast and slow paths land their result in the same host register.  The slow-path call jit_mmu_handler now takes vaddr/type/pc as arguments instead of staging them in rv->jit_mmu memory, and the SysV x86-64 call site keeps %rsp 16-byte aligned through the trampoline.

Static asserts in src/jit.c and src/t2c.c lock the tlb_entry_t layout, TLB_SIZE bounds, and PTE/level constants used by the emitted code.

Validated on aarch64 host with system+JIT+T2C (LLVM 20):
 - defconfig, jit_defconfig make check: all PASS
 - Linux boots to login under system+JIT+T2C
 - 4 MiB sha256+gzip workload, 5 paired iterations: gzip -1: 1.120s -> 0.927s (-17.2%) gzip -9: 1.119s -> 0.937s (-16.3%) sha256:  5.089s -> 5.045s ( -0.9%)

---
## Summary by cubic
Inline dTLB fast path for system-mode JIT loads/stores so RAM hits execute inline; misses and MMIO fall back to C. Also reduces translation overhead by clearing only used jump slots.

- **New Features**
  - Inline dTLB probe on aarch64 and x86-64 SysV with alignment/VPN/valid/perm/dirty/range checks; RAM hits execute inline; superpages handled via CSEL/CMOVcc.
  - T2C adds a matching fast path for 4 KiB pages; superpages remain on the slow path due to an intermittent crash with a select-based variant.
  - Increased jump capacity for probe branches (`MAX_JUMPS` 65536; `JUMPS_PER_INSN` 16).

- **Bug Fixes**
  - Initialize `state->n_jumps` in `jit_state_init` to avoid first-translation overruns; switch slow-path ABI to pass `vaddr`/`type`/`pc` to `jit_mmu_handler` and keep x86-64 SysV `%rsp` 16-byte aligned through the trampoline.

<sup>Written for commit b5ea066889e68498165cb734f8d04edd6dce1437. Summary will update on new commits. <a href="https://cubic.dev/pr/sysprog21/rv32emu/pull/746?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/jit.c
+++ b/src/jit.c
@@ -50,6 +50,53 @@
 #include "system.h"
 #endif
 
+#if RV32_HAS(SYSTEM_MMIO)
+/* JIT MMU fast-path ABI invariants.
+ *
+ * The inline dTLB probe emitted by the per-arch fast path hard-codes field
+ * offsets and sizes of tlb_entry_t and memory_t. Any layout change must keep
+ * these asserts true, or the emitted asm/IR will silently access the wrong
+ * bytes. These mirror the structures in riscv_private.h and io.h.
+ *
+ * The fast path also assumes:
+ *   - TLB_SIZE is a power of two so VPN -> index is a simple AND.
+ *   - T1 handles both TLB_PAGE_LEVEL_4K and TLB_PAGE_LEVEL_SUPER inline:
+ *     the offset mask is chosen by CSEL (aarch64) / CMOVcc (x86-64) based on
+ *     the level byte (12 bits for 4 KiB, 22 bits for 4 MiB).  T2C still
+ *     bails to C on superpage entries — an LLVM-IR select variant showed
+ *     intermittent crashes after T2C compilation, root cause not isolated.
+ *   - dirty == 1 on writes; clean PTE writes fall back to C so PTE_D update
+ *     stays in dtlb_lookup() / mmu_translate().
+ *   - paddr range is checked against memory_t mem_size to keep MMIO mappings
+ *     on the slow path.
+ */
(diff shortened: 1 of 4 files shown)


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: b5ea066889e68498165cb734f8d04edd6dce1437 | Previous: 6780cf1ccb7e75227116c68cdc88028f5c3fec52 | Ratio |
|-|-|-|-|
| `Dhrystone` | `1469.333` DMIPS | `1523` DMIPS | `1.04` |
| `CoreMark` | `1116.573` iterations/sec | `1115.919` iterations/sec | `1.00` |

</details>

PR #638, opened 2025-12-14
Fix system emulation reboot

ChinYikMing:
## Description

At 7d13e82, the SBI RST extension did not distinguish between reboot and shutdown type. When a userspace reboot command was issued, the hart was incorrectly halted as if it were a shutdown.

These changes fix the issue for both modes (interpreter and JIT) by properly detecting and handling two reboot types: cold and warm. Each type has its own handler: rv_cold_reboot() and rv_warm_reboot(). Since the initial system power-on is treated as a cold reboot, the initialization code has been refactored to share logic with the reboot path, adhering to the DRY principle. Additionally, device reset helper functions (plic_reset(), u8250_reset()) are introduced to support peripheral reinitialization during reboot.

Key changes:
1. Rename rv_reset() to rv_cold_reboot() - Full system reset including processor, memory, and all peripherals. The initial power-on is treated as a cold reboot.

2. Introduce rv_warm_reboot() - Faster reboot that only resets processor and memory, skipping peripheral reinitialization. Can be triggered via echo "warm" > /sys/kernel/reboot/mode in guestOS.

3. Refactor boot image loading - Extract load_boot_images() helper to load kernel, DTB, and initrd, reducing code duplication between cold and warm reboot paths.

4. Introduce rv_reset_hart() - Static helper function to reset only hart state (GPRs, CSRs, PC), shared by both reboot modes.

5. Add reboot-safe resource management - Add "check for reboot" comments throughout initialization to reuse already-allocated resources (memory, fd_map, PLIC, UART, vblk, block_map, vrng, etc) instead of re-allocating. For JIT state resources, might free and re-allocate (e.g., calling jit_state_exit()) to reuse the existing APIs.

6. Use calloc for vblk/disk arrays - Changed from malloc to calloc so pointers are initialized to NULL, simplifying reboot checks.

7. Use setjmp/longjmp for clean reboot - Reboot rewrites all registers, causing call stack values to become stale. setjmp in rv_step() establishes a return point, longjmp after reboot provides a clean call stack.

8. Add plic_reset() and u8250_reset() - New device reset functions to reinitialize state without free/realloc.

9. Add sbi_rst_type_str() and sbi_rst_reason_str() - Helper functions for human-readable SBI reset type/reason in dmesg.

10. Introduce rv_{spawn,destroy}_t2c(), rv_init_jit() - Helper functions for the lifecycle management of tier-2 JIT background thread and common init logic for tier-1 and tier-2 JIT compilation.

## Testing
1. Clone the repo:
[...]

PR #748, opened 2026-07-01
Migrate virtio-net from semu

Charlie-Tsai1123:
## Summary

This PR migrates virtio-net support from semu and follows the existing virtio-blk and virtio-rng integration in rv32emu.

The implementation adds a TAP-backed virtio-net device model for system emulation mode, including MMIO register handling, queue setup, feature negotiation, RX/TX virtqueue handling, virtio-net header processing, interrupt delivery through the PLIC, dynamic DTB node creation, and a runtime option for enabling the device.

## Implementation notes

* Add a TAP backend helper for host network I/O.
* Add a virtio-net device model with RX and TX virtqueue handling.
* Handle the virtio-net header on both guest TX and RX paths.
* Route virtio-net MMIO accesses through the system MMIO path.
* Update virtio-net interrupts through the PLIC.
* Dynamically create a `virtio,mmio` DTB node for virtio-net.
* Ensure virtio-net can coexist with existing virtio-blk and virtio-rng devices without reusing MMIO bases or IRQs.
* Add a runtime option for enabling virtio-net with a TAP backend.
* Note that opening a TAP device usually requires elevated privileges, so rv32emu should be executed with `sudo` or equivalent permissions when using the TAP backend.

## Test

If don't have build/linux-image/Image
```bash
make build-linux-image
```

Build:

```bash
make system_defconfig
make ENABLE_SYSTEM=1

sudo build/rv32emu \
  -k build/linux-image/Image \
  -i build/linux-image/rootfs.cpio \
  -x vrng \
  -x vnet:tap
```
After run rv32emu with `-x vnet:tap` rv32emu would build TAP, so host linux doesn't need to build  TAP again. 
Host TAP setup:

```bash
ip link | grep tap
sudo ip addr add 192.168.100.1/24 dev tap0
sudo ip link set  tap0 up
sudo tcpdump -i tap0 -n -vv
```

Guest device verification:

```bash
readlink /sys/bus/virtio/devices/virtio0/driver
ip link set eth0 up
ip addr add 192.168.100.2/24 dev eth0
ip addr show eth0
ping -c 3 192.168.100.1
```

Expected Result:
```bash
# readlink /sys/bus/virtio/devices/virtio0/driver
../../../../../bus/virtio/drivers/virtio_net
# ip link set eth0 up
# ip addr add 192.168.100.2/24 dev eth0
# ip addr show eth0
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast qlen 1000
    link/ether aa:e0:56:5b:58:bc brd ff:ff:ff:ff:ff:ff
    inet 192.168.100.2/24 scope global eth0
       valid_lft forever preferred_lft forever
# ping -c 3 192.168.100.1
PING 192.168.100.1 (192.168.100.1): 56 data bytes
64 bytes from 192.168.100.1: seq=0 ttl=64 time=1.200 ms

64 bytes from 192.168.100.1: seq=1 ttl=64 time=1001.085 ms
64 bytes from 192.168.100.1: seq=2 ttl=64 time=1.334 ms

--- 192.168.100.1 ping statistics ---
3 packets transmitted, 3 packets received, 0% packet loss
round-trip min/avg/max = 1.200/334.539/1001.085 ms
# 
```

---
## Summary by cubic
Migrates virtio-net from semu into rv32emu's system emulator as a build-configurable `-x vnet:<backend>` device. Supports Linux TAP, user-mode SLIRP, and macOS vmnet; TAP still requires Linux host privileges.

**Features**
- Implements VirtIO-MMIO feature negotiation, MAC/link/MTU configuration, RX/TX queues, headers, interrupts, reset, DTB wiring, and backend polling.
- Adds user-mode networking through bundled `minislirp` without host setup, plus shared-mode `vmnet` support on macOS with Clang.
- Adds Kconfig and build pruning that matches backend selection to host/compiler capabilities, excludes networking on Emscripten, and rejects invalid backend configurations.
- Documents setup and adds Linux TAP/user and macOS user/vmnet boot coverage using runtime network settings.

**Fixes**
- Refreshes the PLIC promptly and only signals used-ring interrupts when the ring advances, avoiding delayed packets and spurious interrupts.
- Preserves the host backend across cold and warm reboots while resetting guest-visible state, handles permanent backend errors, and throttles polling to 5,000 guest cycles, keeping it active under the GDB stub.
- Tracks all `minislirp` timers so expired timers fire correctly.
- Fixes virtqueue index wraparound validation in virtio-net, virtio-blk, and virtio-rng, and reads device configuration at byte offsets so guests receive the configured MAC.
- Hardens teardown so vmnet state survives until stop completes, cleans up backends before reinitialization, and keeps virtio-net cleanup running after block-image failures.
- Keeps build objects aligned with effective `ENABLE_*` overrides, invalidates stale objects when features or host capabilities change, and fixes `SoftFloat`, `minislirp`, and generated DTB dependency tracking.
- Makes ARM64 dependency setup more resilient with IPv4 apt retries and ensures boot tests have `expect`.

<sup>Written for commit 25152a03f3e3cb9733fb563c4c9f667109f1b934. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/rv32emu/pull/748?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.ci/test-netdev.sh
+++ b/.ci/test-netdev.sh
@@ -1,0 +1,81 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+. "${SCRIPT_DIR}/common.sh"
+
+check_platform
+
+if [[ "${OS_TYPE}" != "Linux" ]]; then
+    print_warning "Skipping virtio-net test on non-Linux host"
+    exit 0
+fi
+
+register_cleanup cleanup_emulator
+
+TIMEOUT=${BOOT_TIMEOUT:-60}
+
+ASSERT expect <<- DONE
+    set timeout ${TIMEOUT}
+    set tap_if ""
+
+    spawn sudo -E build/rv32emu \
+        -k build/linux-image/Image \
+        -i build/linux-image/rootfs.cpio \
+        -x vnet:tap
+
+    expect {
+        -re {allocated TAP interface: (tap[0-9]+)} {
+            set tap_if \$expect_out(1,string)
+            exp_continue
+        }
+        "buildroot login:" {
+            if { "\$tap_if" == "" } {
+                set tap_if [exec sh -c {ip -o link show | awk -F': ' '\$2 ~ /^tap[0-9]+/ {print \$2; exit}'}]
+            }
+
+            exec sudo ip addr replace 192.168.100.1/24 dev \$tap_if
+            exec sudo ip link set \$tap_if up
+
+            send "root\\r"
+        }
+        timeout {
+            exit 1
+        }
+    }
+
+    expect "# "
+    send "readlink /sys/bus/virtio/devices/virtio0/driver\\r"
+    expect {
+        "virtio_net" {}
+        timeout { exit 2 }
+    }
+
+    expect "# "
+    send "ip link set eth0 up\\r"
+
+    expect "# "
+    send "ip addr add 192.168.100.2/24 dev eth0\\r"
+
+    expect "# "
+    send "ip addr show eth0\\r"
+    expect {
+        "192.168.100.2/24" {}
+        timeout { exit 3 }
+    }
+
+    expect "# "
+    send "ping -c 3 -W 5 192.168.100.1\r"
+    expect {
+        -re {3 packets transmitted, 3 packets received|3 packets transmitted, 3 received} {
+            send "\x01"
+            send "x"
+            exit 0
+        }
+        timeout {
+            exit 4
+        }
+DONE
+
+print_success "virtio-net boot test passed"
--- a/src/devices/netdev.c
+++ b/src/devices/netdev.c
@@ -1,0 +1,116 @@
+/*
+ * rv32emu is freely redistributable under the MIT License. See the file
+ * "LICENSE" for information on usage and redistribution of this file.
+ */
+
+#include "netdev.h"
+
+#include <stdbool.h>
+#include <string.h>
+
+#include "utils.h"
+
+#if defined(__linux__) && !defined(__EMSCRIPTEN__)
+
+#include <errno.h>
+#include <fcntl.h>
+#include <linux/if.h>
+#include <linux/if_tun.h>
+#include <stdlib.h>
+#include <sys/ioctl.h>
+#include <unistd.h>
+
+static int net_init_tap(netdev_t *netdev)
+{
+    net_tap_options_t *tap = (net_tap_options_t *) netdev->op;
+
+    tap->tap_fd = open("/dev/net/tun", O_RDWR);
+    if (tap->tap_fd < 0) {
+        rv_log_error("failed to open TAP device: %s", strerror(errno));
+        return -1;
+    }
+
+    struct ifreq ifreq = {
+        .ifr_flags = IFF_TAP | IFF_NO_PI,
+    };
+
+    strncpy(ifreq.ifr_name, "tap%d", sizeof(ifreq.ifr_name) - 1);
+
+    if (ioctl(tap->tap_fd, TUNSETIFF, &ifreq) < 0) {
+        rv_log_error("failed to allocate TAP device: %s", strerror(errno));
+        close(tap->tap_fd);
+        tap->tap_fd = -1;
+        return -1;
+    }
+
+    rv_log_info("allocated TAP interface: %s", ifreq.ifr_name);
+
+    if (fcntl(tap->tap_fd, F_SETFL,
+              fcntl(tap->tap_fd, F_GETFL, 0) | O_NONBLOCK) < 0) {
+        rv_log_error("failed to set TAP non-blocking mode: %s",
+                     strerror(errno));
+        close(tap->tap_fd);
+        tap->tap_fd = -1;
+        return -1;
+    }
+
+    return 0;
+}
+
+bool netdev_init(netdev_t *netdev, const char *net_type)
+{
+    if (!netdev || !net_type || strcmp(net_type, "tap"))
+        return false;
+
+    netdev->name = (char *) net_type;
+    netdev->type = NETDEV_IMPL_tap;
+    netdev->op = calloc(1, sizeof(net_tap_options_t));
+    if (!netdev->op)
+        return false;
+
+    if (net_init_tap(netdev) < 0) {
+        free(netdev->op);
+        netdev->op = NULL;
+        return false;
+    }
+
+    return true;
+}
+
+void netdev_delete(netdev_t *netdev)
+{
+    if (!netdev || !netdev->op)
+        return;
+
+    switch (netdev->type) {
+    case NETDEV_IMPL_tap: {
+        net_tap_options_t *tap = (net_tap_options_t *) netdev->op;
+        if (tap->tap_fd >= 0)
+            close(tap->tap_fd);
+        break;
+    }
+    default:
+        break;
+    }
+
+    free(netdev->op);
+    netdev->op = NULL;
+}
+
+#else
+
+bool netdev_init(netdev_t *netdev, const char *net_type)
+{
+    (void) netdev;
+    (void) net_type;
+
+    rv_log_error("virtio-net TAP backend is only supported on Linux hosts");
+    return false;
+}
+
+void netdev_delete(netdev_t *netdev)
+{
+    (void) netdev;
+}
+
+#endif
--- a/src/devices/netdev.h
+++ b/src/devices/netdev.h
@@ -1,0 +1,28 @@
+/*
+ * rv32emu is freely redistributable under the MIT License. See the file
+ * "LICENSE" for information on usage and redistribution of this file.
+ */
+
+#pragma once
+
+#include <stdbool.h>
+
+typedef struct netdev netdev_t;
+
+typedef enum {
+    NETDEV_IMPL_tap,
+} netdev_impl_t;
+
+typedef struct {
+    int tap_fd;
+} net_tap_options_t;
+
+struct netdev {
+    char *name;
+    netdev_impl_t type;
+    void *op;
+};
+
+bool netdev_init(netdev_t *netdev, const char *net_type);
+
+void netdev_delete(netdev_t *netdev);
--- a/src/devices/virtio-net.c
+++ b/src/devices/virtio-net.c
@@ -1,0 +1,693 @@
+/*
+ * rv32emu is freely redistributable under the MIT License. See the file
+ * "LICENSE" for information on usage and redistribution of this file.
+ */
+
+#include <assert.h>
+#include <errno.h>
+#include <poll.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/uio.h>
+#include <unistd.h>
+
+#include "utils.h"
+#include "virtio.h"
+
+#define VNET_FEATURES_0 0
+#define VNET_FEATURES_1 1 /* VIRTIO_F_VERSION_1 */
+
+#define VNET_QUEUE_NUM_MAX 1024
+#define VNET_QUEUE_COUNT 2
+#define VNET_QUEUE (vnet->queues[vnet->queue_sel])
+
+enum {
+    VNET_QUEUE_RX = 0,
+    VNET_QUEUE_TX = 1,
+};
+
+#define VNET_LINK_UP 1
+#define VNET_HDR_SIZE 12
+
+typedef struct __attribute__((packed)) {
+    uint8_t mac[6];
+    uint16_t status;
+    uint16_t max_virtqueue_pairs;
+    uint16_t mtu;
+} virtio_net_config_t;
+
+static size_t vnet_min(size_t a, size_t b)
+{
+    return a < b ? a : b;
+}
+
+static void virtio_net_set_fail(virtio_net_state_t *vnet)
+{
+    vnet->status |= VIRTIO_STATUS_DEVICE_NEEDS_RESET;
+    if (vnet->status & VIRTIO_STATUS_DRIVER_OK)
+        vnet->interrupt_status |= VIRTIO_INT_CONF_CHANGE;
+}
+
+static bool vnet_guest_range_ok(uint64_t addr, uint64_t len)
+{
+    if (len == 0)
+        return true;
+
+    if (addr >= MEM_SIZE)
+        return false;
+
+    if (len > MEM_SIZE - addr)
+        return false;
+
+    return true;
+}
+
+static bool vnet_word_range_ok(uint32_t word_index, uint64_t words)
+{
+    uint64_t byte_addr = (uint64_t) word_index * sizeof(uint32_t);
+    uint64_t byte_len = words * sizeof(uint32_t);
+
+    return vnet_guest_range_ok(byte_addr, byte_len);
+}
+
+static bool vnet_check_word_range(virtio_net_state_t *vnet,
+                                  uint32_t word_index,
+                                  uint64_t words)
+{
+    if (!vnet_word_range_ok(word_index, words)) {
+        virtio_net_set_fail(vnet);
+        return false;
+    }
+
+    return true;
+}
+
+static inline uint32_t vnet_preprocess(virtio_net_state_t *vnet, uint32_t addr)
+{
+#if MEM_SIZE < 0x100000000ULL
+    if ((addr >= MEM_SIZE) || (addr & 0b11)) {
+#else
+    if (addr & 0b11) {
+#endif
+        virtio_net_set_fail(vnet);
+        return 0;
+    }
+
+    return addr >> 2;
+}
+
+static void virtio_net_update_status(virtio_net_state_t *vnet, uint32_t status)
+{
+    vnet->status |= status;
+    if (status)
+        return;
+
+    /* Reset while preserving environment-owned fields. */
+    netdev_t peer = vnet->peer;
+    uint32_t *ram = vnet->ram;
+    void *priv = vnet->priv;
+
+    memset(vnet, 0, sizeof(*vnet));
+
+    vnet->peer = peer;
+    vnet->ram = ram;
+    vnet->priv = priv;
+    vnet->queues[VNET_QUEUE_RX].fd_ready = true;
+    vnet->queues[VNET_QUEUE_TX].fd_ready = true;
+}
+
+static bool vnet_iovec_write(struct iovec **vecs,
+                             size_t *nvecs,
+                             const uint8_t *src,
+                             size_t n)
+{
+    while (n > 0 && *nvecs > 0) {
+        size_t to_copy = vnet_min(n, (*vecs)->iov_len);
+
+        memcpy((*vecs)->iov_base, src, to_copy);
+        src += to_copy;
+        n -= to_copy;
+
+        (*vecs)->iov_base = (void *) ((uintptr_t) (*vecs)->iov_base + to_copy);
+        (*vecs)->iov_len -= to_copy;
+
+        if ((*vecs)->iov_len == 0) {
+            (*vecs)++;
+            (*nvecs)--;
+        }
+    }
+
+    return n > 0;
+}
+
+static bool vnet_iovec_read(struct iovec **vecs,
+                            size_t *nvecs,
+                            uint8_t *dst,
+                            size_t n)
+{
+    while (n > 0 && *nvecs > 0) {
+        size_t to_copy = vnet_min(n, (*vecs)->iov_len);
+
+        memcpy(dst, (*vecs)->iov_base, to_copy);
+        dst += to_copy;
+        n -= to_copy;
+
+        (*vecs)->iov_base = (void *) ((uintptr_t) (*vecs)->iov_base + to_copy);
+        (*vecs)->iov_len -= to_copy;
+
+        if ((*vecs)->iov_len == 0) {
+            (*vecs)++;
+            (*nvecs)--;
+        }
+    }
+
+    return n > 0;
+}
+
+static bool vnet_build_iovs(virtio_net_state_t *vnet,
+                            virtio_net_queue_t *queue,
+                            uint16_t buffer_idx,
+                            bool device_writes,
+                            struct iovec *iovs,
+                            size_t *niovs)
+{
+    uint32_t *ram = vnet->ram;
+    uint16_t desc_idx = buffer_idx;
+
+    *niovs = 0;
+
+    if (queue->queue_num == 0 || buffer_idx >= queue->queue_num) {
+        virtio_net_set_fail(vnet);
+        return false;
+    }
+
+    for (uint32_t iter = 0; iter < queue->queue_num; iter++) {
+        if (desc_idx >= queue->queue_num) {
+            virtio_net_set_fail(vnet);
+            return false;
+        }
+
+        uint32_t desc_addr = queue->queue_desc + desc_idx * 4;
+        if (!vnet_check_word_range(vnet, desc_addr, 4))
+            return false;
+
+        const struct virtq_desc *desc =
+            (const struct virtq_desc *) &ram[desc_addr];
+
+        bool writable = !!(desc->flags & VIRTIO_DESC_F_WRITE);
+        if (writable != device_writes) {
+            virtio_net_set_fail(vnet);
+            return false;
+        }
+
+        if (!vnet_guest_range_ok(desc->addr, desc->len)) {
+            virtio_net_set_fail(vnet);
+            return false;
+        }
+
+        iovs[*niovs].iov_base = (void *) ((uintptr_t) ram + desc->addr);
+        iovs[*niovs].iov_len = desc->len;
+        (*niovs)++;
+
+        if (!(desc->flags & VIRTIO_DESC_F_NEXT))
+            return true;
+
+        desc_idx = desc->next;
+    }
+
+    /* Descriptor chain loop or too long chain. */
+    virtio_net_set_fail(vnet);
+    return false;
+}
+
+static ssize_t vnet_handle_read(netdev_t *netdev,
+                                virtio_net_queue_t *queue,
+                                struct iovec *iovs,
+                                size_t niovs)
+{
+    switch (netdev->type) {
+    case NETDEV_IMPL_tap: {
+        net_tap_options_t *tap = (net_tap_options_t *) netdev->op;
+        ssize_t plen = readv(tap->tap_fd, iovs, niovs);
+
+        if (plen < 0 && (errno == EWOULDBLOCK || errno == EAGAIN)) {
+            queue->fd_ready = false;
+            return -1;
+        }
+
+        if (plen < 0) {
+            rv_log_error("virtio-net: could not read packet: %s",
+                         strerror(errno));
+            return -1;
+        }
+
+        return plen;
+    }
+    default:
+        return -1;
+    }
+}
+
+static ssize_t vnet_handle_write(netdev_t *netdev,
+                                 virtio_net_queue_t *queue,
+                                 struct iovec *iovs,
+                                 size_t niovs)
+{
+    switch (netdev->type) {
+    case NETDEV_IMPL_tap: {
+        net_tap_options_t *tap = (net_tap_options_t *) netdev->op;
+        ssize_t plen = writev(tap->tap_fd, iovs, niovs);
+
+        if (plen < 0 && (errno == EWOULDBLOCK || errno == EAGAIN)) {
+            queue->fd_ready = false;
+            return -1;
+        }
+
+        if (plen < 0) {
+            rv_log_error("virtio-net: could not write packet: %s",
+                         strerror(errno));
+            return -1;
+        }
+
+        return plen;
+    }
+    default:
+        return -1;
+    }
+}
+
+static bool vnet_get_avail_idx(virtio_net_state_t *vnet,
+                               virtio_net_queue_t *queue,
+                               uint16_t queue_idx,
+                               uint16_t *buffer_idx)
+{
+    uint32_t avail_addr = queue->queue_avail + 1 + queue_idx / 2;
+
+    if (!vnet_check_word_range(vnet, avail_addr, 1))
+        return false;
+
+    *buffer_idx = vnet->ram[avail_addr] >> (16 * (queue_idx % 2));
+
+    return true;
+}
+
+static bool vnet_put_used_elem(virtio_net_state_t *vnet,
+                               virtio_net_queue_t *queue,
+                               uint16_t new_used,
+                               uint16_t buffer_idx,
+                               uint32_t len)
+{
+    uint32_t used_addr =
+        queue->queue_used + 1 + (new_used % queue->queue_num) * 2;
+
+    if (!vnet_check_word_range(vnet, used_addr, 2))
+        return false;
+
+    vnet->ram[used_addr] = buffer_idx;
+    vnet->ram[used_addr + 1] = len;
+
+    return true;
+}
+
+static void virtio_net_try_rx(virtio_net_state_t *vnet)
+{
+    uint32_t *ram = vnet->ram;
+    virtio_net_queue_t *queue = &vnet->queues[VNET_QUEUE_RX];
+
+    if (vnet->status & VIRTIO_STATUS_DEVICE_NEEDS_RESET)
+        return;
+
+    if (!queue->fd_ready)
+        return;
+
+    if (!((vnet->status & VIRTIO_STATUS_DRIVER_OK) && queue->ready))
+        return virtio_net_set_fail(vnet);
+
+    if (queue->queue_num == 0)
+        return virtio_net_set_fail(vnet);
+
+    if (!vnet_check_word_range(vnet, queue->queue_avail, 1))
+        return;
+    if (!vnet_check_word_range(vnet, queue->queue_used, 1))
+        return;
+
+    uint16_t new_avail = ram[queue->queue_avail] >> 16;
+    if (new_avail - queue->last_avail > (uint16_t) queue->queue_num) {
+        rv_log_error("virtio-net: RX available ring size check failed");
+        return virtio_net_set_fail(vnet);
+    }
+
+    if (queue->last_avail == new_avail)
+        return;
+
+    uint16_t new_used = ram[queue->queue_used] >> 16;
+
+    while (queue->last_avail != new_avail) {
+        uint16_t queue_idx = queue->last_avail % queue->queue_num;
+        uint16_t buffer_idx;
+
+        if (!vnet_get_avail_idx(vnet, queue, queue_idx, &buffer_idx))
+            return;
+
+        struct iovec iovs[VNET_QUEUE_NUM_MAX];
+        size_t niovs = 0;
+
+        if (!vnet_build_iovs(vnet, queue, buffer_idx, true, iovs, &niovs))
+            return;
+
+        struct iovec *cursor = iovs;
+        size_t ncursor = niovs;
+        uint8_t virtio_header[VNET_HDR_SIZE] = {0};
+
+        /* num_buffers = 1, matching semu's simple RX path. */
+        virtio_header[10] = 1;
+
+        if (vnet_iovec_write(&cursor, &ncursor, virtio_header,
+                             sizeof(virtio_header))) {
+            rv_log_error("virtio-net: RX buffer too small for virtio header");
+            return virtio_net_set_fail(vnet);
+        }
+
+        ssize_t plen = vnet_handle_read(&vnet->peer, queue, cursor, ncursor);
+        if (plen < 0)
+            break;
+
+        if (!vnet_put_used_elem(vnet, queue, new_used, buffer_idx,
+                                (uint32_t) plen + sizeof(virtio_header)))
+            return;
+
+        queue->last_avail++;
+        new_used++;
+    }
+
+    ram[queue->queue_used] &= MASK(16);
+    ram[queue->queue_used] |= ((uint32_t) new_used) << 16;
+
+    if (!(ram[queue->queue_avail] & 1))
+        vnet->interrupt_status |= VIRTIO_INT_USED_RING;
+}
+
+static void virtio_net_try_tx(virtio_net_state_t *vnet)
+{
+    uint32_t *ram = vnet->ram;
+    virtio_net_queue_t *queue = &vnet->queues[VNET_QUEUE_TX];
+
+    if (vnet->status & VIRTIO_STATUS_DEVICE_NEEDS_RESET)
+        return;
+
+    if (!queue->fd_ready)
+        return;
+
+    if (!((vnet->status & VIRTIO_STATUS_DRIVER_OK) && queue->ready))
+        return virtio_net_set_fail(vnet);
+
+    if (queue->queue_num == 0)
+        return virtio_net_set_fail(vnet);
+
+    if (!vnet_check_word_range(vnet, queue->queue_avail, 1))
+        return;
+    if (!vnet_check_word_range(vnet, queue->queue_used, 1))
+        return;
+
+    uint16_t new_avail = ram[queue->queue_avail] >> 16;
+    if (new_avail - queue->last_avail > (uint16_t) queue->queue_num) {
+        rv_log_error("virtio-net: TX available ring size check failed");
+        return virtio_net_set_fail(vnet);
+    }
+
+    uint16_t new_used = ram[queue->queue_used] >> 16;
+
+    while (queue->last_avail != new_avail) {
+        uint16_t queue_idx = queue->last_avail % queue->queue_num;
+        uint16_t buffer_idx;
+
+        if (!vnet_get_avail_idx(vnet, queue, queue_idx, &buffer_idx))
+            return;
+
+        struct iovec iovs[VNET_QUEUE_NUM_MAX];
+        size_t niovs = 0;
+
+        if (!vnet_build_iovs(vnet, queue, buffer_idx, false, iovs, &niovs))
+            return;
+
+        struct iovec *cursor = iovs;
+        size_t ncursor = niovs;
+        uint8_t virtio_header[VNET_HDR_SIZE];
+
+        if (vnet_iovec_read(&cursor, &ncursor, virtio_header,
+                            sizeof(virtio_header))) {
+            rv_log_error("virtio-net: TX buffer too small for virtio header");
+            return virtio_net_set_fail(vnet);
+        }
+
+        ssize_t plen = vnet_handle_write(&vnet->peer, queue, cursor, ncursor);
+        if (plen < 0)
+            break;
+
+        if (!vnet_put_used_elem(vnet, queue, new_used, buffer_idx, 0))
+            return;
+
+        queue->last_avail++;
+        new_used++;
+    }
+
+    ram[queue->queue_used] &= MASK(16);
+    ram[queue->queue_used] |= ((uint32_t) new_used) << 16;
+
+    if (!(ram[queue->queue_avail] & 1))
+        vnet->interrupt_status |= VIRTIO_INT_USED_RING;
+}
+
+void virtio_net_refresh_queue(virtio_net_state_t *vnet)
+{
+    if (!vnet || !vnet->peer.op)
+        return;
+
+    if (!(vnet->status & VIRTIO_STATUS_DRIVER_OK) ||
+        (vnet->status & VIRTIO_STATUS_DEVICE_NEEDS_RESET))
+        return;
+
+    switch (vnet->peer.type) {
+    case NETDEV_IMPL_tap: {
+        net_tap_options_t *tap = (net_tap_options_t *) vnet->peer.op;
+        struct pollfd pfd = {
+            .fd = tap->tap_fd,
+            .events = POLLIN | POLLOUT,
+        };
+
+        poll(&pfd, 1, 0);
+
+        if (pfd.revents & POLLIN) {
+            vnet->queues[VNET_QUEUE_RX].fd_ready = true;
+            virtio_net_try_rx(vnet);
+        }
+
+        if (pfd.revents & POLLOUT) {
+            vnet->queues[VNET_QUEUE_TX].fd_ready = true;
+            virtio_net_try_tx(vnet);
+        }
+
+        break;
+    }
+    default:
+        break;
+    }
+}
+
+static uint32_t virtio_net_config_read(virtio_net_state_t *vnet, uint32_t addr)
+{
+    virtio_net_config_t *cfg = (virtio_net_config_t *) vnet->priv;
+    uint32_t offset = (addr - VIRTIO_Config) * sizeof(uint32_t);
+    uint32_t value = 0;
+
+    if (offset >= sizeof(*cfg))
+        return 0;
+
+    memcpy(&value, (uint8_t *) cfg + offset,
+           vnet_min(sizeof(value), sizeof(*cfg) - offset));
+
+    return value;
+}
+
+static void virtio_net_config_write(virtio_net_state_t *vnet,
+                                    uint32_t addr,
+                                    uint32_t value)
+{
+    virtio_net_config_t *cfg = (virtio_net_config_t *) vnet->priv;
+    uint32_t offset = (addr - VIRTIO_Config) * sizeof(uint32_t);
+
+    if (offset >= sizeof(*cfg))
+        return;
+
+    memcpy((uint8_t *) cfg + offset, &value,
+           vnet_min(sizeof(value), sizeof(*cfg) - offset));
+}
+
+uint32_t virtio_net_read(virtio_net_state_t *vnet, uint32_t addr)
+{
+    addr = addr >> 2;
+
+#define _(reg) VIRTIO_##reg
+    switch (addr) {
+    case _(MagicValue):
+        return VIRTIO_MAGIC_NUMBER;
+    case _(Version):
+        return VIRTIO_VERSION;
+    case _(DeviceID):
+        return VIRTIO_NET_DEV_ID;
+    case _(VendorID):
+        return VIRTIO_VENDOR_ID;
+    case _(DeviceFeatures):
+        return vnet->device_features_sel == 0
+                   ? VNET_FEATURES_0 | vnet->device_features
+                   : (vnet->device_features_sel == 1 ? VNET_FEATURES_1 : 0);
+    case _(QueueNumMax):
+        return VNET_QUEUE_NUM_MAX;
+    case _(QueueReady):
+        return (uint32_t) VNET_QUEUE.ready;
+    case _(InterruptStatus):
+        return vnet->interrupt_status;
+    case _(Status):
+        return vnet->status;
+    case _(ConfigGeneration):
+        return VIRTIO_CONFIG_GENERATE;
+    default:
+        if (addr >= _(Config))
+            return virtio_net_config_read(vnet, addr);
+        return 0;
+    }
+#undef _
+}
+
+void virtio_net_write(virtio_net_state_t *vnet, uint32_t addr, uint32_t value)
+{
+    addr = addr >> 2;
+
+#define _(reg) VIRTIO_##reg
+    switch (addr) {
+    case _(DeviceFeaturesSel):
+        vnet->device_features_sel = value;
+        break;
+    case _(DriverFeatures):
+        if (vnet->driver_features_sel == 0)
+            vnet->driver_features = value;
+        break;
+    case _(DriverFeaturesSel):
+        vnet->driver_features_sel = value;
+        break;
+    case _(QueueSel):
+        if (value < VNET_QUEUE_COUNT)
+            vnet->queue_sel = value;
+        else
+            virtio_net_set_fail(vnet);
+        break;
+    case _(QueueNum):
+        if (value > 0 && value <= VNET_QUEUE_NUM_MAX)
+            VNET_QUEUE.queue_num = value;
+        else
+            virtio_net_set_fail(vnet);
+        break;
+    case _(QueueReady):
+        VNET_QUEUE.ready = value & 1;
+        if (value & 1) {
+            if (!vnet_check_word_range(vnet, VNET_QUEUE.queue_avail, 1))
+                return;
+            VNET_QUEUE.last_avail = vnet->ram[VNET_QUEUE.queue_avail] >> 16;
+            VNET_QUEUE.fd_ready = true;
+        }
+        break;
+    case _(QueueDescLow):
+        VNET_QUEUE.queue_desc = vnet_preprocess(vnet, value);
+        break;
+    case _(QueueDescHigh):
+        if (value)
+            virtio_net_set_fail(vnet);
+        break;
+    case _(QueueDriverLow):
+        VNET_QUEUE.queue_avail = vnet_preprocess(vnet, value);
+        break;
+    case _(QueueDriverHigh):
+        if (value)
+            virtio_net_set_fail(vnet);
+        break;
+    case _(QueueDeviceLow):
+        VNET_QUEUE.queue_used = vnet_preprocess(vnet, value);
+        break;
+    case _(QueueDeviceHigh):
+        if (value)
+            virtio_net_set_fail(vnet);
+        break;
+    case _(QueueNotify):
+        switch (value) {
+        case VNET_QUEUE_RX:
+            virtio_net_try_rx(vnet);
+            break;
+        case VNET_QUEUE_TX:
+            virtio_net_try_tx(vnet);
+            break;
+        default:
+            virtio_net_set_fail(vnet);
+            break;
+        }
+        break;
+    case _(InterruptACK):
+        vnet->interrupt_status &= ~value;
+        break;
+    case _(Status):
+        virtio_net_update_status(vnet, value);
+        break;
+    default:
+        if (addr >= _(Config))
+            virtio_net_config_write(vnet, addr, value);
+        break;
+    }
+#undef _
+}
+
+bool virtio_net_init(virtio_net_state_t *vnet, const char *net_type)
+{
+    if (!netdev_init(&vnet->peer, net_type)) {
+        rv_log_error("virtio-net: failed to initialize net backend: %s",
+                     net_type ? net_type : "(null)");
+        return false;
+    }
+
+    return true;
+}
+
+virtio_net_state_t *vnet_new(void)
+{
+    virtio_net_state_t *vnet = calloc(1, sizeof(*vnet));
+    assert(vnet);
+
+    virtio_net_config_t *cfg = calloc(1, sizeof(*cfg));
+    assert(cfg);
+
+    cfg->mac[0] = 0x52;
+    cfg->mac[1] = 0x54;
+    cfg->mac[2] = 0x00;
+    cfg->mac[3] = 0x12;
+    cfg->mac[4] = 0x34;
+    cfg->mac[5] = 0x56;
+    cfg->status = VNET_LINK_UP;
+    cfg->max_virtqueue_pairs = 1;
+    cfg->mtu = 1500;
+
+    vnet->priv = cfg;
+    vnet->queues[VNET_QUEUE_RX].fd_ready = true;
+    vnet->queues[VNET_QUEUE_TX].fd_ready = true;
+
+    return vnet;
+}
+
+void vnet_delete(virtio_net_state_t *vnet)
+{
+    if (!vnet)
+        return;
+
+    netdev_delete(&vnet->peer);
+    free(vnet->priv);
+    free(vnet);
+}
--- a/src/devices/virtio.h
+++ b/src/devices/virtio.h
@@ -5,6 +5,11 @@
 
 #pragma once
 
+#include <stdbool.h>
+#include <stdint.h>
+
+#include "netdev.h"
+
 #define VIRTIO_VENDOR_ID 0x12345678
 #define VIRTIO_MAGIC_NUMBER 0x74726976
 #define VIRTIO_VERSION 2
@@ -37,6 +42,9 @@
 #define VIRTIO_BLK_F_RO (1 << 5)
 
 #define VIRTIO_RNG_DEV_ID 4
+
+#define VIRTIO_NET_DEV_ID 1
+#define IRQ_VNET_BIT(base) (1 << (base))
 
 /* VirtIO MMIO registers */
 #define VIRTIO_REG_LIST                  \
@@ -157,3 +165,50 @@ bool virtio_rng_init(virtio_rng_state_t *vrng);
 virtio_rng_state_t *vrng_new(void);
 
 void vrng_delete(virtio_rng_state_t *vrng);
+
+typedef struct {
+    uint32_t queue_num;
+    uint32_t queue_desc;
+    uint32_t queue_avail;
+    uint32_t queue_used;
+    uint16_t last_avail;
+    bool ready;
+    bool fd_ready;
+} virtio_net_queue_t;
+
+typedef struct {
+    /* feature negotiation */
+    uint32_t device_features;
+    uint32_t device_features_sel;
+    uint32_t driver_features;
+    uint32_t driver_features_sel;
+
+    /* queue config */
+    uint32_t queue_sel;
+    virtio_net_queue_t queues[2];
+
+    /* status */
+    uint32_t status;
+    uint32_t interrupt_status;
+
+    /* supplied by environment */
+    uint32_t *ram;
+
+    /* host network backend */
+    netdev_t peer;
+
+    /* implementation-specific */
+    void *priv;
+} virtio_net_state_t;
+
+uint32_t virtio_net_read(virtio_net_state_t *vnet, uint32_t addr);
+
+void virtio_net_write(virtio_net_state_t *vnet, uint32_t addr, uint32_t value);
+
+bool virtio_net_init(virtio_net_state_t *vnet, const char *net_type);
+
+virtio_net_state_t *vnet_new(void);
+
+void vnet_delete(virtio_net_state_t *vnet);
+
+void virtio_net_refresh_queue(virtio_net_state_t *vnet);
--- a/src/main.c
+++ b/src/main.c
@@ -69,6 +69,7 @@ static int opt_virtio_blk_idx = 0;
 
 /* enable virtio-rng device */
 static bool opt_virtio_rng = false;
+static char *opt_virtio_net_backend;
 #endif
 
 static void reset_getopt_state(void)
@@ -111,6 +112,7 @@ static void reset_runtime_options(void)
     memset(opt_virtio_blk_img, 0, sizeof(opt_virtio_blk_img));
     opt_virtio_blk_idx = 0;
     opt_virtio_rng = false;
+    opt_virtio_net_backend = NULL;
 #endif
 
     reset_getopt_state();
@@ -137,6 +139,7 @@ static void print_usage(const char *filename)
         "(default read and write). This option may be specified "
         "multiple times for multiple block devices\n"
         "  -x vrng : enable virtio-rng device\n"
+        "  -x vnet:<backend>: use <backend> as virtio-net backend interface \n"
         "  -b <bootargs> : use customized <bootargs> for the kernel\n"
 #endif
         "  -d [filename]: dump registers as JSON to the "
@@ -192,6 +195,12 @@ static bool parse_args(int argc, char **args)
                 }
                 opt_virtio_blk_img[opt_virtio_blk_idx++] =
                     optarg + 5; /* strlen("vblk:") */
+            } else if (!strncmp("vnet:", optarg, 5)) {
+                if (!optarg[5]) {
+                    rv_log_error("Missing virtio-net backend interface.\n");
+                    return false;
+                }
+                opt_virtio_net_backend = optarg + 5; /* strlen("vnet:") */
             } else if (!strcmp("vrng", optarg)) {
                 opt_virtio_rng = true;
             } else {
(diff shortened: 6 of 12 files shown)


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: 103be25e484c9a69706e44ea4666ff167c35b7d9 | Previous: c48ad2aca1d64def3382207a235cbb991343982f | Ratio |
|-|-|-|-|
| `Dhrystone` | `1557.667` DMIPS | `1550.333` DMIPS | `1.00` |
| `CoreMark` | `1117.948` iterations/sec | `1062.655` iterations/sec | `0.95` |

</details>

shengwen-tw:
Hi @Charlie-Tsai1123,

I’m not the original author of the virtio-net device. Please check the commit history via `git log --follow -- virtio-net.c`; I believe the original author is @jserv.

Charlie-Tsai1123:
Thanks for the clarification, and sorry for the incorrect attribution. 
I checked the semu history with `git log --follow -- virtio-net.c`. The original virtio-net implementation appears to have been introduced by jserv. I will update the commit message accordingly.

jserv (changes requested):
Refine CI pipeline to validate virtio-net.

jserv:
> Thanks for the clarification, and sorry for the incorrect attribution. I checked the semu history with `git log --follow -- virtio-net.c`. The original virtio-net implementation appears to have been introduced by jserv. I will update the commit message accordingly.

Remove "This implementation is based on semu's virtio-net device, originally introduced by Jserv" as you already append "Co-authored-by: Jim Huang".

jserv (changes requested):
Rework semu's network infrastructure, as described in [networking.md](https://github.com/sysprog21/semu/blob/master/docs/networking.md):
- Linux: TAP (kernel-level) and user-mode (SLIRP) networking
- macOS: vmnet.framework (kernel-level NAT; bridge mode planned) and user-mode (SLIRP) networking

For this pull request, both TAP and SLIRP should be landed.

.ci/test-netdev.sh
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env bash
```
ChinYikMing: Could you reuse the `boot-linux.sh` and `boot-linux-prepare.sh`? Specifically, boot the guest linux and run the virtio net related validation like virtio block.

src/main.c
```diff
@@ -137,6 +139,7 @@ static void print_usage(const char *filename)
         "(default read and write). This option may be specified "
         "multiple times for multiple block devices\n"
         "  -x vrng : enable virtio-rng device\n"
+        "  -x vnet:<backend>: use <backend> as virtio-net backend interface \n"
```
ChinYikMing: Remove the trailing space.

src/main.c
```diff
@@ -192,6 +195,12 @@ static bool parse_args(int argc, char **args)
                 }
                 opt_virtio_blk_img[opt_virtio_blk_idx++] =
                     optarg + 5; /* strlen("vblk:") */
+            } else if (!strncmp("vnet:", optarg, 5)) {
```
ChinYikMing, line 233: Do we need to limit the `vnet` cli option passed by the user like the `vblk`?

src/riscv.c
```diff
@@ -1070,8 +1121,14 @@ void rv_run(riscv_t *rv)
         emscripten_set_main_loop_arg(rv_step, (void *) rv, 0, 1);
 #else
         /* default main loop */
-        for (; !rv_has_halted(rv);) /* run until the flag is done */
-            rv_step(rv);            /* step instructions */
+        for (; !rv_has_halted(rv);) { /* run until the flag is done */
+            rv_step(rv);              /* step instructions */
+
+#if RV32_HAS(SYSTEM_MMIO)
```
ChinYikMing: Can we do the check in `rv_step`?

src/riscv.c
```diff
@@ -501,6 +504,34 @@ static void load_dtb(char **ram_loc, vm_attr_t *attr)
...
+            assert(fdt_setprop_string(dtb_buf, subnode, "compatible",
+                                      "virtio,mmio") == 0);
+
+            uint32_t reg[2] = {cpu_to_fdt32(new_addr), cpu_to_fdt32(size)};
+            assert(fdt_setprop(dtb_buf, subnode, "reg", reg, sizeof(reg)) == 0);
+
+            uint32_t irq = cpu_to_fdt32(new_irq);
+            assert(fdt_setprop(dtb_buf, subnode, "interrupts", &irq,
+                               sizeof(irq)) == 0);
+            dev_idx++;
```
ChinYikMing: Is this necessary since no more devices?

.ci/test-netdev.sh
ChinYikMing: For modularity, agree to create a new file for maintainance just like `rtc.sh`.

src/main.c
Charlie-Tsai1123: Fixed. Thanks for reminder.

src/main.c
Charlie-Tsai1123: I added validation for the vnet option.
At this stage, only one virtio-net device is supported, so repeated vnet options are rejected. The backend is also validated and currently restricted to tap. The user backend will be enabled in the follow-up SLIRP commit.

src/riscv.c
Charlie-Tsai1123: Fixed. The final dev_idx++ after adding the virtio-net DTB node is unnecessary.

.ci/test-netdev.sh
Charlie-Tsai1123: Done. The workflow now runs the virtio-net tests through .ci/boot-linux.sh by setting VNET_BACKEND. This allows the virtio-net checks to reuse the existing boot-linux.sh and boot-linux-prepare.sh flow.

src/riscv.c
Charlie-Tsai1123: I moved the virtio-net refresh and interrupt propagation into the per-step execution path because the current backend uses non-blocking polling to receive packets from the host side. I plan to improve this in a follow-up with an event-driven wakeup path, eg. using eventfd.

Charlie-Tsai1123:
This update mainly covers two parts:

1. virtio-net interrupt delivery
Related commit: cf424445fc33a9212248c03ad6f52a8796c3ad58

I moved virtio-net refresh and interrupt propagation into the per-step execution path. After queue refresh, the virtio-net interrupt state is pushed to the PLIC. This avoids delaying completed RX/TX work until a later interrupt update point.

I also updated virtio_net_try_rx() and virtio_net_try_tx() so they only raise used-ring interrupts when the used ring index actually advances. This avoids repeated interrupts when no virtqueue progress was made, and fixes the observed TAP ping latency spike.

Before this change, TAP ping could show delayed replies such as:
```
PING 192.168.100.1 (192.168.100.1): 56 data bytes
64 bytes from 192.168.100.1: seq=0 ttl=64 time=3.831 ms
64 bytes from 192.168.100.1: seq=1 ttl=64 time=1001.088 ms
64 bytes from 192.168.100.1: seq=2 ttl=64 time=1.389 ms
```
After the change, replies are delivered promptly:
```
PING 192.168.100.1 (192.168.100.1): 56 data bytes
64 bytes from 192.168.100.1: seq=0 ttl=64 time=1.181 ms
64 bytes from 192.168.100.1: seq=1 ttl=64 time=0.546 ms
64 bytes from 192.168.100.1: seq=2 ttl=64 time=0.565 ms
```
2. user-mode SLIRP backend
Related commit: 103be25e484c9a69706e44ea4666ff167c35b7d9

I added a minislirp-based user-mode backend for virtio-net, adapted from semu. The backend connects the virtio-net RX/TX paths with libslirp through non-blocking socketpairs, so guest networking can work without TAP, root rivileges, or host network setup.

The current backend support is:

- Linux : vnet:tap and vnet:user
- macOS : vnet:user
- Emscripten: no tap, no user, no vmnet

Emscripten is handled separately because rv32emu supports emcc builds, unlike semu's original networking setup. For emcc, the networking backends are disabled and unsupported vnet backends are rejected during argument parsing.

CI coverage is arranged as follows:

- Linux x64 interpreter matrix runs both VNET_BACKEND=user and VNET_BACKEND=tap.
- macOS runs the user-mode SLIRP boot test.
- TAP is only tested on Linux because it depends on the Linux TAP/TUN interface.
- virtio-net tests are not run for every JIT/T2C/MOP-fusion matrix entry, only run in the interpreter matrix.

Future PR may work:

- vmnet.framework backend:
  I looked at semu's vmnet implementation. It uses Apple's vmnet.framework and Blocks syntax for callbacks. That path is suitable for Clang with Blocks support, but it is not suitable for the current macOS gcc-15 CI job. vmnet also has macOS-specific privilege / entitlement requirements. A follow-up could add vmnet through a clang-only macOS path or a more portable wrapper.

- Event-driven wakeup:
  The current backend follows rv32emu's existing polling-based device refresh model and uses non-blocking I/O. This keeps the initial migration simple and consistent with the current virtio-net execution path. In a follow-up, this can be improved with an event-driven wakeup mechanism, such as  eventfd on Linux or a pipe/condvar-style notification path, so host-side activity can wake the emulator loop instead of relying on frequent polling.

- Emscripten support for virtio-net

src/devices/slirp.c
```diff
@@ -0,0 +1,399 @@
...
+                              slirp_add_poll_socket, usr);
+
+    if (usr->host_to_guest_channel[SLIRP_READ_SIDE] >= 0)
+        slirp_add_poll_socket(usr->host_to_guest_channel[SLIRP_READ_SIDE],
+                              SLIRP_POLL_IN | SLIRP_POLL_HUP, usr);
+
+    if (usr->pfd_len == 0)
+        return 0;
+
+    int ret = poll(usr->pfd, (nfds_t) usr->pfd_len, 0);
```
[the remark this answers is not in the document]
Charlie-Tsai1123: The current implementation uses non-blocking I/O and polls SLIRP from the device refresh path, which is simple and keeps the migration consistent with the existing execution loop. I plan to address this in a follow-up PR with a proper event-driven wakeup path.

src/devices/netdev.h
```diff
@@ -0,0 +1,91 @@
...
+#else
+#define RV32EMU_NET_HAS_SLIRP 0
+#endif
+
+#define RV32EMU_NET_HAS_VMNET 0
+
+typedef struct netdev netdev_t;
+
+typedef enum {
+    NETDEV_IMPL_none = 0,
```
ChinYikMing: Prefer all in upper case: `NETDEV_IMPL_NONE`.

src/devices/netdev.h
```diff
@@ -0,0 +1,91 @@
...
+#if RV32EMU_NET_HAS_TAP
+    NETDEV_IMPL_tap,
```
ChinYikMing: Ditto.

src/devices/netdev.h
```diff
@@ -0,0 +1,91 @@
...
+#endif
+#if RV32EMU_NET_HAS_SLIRP
+    NETDEV_IMPL_user,
```
ChinYikMing: Ditto.

ChinYikMing (commented):
Please help adding backend selection into Kconfig. This can reduce the binary size of rv32emu exe if user might not need the functionality of vnet.

ChinYikMing:
Hi @Charlie-Tsai1123 , could you document how to test the user mode vnet backend as I only see the `tap` backend is listed at [here](https://github.com/sysprog21/rv32emu/pull/748#issue-4785530810)? It will be easier for reviewer to quick test . Thanks.

src/main.c
```diff
@@ -137,6 +139,7 @@ static void print_usage(const char *filename)
         "(default read and write). This option may be specified "
         "multiple times for multiple block devices\n"
         "  -x vrng : enable virtio-rng device\n"
+        "  -x vnet:<backend>: use <backend> as virtio-net backend interface\n"
```
ChinYikMing: Add possible vnet backend here.

E.g., `-x vnet:<backend>: use <backend> as virtio-net backend interface(supported backend: user/tap)\n`

But if vnet is togglable in Kconfig, this should be adjusted again.

src/devices/netdev.c
```diff
@@ -0,0 +1,204 @@
...
+#endif
+
+#if defined(__APPLE__) && RV32EMU_NET_HAS_SLIRP
+static const char *netdev_macos_default(void)
+{
+    return "user";
+}
+#endif
+
+bool netdev_init(netdev_t *netdev, const char *net_type)
```
ChinYikMing, line 129: Initialization for `NETDEV_IMPL_user` has duplicated logic. Please rewrite.

Something like this might be cleaner:
```c

/* Linux specify */
if defined(__linux__) && RV32EMU_NET_HAS_TAP
init_tap()
else /* user mode for macOS and Linux */
init_user()
endif
```

src/main.c
Charlie-Tsai1123: Done. I've updated the help message to include tap and user. I also made it dynamic based on Kconfig: the `-x vnet` help text will now only display the specific backends that are actually enabled in the current build configuration.

src/devices/netdev.c
Charlie-Tsai1123: Done. I write `netdev_setup` for duplicated logic.

Charlie-Tsai1123:
Thanks for the review. I have added build-time backend selection through Kconfig and documented how to test the user-mode backend.

# test user-mode
Build (Host):
```
make system_defconfig
make ENABLE_SYSTEM=1

build/rv32emu \
  -k build/linux-image/Image \
  -i build/linux-image/rootfs.cpio \
  -x vnet:user
```
Guest device verification (Guest):
```
readlink /sys/bus/virtio/devices/virtio0/driver
ip link set eth0 up
ip addr flush dev eth0
ip addr add 10.0.2.15/24 dev eth0
ip route add default via 10.0.2.2
ip addr show eth0
ip route
ping -c 3 -W 5 10.0.2.2
```
Expected result (Guest):
```
# readlink /sys/bus/virtio/devices/virtio0/driver
../../../../../bus/virtio/drivers/virtio_net
# ip link set eth0 up
# ip addr flush dev eth0
# ip addr add 10.0.2.15/24 dev eth0
# ip route add default via 10.0.2.2
# ip addr show eth0
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast qlen 1000
    link/ether c2:84:a2:01:23:bb brd ff:ff:ff:ff:ff:ff
    inet 10.0.2.15/24 scope global eth0
       valid_lft forever preferred_lft forever
# ip route
default via 10.0.2.2 dev eth0 
10.0.2.0/24 dev eth0 scope link  src 10.0.2.15 
# ping -c 3 -W 5 10.0.2.2
PING 10.0.2.2 (10.0.2.2): 56 data bytes
64 bytes from 10.0.2.2: seq=0 ttl=255 time=1.125 ms
64 bytes from 10.0.2.2: seq=1 ttl=255 time=0.526 ms
64 bytes from 10.0.2.2: seq=2 ttl=255 time=0.544 ms

--- 10.0.2.2 ping statistics ---
3 packets transmitted, 3 packets received, 0% packet loss
round-trip min/avg/max = 0.526/0.731/1.125 ms
```
or test with the existing automated user-mode network test (Host):
```
VNET_BACKEND=user bash .ci/boot-linux.sh
```

# Kconfig changes
The following options were added:
```
CONFIG_VIRTIO_NET
CONFIG_VIRTIO_NET_TAP
CONFIG_VIRTIO_NET_USER
```
Default `system_defconfig` still contain virtio-net setting, user could disable it by using `make config` and cancel it:
```
make config
```
select `Execution Modes`:
image

Under the system-emulation configuration, users can independently enable or disable:
image
Disabling `CONFIG_VIRTIO_NET` excludes  `virtio-net.o` `netdev.o`  `slirp.o` from the build:

When `CONFIG_VIRTIO_NET` is enabled but neither TAP nor USER is selected, virtio-net is also treated as unavailable and the network objects are excluded.

When user cancel virtio-net, the binary size of rv32emu exe decrease. I copy the build file and test

copy file after `make` for different config setting:

```
cp build/rv32emu /tmp/rv32emu-no-vnet
```
```
cp build/rv32emu /tmp/rv32emu-vnet-full
```
Then their size:
```
charlie-tsai:/media/charlie-tsai/DATA/road_to_master/rv32emu$stat -c '%n: %s bytes'     /tmp/rv32emu-vnet-full     /tmp/rv32emu-no-vnet
/tmp/rv32emu-vnet-full: 344416 bytes
/tmp/rv32emu-no-vnet: 229560 bytes
```

ChinYikMing:
The boot linux CI is refactored in bda46496636cfaa3bd6ab9dcd0562ac70f22abe7 of #638, I would like to see the newly added virtio-net CI test rebase on that.

Charlie-Tsai1123:
Thanks for the suggestion. I rebased the virtio-net changes onto the latest version of PR #638 and updated the CI integration accordingly.

Previously, the virtio-net user-mode and TAP tests were defined as separate GitHub Actions steps. They are now integrated into `.ci/boot-linux.sh`, following the same structure as the RTC, reboot, and virtio-blk tests:

- virtio-net user-mode SLIRP is tested on Linux and macOS.
- virtio-net TAP is tested on Linux and skipped on macOS hosts.
- The separate virtio-net workflow steps were removed from `main.yml`  to avoid duplicate execution.
- The virtio-net boot timeout on macOS was extended because the default timeout was insufficient under the macOS GitHub runner load.

Previous:
image
Now:
image


All relevant build and boot tests passed. The only failing job is `docker-hub-build-and-publish`, which requires Docker Hub credentials that are not available for this temporary branch/fork, so it appears unrelated to these changes.

Temporary integration branch:
https://github.com/Charlie-Tsai1123/rv32emu/tree/test/virtio-net-on-pr638
CI test result:
https://github.com/Charlie-Tsai1123/rv32emu/actions/runs/30758331124/job/91524331778
Small modification after rebase:
https://github.com/Charlie-Tsai1123/rv32emu/commit/68e5e4b43127152f4149f252c3155fd9af4ade10

I have not updated the branch of PR #748 yet because PR #638 is still open.

jserv (changes requested):
Rebase latest `master` branch and resolve conflicts.

ChinYikMing:
Clean build has parallel build regression. @Charlie-Tsai1123 Please fix this.

### Reproduce
```
git clone https://github.com/Charlie-Tsai1123/rv32emu.git -b research/virtio-net

git branch
* research/virtio-net

make system_defconfig
Downloading Kconfig tools...
make -j8Kconfig tools installed to tools/kconfig
Loaded configuration 'configs/defconfig'
Configuration saved to '.config'
Generated .config from default configuration
Applying configs/system_defconfig...
Loaded configuration 'configs/system_defconfig'
Configuration saved to '.config'
Configuration applied.

make -j8
 DTC	build/minimal.dtb
  CC	build/devices/netdev.o
  CC	build/devices/rtc.o
Submodule 'softfloat' (https://github.com/ucb-bar/berkeley-softfloat-3) registered for path 'src/softfloat'
  CC	build/devices/virtio-blk.o
Cloning into '/usr/local/google/home/yikming/dev/review/rv32emu/src/softfloat'...
  CC	build/devices/virtio-net.o
Submodule 'src/minislirp' (https://github.com/sysprog21/minislirp) registered for path 'src/minislirp'
  CC	build/devices/virtio-rng.o
  BIN2C	src/minimal_dtb.h
Cloning into '/usr/local/google/home/yikming/dev/review/rv32emu/src/minislirp'...
In file included from src/decode.h:12,
                 from src/riscv_private.h:22,
                 from src/system.h:14,
                 from src/devices/virtio-net.c:16:
src/riscv.h:30:10: fatal error: softfloat/source/include/softfloat.h: No such file or directory
   30 | #include "softfloat/source/include/softfloat.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make: *** [mk/system.mk:52: build/devices/virtio-net.o] Error 1
make: *** Waiting for unfinished jobs....
Submodule path 'src/minislirp': checked out '1a307c8f18e9eb1c8c12c5eeaa1f76e66a27aad2'
remote: Enumerating objects: 2518, done.
remote: Counting objects: 100% (2518/2518), done.
remote: Compressing objects: 100% (225/225), done.
remote: Total 2201 (delta 2072), reused 2093 (delta 1974), pack-reused 0 (from 0)
Receiving objects: 100% (2201/2201), 573.37 KiB | 2.82 MiB/s, done.
Resolving deltas: 100% (2072/2072), completed with 85 local objects.
From https://github.com/ucb-bar/berkeley-softfloat-3
 * branch            3b70b5d8147675932c38b36cd09af6df4eedd919 -> FETCH_HEAD
Submodule path 'src/softfloat': checked out '3b70b5d8147675932c38b36cd09af6df4eedd919'
```

src/devices/netdev.c
```diff
@@ -0,0 +1,188 @@
...
+    ifreq.ifr_name[sizeof(ifreq.ifr_name) - 1] = '\0';
+
+    if (ioctl(tap->tap_fd, TUNSETIFF, &ifreq) < 0) {
+        rv_log_error("failed to allocate TAP device: %s", strerror(errno));
+        close(tap->tap_fd);
+        tap->tap_fd = -1;
+        return -1;
+    }
+
+    rv_log_info("allocated TAP interface: %s", ifreq.ifr_name);
```
ChinYikMing: The default log level is `WARN`, this line will not show by default. Use `rv_log_warn` instead.

https://github.com/sysprog21/rv32emu/blob/045d031f319020cce48e5f3072db77b7520196e0/src/main.c#L378

ChinYikMing:
The upstream has merged reboot fix (#638), please help to extend the `rv_cold_reboot` to reset the virtio-net device state when reboot if virtio-net device is enabled. 
https://github.com/sysprog21/rv32emu/blob/045d031f319020cce48e5f3072db77b7520196e0/src/riscv.c#L1258

You can take virtio block as reference https://github.com/sysprog21/rv32emu/blob/045d031f319020cce48e5f3072db77b7520196e0/src/riscv.c#L1495

Charlie-Tsai1123:
> The upstream has merged reboot fix (#638), please help to extend the `rv_cold_reboot` to reset the virtio-net device state when reboot if virtio-net device is enabled.
> 
> https://github.com/sysprog21/rv32emu/blob/045d031f319020cce48e5f3072db77b7520196e0/src/riscv.c#L1258
> 
> You can take virtio block as reference
> 
> https://github.com/sysprog21/rv32emu/blob/045d031f319020cce48e5f3072db77b7520196e0/src/riscv.c#L1495

Fixed. I also verified the reboot behavior manually. I temporarily added `rv_log_warn("virtio-net backend initialized");` in `virtio_net_init()` and confirmed that it was printed only during the initial boot and not after the cold reboot, showing that the backend was reused instead of being re-created. The temporary log was removed after verification.

jserv (changes requested):
Rebase latest `master` branch and rework commits, making each of them comply with https://chris.beams.io/git-commit while `CI:` is the only tag allowed.

ChinYikMing:
@Charlie-Tsai1123 

If the virtio network device is enabled without selecting a backend, the build still completes successfully. Is this the expected behavior? I think require the user (show some error message) to select one of the backend might be more appropriate.

image

Charlie-Tsai1123:
> @Charlie-Tsai1123
> 
> If the virtio network device is enabled without selecting a backend, the build still completes successfully. Is this the expected behavior? I think require the user (show some error message) to select one of the backend might be more appropriate.
> 
> image

Currently, if virtio-net is enabled without selecting any backend, the build silently excludes the virtio-net device. I can change this behavior to report an error and require at least one backend to be selected.

src/devices/virtio-net.c
```diff
@@ -0,0 +1,803 @@
...
+    if (queue->queue_num == 0)
+        return virtio_net_set_fail(vnet);
+
+    if (!vnet_check_word_range(vnet, queue->queue_avail, 1))
+        return;
+    if (!vnet_check_word_range(vnet, queue->queue_used, 1))
+        return;
+
+    uint16_t new_avail = ram[queue->queue_avail] >> 16;
+    if (new_avail - queue->last_avail > (uint16_t) queue->queue_num) {
```
jserv: new_avail and last_avail both promote to int here, so a backwards jump (last_avail 100, new_avail 50) yields a negative difference that passes the check, and the loop below then iterates about 65,000 times, issuing a syscall and retiring a descriptor per pass on ring entries the driver never filled. Compute the pending count in uint16_t and test that against queue_num. The TX copy at line 476 and the existing virtio-blk/virtio-rng checks have the same shape, so the follow-up fix belongs in all of them.

```suggestion
    uint16_t pending = new_avail - queue->last_avail;
    if (pending > (uint16_t) queue->queue_num) {
```

src/devices/virtio-net.c
```diff
@@ -0,0 +1,803 @@
...
+        virtio_header[10] = 1;
+
+        if (vnet_iovec_write(&cursor, &ncursor, virtio_header,
+                             sizeof(virtio_header))) {
+            rv_log_error("virtio-net: RX buffer too small for virtio header");
+            return virtio_net_set_fail(vnet);
+        }
+
+        ssize_t plen = vnet_handle_read(&vnet->peer, queue, cursor, ncursor);
+        if (plen < 0)
```
jserv: vnet_handle_read returns 0 for a genuine zero-length result and also when readv is handed an empty iovec list, which happens when the RX chain holds only the 12 header bytes and ncursor drops to 0. Only negative values break out, so the descriptor is retired with a used length of 12 and the loop keeps going until every posted RX buffer is consumed. Treat non-positive as no packet.

```suggestion
        if (plen <= 0)
```

src/devices/virtio-net.c
```diff
@@ -0,0 +1,803 @@
...
+        uint8_t virtio_header[VNET_HDR_SIZE];
+
+        if (vnet_iovec_read(&cursor, &ncursor, virtio_header,
+                            sizeof(virtio_header))) {
+            rv_log_error("virtio-net: TX buffer too small for virtio header");
+            return virtio_net_set_fail(vnet);
+        }
+
+        ssize_t plen = vnet_handle_write(&vnet->peer, queue, cursor, ncursor);
+        if (plen < 0)
```
jserv: A write error that is not EAGAIN never retires the descriptor: last_avail stays put, so the next refresh rebuilds the same iovecs and repeats the failing writev, one rv_log_error per emulator step for as long as the guest leaves it queued. EMSGSIZE from an oversized TX chain reaches this path (nothing bounds the chain total against the 1500 byte MTU), as does EINTR, which the EAGAIN test above does not cover. Drop the frame and complete the descriptor on a permanent error, and retry only for EAGAIN, EWOULDBLOCK, and EINTR.

src/devices/virtio-net.c
```diff
@@ -0,0 +1,803 @@
...
+    case _(Version):
+        return VIRTIO_VERSION;
+    case _(DeviceID):
+        return VIRTIO_NET_DEV_ID;
+    case _(VendorID):
+        return VIRTIO_VENDOR_ID;
+    case _(DeviceFeatures):
+        return vnet->device_features_sel == 0
+                   ? VNET_FEATURES_0 | vnet->device_features
+                   : (vnet->device_features_sel == 1 ? VNET_FEATURES_1 : 0);
```
jserv, line 674: Only VIRTIO_F_VERSION_1 is offered, so VIRTIO_NET_F_MAC, _STATUS, and _MTU are absent and the whole virtio_net_config_t built in vnet_new is dead: the guest ignores 52:54:00:12:34:56 and calls eth_hw_addr_random() instead. The test output in the description shows aa:e0:56:5b:58:bc, and that address changes on every boot, which breaks static ARP entries and DHCP reservations against the TAP gateway. Advertise bit 5 (F_MAC) in VNET_FEATURES_0, plus bit 16 and bit 3 if status and mtu are meant to be read, or drop the unused fields.

src/riscv.c
```diff
@@ -764,8 +809,16 @@ void rv_run(riscv_t *rv)
...
 #else
         /* default main loop */
-        for (; !rv_has_halted(rv);) /* run until the flag is done */
-            rv_step(rv);            /* step instructions */
+        for (; !rv_has_halted(rv);) { /* run until the flag is done */
+            rv_step(rv);              /* step instructions */
+
+#if RV32_HAS(VIRTIO_NET)
+            if (attr->vnet) {
+                virtio_net_refresh_queue(attr->vnet);
```
jserv: This runs after every rv_step, which is 100 guest instructions by default, so the TAP path issues one poll syscall per 100 instructions and the SLIRP path issues four zero-timeout polls plus two slirp_pollfds_fill_socket walks. On an otherwise idle link that host syscall traffic dominates the emulator loop, and it is a plausible source of both the 1001 ms ping outlier in the description and the CI budget this PR needs (macOS 60 to 80 minutes, NETDEV_BOOT_TIMEOUT 900). Gate the refresh on a cycle counter so it fires every few thousand instructions instead.

src/devices/slirp.c
```diff
@@ -0,0 +1,404 @@
...
+
+    if (!timer)
+        return NULL;
+
+    timer->slirp = (Slirp *) usr->slirp;
+    timer->id = id;
+    timer->cb_opaque = cb_opaque;
+    timer->expire_time_ms = -1;
+
+    usr->timer = timer;
```
jserv: net_user_options_t holds one timer slot, so a second slirp_timer_new_opaque silently displaces the first, net_slirp_handle_timer never fires the displaced one, and net_slirp_timer_free leaves usr->timer pointing at the survivor. It works today only because minislirp arms just the IPv6 router advertisement timer, so a future timer would show up as silently dead rather than as a crash. Keep the timers on a list and walk it in net_slirp_handle_timer.

src/devices/slirp.c
```diff
@@ -0,0 +1,404 @@
...
+}
+
+static ssize_t net_slirp_send_packet(const void *buf, size_t len, void *opaque)
+{
+    net_user_options_t *usr = (net_user_options_t *) opaque;
+
+    if (!usr || usr->guest_to_host_channel[SLIRP_WRITE_SIDE] < 0)
+        return 0;
+
+    return write(usr->guest_to_host_channel[SLIRP_WRITE_SIDE], buf, len);
```
jserv: The two channel names are inverted with respect to the traffic they carry: guest_to_host_channel is written here by SLIRP and read by the guest RX path in vnet_handle_read, while host_to_guest_channel is written by the guest TX path in vnet_handle_write and drained into slirp_input by net_slirp_read. Reading the RX and TX handlers against these names costs a double take every time. Swap the two names so each describes its actual direction.

src/riscv.c
Charlie-Tsai1123: Fixed. Backend refresh is now gated by the guest cycle counter and runs every 5000 cycles.

src/devices/virtio-net.c
Charlie-Tsai1123: Fixed. The device now advertises `VIRTIO_NET_F_MAC`, `VIRTIO_NET_F_STATUS`, and `VIRTIO_NET_F_MTU`. I also fixed the device config access to preserve byte offsets before converting MMIO register addresses to word indices.

src/devices/virtio-net.c
Charlie-Tsai1123: Fixed. Retryable errors (`EAGAIN`, `EWOULDBLOCK`, and `EINTR`) now keep the descriptor pending, while permanent errors drop the packet and complete the descriptor so the TX queue can continue.

.ci/netdev.sh
```diff
@@ -0,0 +1,133 @@
...
+COLOR_R='\e[31;01m' # Red
+COLOR_Y='\e[33;01m' # Yellow
+COLOR_N='\e[0m'     # No color
+
+MESSAGES=("${COLOR_G}OK!"
+    "${COLOR_R}Fail to boot"
+    "${COLOR_R}Fail to login"
+    "${COLOR_R}Fail to run commands"
+    "${COLOR_R}Fail to ping gateway"
+)
```
jserv, line 52: Include `.ci/common.sh` as `.ci/virtio-blk.sh` does.

docs/networking.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Future work
+
+### vmnet.framework
+
+macOS vmnet.framework support is not implemented in rv32emu yet.
+
+semu has a vmnet backend, but that implementation uses Apple's vmnet.framework and Blocks syntax for callbacks. This is suitable for Clang with Blocks support, but it is not suitable for the current macOS gcc-15 CI job. vmnet also has macOS-specific privilege and entitlement requirements.
+
+A future implementation could add vmnet through a Clang-only macOS path or a more portable wrapper.
```
jserv, line 171: Implement this within this pull request.

Charlie-Tsai1123: Implemented in this pull request. Added a macOS vmnet.framework backend for VirtIO networking, with shared, host-only, and bridged modes implemented internally. The current vnet:vmnet frontend uses shared mode.
The backend is enabled for macOS Clang builds because vmnet.framework uses Apple Blocks-based callbacks, while GCC builds keep the existing networking backends.
I also added a macOS vmnet boot test, and the full CI now passes.

.ci/boot-linux.sh
```diff
@@ -27,4 +27,28 @@ RET=$((${RET} + $?))
 (. "${SCRIPT_DIR}/virtio-blk.sh")
 RET=$((${RET} + $?))
 
+# Virtio-net user-mode backend test
```
jserv, line 30: The `vmnet` block below checks `.config` before running, but the `user` and `tap` blocks do not. With `CONFIG_VIRTIO_NET_USER=n` (or `CONFIG_VIRTIO_NET_TAP=n`), `virtio_net_backend_supported()` rejects `-x vnet:user` during argument parsing, so `rv32emu` exits immediately and the whole boot suite fails instead of skipping the test. Guard all three the same way, for example with `grep -q '^CONFIG_VIRTIO_NET_USER=y$' .config`.

.ci/netdev.sh
```diff
@@ -0,0 +1,162 @@
...
+        guest_ip=""
+        gateway_ip=""
+        test_name="vmnet shared"
+        ;;
+esac
+
+register_cleanup cleanup_emulator
+
+TIMEOUT=${NETDEV_BOOT_TIMEOUT:-${TIMEOUT}}
+MESSAGES+=("${COLOR_R}Fail to ping gateway")
```
jserv, line 58: The appended entry covers exit code 4, but the expect block also exits 2 when `readlink /sys/bus/virtio/devices/virtio0/driver` does not report `virtio_net`, and that indexes the inherited `Fail to login` message. A guest that boots and logs in fine but never binds the driver gets reported as a login failure. Give that case its own message, or move the exits past the end of the inherited array.

src/devices/netdev-vmnet.c
```diff
@@ -0,0 +1,552 @@
...
+
+    pthread_mutex_lock(&state->lock);
+
+    /*
+     * Preserve the packet boundary while passing packets through a byte
+     * stream pipe. Store the packet length first and then the packet data.
+     */
+    uint32_t pkt_len = (uint32_t) len;
+
+    if (write(state->pipe_fds[1], &pkt_len, sizeof(pkt_len)) !=
```
jserv, line 54: The length prefix and the payload are two separate writes, and only `pipe_fds[0]` is set non-blocking, so the write side blocks: once the guest stops posting RX buffers the pipe fills and this callback stalls the serial vmnet queue while holding `state->lock`. If the length write lands and the payload write is short or fails, the code only logs and returns, leaving the stream permanently desynchronized so `net_vmnet_read` consumes payload bytes as the next length. Use a non-blocking `SOCK_DGRAM` socketpair so each frame is one atomic message and a full buffer drops the packet instead of blocking.

src/devices/netdev-vmnet.c
```diff
@@ -0,0 +1,552 @@
...
+          if (ret == VMNET_SUCCESS) {
+              vmnet_store_mac(state, param);
+
+              rv_log_warn(
+                  "vmnet: %s mode started, MAC "
+                  "%02x:%02x:%02x:%02x:%02x:%02x",
+                  mode_name, state->mac[0], state->mac[1], state->mac[2],
+                  state->mac[3], state->mac[4], state->mac[5]);
+
+              vmnet_register_packet_callback(state, iface);
```
jserv, line 179: `iface` is read here inside the completion block, but it is only assigned when `vmnet_start_interface` returns on the calling thread. The block runs on `state->queue` with no ordering against that assignment, so the packet callback can be registered on a NULL interface and no packet ever arrives. Register it after `dispatch_semaphore_wait` returns and `state->iface` has been set.

src/devices/netdev-vmnet.c
```diff
@@ -0,0 +1,552 @@
...
+          }
+
+          dispatch_semaphore_signal((dispatch_semaphore_t) state->sem);
+        });
+
+    dispatch_semaphore_wait((dispatch_semaphore_t) state->sem,
```
jserv, line 185: `vmnet_start_interface` returns NULL when it fails outright, and in that case the completion handler that signals `state->sem` never runs, so this wait blocks forever and `rv32emu` hangs at startup with no diagnostic. Check the return value before waiting, and give the wait a finite deadline so a framework that accepts the request but never calls back still unwinds.

src/devices/netdev-vmnet.c
```diff
@@ -0,0 +1,552 @@
...
+
+void net_vmnet_cleanup(net_vmnet_state_t *state)
+{
+    if (!state)
+        return;
+
+    state->running = false;
+
+    if (state->iface) {
+        vmnet_stop_interface(
```
jserv, line 518: `vmnet_stop_interface` only schedules its completion handler on `state->queue`, yet the code below releases that queue, closes `pipe_fds`, and destroys `state->lock` right away, and `netdev_delete` then frees `state`. A packet callback already queued or still running will touch a destroyed mutex, a closed fd, and freed memory on every shutdown and cold reboot. Clear the event callback with `vmnet_interface_set_event_callback(iface, VMNET_INTERFACE_PACKETS_AVAILABLE, NULL, NULL)`, wait for the stop completion, then drain the queue before releasing anything.

src/devices/slirp.c
```diff
@@ -0,0 +1,432 @@
...
+static int64_t monotonic_clock_ms(void)
+{
+    return monotonic_clock_ns() / 1000000LL;
+}
+
+static ssize_t net_slirp_send_packet(const void *buf, size_t len, void *opaque)
+{
+    net_user_options_t *usr = (net_user_options_t *) opaque;
+
+    if (!usr || usr->guest_to_host_channel[SLIRP_WRITE_SIDE] < 0)
```
jserv, line 49: The guard tests `guest_to_host_channel[SLIRP_WRITE_SIDE]`, but the `write` below uses `host_to_guest_channel[SLIRP_WRITE_SIDE]`. The descriptor actually written is never validated; the check only holds today because `net_slirp_init` creates and `net_slirp_cleanup` closes both pairs together.

```suggestion
    if (!usr || usr->host_to_guest_channel[SLIRP_WRITE_SIDE] < 0)
```

src/devices/virtio-net.c
```diff
@@ -0,0 +1,934 @@
...
+        if (desc_idx >= queue->queue_num) {
+            virtio_net_set_fail(vnet);
+            return false;
+        }
+
+        uint32_t desc_addr = queue->queue_desc + desc_idx * 4;
+        if (!vnet_check_word_range(vnet, desc_addr, 4))
+            return false;
+
+        const struct virtq_desc *desc =
```
jserv, line 208: `vnet_preprocess` only rejects addresses with the low two bits set, so `queue_desc` is guaranteed 4-byte alignment while `struct virtq_desc` needs 8 for its `uint64_t addr`. A guest that programs `QueueDescLow` to an address that is 4-byte but not 8-byte aligned makes every field access through this pointer undefined, which faults on strict-alignment hosts. `virtio_blk_handle_request` copies the entry with `memcpy` instead of casting for exactly this reason, and carries a comment saying so; do the same here.

jserv (changes requested):
Rebase latest `master` branch and resolve conflicts.

Makefile
```diff
@@ -222,12 +223,76 @@ $(OUT)/emulate.o: CFLAGS += -foptimize-sibling-calls -fomit-frame-pointer -fno-s
...
 include mk/http.mk
 
+# VirtIO networking is available only for kernel system emulation and requires
+# at least one host backend. CONFIG_* values may be overridden by legacy
+# ENABLE_* flags after loading .config, so derive the effective build state here
+# instead of relying only on the Kconfig dependency graph.
+VIRTIO_NET_BUILD_ENABLED := n
+ifeq ($(CONFIG_SYSTEM),y)
+ifneq ($(CONFIG_ELF_LOADER),y)
+ifeq ($(CONFIG_VIRTIO_NET),y)
```
jserv: `VIRTIO_NET_BUILD_ENABLED` tests `$(CONFIG_VIRTIO_NET)` directly, but the `-DRV32_FEATURE_VIRTIO_NET` value comes from `$(call set-features, ...)`, which routes through `has` and lets a legacy `ENABLE_VIRTIO_NET` override the `.config` value. `make system_defconfig && make ENABLE_VIRTIO_NET=0` then compiles `virtio-net.c` with `RV32_FEATURE_VIRTIO_NET=0`, so `virtio.h` never declares `virtio_net_state_t` and the file fails to compile; the same split applies per backend, where `ENABLE_VIRTIO_NET_USER=1` leaves `netdev.c` calling `net_slirp_init` with `slirp.o` filtered out of `DEV_OBJS`. Derive these from `$(call has, ...)` here and in `check-vnet-config`, which is what the comment above this block already promises.

src/devices/slirp.c
```diff
@@ -0,0 +1,432 @@
...
+
+        rv_log_error("SLIRP poll failed: %s", strerror(errno));
+        return -1;
+    }
+
+    if (usr->guest_to_host_channel[SLIRP_READ_SIDE] >= 0) {
+        for (int i = 0; i < usr->pfd_len; i++) {
+            if (usr->pfd[i].fd == usr->guest_to_host_channel[SLIRP_READ_SIDE] &&
+                (usr->pfd[i].revents & POLLIN)) {
+                net_slirp_read(usr);
```
jserv: `net_slirp_read` consumes exactly one datagram and this loop calls it once, so `net_slirp_poll` moves at most one guest TX frame into SLIRP per call. With two polls per `virtio_net_refresh_queue` and a refresh every `VNET_REFRESH_INTERVAL` cycles, guest to host throughput is capped at two frames per 5000 guest cycles no matter how many descriptors `virtio_net_try_tx` already drained into the socketpair. Drain until the channel is empty.

```suggestion
                while (net_slirp_read(usr) > 0)
                    ;
```

src/devices/virtio-net.c
```diff
@@ -0,0 +1,939 @@
...
+static void virtio_net_set_fail(virtio_net_state_t *vnet)
+{
+    vnet->status |= VIRTIO_STATUS_DEVICE_NEEDS_RESET;
+    if (vnet->status & VIRTIO_STATUS_DRIVER_OK)
+        vnet->interrupt_status |= VIRTIO_INT_CONF_CHANGE;
+}
+
+static bool vnet_guest_range_ok(uint64_t addr, uint64_t len)
+{
+    if (len == 0)
```
jserv: `vnet_guest_range_ok` accepts any `addr` once `len` is zero, and `vnet_build_iovs` then stores `(uintptr_t) ram + desc.addr` as `iov_base`. A guest publishing a descriptor with `len = 0` and `addr = 0xFFFFFFFFFFFFFFFF` produces a wild pointer that is handed to `readv`, `writev`, and `vmnet_write`, safe only for as long as every one of those consumers honors the zero length. Drop this early return: the two checks below already accept a zero length inside the window, and `virtio_blk_check_range` validates the address unconditionally for the same reason.

src/devices/virtio-net.c
```diff
@@ -0,0 +1,939 @@
...
+        virtio_header[10] = 1;
+
+        if (vnet_iovec_write(&cursor, &ncursor, virtio_header,
+                             sizeof(virtio_header))) {
+            rv_log_error("virtio-net: RX buffer too small for virtio header");
+            return virtio_net_set_fail(vnet);
+        }
+
+        ssize_t plen = vnet_handle_read(&vnet->peer, queue, cursor, ncursor);
+        if (plen <= 0)
```
jserv, line 533: When `vnet_handle_read` fails for a reason other than `EAGAIN` it logs and returns -1 without clearing `queue->fd_ready`, so this `break` leaves the RX queue armed. A TAP fd in an error state keeps reporting `POLLIN`, so the next `virtio_net_refresh_queue` calls straight back in and the error repeats once per refresh for the rest of the run. Give a permanent read error the handling the TX path gives `VNET_TX_DROP`: clear `fd_ready` or fail the device so the loop stops.

src/devices/virtio-net.c
```diff
@@ -0,0 +1,939 @@
...
+void virtio_net_write(virtio_net_state_t *vnet, uint32_t addr, uint32_t value)
+{
+    addr = addr >> 2;
+
+#define _(reg) VIRTIO_##reg
+    switch (addr) {
+    case _(DeviceFeaturesSel):
+        vnet->device_features_sel = value;
+        break;
+    case _(DriverFeatures):
```
jserv, line 815: The upper driver feature word is dropped here, so nothing records whether the driver accepted `VIRTIO_F_VERSION_1`, yet `VNET_HDR_SIZE` is fixed at 12. A driver that negotiates without it (and without `VIRTIO_NET_F_MRG_RXBUF`, which is not offered) uses the 10 byte `virtio_net_hdr`, which shifts every RX and TX frame by the two `num_buffers` bytes and silently corrupts the link. Keep the high word and fail the device when `FEATURES_OK` is written without `VIRTIO_F_VERSION_1` set.

src/riscv.c
```diff
@@ -906,8 +962,21 @@ void rv_run(riscv_t *rv)
...
+#if RV32_HAS(VIRTIO_NET)
+            if (attr->vnet) {
+                if (rv->csr_cycle - last_vnet_refresh >=
+                    VNET_REFRESH_INTERVAL) {
+                    virtio_net_refresh_queue(attr->vnet);
+                    last_vnet_refresh = rv->csr_cycle;
+                }
+
+                emu_update_vnet_interrupts(rv);
+            }
```
jserv, line 980: `emu_update_vnet_interrupts()` runs after every `rv_step()` rather than after the refresh, so `plic_update_interrupts()` is called once per translated block for the whole boot even on an idle link. `vnet->interrupt_status` only changes inside `virtio_net_refresh_queue()` and inside `virtio_net_write()`, and the MMIO path in `system.h` already updates the PLIC for the latter.

```suggestion
                if (rv->csr_cycle - last_vnet_refresh >=
                    VNET_REFRESH_INTERVAL) {
                    virtio_net_refresh_queue(attr->vnet);
                    emu_update_vnet_interrupts(rv);
                    last_vnet_refresh = rv->csr_cycle;
                }
            }
```

src/system.c
```diff
@@ -48,6 +48,23 @@ void emu_update_vrng_interrupts(riscv_t *rv)
...
+#if RV32_HAS(VIRTIO_NET)
+void emu_update_vnet_interrupts(riscv_t *rv)
+{
+    vm_attr_t *attr = PRIV(rv);
+
+    if (!attr->vnet)
+        return;
+
+    if (attr->vnet->interrupt_status)
+        attr->plic->active |= IRQ_VNET_BIT(attr->vnet_irq);
```
jserv, line 60: `IRQ_VNET_BIT` is `1U << vnet_irq` and `plic_t.active` is 32 bits, but `load_dtb` gives virtio-net the highest interrupt of the three virtio devices, `next_irq + vblk_cnt + vrng`. `VBLK_DEV_MAX` is 100, so a command line with 30 or more `-x vblk:` images pushes `vnet_irq` to 32 and the shift is undefined; `IRQ_VBLK_BIT` carries the same exposure today. Worth a follow-up that rejects an assigned interrupt above 31 while the DTB nodes are built, instead of bringing the device up on an interrupt the PLIC cannot represent.

src/devices/virtio-net.c
Charlie-Tsai1123: Fixed. Permanent RX backend errors now mark the device as needing reset, so an errored backend is not retried on every refresh. EAGAIN/EWOULDBLOCK remain transient, and EINTR is also kept retryable. I implement on commit [Handle permanent virtio-net I/O errors](https://github.com/sysprog21/rv32emu/pull/748/changes/7927110701f09d88e3665d7f72d86d5caf1d32c1)

src/system.c
Charlie-Tsai1123: Agreed. Since this also affects the existing virtio-blk irq bitmap, I’ll address it in a follow-up by validating the assigned virtio irq range while building the DTB, rather than expanding this PR further.

Makefile
Charlie-Tsai1123: Fixed. The virtio-net build state, backend selection, and `check-vnet-config` now use `$(call has, ...)`, so object selection and `RV32_FEATURE_*` follow the same effective configuration, including legacy `ENABLE_*` overrides.

While testing the override cases, I also found two related incremental-build issues. For example, after building with the TAP backend enabled, I rebuilt with `ENABLE_VIRTIO_NET_TAP=0 ENABLE_VIRTIO_NET_USER=1` without running `make clean`. The usage output (./build/rv32emu -h) reflected the USER backend, but `vnet:user` was still rejected at runtime because some device objects from the previous configuration had not been rebuilt.

This exposed two problems: the effective-config stamp was not available when the device object rules were defined and did not record the effective virtio-net feature values, and `make clean` only removed the currently filtered `DEV_OBJS`, leaving objects from previously enabled backends behind.

I moved the effective-config stamp definition before `mk/system.mk`, added the effective virtio-net feature values to the stamp, and made cleanup remove all possible device objects and dependency files.

.ci/netdev.sh
```diff
@@ -0,0 +1,206 @@
...
+        test_name="vmnet shared"
+        ;;
+esac
+
+register_cleanup cleanup_emulator
+
+TIMEOUT=${NETDEV_BOOT_TIMEOUT:-${TIMEOUT}}
+MESSAGES[2]="${COLOR_R}Fail to bind virtio-net driver"
+MESSAGES+=("${COLOR_R}Fail to ping gateway")
+MESSAGES+=("${COLOR_R}Fail to identify backend interface")
```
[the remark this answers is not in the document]
Charlie-Tsai1123: The current indexing is already correct. `MESSAGES` starts with four entries at indices 0–3. `MESSAGES[2]` only replaces the existing entry at index 2, the first `MESSAGES+=` appends the gateway failure message at index 4, and the second `MESSAGES+=` therefore appends the backend-interface failure message at index 5. As a result, the `exit 5` paths already select the intended diagnostic, so no code change is needed here.

src/riscv.c
```diff
@@ -1712,6 +1780,24 @@ bool rv_cold_reboot(riscv_t *rv, riscv_word_t pc)
         }
     }
 
+#if RV32_HAS(VIRTIO_NET)
+    if (attr->data.system.vnet_backend) {
```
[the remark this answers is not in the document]
Charlie-Tsai1123: I checked the surrounding preprocessor branches. This block is after the `#else` corresponding to `#if !RV32_HAS(SYSTEM_MMIO)`, so it is compiled specifically for the SYSTEM_MMIO/kernel-boot path. The DTB/MMIO path initializes virtio-net here already, so no code change is needed for this item.

src/devices/netdev-vmnet.c
```diff
@@ -0,0 +1,638 @@
...
+    rv_log_warn("vmnet: %s mode started, MAC %02x:%02x:%02x:%02x:%02x:%02x",
+                mode_name, state->mac[0], state->mac[1], state->mac[2],
+                state->mac[3], state->mac[4], state->mac[5]);
+
+    vmnet_return_t ret = vmnet_register_packet_callback(state, iface);
+
+    if (ret != VMNET_SUCCESS) {
+        rv_log_error("vmnet: failed to register packet callback: %d", ret);
+
+        vmnet_stop_interface_sync(state);
```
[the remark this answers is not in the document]
Charlie-Tsai1123: Fixed. `vmnet_stop_interface_sync()` now clears `state->iface` only after stop completion succeeds. If the stop cannot be scheduled or completes with an error, `net_vmnet_cleanup()` retains the interface, dispatch queue, socketpair, and backend state. Cleanup failure is propagated through `netdev_delete()/vnet_delete()` so the state can be retried safely instead of being freed while vmnet may still reference it. Partial initialization failures now use the same cleanup path.

src/riscv.c
```diff
@@ -823,6 +869,13 @@ static void rv_fsync_device(void)
         vrng_delete(attr->vrng);
         attr->vrng = NULL;
     }
+
+#if RV32_HAS(VIRTIO_NET)
+    if (attr->vnet) {
```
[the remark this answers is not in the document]
Charlie-Tsai1123: Fixed. Block-image `pwrite()`/`fsync()` failures now go through the per-device cleanup path instead of returning from `rv_fsync_device()`, so vrng and virtio-net teardown still runs.

src/system.c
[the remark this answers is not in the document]
Charlie-Tsai1123: Agreed. This is the same 32-bit PLIC interrupt-mask limitation that also affects the existing virtio-blk allocation path. I am keeping it as a separate follow-up and will validate/reject out-of-range optional VirtIO IRQ assignments while building the DTB rather than expanding this PR further.

Charlie-Tsai1123:
### Refined the macOS vmnet CI path as well.
The previous test selected the first `bridge*` interface on the host to derive the vmnet gateway. That is ambiguous when the runner already has other bridge interfaces.
The test now snapshots the existing bridges before starting rv32emu and selects the bridge created by the current vmnet invocation. It derives the gateway from that bridge, configures the guest with another address from the same `/24` subnet, and then verifies virtio-net binding and packet delivery by pinging the gateway.
I also tried using DHCP for the guest, but the current rv32emu Linux image has `CONFIG_PACKET` disabled, so BusyBox `udhcpc` cannot create its packet socket. Rather than expanding this PR to modify the guest kernel configuration, the CI keeps the existing static-address approach while deriving the subnet from the runtime vmnet bridge.
This also removes the previous fixed `192.168.2.0/24` assumption and avoids accidentally configuring an unrelated host bridge.

### Test vmnet
This part describes how to test vmnet on macOS.

host os:
```
spawn sudo -E build/rv32emu -k build/linux-image/Image -i build/linux-image/rootfs.cpio -x vnet:vmnet
```

guest os:
```
readlink /sys/bus/virtio/devices/virtio0/driver
ip link set eth0 up
ip addr flush dev eth0
ip addr add 192.168.2.10/24 dev eth0
ip addr show eth0
ping -c 3 -W 5 192.168.2.1
```
Expected Result:
```
# readlink /sys/bus/virtio/devices/virtio0/driver

../../../../../bus/virtio/drivers/virtio_net

# ip link set eth0 up

# ip addr flush dev eth0

# ip addr add 192.168.2.10/24 dev eth0

# ip addr show eth0

2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast qlen 1000

    link/ether 0e:0d:22:e0:72:cb brd ff:ff:ff:ff:ff:ff

    inet 192.168.2.10/24 scope global eth0

       valid_lft forever preferred_lft forever

# ping -c 3 -W 5 192.168.2.1

PING 192.168.2.1 (192.168.2.1): 56 data bytes

64 bytes from 192.168.2.1: seq=0 ttl=64 time=1.172 ms

64 bytes from 192.168.2.1: seq=1 ttl=64 time=0.542 ms

64 bytes from 192.168.2.1: seq=2 ttl=64 time=0.561 ms
```

Makefile
```diff
@@ -226,12 +233,97 @@ $(OUT)/emulate.o: CFLAGS += -foptimize-sibling-calls -fomit-frame-pointer -fno-s
...
+
+ifeq ($(UNAME_S),Darwin)
+MINISLIRP_CFLAGS := MYCFLAGS="-D_DARWIN_C_SOURCE"
+LDFLAGS += -lresolv
+endif
+
+$(MINISLIRP_DIR)/src/Makefile:
+	$(Q)git submodule update --init $(MINISLIRP_DIR)
+
+$(MINISLIRP_LIB): $(MINISLIRP_DIR)/src/Makefile
```
jserv, line 319: `libslirp.a` depends only on the submodule's `Makefile`, so editing minislirp sources or bumping the submodule without touching that file relinks the stale archive, and `clean` never removes it. A clean rebuild after switching compiler or target architecture still links the old objects. Recurse into the submodule build unconditionally (phony prerequisite) and add the archive plus its objects to `clean`.

src/devices/netdev-vmnet.c
```diff
@@ -0,0 +1,657 @@
...
+        return -1;
+
+    return fcntl(fd, F_SETFL, flags | O_NONBLOCK);
+}
+
+static void vmnet_packet_handler(net_vmnet_state_t *state,
+                                 const uint8_t *buf,
+                                 ssize_t len)
+{
+    if (!state || !state->running || len <= 0)
```
jserv, line 60: `state->running` is read here on vmnet's dispatch queue and written by the emulator thread after callback registration and again at the top of `net_vmnet_cleanup`, with no synchronization on either side. That is a data race. `net_vmnet_cleanup` already clears the event callback and drains the queue before closing `rx_fds`, so this check can go; if it stays, the flag needs to be `_Atomic`.

src/devices/netdev-vmnet.c
```diff
@@ -0,0 +1,657 @@
...
+     */
+    dispatch_retain(start_sem);
+
+    interface_ref iface =
+        vmnet_start_interface(iface_desc, (dispatch_queue_t) state->queue,
+                              ^(vmnet_return_t ret, xpc_object_t param) {
+                                status = ret;
+
+                                if (ret == VMNET_SUCCESS)
+                                    vmnet_store_mac(mac.bytes, param);
```
jserv, line 287: The start completion takes only `vmnet_mac_address_key` out of `param` and drops `vmnet_max_packet_size_key` and `vmnet_mtu_key`. RX buffers are then fixed at `VMNET_PKT_MAX` and the guest is told MTU 1500 by `vnet_new` regardless of what the interface reported. An interface whose maximum packet size exceeds 2048 makes `vmnet_read` fail, and one with a different MTU leaves the guest misconfigured. Read both keys here, size the RX buffers from the reported maximum, and publish the reported MTU in the virtio-net config.

src/devices/virtio-net.c
```diff
@@ -0,0 +1,961 @@
...
+        struct iovec *vecs = iovs;
+        size_t nvecs = niovs;
+
+        if (vnet_iovec_write(&vecs, &nvecs, buf, (size_t) plen)) {
+            rv_log_error(
+                "virtio-net: vmnet packet exceeds "
+                "guest RX buffer");
+
+            virtio_net_set_fail(vnet);
+            return -1;
```
jserv, line 342: An inbound frame larger than the descriptor chain the guest posted puts the device into `VIRTIO_STATUS_DEVICE_NEEDS_RESET`, which kills guest networking until the driver reprobes. The trigger is host traffic, not a driver bug, and the TAP and user paths let `readv` truncate instead. `net_vmnet_read` has already consumed the datagram, so dropping it here is enough.

```suggestion
        rv_log_warn(
            "virtio-net: dropping vmnet packet larger than "
            "guest RX buffer");

        return 0;
```

src/riscv.c
```diff
@@ -908,8 +966,20 @@ void rv_run(riscv_t *rv)
...
-        for (; !rv_has_halted(rv);) /* run until the flag is done */
-            rv_step(rv);            /* step instructions */
+        for (; !rv_has_halted(rv);) { /* run until the flag is done */
+            rv_step(rv);              /* step instructions */
+
+#if RV32_HAS(VIRTIO_NET)
+            if (attr->vnet) {
+                if (rv->csr_cycle - last_vnet_refresh >=
+                    VNET_REFRESH_INTERVAL) {
+                    virtio_net_refresh_queue(attr->vnet);
```
jserv, line 976: The network tick lives only in the default run loop, so the `rv_debug` branch taken by `-g` and the trace branch never call `virtio_net_refresh_queue`. With `-g -x vnet:user` guest TX reaches the socketpair but `net_slirp_poll` never runs, and TAP and vmnet RX is never drained, so networking is silently dead under the GDB stub. Move the tick into a helper both execution paths call, or invoke it from the debugger's continue and single-step loops.

src/riscv.c
```diff
@@ -1712,6 +1782,24 @@ bool rv_cold_reboot(riscv_t *rv, riscv_word_t pc)
         }
     }
 
+#if RV32_HAS(VIRTIO_NET)
+    if (attr->data.system.vnet_backend) {
+        if (attr->vnet) { /* check for reboot */
+            virtio_net_reset(attr->vnet);
```
jserv, line 1788: `virtio_net_reset` is wired into the cold reboot path only. `rv_warm_reboot` zeroes guest RAM and reloads the kernel while leaving `status`, `ready`, `last_avail` and the ring word indices intact. Unlike virtio-blk and virtio-rng, this device is driven by the periodic refresh rather than only by guest notifies, so it keeps walking the old ring addresses over memory that now holds the freshly loaded image, and can land a used-ring entry in it before the new kernel resets the device. Reset it from `rv_warm_reboot` too, or gate the refresh on a flag that warm reboot clears.

Makefile
```diff
@@ -327,6 +429,20 @@ distclean: cleanconfig
 	$(Q)$(RM) $(OUT)/rv32emu-prebuilt*.tar.gz $(OUT)/rv32emu-prebuilt-sail-*
 	$(Q)$(call notice, [OK])
 
-.PHONY: all tool clean cleanconfig distclean gdbstub-test FORCE
+# Reject a build that enables virtio-net without a host backend. Keep this as a
+# build prerequisite rather than a parse-time error so maintenance targets such
+# as clean and config remain usable with an incomplete configuration.
+check-vnet-config:
+	$(Q)if [ "$(call has,SYSTEM)" = "1" ] && \
```
ChinYikMing, line 436: Shorten the `SYSTEM` and `ELF_LOADER` by `SYSTEM_MMIO`.

Makefile
```diff
@@ -226,12 +233,97 @@ $(OUT)/emulate.o: CFLAGS += -foptimize-sibling-calls -fomit-frame-pointer -fno-s
 # HTTP Utilities (shared by external.mk and artifact.mk)
 include mk/http.mk
 
+# VirtIO networking is available only for kernel system emulation and requires
```
ChinYikMing, line 236: Create a new Makefile in mk/ for VIRTIO_NET specific rules and include it.

Makefile
ChinYikMing: Can we move this check to the virtio_net specific Makefile under mk/?

src/feature.h
```diff
@@ -163,6 +163,61 @@
...
+#define RV32_FEATURE_VIRTIO_NET 0
+#endif
+
+/* Backends cannot be enabled without the VirtIO network device. */
+#if !RV32_FEATURE_VIRTIO_NET
+#undef RV32_FEATURE_VIRTIO_NET_TAP
+#define RV32_FEATURE_VIRTIO_NET_TAP 0
+#undef RV32_FEATURE_VIRTIO_NET_USER
+#define RV32_FEATURE_VIRTIO_NET_USER 0
+
```
ChinYikMing, line 215: Remove the blank line.

src/feature.h
```diff
@@ -163,6 +163,61 @@
...
+#undef RV32_FEATURE_VIRTIO_NET_VMNET
+#define RV32_FEATURE_VIRTIO_NET_VMNET 0
+
```
ChinYikMing, line 218: Ditto.

src/feature.h
```diff
@@ -163,6 +163,61 @@
...
+#endif
+
+/* macOS vmnet.framework backend for VirtIO networking */
+#ifndef RV32_FEATURE_VIRTIO_NET_VMNET
+#define RV32_FEATURE_VIRTIO_NET_VMNET 0
+#endif
+
+/*
+ * vmnet.framework is supported only on macOS with Clang because its
+ * asynchronous callback API uses Apple Blocks.
```
ChinYikMing, line 188: Do you have any reference link to put here?

src/devices/virtio.h
```diff
@@ -5,12 +5,22 @@
 
 #pragma once
 
+#include <stdbool.h>
```
ChinYikMing, line 8: Remove redundant dependency.

src/devices/virtio.h
```diff
@@ -5,12 +5,22 @@
...
+#include <stdint.h>
```
ChinYikMing, line 9: Ditto.

src/devices/netdev.h
```diff
@@ -0,0 +1,163 @@
...
+ * "LICENSE" for information on usage and redistribution of this file.
+ */
+
+#pragma once
+
+#include <poll.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <sys/types.h>
+#include <sys/uio.h>
```
ChinYikMing, line 12: Please ensure these dependency headers are mandatory.

Makefile
```diff
@@ -226,12 +233,97 @@ $(OUT)/emulate.o: CFLAGS += -foptimize-sibling-calls -fomit-frame-pointer -fno-s
...
+VIRTIO_NET_BUILD_ENABLED := y
+endif
+
 # External Dependencies & System Emulation
 include mk/external.mk
 include mk/artifact.mk
 include mk/system.mk
 include mk/wasm.mk
 
+ifeq ($(VIRTIO_NET_VMNET_BUILD_ENABLED),y)
```
ChinYikMing, line 295: Maybe move this to virtio_net dedicated makefile under mk/?

src/devices/netdev-vmnet.c
Charlie-Tsai1123: Removed the unsynchronized state->running check from the vmnet receive callback. Cleanup already removes the callback and drains the serial dispatch queue before closing the socketpair.

src/devices/netdev-vmnet.c
Charlie-Tsai1123: Updated. The vmnet start completion now records the reported MTU and maximum packet size. RX buffers use the reported maximum packet size, and the vmnet MTU is exposed through the virtio-net configuration. When RX buffers too large, it would drop buffer and write used ring.

src/feature.h
Charlie-Tsai1123: Added references to the [Apple vmnet API](https://developer.apple.com/documentation/vmnet/vmnet_start_interface(_:_:_:)) showing the [pcompletion-handler Block type](https://developer.apple.com/documentation/vmnet/vmnet_start_interface_completion_handler_t), together with the [Clang Blocks documentation](https://clang.llvm.org/docs/CommandGuide/clang.html#cmdoption-clang-fblocks).

src/riscv.c
Charlie-Tsai1123: Updated. rv_warm_reboot() now resets the guest-visible virtio-net state while keeping the host backend alive. I verified that networking works again after a warm reboot.

src/riscv.c
Charlie-Tsai1123: Updated. The periodic virtio-net refresh is now shared with the GDB execution paths, including continue and single-step. I also verified -g -x vnet:user with successful guest-to-gateway ping.

Charlie-Tsai1123:
A couple of additional notes from the latest changes and testing:

### mk/virtio-net.mk

The VirtIO-net specific build logic has been moved out of the top-level Makefile into mk/virtio-net.mk. It now contains the build-time logic specific to VirtIO networking, including:
- determining whether VirtIO-net is available for the current system/host/compiler configuration;
- selecting the TAP, SLIRP, and vmnet backends;
- applying vmnet-specific compiler/linker flags;
- integrating and rebuilding the bundled minislirp dependency;
- validating that an enabled VirtIO-net configuration has at least one usable backend;
- tracking VirtIO-net related generated files for cleanup.

This keeps the top-level Makefile focused on common build orchestration while keeping the VirtIO-net backend and host-specific logic in one place.

### Reboot under the GDB stub

While verifying the VirtIO-net refresh path under the GDB stub, I also noticed a separate reboot issue.

Networking itself works with `-g -x vnet:user` (including guest-to-SLIRP gateway traffic), but issuing a reboot from the guest causes rv32emu to terminate with a segmentation fault.

It can be reproduced with:

Terminal 1:
```
make system_defconfig
make ENABLE_SYSTEM=1 ENABLE_GDBSTUB=1 -j8
build/rv32emu -k build/linux-image/Image -i build/linux-image/rootfs.cpio -g -x vnet:user
```

Terminal 2:
```
export PATH="$(pwd)/toolchain/bin:$PATH"
riscv-none-elf-gdb
```
In Terminal 2 gdb
```
target remote :1234
c
```
```
(gdb) target remote :1234
Remote debugging using :1234
warning: No executable has been specified and target does not support
determining executable automatically.  Try using the "file" command.
0x00000000 in ?? ()
(gdb) continue
Continuing.
```

Then, Terminal 1 rv32emu would start. Then, inside the guest:
```
reboot
```
The observed result is:
```
Welcome to Buildroot
buildroot login: root
# reboot
# Stopping network: OK
Stopping klogd: OK
Stopping syslogd: stopped /sbin/syslogd (pid 35)
OK
umount: devtmpfs busy - remounted read-only
umount: can't unmount /: Invalid argument
The system is going down NOW!
Sent SIGTERM to all processes
Sent SIGKILL to all processes
Requesting system reboot
[    9.036765] reboot: Restarting system
20:50:49 WARN  src/syscall.c:541: System reset: type=cold reboot, reason=no reason
Segmentation fault         (core dumped) build/rv32emu -k build/linux-image/Image -i build/linux-image/rootfs.cpio -g
```

This appears to be independent of the VirtIO-net backend itself (without `-x vnet:user` or `-x vnet:tap` have the same problem): the networking path works correctly under the GDB stub before the reboot. I am leaving this issue outside the scope of this PR and documenting it here for follow-up.

jserv (changes requested):
Rework the series into functionally minimal commits, folding similar ones and enforcing the project's commit message rules.

src/devices/netdev.h
Charlie-Tsai1123: Updated. `netdev.h` now explicitly includes the headers required by the types and feature macros it uses, instead of relying on transitive includes. `struct iovec` is forward-declared in the header, with `<sys/uio.h>` included only in `netdev-vmnet.c` where the complete definition is required.

jserv:
Thank @Charlie-Tsai1123 for contributing!

PR #754, opened 2026-07-24
Enable to generate rootfs.cpio.gz

ChinYikMing:
This is friendly for user to store smaller size images locally.

| Image Type | Uncompressed | Compressed (.gz) | Reduction |
| :--- | ---: | ---: | ---: |
| Kernel | 4.6M | 2.2M | 52.17% |
| Rootfs (cpio) | 32.0M | 15.0M | 53.12% |

Also, once the prebuilt gzipped linux images are available, the
rv32emu's bootloader can be extended to handle these images to have
more capability for wider ranges of linux images.

---
## Summary by cubic
Adds gzip-compressed Linux images (`rootfs.cpio.gz`, `Image.gz`) to cut local storage by ~50% and pave the way for future bootloader support. The compressed files are built, copied to the output, and added to the prebuilt bundle with checksums.

- **New Features**
  - Enable `BR2_TARGET_ROOTFS_CPIO_GZIP` in `buildroot` to produce `rootfs.cpio.gz`.
  - Copy `rootfs.cpio.gz` and `Image.gz` to output; include both in prebuilt artifacts and `sha1sum-linux-image`.

<sup>Written for commit 831ec5b786fc4621f4327fdee1a80ec5d1123fd4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/rv32emu/pull/754?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #755, opened 2026-07-24
Enable to boot Linux guestOS with gzip images

ChinYikMing:
The zlib menu design is to natively support the gzip-compressed Image
since modern Linux/macOS machines have pre-installed the zlib.
If the target platform does not have the zlib and users use
gzip-compressed Image, the emulator will simply exit with error.
The zlib config option will not able to be toggled on/off in the
config menu and it is detected depends on the target platform via
tools/detect-env.py.

This enables the user to pass compressed kernel and rootfs images
to rv32emu which means the user can store kernel and rootfs image in
compression to save local disk storage.

The emcc toolchain has ported the zlib and exposed it via -sUSE_ZLIB=1
[1]. Mark it as a TODO for future support.

Furthermore, introduce a new ci script called .ci/boot-gzip-linux.sh and append
it into the driver script .ci/boot-linux.sh.

[1] https://emscripten.org/docs/tools_reference/settings_reference.html#use-zlib

---
## Summary by cubic
Adds gzip image support by auto-detecting and decompressing `.gz` kernel and rootfs images in memory, so Linux guests can boot from compressed images without changing how uncompressed images are loaded.

**New Features**
- Detects gzip via magic bytes and stream-decompresses to RAM with enforced size limits; uncompressed images keep the existing mmap/read path.
- Kernel images are now validated against available memory too; previously only the rootfs had a size check.
- CI covers both-gzipped, kernel-only, and rootfs-only cases via `.ci/boot-gzip-linux.sh`.
- Docs: gzipped image usage and examples added in `docs/system.md`.

**Dependencies**
- New `CONFIG_LINK_ZLIB` auto-detected via `tools/detect-env.py --have-zlib`; not user togglable and disabled for Emscripten (TODO).
- `mk/system.mk` links `-lz` when set, for `SYSTEM` targets only.

<sup>Written for commit 7aa2c3de30eb4321778711817a3399b8833b0e8c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/rv32emu/pull/755?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ChinYikMing:
Can do the test once #754 is merged and golden prebuilt images are available. Mark as draft now.

jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: 501f54c4389a1ae4ccfa212b5ecca314297c67d8 | Previous: eac2b9fae201a4d766e23a174acbe37ba8469ac5 | Ratio |
|-|-|-|-|
| `Dhrystone` | `1507.333` DMIPS | `1565` DMIPS | `1.04` |
| `CoreMark` | `1058.883` iterations/sec | `1117.57` iterations/sec | `1.06` |

</details>

ChinYikMing:
### Testing procedue

1. Clone repo
```
git clone https://github.com/ChinYikMing/rv32emu.git -b boot-linux-with.gz --depth 1 && cd rv32emu
```

2. Fetch prebuilt Linux images
```
make system_defconfig
make artifact ENABLE_PREBUILT=1
```

3. Build system emulator
```
make -j8
```

4. Boot Linux guestOS with gzipped images (4 combination)

- a. Both gzipped

```
build/rv32emu -k build/linux-image/Image.gz -i build/linux-image/rootfs.cpio.gz
```

- b. Gzipped kernel
```
build/rv32emu -k build/linux-image/Image.gz -i build/linux-image/rootfs.cpio
```

- c. Gzipped rootfs.cpio
```
build/rv32emu -k build/linux-image/Image -i build/linux-image/rootfs.cpio.gz
```

- d. Legacy
```
build/rv32emu -k build/linux-image/Image -i build/linux-image/rootfs.cpio
```

### Expected Result
Boot the Linux guestOS successfully across all 4 combinations (login with `root`).
```
Welcome to Buildroot
buildroot login:
# uname a
Linux buildroot 6.1.179 #1 Thu Jul 30 09:13:45 UTC 2026 riscv32 GNU/Linux
```

mk/system.mk
```diff
@@ -125,4 +125,9 @@ CFLAGS := $(filter-out -DMEM_SIZE=%,$(CFLAGS))
 CFLAGS += -DMEM_SIZE=0x20000000ULL
 endif
 
+# Dynamic link zlib
+ifeq ($(CONFIG_LINK_ZLIB),y)
+    LDFLAGS += -lz
```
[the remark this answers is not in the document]
ChinYikMing: zlib has been built right into macOS since 2000 (Mac OS X Beta) and major Linux distros like Slackware, Red Hat, and Debian since late 1995.

ChinYikMing:
@cubic-dev-ai

mk/system.mk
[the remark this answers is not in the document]
ChinYikMing: zlib is available at modern Uniz-like OS. So, it might not the issue when linking on non-emscripten platform.

.ci/boot-gzip-linux.sh
```diff
@@ -0,0 +1,67 @@
...
+COLOR_G='\e[32;01m' # Green
+COLOR_R='\e[31;01m' # Red
+COLOR_Y='\e[33;01m' # Yellow
+COLOR_N='\e[0m'     # No color
+
+MESSAGES=("${COLOR_G}OK!"
+    "${COLOR_R}Fail to boot"
+    "${COLOR_R}Fail to login"
+    "${COLOR_R}Fail to run commands"
+)
```
jserv: Move these  to a common shell script, shared by all scripts.

ChinYikMing: > Move these to a common shell script, shared by all scripts.

Done.

src/riscv.c
```diff
@@ -334,55 +340,139 @@ void rv_destroy_t2c(riscv_t *rv)
...
+    is_gzip_image = is_gzip_file(name);
+    if (is_gzip_image) {
+        gzfile = gzopen(name, "rb");
+        if (!gzfile)
+            goto fail;
+
+        int bytes_read;
+        while ((bytes_read = gzread(gzfile, buf, sizeof(buf))) > 0) {
+            /* Check overflow for rootfs.cpio.gz */
+            if (max_size > 0 && dec_buf_off + bytes_read > max_size) {
```
jserv, line 393: load_boot_images() calls map_file() for the kernel with max_size == 0, so this bound check is skipped entirely and the loop below memcpy's the whole decompressed stream into guest RAM with no ceiling. An Image.gz that expands past dtb_addr - INITRD_SIZE silently overwrites the initrd and DTB regions that get populated immediately after, and one that expands past mem_size runs off the end of the allocation. Pass the kernel's real capacity from the caller (mem_size - DTB_SIZE - (initrd ? INITRD_SIZE : 0)) and treat max_size == 0 here as an error rather than as unlimited.

src/riscv.c
```diff
@@ -334,55 +340,139 @@ void rv_destroy_t2c(riscv_t *rv)
...
+                rv_log_error(
+                    "Decompression overflows, required: %d, quota: %ld",
+                    bytes_read, (long) (max_size - dec_buf_off));
+                gz_err_str = strdup("Decompression overflows");
+                goto gz_cleanup;
```
jserv: The contract documented above map_file() says it returns -1 when the content exceeds max_size so the caller can print the actionable message. The uncompressed branch at line 430 honors that; this branch jumps to fail and exit(EXIT_FAILURE), so an oversized rootfs.cpio.gz loses the "rebuild with a larger INITRD_SIZE" hint that an uncompressed one still gets. The logged "required: %d" is also the chunk size, not the amount actually needed.

```suggestion
                gzclose(gzfile);
                return -1; /* Caller handles the error message */
```

src/riscv.c
```diff
@@ -334,55 +340,139 @@ void rv_destroy_t2c(riscv_t *rv)
...
-        goto fail;
+    int fd;
+    off_t file_size;
+
+#if !defined(__EMSCRIPTEN__)
+    bool is_gzip_image;
+    gzFile gzfile = NULL;
+    int gzerrno;
+    char *gz_err_str = NULL;
+    uint32_t dec_buf_off = 0;
```
jserv: dec_buf_off is uint32_t while max_size and the return type are off_t, so the running total wraps at 4 GiB and file_size at line 409 truncates the same way. Nothing bounds it on the kernel path (max_size == 0), so reaching the wrap is possible rather than theoretical.

```suggestion
    off_t dec_buf_off = 0;
```

src/riscv.c
```diff
@@ -334,55 +340,139 @@ void rv_destroy_t2c(riscv_t *rv)
...
+            memcpy(*ram_loc + dec_buf_off, buf, bytes_read);
+            dec_buf_off += bytes_read;
+        }
 
-    /* Validate file size if max_size constraint is specified */
-    if (max_size > 0 && st.st_size > max_size) {
-        close(fd);
-        return -1; /* Caller handles the error message */
-    }
+        if (bytes_read < 0) {
```
jserv: gzread() does not report a truncated stream through its return value: when the input ends mid-deflate, zlib records Z_BUF_ERROR internally and gzread returns a short count, so the loop exits normally with bytes_read == 0 and this check never fires. A truncated Image.gz then loads as a partial kernel and boots into garbage. Check gzerror() after the loop, or take the return value of gzclose() and reject anything other than Z_OK.

src/riscv.c
```diff
@@ -15,6 +15,12 @@
 #if RV32_HAS(SYSTEM_MMIO)
 #include <termios.h>
 #include "dtc/libfdt/libfdt.h"
+
+/* TODO: support WASM target by leveraging -sUSE_ZLIB=1 */
+#if !defined(__EMSCRIPTEN__)
+#include <zlib.h>
```
jserv, line 21: This include and the gz* calls below are gated only on `__EMSCRIPTEN__`, but -lz is added only when CONFIG_LINK_ZLIB=y, and that symbol is false on any host where the probe does not find zlib. A native SYSTEM build on such a host fails right here on the missing header. Register LINK_ZLIB with set-features in the Makefile and gate this include, is_gzip_file(), and the gzip branch on RV32_HAS(LINK_ZLIB), the same condition that controls -lz.

tools/detect-env.py
```diff
@@ -172,6 +173,14 @@ def have_riscv_toolchain():
     return False
 
 
+def have_zlib():
+    """Check if ZLIB is available."""
+    ret, _, _ = run_cmd(
+        [sys.executable, "-c", "import zlib; print(zlib.ZLIB_VERSION)"]
```
jserv: Importing the host interpreter's zlib module says nothing about whether the configured CC can find zlib.h or link -lz: a distro Python ships the module while zlib1g-dev is absent, and a cross build probes the host sysroot rather than the target one. A false positive here turns into the compile or link failure described on src/riscv.c:21, and a false negative silently drops gzip support. Use check_pkg_config("zlib"), or compile and link a minimal <zlib.h> program with the configured compiler.

src/riscv.c
ChinYikMing: https://zlib.net/manual.html

> Z_BUF_ERROR indicates that the input file ended in the middle of a gzip stream. Note that gzread does not return –1 in the event of an incomplete gzip stream. This error is deferred until gzclose(), which will return Z_BUF_ERROR if the last gzread ended in the middle of a gzip stream. Alternatively, gzerror can be used before gzclose to detect this case.

src/riscv.c
```diff
@@ -334,55 +340,149 @@ void rv_destroy_t2c(riscv_t *rv)
...
+         * Z_BUF_ERROR indicates that the input file ended in the middle of a
+         * gzip stream. Note that gzread does not return -1 in the event of an
+         * incomplete gzip stream. This error is deferred until gzclose(), which
+         * will return Z_BUF_ERROR if the last gzread ended in the middle of a
+         * gzip stream.  Alternatively, gzerror can be used before gzclose to
+         * detect this case.
+         *
+         * https://zlib.net/manual.html
+         */
+        const char *tmp_gz_err_str = gzerror(gzfile, &gzerrno);
```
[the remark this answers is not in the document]
ChinYikMing: The comment also mention that gzerror can be used before gzclose to detect the error.

jserv:

Issue #735, opened 2026-05-17
RVV trap-path tests, arch-test V coverage, and decode layout guard


jserv:
Three follow-ups bundled because they are all about preventing RVV regressions; subdivide later if any one of them grows.

### 1. Trap-path smoke coverage

The current smoke harness (`mk/tests.mk` `check-rvv-smoke` matches `tail -n 1 == \"RVV smoke OK\"`) cannot distinguish an intentional illegal-instruction trap from a generic abort, and `tests/rvv-smoke.S` already documents this gap at line ~739. Targets:

- `vd`-vs-`v0` overlap traps in masked segmented loads.
- `vrgather` source/destination overlap traps.
- `vcompress` non-zero `vstart` traps.
- Fault-only-first mid-segment fault behavior (vl truncation vs trap on first active element).

Needs an `mtvec` / trap-handler scaffold in `tests/rvv-smoke.S` (or sibling) AND an updated `check-test` mechanism that asserts more than the last stdout line — exit-code assertions or marker strings.

### 2. `riscv-arch-test` V battery in CI

The existing arch-test plumbing (`tests/arch-test-target/setup.py`, `tests/arch-test-target/rv32emu/rv32emu_isa.yaml`) advertises `RV32IMCZicsr_Zifencei` and does not parse `V`. Work item is broader than \"wire it up\":

- Teach the target plugin about the V extension.
- Pull in the matching upstream test suite (commits/tag to be determined).
- Add a CI job gated on `CONFIG_EXT_V=y`.

This catches spec-compliance regressions the smoke test cannot.

### 3. Cross-struct opcode-offset `static_assert`

`try_fuse_sequence` does `memcpy(ir->fuse, ir, sizeof(opcode_fuse_t))`, so `rv_insn_t.opcode` and `opcode_fuse_t.opcode` share their byte position by contract. Add `static_assert(offsetof(rv_insn_t, opcode) == offsetof(opcode_fuse_t, opcode), \"...\")` in `src/decode.h`. `CHANGELOG.md` (2026-05-16: RVV Build and Opcode Layout Fixes) documents the regression that motivated this guard.

Tracked under the umbrella issue #504.

PR #766, opened 2026-09-10
Pin the rv_insn_t/opcode_fuse_t shared prefix

alanhc:
Addresses the third item of #735.

`try_fuse_sequence()` populates the fuse array with
`memcpy(ir->fuse + j, next_ir, sizeof(opcode_fuse_t))`, reinterpreting the
leading bytes of `rv_insn_t` as an `opcode_fuse_t`. That is only valid while
both structures agree on the offset of every field the fused handlers read
back, but the existing `_Static_assert`s in `decode.h` pin the layout of
`opcode_fuse_t` alone. Nothing catches a change on the `rv_insn_t` side.

The issue mentions `opcode`, but since the memcpy copies the whole 12-byte
prefix, the contract covers `imm`, `rd`, `rs1`, `rs2` and `opcode`, so this
pins all five plus a size relation.

Note that `rv_insn_t` gains members under `EXT_C`, `EXT_V` and `EXT_F`, so
the prefix layout is configuration-dependent; the asserts were checked with
`EXT_V` both enabled and disabled.

Verified by inserting a layout-shifting member ahead of `opcode`, which the
new assert rejects at compile time. Note that a `uint8_t` inserted there is
absorbed by existing alignment padding and does not shift `opcode`, so the
negative test needs a wider type.

`make check` passes; clang-format 20.1.7 clean.

---
## Summary by cubic
Pins the shared prefix layout between `rv_insn_t` and `opcode_fuse_t` so the `memcpy` in `try_fuse_sequence()` can't silently corrupt fused sequences when the structs drift. Addresses the third item of #735.

- Adds compile-time offset asserts for `imm`, `rd`, `rs1`, `rs2`, and `opcode`, plus a size check.

<sup>Written for commit acd830a86029cfeec3486c164e5dca1fb82200ca. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/rv32emu/pull/766?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jserv (changes requested):
Avoid `decode` tag in the subject of both pull request and git commits.

alanhc:
Dropped the `decode:` tag from the pull request subject and from the commit. The reword is all that changed: `acd830a8` and the previous `9b6058c8` share tree `b19ad86d` and parent `5e4686a6`.

src/decode.h
```diff
@@ -1078,5 +1078,32 @@ typedef struct rv_insn {
...
+_Static_assert(offsetof(rv_insn_t, rd) == offsetof(opcode_fuse_t, rd),
+               "rv_insn_t.rd must match opcode_fuse_t.rd");
+_Static_assert(offsetof(rv_insn_t, rs1) == offsetof(opcode_fuse_t, rs1),
+               "rv_insn_t.rs1 must match opcode_fuse_t.rs1");
+_Static_assert(offsetof(rv_insn_t, rs2) == offsetof(opcode_fuse_t, rs2),
+               "rv_insn_t.rs2 must match opcode_fuse_t.rs2");
+_Static_assert(offsetof(rv_insn_t, opcode) == offsetof(opcode_fuse_t, opcode),
+               "rv_insn_t.opcode must match opcode_fuse_t.opcode");
+_Static_assert(sizeof(rv_insn_t) >= sizeof(opcode_fuse_t),
+               "rv_insn_t must be large enough to copy an opcode_fuse_t from");
```
jserv, line 1106: `sizeof(rv_insn_t) >= sizeof(opcode_fuse_t)` cannot fail: `rv_insn_t` carries `imm2`, `fuse`, `pc` and several pointers past the prefix, so this adds no coverage. The width that is genuinely unpinned is `opcode`'s. Widening `rv_insn_t.opcode` to `uint32_t` leaves its offset at 8, so every assert above still passes while `opcode_fuse_t.opcode` reads back only the low half of the copied value. A width change to `rd`, `rs1` or `rs2` shifts `opcode` and gets caught; nothing sits after `opcode` to catch it, and that field already outgrew `uint8_t` once for the `EXT_V` fuse IDs.

```suggestion
_Static_assert(sizeof(((rv_insn_t *) 0)->opcode) ==
                   sizeof(((opcode_fuse_t *) 0)->opcode),
               "rv_insn_t.opcode must match opcode_fuse_t.opcode width");
```

src/decode.h
```diff
@@ -1078,5 +1078,32 @@ typedef struct rv_insn {
     branch_history_table_t *branch_table;
 } rv_insn_t;
 
+/* Compile-time layout verification of the rv_insn_t/opcode_fuse_t contract.
+ *
+ * try_fuse_sequence() populates the fuse array with
+ * memcpy(ir->fuse + j, next_ir, sizeof(opcode_fuse_t)), reinterpreting the
```
jserv, line 1084: Worth noting as a follow-up rather than a change here: the asserts pin the layout but not the type. `memcpy` into the untyped `mpool_alloc()` slot gives that storage the effective type of `rv_insn_t`, and the fused handlers then read it back through `opcode_fuse_t *`, which the default strict aliasing in the `-std=gnu11` build does not sanction. Building each entry by field assignment (`ir->fuse[j] = (opcode_fuse_t){.imm = next_ir->imm, ...}`) would retire both that question and the need for this assert block.

Issue #734, opened 2026-05-17
Implement remaining RVV 1.0 instruction families


jserv:
After PR #733 landed the bulk of the RISC-V V Vector extension, several V 1.0 instruction families are still unimplemented; the decoder currently rejects them with illegal-instruction. Implement each, with EMUL / overlap / alignment validation matching the existing helpers, plus the corresponding execution path (not just decode wiring):

- `vmv<nr>r.v` whole-register move (V 1.0 §16.6) — `src/decode_v.c` `op_100111` case 3. Helpers `rvv_whole_reg_load` / `rvv_whole_reg_store` in `src/rv32_v_template.c` already exist; this op needs its own register-to-register move path with `vstart` / `evl` handling per spec.
- `vfmv.s.f` / `vfmv.f.s` scalar-FP move (V 1.0 §16.2) — `src/decode_v.c` `op_010000` cases 1 and 5 (VWFUNARY0 / VRFUNARY0).
- `vzext.vf{2,4,8}` / `vsext.vf{2,4,8}` integer extension (V 1.0 §11.3) — `src/decode_v.c` `op_010010` case 2 (VXUNARY0).
- `vfcvt.*` / `vfwcvt.*` (V 1.0 §13.17, §13.18) and `vfncvt.*` (§13.19) FP conversions — `src/decode_v.c` `op_010010` case 1 (VFUNARY0); reuse scalar softfloat conversion paths.
- `vfsqrt.v` (§13.8), `vfrsqrt7.v` (§13.9), `vfrec7.v` (§13.10), `vfclass.v` (§13.14) — `src/decode_v.c` `op_010011` case 1 (VFUNARY1).

Each should land with smoke coverage in `tests/rvv-smoke.S`.

Tracked under the umbrella issue #504.

yc199911:
I've implemented the RVV 1.0 instructions tracked by this issue. The work is on
my branch `yc199911:issue-734-vmv-whole-reg` and covers:

- `vmv<nr>r.v`            — whole-register move
- `vfmv.s.f` / `vfmv.f.s` — scalar ↔ vector FP move
- `vzext.*` / `vsext.*`   — integer zero/sign extension
- `vfcvt.*`              — single-width FP ↔ integer conversion
- `vfwcvt.*` / `vfncvt.*` — widening / narrowing FP conversion (incl. rod)
- `vfsqrt.v`, `vfclass.v`, `vfrsqrt7.v`, `vfrec7.v` (VFUNARY1)

[...]

PR #769, opened 2026-09-10
Implement RVV floating-point scalar moves

alanhc:
Implements the floating-point scalar moves from #734.

Adds `vfmv.f.s` and `vfmv.s.f` (V 1.0 §16.3) at the VWFUNARY0 and VRFUNARY0
sites of funct6=0x10, both previously rejected with a FIXME so they would
not be misdecoded as `vmv.x.s` / `vmv.s.x`.

Both transfer element 0 between a vector register and an `f` register and
ignore LMUL. Two spec details are easy to miss:

- `vfmv.f.s` performs its operation even when `vstart >= vl` or `vl` is 0;
- `vfmv.s.f` writes nothing when `vstart >= vl`, which with `vl == 0` covers
  every `vstart`.

The masked encodings of both are reserved, and the sub-opcode field must be
zero; the decode tests cover those rejections and confirm `vmv.x.s` still
decodes from the same funct6.

Smoke coverage includes the `vl == 0` case, which a mutation that drops the
`vstart < vl` guard is caught by.

clean.

---
## Summary by cubic
Implements the RVV floating-point scalar moves `vfmv.f.s` and `vfmv.s.f` (V 1.0 §16.3). These transfer element 0 between a vector register and an `f` register, ignoring LMUL; previously their encodings were rejected to avoid misdecoding as `vmv.x.s` / `vmv.s.x`, and now they decode and execute.

**Implementation notes**
- `vfmv.f.s` performs its operation even when `vstart >= vl` or `vl == 0`.
- `vfmv.s.f` writes nothing when `vstart >= vl`, and with `vl == 0` that covers every `vstart`.
- Both instructions require SEW=32; masked encodings are reserved and rejected.
- Added constopt entries and smoke tests covering the `vl == 0` case.

<sup>Written for commit 8b0a2a773e9a7dc70a38f27a4a8b60a4ace8ac57. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/rv32emu/pull/769?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jserv:
Merge into #770

PR #770, opened 2026-09-10
Implement RVV single-width floating-point conversions

alanhc:
Implements the single-width conversions from #734. First of three PRs that
share the VFUNARY0 decode site (funct6=0x12, funct3=0x1); see the note at
the end.

Adds the six same-width `vfcvt` forms (V 1.0 §13.17). The `.rtz` variants
pin round-towards-zero through the `_r_minMag` SoftFloat entry points, while
the plain forms follow the dynamic rounding mode installed by
`rvv_fp_begin_round()`.

The widening (`vfwcvt`, vs1 0x08-0x0f) and narrowing (`vfncvt`, vs1
0x10-0x17) members are not decoded here and continue to be rejected rather
than misdecoded; the decode tests assert that.

One note on the smoke coverage: the test uses 1.5 as an input because it
rounds to 2 under RNE but truncates to 1 under RTZ. A value such as 2.5
gives 2 under both because of ties-to-even and would not distinguish an
implementation that ignored the `.rtz` suffix. Signed vs unsigned is pinned
with -3, which differs in both directions.

clean.

**Ordering:** this touches the same `case 1:` block as the `vfwcvt` and
`vfncvt` PRs. They are independent in substance but will conflict textually,
so they are best taken in order (this one, then vfwcvt, then vfncvt); I will
rebase the later two as each lands and update the "not implemented yet"
comments accordingly.

---
## Summary by cubic
Implements the six same-width `vfcvt` conversion instructions (V 1.0 §13.17), which were previously rejected. The `.rtz` variants pin round-towards-zero; the plain forms follow the dynamic rounding mode. Widening and narrowing forms remain unimplemented and are still rejected.

- Adds smoke tests that use 1.5 to distinguish dynamic rounding from `.rtz` behavior.
- Shares the VFUNARY0 decode site with future `vfwcvt`/`vfncvt` PRs; those will need rebasing in sequence.

<sup>Written for commit 0c25662798f2e0f548b69ffe2266d7a9f32695d9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/rv32emu/pull/770?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jserv:
Instead of reviewing/merging #771, #772, and #773, I would like to aggregate their commits in this pull request. Rework subject/description of this PR accordingly.

PR #772, opened 2026-09-10
Implement RVV narrowing floating-point conversions

alanhc:
Implements the narrowing conversions from #734. Third of three PRs sharing
the VFUNARY0 decode site; best taken after #770 and #771.

Adds the eight `vfncvt` forms (V 1.0 §13.19), producing a 32-bit destination
element from a 64-bit source.

Unlike the widening direction, narrowing shares the overlap constraints of
the other narrowing instructions, so the existing
`rvv_cross_eew_overlap_illegal()` applies unchanged and this follows the
`vnsrl_wv` pattern.

`vfncvt.rod.f.f.w` rounds towards odd. SoftFloat is already built with
`-D SOFTFLOAT_ROUND_ODD` (mk/softfloat.mk), so `softfloat_round_odd` can be
selected directly; that was confirmed against the library before relying on
it. The three rounding behaviours (dynamic frm, RTZ, ROD) are dispatched by
a parameter rather than duplicating the RVOP body.

The rod smoke test uses a double that lies exactly halfway between two
floats (1.0 + 2^-24): RNE gives 0x3f800000 and ROD gives 0x3f800001. An
ordinary value produces the same result under both modes and would not
exercise the path at all.

clean.

---
## Summary by cubic
Implements the eight `vfncvt` narrowing floating-point conversions from V 1.0 §13.19, so RVV code can convert 64-bit source elements to 32-bit destination elements. These encodings were previously rejected; they now execute using the same overlap constraints as the other narrowing instructions.

**Details**
- Rounding follows `frm` except `vfncvt.rod.f.f.w` (round towards odd) and the `.rtz` forms (truncate).
- The single-width and widening VFUNARY0 forms remain unimplemented and are still rejected.
- Smoke tests cover tie-to-even vs round-towards-odd behavior and signed vs unsigned 64-bit integer sources.

<sup>Written for commit 67395906490afb8240f7ec1faaacbd35fde6c78b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/rv32emu/pull/772?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jserv:
Merge into #770

PR #773, opened 2026-09-10
Implement RVV floating-point unary instructions

alanhc:
Implements the last of the five families in #734.

Adds `vfsqrt.v`, `vfrsqrt7.v`, `vfrec7.v` and `vfclass.v` (V 1.0 §13.8-13.14)
at the VFUNARY1 decode site (funct6=0x13, funct3=0x1), previously rejected as
reserved.

`vfrsqrt7` and `vfrec7` are the table-driven 7-bit estimates. The two
128-entry tables are transcribed from the specification's `vfrsqrt7.adoc` and
`vfrec7.adoc`, and kept in the specification's row layout by disabling
clang-format around them, as several existing files already do.

Both instructions were developed against the worked examples in the spec
before being written in C:

- `vfrsqrt7(0x00718abc) = 0x5f080000`, `vfrsqrt7(0x7f765432) = 0x1f820000`
- `vfrec7(0x00718abc) = 0x7e900000`, `vfrec7(0x7f765432) = 0x00214000`

All four are covered by the smoke test. Two details that cost some time and
may be worth noting for review:

- for a subnormal `vfrec7` output the implicit leading one has to be
  restored before the right shift, otherwise the low bit of the result is
  lost (the second worked example catches this);
- the two published `vfrsqrt7` examples both have an odd normalized
  exponent, so the floor in `(3*B - 1 - e) / 2` is absorbed by integer
  division and an off-by-one there passes both of them. The smoke test adds
  an even-exponent input (4.0 -> 0x3eff0000) which does catch it.

clean.

---
## Summary by cubic
Implements the last of the five RVV floating-point unary instruction families in #734 by adding `vfsqrt.v`, `vfrsqrt7.v`, `vfrec7.v`, and `vfclass.v` (V 1.0 §13.8–13.14) at the VFUNARY1 decode site, which previously rejected the encoding as reserved.

- The four instructions are distinguished by the `vs1` field (00000, 00100, 00101, 10000); both masked and unmasked forms are permitted.
- The 7-bit estimate instructions use 128-entry lookup tables transcribed from the spec, with clang-format disabled around them to preserve the spec's row layout.
- Smoke tests cover both spec worked examples; an even-exponent `vfrsqrt7` input (4.0 → 0x3eff0000) is added because both published examples share an odd normalized exponent and would not catch a floor off-by-one in the output-exponent formula.
- Subnormal `vfrec7` output restores the implicit leading one before the right shift, otherwise the low bit is lost (the second worked example exercises this path).
- `vfrsqrt7` results are independent of the dynamic rounding mode; only `vfrec7` overflow depends on `frm`, so only `vfsqrt` and `vfrec7` invoke `rvv_fp_begin_round`.
- All four are non-foldable in constopt, run as `V_NOP` without `EXT_V`, and `make check` passes in default, `EXT_V`, JIT+`EXT_V`, and `EXT_V`-without-`EXT_F` builds.

<sup>Written for commit 08bdc4dfa1f7330b4aef3025bf72c6a6a1de8f14. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/rv32emu/pull/773?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jserv:
Merge into #770

PR #776, opened 2026-09-16
Rebuild the map on parent-linked red-black nodes

jserv:
The map paired a left-leaning insert derived from jemalloc's rb.h with a general parent-pointer erase, and the two disagree about what a valid tree looks like. Erasing leaves a right-leaning red link, which is fine for a red-black tree but not for a left-leaning one, and the next insertion descending through that link applies its lean-left rotation and produces a red-red violation. Inserting 8, 747 and 776, erasing 8, then inserting 880 is enough to corrupt the tree. Both `breakpoint_map` and `fd_map` insert after erasing, so this was reachable in normal use, and the existing tests missed it because none of them inserted after an erase.

Both halves now use the textbook bottom-up fixups so they share one invariant. The parent links that erase already required make the path array, its depth ceiling, and the silent insert failure when a tree outgrew it unnecessary, so those go away with it. Nodes come from map-owned chunks with a free list, and keys and values live inline at offsets computed once per map rather than behind per-node pointers, which brings an int-keyed int-valued node from 48 to 40 bytes. The iterator now carries its owning map, so `map_erase` cannot be handed an iterator belonging to a different map, and `map_iter_value_ptr` gives callers the address of a stored value: `breakpoint_map_find` had been reinterpreting a stored `breakpoint_t` as a pointer for want of that accessor.

The other three commits are separate concerns that surfaced while working here, kept apart so they can be read or reverted on their own. `__UNREACHABLE` expanded to an empty statement on compilers with neither `__builtin_unreachable` nor `__assume`, which let about ten non-void functions run off the end; it now aborts. The guest stack wrote eight bytes into each four-byte argv slot and came out right only because the host is little-endian. And six functions implemented in `syscall_sdl.c` were declared by hand at each use site across three files, two of them with an empty parameter list, so they move to a header and the build now enforces `-Wmissing-prototypes` and `-Wstrict-prototypes`, which immediately caught five more instances that only the system build compiles.

Verified with `make check` and `make tests` on default, SDL, gdbstub, and system builds, and each of the four commits builds and passes `make check` on its own. The map was checked against a shadow model under randomized insert and erase interleaving, exhaustively over every insert permutation crossed with every erase permutation for n up to 6, and with malloc failure injected at each allocation, all under ASan and UBSan. argv delivery was confirmed with a guest that prints its own `argv` and checks the terminator. No timing numbers are quoted here: the machine available for measurement was under heavy unrelated load, so any A/B I ran would have been noise.

Deliberately left out: the mirrored left and right arms of the fixups could collapse behind a `child[2]` array, which is the largest remaining cleanup in this file, but it changes the public node layout and belongs in its own change rather than riding along with a bug fix. The 25 remaining `-Wcast-align` warnings are all the `container_of` idiom in `cache.c`, `mpool.c`, and `elf.c`, and enabling that warning project-wide needs those sites annotated first.

---
## Summary by cubic
Rebuilds the map on parent-linked red-black nodes so inserting after an erase no longer corrupts `breakpoint_map` or `fd_map`. The old left-leaning insert disagreed with general erase; the new code shares textbook bottom-up fixups, and nodes now come from map-owned chunks with inline key/value storage, shrinking an int/int node from 48 to 40 bytes.

**Map API**

- Iterators carry their owning map, so `map_at_end` and `map_erase` no longer take a map argument.
- `map_iter_key_ptr` and `map_iter_value_ptr` expose stored addresses; keys are read-only.
- `map_ceil` and `map_floor` replace the bound helpers; comparators may return any negative, zero, or positive value, so `memcmp`/`strcmp` work directly.
- `map_new_aligned` supports power-of-two alignment for over-aligned payloads.

**Other fixes**

- Writes guest argument entries as `uint32_t`, removing host-endian dependence from stack setup.
- Fallback `__UNREACHABLE` now calls `abort()` instead of letting non-void functions fall through.
- Cross-TU entry points move to `syscall_sdl.h`; `-Wmissing-prototypes` and `-Wstrict-prototypes` are enabled (vendored SoftFloat excluded), and T2C-only `rv_destroy_t2c` plus JIT wrappers are now static.
- Standard-stream remap failures are logged instead of leaving descriptors silently unmapped.
- Adds randomized, exhaustive, allocation-failure, alignment, and lifecycle coverage for maps, including the erase-then-insert regression.

<sup>Written for commit b654564fbb8e73ecc1f74676d56cb29b1b1d060f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/rv32emu/pull/776?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: b654564fbb8e73ecc1f74676d56cb29b1b1d060f | Previous: f7ed516c033125bc8f60899542546dcc1d664f39 | Ratio |
|-|-|-|-|
| `Dhrystone` | `2872` DMIPS | `2166.25` DMIPS | `0.75` |
| `CoreMark` | `2113.674` iterations/sec | `1752.571` iterations/sec | `0.83` |

</details>

PR #777, opened 2026-09-19
Fix wasm system music

### LINKED ISSUE #751: WASM target unable to play music when running SDL games on system emulation
Error: `ERROR src/syscall_sdl.c:806: Mix_LoadMUSType_RW failed: Couldn't open /etc/timidity/freepat`

Root cause: `missing /etc/timitidy` bundle.

Solution: Adding `/etc/timidity` bundle during wasm [compilation](https://github.com/sysprog21/rv32emu/blob/664fdd042703237436c3245bc510b78dd3fe2e06/mk/wasm.mk#L73):
```
...
--embed-file build/timidity@/etc/timidity
...
```

ChinYikMing:
- Update WASM README
To run wasm_defconfig before any build.

- Resolve the web workers path correctly
So emcc 3.1.51 works.

- Fix music unable to play when playing video games 
Close #751 

- CI: Deploy rv32emu-demo with music enabled in system emulation
During rv32emu-demo deployment, the timitidy.tar.gz should be commited.

- CI: Reuse target when building wasm for user/system

- CI: Commit common.js during demo page deployment

- Reject unsafe path
Reject absolute names and any .. path component before creating
directories or writing files.

---
## Summary by cubic
Fixes the system-mode WASM demo so Doom and Quake music plays again and the web worker path resolves correctly under emcc 3.1.51. Closes #751.

**Details**
- Timidity instruments are now packaged as `timidity.tar` and `timidity.tar.gz`, fetched at runtime, and unpacked into `/etc/timidity`; browsers without `DecompressionStream` use the plain tar.
- If those files are missing from the deployment, the guest still boots but logs a warning and starts without MIDI music.
- The web worker copy step uses `basename` so the staged worker file matches what the browser requests.
- The tar extraction and error helpers are shared from `assets/wasm/js/common.js` by both demo pages; extraction rejects absolute paths and `..` components so entries can't escape `/etc/timidity`.
- `docs/wasm.md` now instructs running `make wasm_defconfig` before building, and CI stages the timidity files in both user and system demo builds.

<sup>Written for commit b5b0e19321d1334fbc788801979407495822e167. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/rv32emu/pull/777?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


assets/wasm/html/system.html
```diff
@@ -692,10 +753,21 @@ <h1>Linux System Emulation</h1>
           fetchWithProgress('rootfs.cpio', 'Rootfs'),
           fetchOptionalDisk(),
         ]);
+        const timidityGzip = await fetch("timidity.tar.gz");
 
         if (term) term.write('\x1b[33mWriting artifacts to VFS...\x1b[0m\r\n');
         Module.FS.writeFile('/Image', kernelData);
         Module.FS.writeFile('/rootfs.cpio', rootfsData);
+        Module.FS.mkdirTree('/etc/timidity');
```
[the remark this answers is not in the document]
ChinYikMing: The timidity.tar.gz is extracted and written to the WASM sandbox FS's /etc/timidity. Without this, the guestOS cannot access the timidity instruments and configuration on wasm platform.

assets/wasm/html/system.html
```diff
@@ -692,10 +753,21 @@ <h1>Linux System Emulation</h1>
           fetchWithProgress('rootfs.cpio', 'Rootfs'),
           fetchOptionalDisk(),
         ]);
+        const timidityGzip = await fetch("timidity.tar.gz");
```
jserv: `fetch` resolves for 404 and 500 as well, so a deployment that has not been refreshed with `timidity.tar.gz` reaches `DecompressionStream` with an HTML error page and fails with a gzip decode error that names nothing useful. Check the status and report it, the way `ensureProgramAssets` in `user.html` already does.

assets/wasm/html/system.html
```diff
@@ -692,10 +753,21 @@ <h1>Linux System Emulation</h1>
...
 
         if (term) term.write('\x1b[33mWriting artifacts to VFS...\x1b[0m\r\n');
         Module.FS.writeFile('/Image', kernelData);
         Module.FS.writeFile('/rootfs.cpio', rootfsData);
+        Module.FS.mkdirTree('/etc/timidity');
+	try {
+          const gunzipped = timidityGzip.body.pipeThrough(new DecompressionStream('gzip'));
```
jserv: No feature detection here, so on a browser without `DecompressionStream` (Firefox before 113, Safari before 16.4) the constructor throws and the kernel never boots. The plain `timidity.tar` that `.ci/wasm-build.sh` stages right next to the gzip is never fetched, which is the exact case it was staged for. Either mirror the `supportsGzipDecompression` and `fallbackUrl` handling in `user.html`, or stop copying the plain tar into the system stage so the deploy is not carrying an unused copy of the instrument set. `timidityGzip.body` can also be `null`, in which case this is a `TypeError` rather than a decode failure.

assets/wasm/html/system.html
```diff
@@ -692,10 +753,21 @@ <h1>Linux System Emulation</h1>
...
+          let tarBuf = await new Response(gunzipped).arrayBuffer();
+          const n = extractTarToMemfs(new Uint8Array(tarBuf), "/etc/timidity");
+          if (term) term.write(
+            `\x1b[36m> Unpacked ${n} files to /etc/timidity\x1b[0m\r\n`);
+	} catch (e) {
+          throw new Error(`Processing timitidy.tar.gz failed: ${formatError(e)}`);
```
jserv: Rethrowing here turns a missing or unreadable instrument set into a boot failure: this sits inside the `runSysButton` handler ahead of `callMain`, so nothing after it runs. `/etc/timidity` only feeds host SDL_mixer MIDI, and system mode booted fine without it before this change. Write a warning to `term` and fall through so the guest still starts. The message also spells the file `timitidy.tar.gz`.

assets/wasm/html/system.html
```diff
@@ -692,10 +753,21 @@ <h1>Linux System Emulation</h1>
           fetchWithProgress('rootfs.cpio', 'Rootfs'),
           fetchOptionalDisk(),
         ]);
+        const timidityGzip = await fetch("timidity.tar.gz");
 
         if (term) term.write('\x1b[33mWriting artifacts to VFS...\x1b[0m\r\n');
         Module.FS.writeFile('/Image', kernelData);
         Module.FS.writeFile('/rootfs.cpio', rootfsData);
+        Module.FS.mkdirTree('/etc/timidity');
+	try {
```
jserv: This line, its `} catch (e) {`, and the closing `}` are indented with tabs while the rest of the script uses spaces, so the block does not line up with the code it wraps.

assets/wasm/html/system.html
```diff
@@ -633,6 +633,54 @@ <h1>Linux System Emulation</h1>
     const progressElement = document.getElementById('progress');
     const runSysButton = document.getElementById('runSysButton');
 
+    function extractTarToMemfs(data, destPrefix) {
```
jserv: `extractTarToMemfs` and `formatError` are byte-identical copies of the ones in `user.html`. Two copies of a tar parser drift, and the copy that landed here already lost the status check and the non-gzip path that the original has. Worth a follow-up that hoists both into a shared script under `assets/wasm/js/` and pulls it into both pages.

mk/wasm.mk
```diff
@@ -195,10 +195,10 @@ $(XTERM_CSS): | $(XTERM_VENDOR)
 XTERM_DATA := $(XTERM_JS) $(XTERM_CSS)
 
 # Dependencies for WASM build
-# System mode: kernel image only (no audio/games)
+# System mode: kernel image games, and timidity for MIDI audio
```
jserv: "kernel image games" lost a word. System mode now pulls the kernel image and timidity, so say that.

assets/wasm/html/system.html
ChinYikMing: Create a new js file called assets/wasm/js/common.js for the purpose.

jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: 9165c776c0bc88100692d0aa2050f3899847dc6f | Previous: e8b557fe4f390671691b29c1e24b39f5d1c90b55 | Ratio |
|-|-|-|-|
| `Dhrystone` | `4744.8` DMIPS | `3468.2` DMIPS | `0.73` |
| `CoreMark` | `3098.873` iterations/sec | `2676.518` iterations/sec | `0.86` |

</details>

jserv:

PR #778, opened 2026-09-19
Support zlib on wasm platform

ChinYikMing:
**Add wasm_system_defconfig**: Introduces a new configuration for WASM system emulation that correctly detects and links against zlib during the Kconfig evaluation phase.
**Optimize demo payload downloads**: Replaces the uncompressed kernel and rootfs images with gzipped versions. This significantly reduces the payload size and improves the loading speed(> 2x gain) on the rv32emu-demo GitHub page.

---
## Summary by cubic
Adds `zlib` support to Emscripten builds so the WASM system demo boots `Image.gz` and `rootfs.cpio.gz`, cutting downloads by more than half. System emulation builds now use a dedicated `wasm_system_defconfig` instead of `wasm_defconfig` plus deferred `ENABLE_SYSTEM=1`, which skipped `zlib` detection.

- Treats Emscripten as `zlib`-capable in `tools/detect-env.py` and `configs/Kconfig`, and passes `-sUSE_ZLIB=1` when `CONFIG_LINK_ZLIB=y`.
- Updates `system.html`, docs, build scripts, CI, and the Emscripten cache warm-up to serve and link the gzipped artifacts.
- Fails with a clear fatal error when a gzipped image is loaded without `zlib` support.

<sup>Written for commit bdff57fc2bcf845a2381567be44de9aba6ef370a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/rv32emu/pull/778?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ChinYikMing:
Converting to draft since it is on top of #777.

jserv (commented):
# Benchmarks

<details>

| Benchmark suite | Current: bdff57fc2bcf845a2381567be44de9aba6ef370a | Previous: 002a256381e2114e888300adfc133d530fbb1a1c | Ratio |
|-|-|-|-|
| `Dhrystone` | `4500.25` DMIPS | `2721.6` DMIPS | `0.60` |
| `CoreMark` | `3384.984` iterations/sec | `2086.718` iterations/sec | `0.62` |

</details>

tools/detect-env.py
```diff
@@ -175,7 +175,9 @@ def have_riscv_toolchain():
 
 def have_zlib():
     """Check if ZLIB is available."""
-    return check_pkg_config("zlib") and _compiler_type() != "Emscripten"
+    if _compiler_type() == "Emscripten":
+        return True
```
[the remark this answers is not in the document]
ChinYikMing: have_emcc() does not mean want to use emcc, the _compiler_type() is the user current using compiler.

ChinYikMing:
@cubic-dev-ai

ChinYikMing:
@cubic-dev-ai

configs/Kconfig
```diff
@@ -430,7 +430,7 @@ config DEBUG_SYMBOLS
 
 config LINK_ZLIB
     def_bool $(python,assert run(sys.executable, 'tools/detect-env.py', '--have-zlib'))
-    depends on SYSTEM && !ELF_LOADER && !CC_IS_EMCC
+    depends on SYSTEM && !ELF_LOADER
```
jserv, line 487: Dropping `!CC_IS_EMCC` here moves the Emscripten decision into `tools/detect-env.py`, but that script never sees `emcc`. `make wasm_system_defconfig` runs `defconfig.py` before any `.config` exists, and `mk/toolchain.mk` sets `CC := emcc` with a plain assignment that make does not export, so `_compiler_type()` probes the host `cc` and the new Emscripten branch is dead code on that path. `CONFIG_LINK_ZLIB` for the WASM build therefore comes from the host's `pkg-config --exists zlib`: on a host with no `zlib.pc` the demo links without zlib and then boots `-k Image.gz`, which it cannot read. `SDL` already handles the Emscripten ports through the Kconfig symbol; do the same here.

```kconfig
config LINK_ZLIB
    bool
    default y if CC_IS_EMCC
    default $(python,assert run(sys.executable, 'tools/detect-env.py', '--have-zlib'))
    depends on SYSTEM && !ELF_LOADER
```

mk/wasm.mk
```diff
@@ -297,8 +302,8 @@ prepare-web: $(start_web_deps)
 	$(foreach T, $(STATIC_WEB_FILES), $(call cp-web-file, $(T)))
 	$(call cp-web-worker)
 ifeq ($(CONFIG_SYSTEM),y)
-	$(Q)cp build/linux-image/Image $(DEMO_DIR)/
-	$(Q)cp $(OUT)/linux-image/rootfs.web.cpio $(DEMO_DIR)/rootfs.cpio
+	$(Q)cp build/linux-image/Image.gz $(DEMO_DIR)/
+	$(Q)gzip -9 -c $(OUT)/linux-image/rootfs.web.cpio > $(DEMO_DIR)/rootfs.cpio.gz
```
jserv, line 306: `compress-web` further down still copies `build/linux-image/Image` and `rootfs.web.cpio` to `rootfs.cpio`, and its `rm -f $(DEMO_DIR)/*.gz` wipes anything gzipped, so a demo staged with that target 404s on both fetches `system.html` now makes. It needs the same two lines as here. While touching them: `cp build/linux-image/Image.gz` hardcodes `build/` where the prerequisite is `$(OUT)/linux-image/Image.gz`, so a build with `OUT=` set copies from the wrong tree.

configs/wasm_system_defconfig
```diff
@@ -0,0 +1,8 @@
+# WebAssembly build with Emscripten
```
jserv, line 1: `.github/workflows/main.yml` still builds the WASM system binary with `make wasm_defconfig` followed by `ENABLE_SYSTEM=1 ENABLE_GOLDFISH_RTC=1`. `ENABLE_SYSTEM` overrides `CONFIG_SYSTEM` at make level but not `CONFIG_LINK_ZLIB`, which `wasm_defconfig` already resolved to `n` because `SYSTEM` was unset there, so PR CI compiles neither the zlib path nor `-sUSE_ZLIB=1`. Pointing that step at this defconfig is what keeps the deployed configuration under test.

.github/workflows/deploy-wasm.yml
```diff
@@ -43,7 +43,7 @@ jobs:
               # Build system configuration
               mk/kconfig.mk
               mk/wasm.mk
-              configs/wasm_defconfig
+              configs/wasm_system_defconfig
```
jserv, line 48: `configs/Kconfig` belongs in this list too: `LINK_ZLIB` is resolved there, so a merge that changes only that file can flip the system demo between a zlib build and a non-zlib one while this job stays skipped.

assets/wasm/html/system.html
```diff
@@ -739,7 +739,7 @@ <h1>Linux System Emulation</h1>
             `\x1b[33m> ${formatError(e)}; booting without MIDI music\x1b[0m\r\n`);
         }
 
-        let bootArgs = '-k Image -i rootfs.cpio';
+        let bootArgs = '-k Image.gz -i rootfs.cpio.gz';
```
jserv, line 742: Passing gzipped images turns a truncated or corrupt download into a decompression failure inside the emulator rather than a short read, and `map_file()` handles that with `exit(EXIT_FAILURE)`. That lands in the same `catch` as the fetches, so the page reports `Download failed` and prints `err.message`, which Emscripten's `ExitStatus` does not carry. Give `Module['run_system'](bootArgs)` its own `try`/`catch` that names the boot as the failure and formats the error with `formatError`.

src/riscv.c
```diff
@@ -16,7 +16,6 @@
 #include <termios.h>
 #include "dtc/libfdt/libfdt.h"
 
-/* TODO: support WASM target by leveraging -sUSE_ZLIB=1 */
 #if RV32_HAS(LINK_ZLIB)
```
jserv, line 20: `is_gzip_file()` lives inside this guard, so a build with `LINK_ZLIB` off reads a gzip stream straight into guest RAM and the kernel never starts, with nothing printed. Now that the WASM demo hands the emulator `Image.gz` unconditionally, that combination is reachable whenever zlib detection resolves the wrong way. Worth a follow-up here: check the magic outside the guard and fail with a message naming the file as gzip-compressed when the build has no zlib.

assets/wasm/html/system.html
ChinYikMing: DecompressionStream is used for timidity.tar.gz, the Image.gz and rootfs.cpio.gz is ungzipped via emcc's zlib port.

src/riscv.c
ChinYikMing: Now if the given images is gzipped but no ZLIB is linked. Error will be shown to user:
> FATAL src/riscv.c:391: Image.gz is gzipped file and zlib is not linked, fail to map file

jserv:

PR #779, opened 2026-09-19
Add a new formatter for HTML/JS files

ChinYikMing:
## Summary by cubic
Adds `Prettier` as the standard formatter for HTML and JS files and applies it to the `assets/wasm/html` pages and `assets/wasm/js/common.js`, so these files now follow a consistent style with no behavior changes.

- CI installs Prettier 3.9.8 and checks tracked HTML and JS files.
- `make format` formats HTML and JS files, and `CONTRIBUTING.md` documents the requirement and commands.
- The generated `coi-serviceworker.min.js` is excluded from formatting via `.prettierignore`.

<sup>Written for commit 4a126fd034222f07ad1d88d89e9d3aed66335b0e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/rv32emu/pull/779?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ChinYikMing:
Depends on #778 because it modifies html files.

ChinYikMing:
@cubic-dev-ai

ChinYikMing:
@cubic-dev-ai

.ci/check-format.sh
```diff
@@ -76,4 +76,40 @@ else
...
+HTML_SOURCES=()
+while IFS= read -r file; do
+    [ -n "$file" ] && HTML_SOURCES+=("$file")
+done < <(git ls-files -- '*.html')
+
+if [ ${#HTML_SOURCES[@]} -gt 0 ]; then
+    echo "Checking HTML files..."
+    HTML_FORMAT_EXIT=0
+    for html_src in "${HTML_SOURCES[@]}"; do
+        npx --no-install prettier --check "${html_src}"
```
jserv: The C block hands every path to a single `clang-format` run; this spawns a `node` process per file and adds the per-file statuses together, and `exit` truncates that sum to 8 bits, so a run whose statuses happen to total a multiple of 256 reports success. `prettier --check` accepts a list: pass `"${HTML_SOURCES[@]}"` in one invocation and set the flag to 1 on failure. Same for the JS loop below.

.ci/check-format.sh
```diff
@@ -76,4 +76,40 @@ else
...
+while IFS= read -r file; do
+    [ -n "$file" ] && JS_SOURCES+=("$file")
+done < <(git ls-files -- '*.js')
+
+if [ ${#JS_SOURCES[@]} -gt 0 ]; then
+    echo "Checking JS files..."
+    JS_FORMAT_EXIT=0
+    for js_src in "${JS_SOURCES[@]}"; do
+        # Skip coi-serviceworker.min.js
+        if [[ "${js_src}" == *"coi-serviceworker.min.js" ]]; then
```
jserv: This exclusion also lives as a `find ! -name` clause in `mk/tools.mk` and again in the `CONTRIBUTING.md` command, so the next vendored or minified asset has to be added in three places, and an editor running `prettier --write` honors none of them. A `.prettierignore` at the repo root states it once and both `--check` and `--write` respect it, which lets all three call sites just name the directories.

.ci/install-formatters.sh
```diff
@@ -26,3 +26,6 @@ pip3 install --break-system-packages black==25.1.0
 DTSFMT_URL="https://github.com/mskelton/dtsfmt/releases/download/${DTSFMT_VERSION}/dtsfmt-x86_64-unknown-linux-musl.tar.gz"
 curl -fsSL --retry 3 "${DTSFMT_URL}" | sudo tar -xz -C /usr/local/bin dtsfmt
 dtsfmt --version
+
+# HTML formatter
+npm install --save-dev --save-exact prettier@3.9.8
```
jserv: Every other formatter here lands outside the checkout (apt, pip, a release binary in `/usr/local/bin`). `--save-dev` writes `package.json`, `package-lock.json`, and `node_modules/` into the working tree, and `.gitignore` covers none of the three. `node_modules` then sits directly in the path that `make format` walks with `find`. Install it the way the rest are installed: `sudo npm install -g prettier@3.9.8`.

CONTRIBUTING.md
```diff
@@ -51,13 +52,16 @@ Software requirements:
...
+* [prettier](https://github.com/prettier/prettier) version 3.9.8.
 
 To maintain a uniform style across languages, run:
 * `clang-format -i *.{c,h}` to apply the project’s C/C++ formatting rules from the up-to-date .clang-format file.
 * `shfmt -w $(find . -type f -name "*.sh")` to clean and standardize all shell scripts.
 * `black .` to enforce a consistent, idiomatic layout for Python code.
 * `dtsfmt ./src/devices/` to enforce a consistent layout for device tree source (and include).
-* `make format` to automatically run all four formatters.
+* `npx --no-install prettier --write ./assets/wasm/html/` to enforce a consistent layout for HTML files.
+* `npx --no-install prettier --write $(find assets/wasm/js -maxdepth 1 -type f ! -name "coi-serviceworker.min.js")` to enforce a consistent layout for JS files.
```
jserv: `--no-install` resolves `prettier` from a local `node_modules`, and the repo ships no `package.json`, so on a fresh clone this command and the HTML one above report a missing package instead of formatting anything. State how a contributor gets `prettier` 3.9.8 onto their machine, then both commands can simply name the directories. The `$(find ...)` substitution here is also unquoted, so it splits on any path containing a space.

CONTRIBUTING.md
```diff
@@ -95,6 +99,10 @@ Python scripts must be clean, consistent, and adhere to modern Python best pract
 * Remove trailing whitespace at the end of lines.
 * Ensure files end with a newline.
 
+## Coding Style for HTML/JS
+
+Prettier is an opinionated code formatter.
```
jserv, line 105: The shell and Python sections list rules a contributor can check a file against; this one states none. No `.prettierrc` is committed either, so what CI enforces is whatever defaults 3.9.8 ships (80 column width, double quotes, two space indent, trailing commas), and a later major release can change them under the project. Either write the settings down here or commit a config so they live in the repo rather than in a tool version.

mk/tools.mk
```diff
@@ -53,18 +53,26 @@ format:
 	SHFMT=$$(which shfmt 2>/dev/null); \
 	DTSFMT=$$(which dtsfmt 2>/dev/null); \
 	BLACK=$$(which black 2>/dev/null); \
+	NODE=$$(which node 2>/dev/null); \
+	NPX=$$(which npx 2>/dev/null); \
+	PRETTIER=$$(npx prettier --version 2>/dev/null); \
```
jserv: Every other formatter here is located with `which`; this one runs `npx` without `--no-install`, so on a machine with no local `prettier` it reaches for the registry and fetches whatever is latest, unpinned against the 3.9.8 that CI checks. `make format` can then rewrite files in a style `check-format.sh` rejects. With `2>/dev/null` on the probe, npm's confirmation prompt is hidden while it waits on stdin. The `npx` on the `xargs` line has the same gap, and `NODE` is detected and checked but never used.

```suggestion
	PRETTIER=$$(npx --no-install prettier --version 2>/dev/null); \
```

mk/tools.mk
```diff
@@ -53,18 +53,26 @@ format:
...
 	SUBMODULES=$$(git config --file .gitmodules --get-regexp path 2>/dev/null | awk '{ print $$2 }') && \
 	PRUNE_PATHS="./$(OUT)" && \
 	for subm in $$SUBMODULES; do PRUNE_PATHS="$$PRUNE_PATHS ./$$subm"; done && \
 	PRUNE_ARGS=$$(echo "$$PRUNE_PATHS" | tr ' ' '\n' | sed 's/^/-path /;s/$$/ -o/' | tr '\n' ' ' | sed 's/ -o $$//') && \
 	find . \( $$PRUNE_ARGS \) -prune -o -name '*.[ch]' -print0 | xargs -0 $$CLANG_FORMAT -i && \
 	find . \( $$PRUNE_ARGS \) -prune -o -name '*.sh' -print0 | xargs -0 $$SHFMT -w && \
 	find . \( $$PRUNE_ARGS \) -prune -o \( -name '*.dts' -o -name '*.dtsi' \) -print0 | xargs -0 -I{} $$DTSFMT {} && \
-	find . \( $$PRUNE_ARGS \) -prune -o \( -name '*.py' -o -name '*.pyi' \) -print0 | xargs -0 $$BLACK --quiet
+	find . \( $$PRUNE_ARGS \) -prune -o \( -name '*.py' -o -name '*.pyi' \) -print0 | xargs -0 $$BLACK --quiet && \
+	find . \( $$PRUNE_ARGS \) -prune -o \( -name '*.html' -o \( -name '*.js' ! -name 'coi-serviceworker.min.js' \) \) -print0 | \
```
jserv: `PRUNE_PATHS` covers `$(OUT)` and the submodules only. Once `.ci/install-formatters.sh` has put a `node_modules/` at the root, this `find` descends into it and hands every dependency's `.js` to `prettier`, and the `clang-format`, `shfmt`, and `black` lines above do the same for their own extensions. Add `./node_modules` to `PRUNE_PATHS`.

ChinYikMing: Install the prettier via: `npm install -g prettier@3.9.8` does not create `node_modules/` at the project root.

```
npm install -g prettier@3.9.8

npx --no-install prettier --version
npm notice run npx
npm notice run 'prettier' --version
3.9.8
```

jserv:

